Issue #31, opened 2019-12-26
phones: do we need session reset, or can we support multiple sessions per user?


groovecoder:
When a user texts for a new relay number and they already have a session, could we just give them a(nother) open session? Per Twilio docs, the 2nd open session will be a different proxy number. So ...

1. User A opens session X and gets proxy number X
2. User A gives proxy number X to 3rd-party B
3. User A opens session Y and gets proxy number Y
4. User A gives proxy number X to 3rd-party C
5. 3rd-party B sends a message to X
  * Updates session X from `open` to `in-progress`
6. 3rd-party C sends a message to Y
  * Updates session Y from `open` to `in-progress`

There may be a multi-user race condition where sessions could be crossed ...

1. User A opens session X and gets proxy number X
2. User A gives proxy number X to 3rd-party B
3. User A opens session Y and gets proxy number Y
4. User A gives proxy number X to 3rd-party C
5. Unrelated 3rd-party D sends a message to X
  * Updates session X from `open` to `in-progress` between A and D
6. Unrelated 3rd-party E sends a message to Y
  * Updates session Y from `open` to `in-progress` between A and E

But I believe this multi-user race condition exists even without multiple sessions? I.e.,

1. User A opens session X and gets proxy number X
2. User A gives proxy number X to 3rd-party B
3. Unrelated 3rd-party D sends a message to X
  * Updates session X from `open` to `in-progress` between A and D

So ... it may be that multi-session makes the multi-user race condition worse.

groovecoder:
Confirmed the race condition exists and that multi-session will make it worse. So, before supporting this, need to figure out how to prevent this. Options I can think of, in order of simplicity+preference:

1. Before we make a session, can [check our list of phone numbers](https://www.twilio.com/docs/proxy/api/phone-number#get-the-list-of-phone-numbers-associated-with-a-proxy-service) for a number with `in_use`: 0, and rely on Twilio to prefer to allocate numbers not already in a session? If there are no `in_use`: 0 numbers, then we don't create a new session.
2. When opening a session, can we make [a "reserved" number](https://www.twilio.com/docs/proxy/reserved-phone-numbers) for the first participant in the session? And then "un-reserve" it after the session closes?
3. When we open a session, can we detect if the proxy number assigned is already in an open session? If so, can we manually destroy that session and try again until we make a session that has an number not involved in any other open session? Note: could be expensive, because Twilio charges $0.10 for "additional active sessions".

groovecoder:
[...]

PR #32, opened 2019-12-27
for #31: only use available numbers for sessions

groovecoder:
To help prevent sessions from being crossed:

1. When a request for a new session number comes in, check that there's an available proxy number not in use by any session.
2. Since Twilio doesn't automatically assign not-in-use numbers, repeatedly try to create the new session until it's created with the not-in-use number.


groovecoder:
Pushed this branch up to heroku and it seems to be working fine. So merging to master.

PR #52, opened 2020-03-08
Switch to socketlabs 49

groovecoder:


emails/views.py
```diff
@@ -61,27 +74,26 @@ def inbound(request):
...
-
-    return HttpResponse(message)
+    sl_message = BasicMessage()
+    sl_message.subject = 'Forwarding email from %s sent to %s' % (
+        from_address, local_portion
+    )
+    sl_message.html_body = html
+    sl_message.plain_text_body = text
+    sl_message.from_email_address = EmailAddress(
+        'inbound@privaterelay.groovecoder.com'
```
oremj: Should we make this configurable in the PR?

groovecoder: Good catch, yes!

Issue #77, opened 2020-03-16
Add browserAction icon and panel


groovecoder:
* Create from the addon (browser action icon) is the primary use case of the addon pop-up panel, along with a link out to the web property to manage existing entries
* Should be able to click on an addon icon which opens a pop-up panel
* Should be able to generate a new email address in that panel with less than 2 clicks
* Should be able to copy a generated email address

groovecoder:
We have a browserAction icon & panel now. We'll file follow-up issues to enhance it.

PR #78, opened 2020-03-16
for #77: start browserAction

groovecoder:


extension/js/popup.js
```diff
@@ -0,0 +1,10 @@
+function popup() {
+  document.querySelector("#generate-alias").addEventListener("click", async (e) => {
+    const newRelayAddressResponse = await browser.runtime.sendMessage({
```
groovecoder: @bakulf this part doesn't seem to wait for the background script to respond - it goes straight ahead to the `alert` call before it has the `newRelayAddressResponse`

bakulf: This doesn't seem correct in my tests. I receive "undefined" because, in `makeRelayAddress`, I don't have the apiToken. Then, this code runs:
```

  if (!apiToken.apiToken) {
    browser.tabs.create({
      url: RELAY_SITE_ORIGIN,
    });
    return;
  }
```

extension/js/background.js
```diff
@@ -41,10 +39,28 @@ async function makeRelayAddressForTargetElement(info, tab) {
...
+browser.runtime.onMessage.addListener(async (m) => {
+  let response;
+
+  switch (m.method) {
+    case "makeRelayAddress":
+      response = await makeRelayAddress();
+      break;
+  }
+
+  return response;
```
bakulf: Let's assueme apiToken is not null, and `makeRelayAddress` returns a `Response` object.
`Response` is not serializable and it cannot be sent from the background.js to the popup.js.
What you want is probably:

`return await response.text()`;

groovecoder: This was it; thanks!!

extension/js/popup.js
groovecoder: I still have `makeRelayAddress` return early if there's no apiToken, but I added code so that `popup.js` won't try to display the response.

Issue #81, opened 2020-03-18
Log FxA "change of profile data" events and message-forwarding events to detect potential errors


groovecoder:
When a user changes their FxA email address, we want to log it alongside the logging of relayed message activity, so we don't accidentally trigger a fraudulent alarm for a legitimate change of a relay recipient's email address.

1. Register an FxA Attached Service Notification handler for [change of profile data](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/docs/service_notifications.md#change-of-profile-data).
2. Log the `profileDataChange` event: `H(real address)`
3. When relaying a message, log an `email_relay` event: `relay_address.id, H(real address), H(relay address)`
4. In the fraud-detection pipeline, detect when:
   * An `email_relay` event contains a different `H(real address)` value than the previous `email_relay` for the same `H(relay address)` **AND**
   * There is no preceding `profileDataChange` event for the `H(real address)`
     * This signals that we are relaying emails to a new real address recipient, but there wasn't a change in the FXA primary email. Which could be fraudulent activity.

PR #82, opened 2020-03-19
for #81: log email_relay events

groovecoder:


emails/views.py
```diff
@@ -144,6 +149,15 @@ def _inbound_logic(json_body):
...
 
+    log_string = json.dumps({
+        'event_type': 'email_relay',
+        'relay_address_id': relay_address.id,
+        'relay_address': sha256(local_portion.encode('utf-8')).hexdigest(),
+        'real_address': sha256(
+            relay_address.user.email.encode('utf-8')
+        ).hexdigest(),
+    })
+    logger.info(log_string)
```
groovecoder: This seems like ti should work, but the `Fields` shows as an empty `{}` ?

```
{"Timestamp": 1584630352479602944, "Type": "events", "Logger": "fx-private-relay", "Hostname": "groovetop", "EnvVersion": "2.0", "Severity": 6, "Pid": 98734, "Fields": {}}
```

oremj: What happens if you pass in the dict directly instead of json dumping it first?

groovecoder:
```
    logger.info({
        'event_type': 'email_relay',
        'relay_address_id': relay_address.id,
        'relay_address': sha256(local_portion.encode('utf-8')).hexdigest(),
        'real_address': sha256(
            relay_address.user.email.encode('utf-8')
        ).hexdigest(),
    })
```

has the same result. :( `Fields` comes up empty:

```
{"Timestamp": 1584632026345210112, "Type": "events", "Logger": "fx-private-relay", "Hostname": "groovetop", "EnvVersion": "2.0", "Severity": 6, "Pid": 98998, "Fields": {}}
```

groovecoder: Note: I filed https://github.com/mozilla-services/python-dockerflow/issues/40 for this.

groovecoder: Unless/until https://github.com/mozilla-services/python-dockerflow/issues/40 is fixed, I updated this to include a `json: ` prefix so the log formatter won't skip it.

If the fraud pipeline log ingestion can parse this, I think we can merge this as-is, and if that up-stream issue is fixed, I'll update the code here later.

groovecoder: Output is now:
```
{"Timestamp": 1584716908799840000, "Type": "events", "Logger": "fx-private-relay", "Hostname": "groovetop", "EnvVersion": "2.0", "Severity": 6, "Pid": 7487, "Fields": {"msg": "json: {\"event_type\": \"email_relay\", \"relay_address_id\": 18, \"relay_address\": \"cf80cd8aed482d5d1527d7dc72fceff84e6326592848447d2dc0b0e87dfc9a90\", \"real_address\": \"f46fbc15047cae1e3b8f098d125372f713e99d974f57d7c0e7d8a1de691c1733\"}"}}
```

groovecoder: Okay, learned I can pass `extra={}` in the `logger.info` call to make the dictionary values show up in `Fields` ...

```
{"Timestamp": 1584720542840398080, "Type": "events", "Logger": "fx-private-relay", "Hostname": "groovetop", "EnvVersion": "2.0", "Severity": 6, "Pid": 9438, "Fields": {"relay_address_id": 18, "relay_address": "cf80cd8aed482d5d1527d7dc72fceff84e6326592848447d2dc0b0e87dfc9a90", "real_address": "f46fbc15047cae1e3b8f098d125372f713e99d974f57d7c0e7d8a1de691c1733", "msg": "email_relay"}}
```

PR #113, opened 2020-04-04
Avoid 'executeScript' when filling relay address, use content-script file

### LINKED ISSUE #112: Replace `executeScript` by actual content script
https://github.com/mozilla/fx-private-relay/commit/9c574ed19503fd5d87b08530c023dadbeacf6a66 added sanitization for the server response before injecting into the DOM, which is good.

Even better and less code (also less fragile code), would be to remove the `executeScript` call. Instead, define a content-script file in the manifest, that listens for a message from the background script to execute the same code. With that you can get rid of the sanitization because you are already making sure the content menu only works on `editable` elements, and thus the code like should be cleaner.

wagnerand:
Fix #112.

I haven't been able to test this code yet, please let me know what you think of it.

### DIFF
--- a/extension/js/background.js
+++ b/extension/js/background.js
@@ -1,42 +1,5 @@
 const RELAY_SITE_ORIGIN = "http://127.0.0.1:8000";
 
-/**
- * Escapes any occurances of &, ", <, > or / with XML entities.
- *
- * @param {string} str
- *        The string to escape.
- * @return {string} The escaped string.
- */
-function escapeXML(str) {
-  const replacements = { "&": "&amp;", "\"": "&quot;", "'": "&apos;", "<": "&lt;", ">": "&gt;", "/": "&#x2F;" };
-  return String(str).replace(/[&"'<>/]/g, m => replacements[m]);
-}
-
-/**
- * A tagged template function which escapes any XML metacharacters in
- * interpolated values.
- *
- * @param {Array<string>} strings
- *        An array of literal strings extracted from the templates.
- * @param {Array} values
- *        An array of interpolated values extracted from the template.
- * @returns {string}
- *        The result of the escaped values interpolated with the literal
- *        strings.
- */
-
-function escaped(strings, ...values) {
-  const result = [];
-
-  for (const [i, string] of strings.entries()) {
-    result.push(string);
-    if (i < values.length)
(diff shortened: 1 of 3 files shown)


extension/manifest.json
```diff
@@ -39,7 +39,7 @@
       "matches": [
           "<all_urls>"
       ],
-      "js": ["js/add_input_icon.js"]
+      "js": ["js/add_input_icon.js", "js/fill_relay_address.js"]
```
groovecoder, line 42: Is it better (performance-wise?) for add-ons to inject multiple content scripts like this, or combine functionality into a single content script like `add_input_icon_and_fill_addresses.js` ?

groovecoder:
Were you able to test this? It doesn't seem to work on my local.

extension/js/background.js
```diff
@@ -72,9 +35,10 @@ async function makeRelayAddressForTargetElement(info, tab) {
     return;
   }
 
-  browser.tabs.executeScript(tab.id, {
-    frameId: info.frameId,
-    code: escaped`let inputElement = browser.menus.getTargetElement(${info.targetElementId});inputElement.value = "${newRelayAddressResponse}";`,
+  browser.tabs.sendMessage(tab.id, {
```
groovecoder, line 38: `Error: Could not establish connection. Receiving end does not exist.`

wagnerand:
Is there a way for me to test this without having to set up my own server? Are there testing instructions?

extension/manifest.json
wagnerand: I am not sure. I created a new file because the existing content script has a functionality-specific filename, and I didn't want to put something unrelated into that. We could merge the two together and rename the file to something more generic.

@rpl, you know know which is more performant?

groovecoder:
If you run `npm run build:dev` in the `extension/` directory, it should build the add-on configured to use https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/ instead of 127.0.0.1

extension/js/background.js
wagnerand: Fixed in 155045b.

wagnerand:
Thanks! I fixed the issue and tested it, should work fine now.

PR #148, opened 2020-04-10
Add number of remaining free aliases.

### LINKED ISSUE #137: Dashboard: Should also have a “__/5 free burners still remaining” line of text somewhere on the dashboard for management purposes
Demo feedback

lesleyjanenorton:
Fixes #137

### DIFF
--- a/privaterelay/templates/profile.html
+++ b/privaterelay/templates/profile.html
@@ -2,7 +2,7 @@
 
 {% load socialaccount %}
 {% load static %}
-{% load custom_tags %}
+{% load relay_tags %}
 
 {% block content %}
 
--- a/privaterelay/templates/profile.html
+++ b/privaterelay/templates/profile.html
@@ -2,6 +2,7 @@
 
 {% load socialaccount %}
 {% load static %}
+{% load custom_tags %}
 
 {% block content %}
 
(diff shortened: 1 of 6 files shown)


privaterelay/templatetags/relay_tags.py
```diff
@@ -1,8 +1,9 @@
 from django import template
+from django.conf import settings
 
 register = template.Library()
 
 
 @register.simple_tag
 def remaining_free_aliases(aliases):
-  return 5-len(aliases)
+  return settings.MAX_NUM_BETA_ALIASES-len(aliases)
```
groovecoder, line 9: Can you make this same change in [`emails.views._index_POST`](https://github.com/mozilla/fx-private-relay/blob/master/emails/views.py#L57) ?

privaterelay/templates/profile.html
```diff
@@ -40,12 +40,14 @@ <h2 class="section-headline dashboard-headline">Manage Email Aliases</h2>
             <span class="remaining-aliases">
               You have <span class="num-remaining-aliases">{% remaining_free_aliases relay_addresses %}</span> remaining free aliases
             </span>
+            {% if not relay_addresses|length == 5 %}
```
groovecoder, line 43: Let's keep using the `settings.MAX_NUM_BETA_ALIASES` instead of hard-coding the limit anywhere. I already added a django context processor that makes all of the `settings` values available in all templates, so it's this easy:

```suggestion
            {% if not relay_addresses|length == settings.MAX_NUM_BETA_ALIASES %}
```

PR #149, opened 2020-04-12
add icon to email input button

groovecoder:
image


groovecoder:
This is still not the final way to do this, but it's a start.

PR #150, opened 2020-04-12
Update home copy; restore "how it works"

groovecoder:
Definitely not 100% committed to this, but it might be a start. It re-uses most [copy from the successful Monitor market test experiment](https://docs.google.com/document/d/1uLROV2egtJ0tuiej2s-SQIrcfiwOjTTop6Q-4uaV-bc/edit#).

I couldn't get the bottom "Add to Firefox" button to look right. 😢 @lesleyjanenorton - can you help here?

image

### DIFF
--- a/privaterelay/templates/home.html
+++ b/privaterelay/templates/home.html
@@ -19,14 +19,51 @@
 <main class="container txt-white flx text-center row-full-width home-hero">
 	<div class="flx flx-row hero-content">
 		<div class="hero-left flx flx-col">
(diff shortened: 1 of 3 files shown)


groovecoder:
I think maybe we can move the line "If the alias receives email you don't want, you can disable it or delete it completely." into the 3rd box and get rid of the "That's it." part?

lesleyjanenorton:
- Restored "Keep your email safe from hackers and trackers" headline
- Small suggested change to subhead, happy to revert.
- Fixed up the How It Works area and the button.

PR #176, opened 2020-04-17
fix #169: handle fxa delete events

### LINKED ISSUE #169: Handle FXA deletes
When a user deletes their FXA account, handle the corresponding FXA [Account Deletion Event](https://github.com/mozilla/fxa/blob/master/packages/fxa-auth-server/docs/service_notifications.md#account-deletion-event) and delete their Relay data too.

groovecoder:


emails/views.py
```diff
@@ -154,6 +157,8 @@ def _inbound_logic(json_body):
             relay_address.save(update_fields=['num_blocked'])
             return HttpResponse("Address does not exist")
     except RelayAddress.DoesNotExist as e:
+        # TODO?: if sha256 of the adress is in DeletedAddresses,
```
jrgm, line 160: address

PR #178, opened 2020-04-18
 fix #100: store domains in add-on and inject them into dashboard 

### LINKED ISSUE #100: Add ability to add notes/comments to aliases
I know this was discussed in Slack, but wanted to put it somewhere more official and documented.

It'd be nice if me (as a user) could optionally assign a note or comment for an alias so I know what service it's for. Currently I have a page of random looking SHAs @relay.firefox.com email addresses but no idea what service each one is for.
I believe the privacy concern was that then we'd know what services a person uses and we don't want to know that, but there was possibly a workaround of storing it in some fancy encoded data storage via FxA so we/Mozilla couldn't decrypt it.

My main concern would be I sign up to service X who promptly sells that information and shares it with all their affiliates and eventually I'm getting email from a bunch of services I never signed up for. It'd be nice to see somewhere on the /accounts/profile page that it was service X who sold me out so I can make sure to never do business with them in the future.
It might be moot, since I don't see anywhere on my incoming email I sent myself from Gmail what alias/address the relayed email was sent TO, I only see my original From email (my personal Gmail) and my To email (my work account). I don't see anywhere in Gmail or the plain text email details where it says what alias it was originally sent to.

I was trying to think of if the personal notes could also be stored locally and never sent to Mozilla, but then that might not sync very well across devices if it lives in local storage or a cookie or the add-on. 🤷‍♂

groovecoder:


extension/js/add_input_icon.js
```diff
@@ -63,6 +63,8 @@ function addRelayIconToInput(emailInput) {
     dismissButton.addEventListener("click", () => {
       errorMessageWrapper.remove();
       buttonElem.classList.remove("relay-icon-disabled");
+      navigator.clipboard.writeText(newRelayAddressResponse);
+      emailInput.value = newRelayAddressResponse;
```
groovecoder, line 67: wait, hmm ... this may have been a bad rebase?

groovecoder: fixed this.

PR #181, opened 2020-04-20
Update input icons

lesleyjanenorton:


extension/js/add_input_icon.js
```diff
@@ -1,33 +1,98 @@
...
+  const relayIconHeight = 30;
+  if (relayIconHeight > inputHeight) {
+    const smallIconSize = "20px";
+    imgEl.style.height = smallIconSize;
+    imgEl.style.width = smallIconSize;
+    imgEl.style.minWidth = smallIconSize;
+    imgEl.style.minHeight = smallIconSize;
+    inputClone.style.paddingRight = "30px";
+    divEl.style.right = "2px";
+  }
```
groovecoder, line 43: Similar concern here - all these `.style.* =` directives might break on a site that has CSP that disallows inline styles.

extension/js/add_input_icon.js
```diff
@@ -1,33 +1,98 @@
...
+
+  createErrorMessage = (buttonElem, content) => {
+    const errorMessageWrapper = document.createElement("div");
+    errorMessageWrapper.classList.add("relay-error-message-wrapper");
+    errorMessageWrapper.style.top = (relayIconHeight + 15)+"px";
```
groovecoder, line 48: inline style.

extension/js/add_input_icon.js
```diff
@@ -1,33 +1,98 @@
...
+      inputClone.value = newRelayAddressResponse;
+  });
+
+  divEl.appendChild(buttonEl);
+  emailInputWrapper.appendChild(divEl);
+  emailInputOriginalParentEl.insertBefore(emailInputWrapper, emailInput);
+  emailInput.remove();
+}
+
+(function () {
```
groovecoder, line 91: Should we call this when the document is done loading? I spot-checked https://www.involve.me/payment-page/?ref=producthunt and the icon didn't show up.

groovecoder: And/or a mutation observer? Spot-check on https://appfollow.io/customer-support-automation-tool didn't show the icon in the pop-up modal.

extension/manifest.json
```diff
@@ -23,7 +23,7 @@
   "browser_action": {
       "browser_style": true,
       "default_icon": {
-          "32": "icons/placeholder-logo-beta.png"
+          "32": "icons/make-new-alias.png"
```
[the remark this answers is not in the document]
lesleyjanenorton: I've been mulling this over and I'm not sure. Since the primary functionality of the addon is to generate new email aliases, I thought it made sense to use the same icon that we're using to represent this functionality in the inputs. We could also modify up the "Generate New Relay Alias" button in the dashboard to use the icon as well, so that anytime you're generating a new alias, you're seeing the same icon. Ideally the logo and this icon would be nearly identical, so maybe once we have a high res version of the icon the brand team proposed, we can use that in all of the places and swap out the "Beta" for a "+" sign when the icon is used in-content to generate new aliases.

extension/js/add_input_icon.js
```diff
@@ -1,33 +1,98 @@
...
+  inputClone.style.paddingRight = "50px !important;"
+
+
+  emailInputWrapper.appendChild(inputClone);
+
+  const inputHeight = emailInput.clientHeight;
+
+  const divEl = document.createElement("div");
+  divEl.classList.add("relay-icon");
+  divEl.style.height = inputHeight+"px";
```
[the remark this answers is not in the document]
lesleyjanenorton: It didn't break on Monitor but I've also been wondering about various CSP conflict possibilities. It's possible that because, at this point in the script, we haven't actually appended this element to anything, it's not in violation with the website's CSP.

extension/js/add_input_icon.js
lesleyjanenorton: Same comment as above. I don't _think_ Monitor allows unsafe-inline (although looking through our CSP policy there, I don't see unsafe-inline explicitly disallowed) yet I was able to get away with this.

extension/js/add_input_icon.js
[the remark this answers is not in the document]
groovecoder: oh that's true! it may be that the add-on script appends style attributes AFTER the CSP has enforced?

extension/js/add_input_icon.js
lesleyjanenorton: Yep will do!

extension/js/add_input_icon.js
```diff
@@ -88,11 +90,37 @@ function addRelayIconToInput(emailInput) {
...
+      if (mutation.target.tagName === "FORM") {
+        const emailInput = mutation.target.querySelector("input[type='email']");
+        if (emailInput) {
+          addRelayIconToInput(emailInput);
+        }
+      }
+    });
+  });
+
+  mutationObserver.observe(document.body, { childList: true, subtree: true });
```
groovecoder, line 124: We should do a performance check on this to see if we can optimize this. There are some tips in [this answer](https://stackoverflow.com/a/39332340/13373794), but I don't know if any of them apply to our use-case?

Doesn't need to block merge, in any case.

groovecoder:
Looking good. 3/5 top sites from producthunt.com work no problem ...

image

image

(neo.tax didn't work fwiw)

image

(graphcms.com didn't work)

PR #186, opened 2020-04-23
Update popup ui

lesleyjanenorton:


extension/js/popup.js
```diff
@@ -1,16 +1,63 @@
...
+  //   showRelayPanel();
+  // }
 
-    if (!newRelayAddressResponse) {
-      // background.js opened a tab to sign in
-      return;
-    }
+  document.querySelectorAll(".generate-alias").forEach(generateAliasBtn => {
+    generateAliasBtn.addEventListener("click", async() => {
+      let newRelayAddressResponse = await browser.runtime.sendMessage({
```
groovecoder, line 25:
```suggestion
      const newRelayAddressResponse = await browser.runtime.sendMessage({
```

extension/js/popup.js
```diff
@@ -1,16 +1,63 @@
...
+  return signUpOrInPanel.classList.remove("hidden");
+}
+
+function showRelayPanel() {
+  const relayPanel = document.querySelector(".signed-in-panel");
+  return relayPanel.classList.remove("hidden");
+}
+
+async function popup() {
+  const browserStorage = await browser.storage.local.get();
```
groovecoder, line 14:
```suggestion
```

extension/js/popup.js
```diff
@@ -1,16 +1,63 @@
...
+          aliasText.classList.remove("show-alias");
+          aliasWrapper.classList.add("hidden");
+        }
+        return;
+      }
+
+      // show newly created alias
+      aliasWrapper.classList.remove("hidden");
+      aliasText.textContent = newRelayAddressResponse;
+      aliasDisplay.classList.add("show-alias");
```
groovecoder, line 54: Suggest setting the `textContent` value first, so there won't be any flash of missing content?

```suggestion
      aliasText.textContent = newRelayAddressResponse;
      aliasWrapper.classList.remove("hidden");
      aliasDisplay.classList.add("show-alias");
```

extension/js/popup.js
```diff
@@ -1,16 +1,61 @@
...
+}
+
+function showRelayPanel() {
+  const relayPanel = document.querySelector(".signed-in-panel");
+  return relayPanel.classList.remove("hidden");
+}
+
+async function popup() {
+  const userApiToken = await browser.storage.local.get("apiToken");
+  if (!userApiToken) {
```
groovecoder, line 15: Based on my struggle with `browser.storage.local.get` in the other parts of the add-on, I don't think checking for `!userApiToken` is enough here. I believe `userApiToken` will be an empty object, which will always evaluate to true. You may need to do something more like `!userApiToken.hasOwnProperty('userApiToken')`?

groovecoder: Yeah, this `userApiToken` check always evaluations to true. Even in a brand new `web-ext run` profile with nothing in the extension storage, I get the relay panel instead of the sign-up panel.

image

PR #233, opened 2020-04-28
Add to fx dashboard prompts

lesleyjanenorton:


static/js/app.js
```diff
@@ -136,6 +158,11 @@ function addEventListeners() {
 			sendGaPing("Create New Relay Alias", "Click", createNewRelayBtn.dataset.analyticsLabel);
 		})
 	});
+
+	const continueWithoutAddonBtn = document.querySelector(".continue-without-addon");
+	if (continueWithoutAddonBtn) {
+		continueWithoutAddonBtn.addEventListener("click", hideInstallCallout());
```
groovecoder, line 164: `hideInstallCallout()` calls the function immediately. Need to drop the `()`.

```suggestion
		continueWithoutAddonBtn.addEventListener("click", hideInstallCallout);
```

groovecoder:
Let's store the hidden state of the add-on install CTA into session storage. Want to let people use relay without it, but also want them to install the add-on because it really is a better experience.

PR #254, opened 2020-05-02
fix #253: add check for browser.menus API calls

gabrielluong:
Fixes #253. We wrap all `browser.menus` calls with a check since this is not supported in Fenix/GeckoView.

PR #259, opened 2020-05-03
fix #258: make the popup page into a mobile responsive page

### LINKED ISSUE #258: Make the popup page into a mobile responsive page
In our Firefox Mobile Platforms, the web extension popup is shown as a normal page in a new tab. To provide a better experience, we will need to make changes to `popup.html` to make it into a responsive page.

Before it is mobile responsive (current):

Screen Shot 2020-05-03 at 9 59 58 AM

gabrielluong:
Fixes #258. To test this, it is either best to run an emulator with the extension as described in https://github.com/mozilla/fx-private-relay/pull/254#issuecomment-622964146 or to run `popup.html` in the browser and toggle on the responsive design mode. If you're doing the latter, you may need to toggle the `hidden` classes manually and fix the relative paths of the fonts and images in `popup.css`.

I took some liberties here to align it with the relay site, and am happy to iterate on the changes either in this PR or as a followup.

Notable changes include:
- Adding a media query that takes into the fact that the popup in the desktop is hardcoded to a height of --panelHeight (300px) and basically checked if the viewport is bigger than 301px, then made everything bigger and aligned for a mobile screen.
- Added the bottom border for the "sign up" link since it is not obvious in mobile (don't feel too strongly about this change)

Before it is mobile responsive (old):

Screen Shot 2020-05-03 at 9 59 58 AM

With the changes (new):

Screen Shot 2020-05-03 at 10 02 43 AM

Screen Shot 2020-05-03 at 10 05 14 AM

Screen Shot 2020-05-03 at 10 07 36 AM

Screen Shot 2020-05-03 at 10 08 20 AM

### DIFF
--- a/extension/css/popup.css
+++ b/extension/css/popup.css
@@ -16,12 +16,14 @@
 
 body {
   --panelWidth: 360px;
+  /* Any changes to --panelHeight should be reflected in the screen media query */
   --panelHeight: 300px;
   --ink: #20123a;
   --Metropolis: "Metropolis", "SF Pro Text", "Segoe UI", Roboto, "Open Sans";
   --fxGradient: linear-gradient(-90deg, #ff9100 0%, #f10366 50%, #6173ff 100%);
   font-family: "Inter UI", "SF Pro Text", 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
   font-size: 17px;
+  margin: 0;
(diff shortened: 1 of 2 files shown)


lesleyjanenorton:
@gabrielluong The underline was needed! Looks like this also fixes #267 .

gabrielluong:
> @gabrielluong The underline was needed! Looks like this also fixes #267 .

I will note that the underline only appears in the mobile responsive mode (under the media query) in case you meant the underline was needed every time.

PR #273, opened 2020-05-04
Fix text color on authentication error page

sirrah23:
If an uninvited user tries to sign in to Firefox Private Relay, all they see is a black envelope because the text on top is also black:
image

In this PR I changed the color of the text to white so that it appears as expected:
image

### DIFF
--- a/privaterelay/templates/socialaccount/authentication_error.html
+++ b/privaterelay/templates/socialaccount/authentication_error.html
@@ -4,7 +4,7 @@
 <main id="profile-main" class="flx env-bg container bg-light dashboard-container" data-api-token="{{ user.profile_set.first.api_token }}">
   <div class="appear-smoothly flx flx-col al-cntr jst-cntr auth-error-wrapper">
     <div class="flx flx-col al-cntr jst-cntr auth-error">
-        <p class="ff-Met no-active-aliases text-center">Sorry, the Private Relay Beta is currently by invitation only. Please check again later.</p>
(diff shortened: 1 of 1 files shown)


lesleyjanenorton:
Hi Sirrah, are you using a high contrast theme? I'm not able to reproduce the first image you shared but running your branch locally, this is what I get...

Without high-contrast: 
Screen Shot 2020-05-05 at 4 30 41 PM
With high-contrast turned on:
Screen Shot 2020-05-05 at 4 36 09 PM

lesleyjanenorton:
Ok looks like I couldn't reproduce this because I was checking it on my local server. I _can_ reproduce this, but only using Nightly, and on `relay.firefox.com`. Not sure what is up here, or what is causing this only on Nightly and only on Production. I will dig back into this tomorrow.

sirrah23:
> Ok looks like I couldn't reproduce this because I was checking it on my local server. I _can_ reproduce this, but only using Nightly, and on `relay.firefox.com`. Not sure what is up here, or what is causing this only on Nightly and only on Production. I will dig back into this tomorrow.

Thanks for taking a look at the issue here @lesleyjanenorton. I was digging around in my settings but I couldn't find why the envelope was black, but per your feedback it sounds like it's not just an issue on my end.

Out of curiosity I tried logging into this out on Chrome and the envelope there was white.

Whatever the case, I don't think that my solution here to simply change the font color will solve the problem.

I think it probably makes sense for me to close this PR and open an issue instead. I'll go ahead and do that today or tomorrow.

groovecoder:
Closing now that #280 is merged.

PR #275, opened 2020-05-05
Consolidate code for filling email inputs

lesleyjanenorton:


groovecoder:
Good call. Looks like a conflict with `fill_relay_address.js` in `master` ? Wanna clean that up so we can merge?

PR #276, opened 2020-05-05
Use modal for extension errors and consolidate shared CSS

lesleyjanenorton:
- Use a modal instead of `alert()` to tell users when they've already created the max number of email addresses (after attempting to generate a new alias from the context menu). Thoughts on this string? Should we change it? I dropped the "Please upgrade" bit since we don't have a path for upgrading yet.
- Start to consolidate the CSS between the modal, input icon, input error messages, and popup.

Screen Shot 2020-05-05 at 2 41 22 PM
Screen Shot 2020-05-05 at 12 19 38 PM


groovecoder:
Some more conflicts in this one. Sorry I think I merged some other extension changes in front of these.

lesleyjanenorton:
@groovecoder no worries!

groovecoder:
Has (new?) conflict with `fill_relay_address.js`.

Issue #272, opened 2020-05-04
Cannot process FXA profile updates after access token has expired


groovecoder:
https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/8296155/?query=is:unresolved

```
KeyError: 'email'
  File "django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "privaterelay/views.py", line 105, in fxa_rp_events
    authentic_jwt, social_account, event_key
  File "privaterelay/views.py", line 148, in _handle_fxa_profile_change
    new_email = extra_data['email']
```

`extra_data` is:
```
{code: 401, errno: 100, error: 'Bad Request', message: [Filtered], reason: 'Invalid token'}
```

PR #277, opened 2020-05-05
wip: for #272: start fxa_refreshing provider to get FXA refresh tokens

groovecoder:
#272 happens when:

1. FXA user signs up for Relay
2. Relay's access token for the user expires
3. FXA user makes a profile change
4. FXA sends a profile change event to Relay
5. Relay tries to use the expired token to refresh the user's profile
6. FXA responds with a `401` `Invalid token` message

This PR creates a new [customized `fxa` `allauth` social account provider](https://django-allauth.readthedocs.io/en/latest/advanced.html#customizing-providers), which includes the `access_type=offline` parameter in the token request. This makes FXA return a `refresh_token` value, which is stored in the `SocialToken` model as `secret`.

To finish the fix, we need to update the `fxa_rp_events` view: If it receives a `401` `Invalid token` response from FXA, it should use the refresh token to get a new access token, and retry the profile fetch.

### DIFF
--- a/fxa_refreshing/provider.py
+++ b/fxa_refreshing/provider.py
@@ -1,0 +1,61 @@
+import requests
+
+from allauth.compat import parse_qsl
+from allauth.socialaccount.providers.oauth2.client import (
+    OAuth2Client, OAuth2Error
+)
+from allauth.socialaccount.providers.fxa.provider import FirefoxAccountsProvider
+
+
+class RefreshingFXAClient(OAuth2Client):
+    # Custom client to include access_type=offline and store refresh token
+    def get_access_token(self, code):
+        data = {
+            'access_type': 'offline',
+            'redirect_uri': self.callback_url,
+            'grant_type': 'authorization_code',
+            'code': code}
+        if self.basic_auth:
+            auth = requests.auth.HTTPBasicAuth(
+                self.consumer_key,
+                self.consumer_secret)
+        else:
+            auth = None
+            data.update({
+                'client_id': self.consumer_key,
+                'client_secret': self.consumer_secret
+            })
+        params = None
+        self._strip_empty_keys(data)
+        url = self.access_token_url
+        if self.access_token_method == 'GET':
+            params = data
+            data = None
+        # TODO: Proper exception handling
+        resp = requests.request(
+            self.access_token_method,
+            url,
+            params=params,
+            data=data,
+            headers=self.headers,
+            auth=auth)
+
+        access_token = None
+        if resp.status_code in [200, 201]:
+            # Weibo sends json via 'text/plain;charset=UTF-8'
+            if (resp.headers['content-type'].split(
+                    ';')[0] == 'application/json' or resp.text[:2] == '{"'):
+                access_token = resp.json()
+            else:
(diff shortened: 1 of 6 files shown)


fxa_refreshing/provider.py
```diff
@@ -0,0 +1,61 @@
...
+            self.access_token_method,
+            url,
+            params=params,
+            data=data,
+            headers=self.headers,
+            auth=auth)
+
+        access_token = None
+        if resp.status_code in [200, 201]:
+            # Weibo sends json via 'text/plain;charset=UTF-8'
```
jezdez, line 45: Is that something you expect here to happen?

fxa_refreshing/provider.py
```diff
@@ -0,0 +1,61 @@
...
+
+from allauth.compat import parse_qsl
+from allauth.socialaccount.providers.oauth2.client import (
+    OAuth2Client, OAuth2Error
+)
+from allauth.socialaccount.providers.fxa.provider import FirefoxAccountsProvider
+
+
+class RefreshingFXAClient(OAuth2Client):
+    # Custom client to include access_type=offline and store refresh token
```
jezdez, line 11: I thought you can do request a refresh token like this:
```
SOCIALACCOUNT_PROVIDERS = {
    'google': {
        'SCOPE': [
            'profile',
            'email',
        ],
        'AUTH_PARAMS': {
            'access_type': 'online',
        }
    }
}
```
Like described in https://django-allauth.readthedocs.io/en/latest/providers.html#django-configuration

jezdez: Can you clarify whta's different to the default OAuth2Client?

fxa_refreshing/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
jezdez, line 3: Not sure if this makes sense to keep 😬

privaterelay/views.py
```diff
@@ -138,12 +135,28 @@ def _get_event_keys_from_jwt(authentic_jwt):
...
     headers = {'Authorization': 'Bearer: {0}'.format(token.token)}
     resp = requests.get(
         FirefoxAccountsOAuth2Adapter.profile_url, headers=headers
     )
+    if resp.status_code == 401:
+        resp_json = resp.json()
+        if resp_json['reason'] == 'Invalid token':
+            # TODO: use the social_account refresh token to get a new access
+            # token and retry
+            pass
```
jezdez, line 158: For this you may want to do it like this: https://github.com/pennersr/django-allauth/issues/420#issuecomment-301805706

fxa_refreshing/provider.py
jezdez: Ah nevermind, this seems to be from the core Oauth2 class.

fxa_refreshing/provider.py
groovecoder: holy crap yes this is what I needed the whole time! Closing this PR to open a dramatically more simple one.

groovecoder:
Superseded by https://github.com/mozilla/fx-private-relay/pull/279

PR #279, opened 2020-05-06
for #272: add access_type=offline to AUTH_PARAMS

groovecoder:
Wow. This 1-line change replaces all of https://github.com/mozilla/fx-private-relay/pull/277

### DIFF
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -1,10 +1,11 @@
+from datetime import timezone, timedelta
 from hashlib import sha256
 import json
 import logging
 import os
 
 from jwt import JWT, jwk_from_dict
-import requests
+from requests_oauthlib import OAuth2Session
 import sentry_sdk
 
 from django.apps import apps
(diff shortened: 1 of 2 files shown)


groovecoder:
Still TODO: use the `social_token.token_secret` to get a new access token and retry the profile fetch with it. ([an example](https://github.com/pennersr/django-allauth/issues/420))

skim1102:
While I was reviewing your work I realized with the` invitations_only`, if the user changes their primary email to their secondary email they will not be able to access the Private Relay since the email address is not inside the `invitations` table. Do you think we can extend 272 to update the email address in the invitations table? Otherwise, I'll file a separate issue for this and continue reviewing this.

privaterelay/views.py
```diff
@@ -139,11 +140,9 @@ def _get_event_keys_from_jwt(authentic_jwt):
 
 
 def _handle_fxa_profile_change(authentic_jwt, social_account, event_key):
-    token = social_account.socialtoken_set.first()
-    headers = {'Authorization': 'Bearer: {0}'.format(token.token)}
-    resp = requests.get(
-        FirefoxAccountsOAuth2Adapter.profile_url, headers=headers
-    )
+    client = _get_oauth2_session(social_account)
+    resp = client.get(FirefoxAccountsOAuth2Adapter.profile_url)
```
skim1102, line 144: Do we not need `headers=headers` on the GET? Or is it handled by the new client being created using `_get_oauth2_session` magically 🧙?

groovecoder: The new client should be handling that part now.

groovecoder:
I deployed this branch manually to the dev server and it works well. So I'm going to go ahead and merge.

PR #281, opened 2020-05-06
ALPHA_INVITE_TOKEN env var to enable invite links

groovecoder:
To test:

1. Leave `ALPHA_INVITE_TOKEN` empty and try to sign up with a non-Mozilla email address.
   * It should NOT work.
2. Set `ALPHA_INVITE_TOKEN` to some string e.g., `unsafe-invite-token-for-local`
3. Restart `runserver`
4. Try to sign up with a non-matching url: http://127.0.0.1:8000/invitation?alpha_token=not-right
   * It should NOT work.
5. Try to sign up with the right url: http://127.0.0.1:8000/invitation?alpha_token=unsafe-invite-token-for-local
   * It should work.
6. Set `MAX_ACTIVE_ACCOUNTS` to a lower number e.g., `1`
7. Restart `runserver`
8. Try to sign up with the right url: http://127.0.0.1:8000/invitation?alpha_token=unsafe-invite-token-for-local
   * It should NOT work.

### DIFF
--- a/privaterelay/signals.py
+++ b/privaterelay/signals.py
@@ -1,6 +1,8 @@
 from datetime import datetime
 
+from django.conf import settings
 from django.core.exceptions import PermissionDenied
+from django.contrib.auth.models import User
 
 from .models import Invitations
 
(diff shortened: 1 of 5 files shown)


privaterelay/signals.py
```diff
@@ -17,10 +19,23 @@ def invitations_only(sender, **kwargs):
     for allowed_domain in ALLOWED_SIGNUP_DOMAINS:
         if domain_part == allowed_domain:
             return True
+
+    active_accounts_count = User.objects.count()
+    if active_accounts_count >= settings.MAX_ACTIVE_ACCOUNTS:
+        raise PermissionDenied
+
+    if kwargs['request'].session['alpha_token'] == settings.ALPHA_INVITE_TOKEN:
```
skim1102, line 27: Being nit-picky, but I am sure there will be people who were invited but did not visit `/invitation` before they tried to login and use the service resulting to a error page with stack trace for `alpha_token` key not found in the session object. My recommendation is have an additional check for the `alpha_token` value in the session to throw a helpful message in the permission denied view like:
```python
if not kwargs['request'].session.has_key('alpha_token'):
    raise PermissionDenied('You must visit the invitation link provided in your invite email before attempting to participate in testing of Private Relay.')
```
(Side note, tried this out and `PermissionDenied` exception does not display the message because I think something is catching the exception and displaying the `authentication_error.html`. If we can pass a context value for the customized message to redirect the user to the correct flow that would be great.)

privaterelay/signals.py
```diff
@@ -17,10 +19,23 @@ def invitations_only(sender, **kwargs):
     for allowed_domain in ALLOWED_SIGNUP_DOMAINS:
         if domain_part == allowed_domain:
             return True
+
+    active_accounts_count = User.objects.count()
+    if active_accounts_count >= settings.MAX_ACTIVE_ACCOUNTS:
+        raise PermissionDenied
```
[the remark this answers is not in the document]
groovecoder: Thanks for the diligence.

The combination of allowed domain and invited accounts is intentional - we want to cap the overall accounts to 1500, which should cover Mozilla staff and our first 300-400 invites together.

Dangit, I wish I could un-see that plural model name. But I'd rather not change it if we don't have to. We should be able to drop the model completely at some point.

groovecoder:
>     * `ALPHA_INVITE_TOKEN` being used in the query param can easily be shared, are we ok with that?

Yup. We want an invite token and a capability url so we're not loading invitees email addresses into a new database that they didn't explicitly opt into. The max account cap is there to mitigate the risk of link sharing sending us too many users.

>     * Currently any alpha invite token users with expired session who have already visited `/invitation` MUST visit the `/invitation` again before being able to signin. Is this intended? Otherwise, we should change the `invitations_only` signal to check the active invitation before checking for `alpha_token` in session.

Great catch! Updated.

privaterelay/signals.py
groovecoder: I updated and cleaned up the logic in the signal handler.

I tried a bit to see if we could return or raise a better error message for the page, but that turned into a bit of a rabbit-hole.

If we want to improve it for the invite-only beta, I think we'll use [messages](https://docs.djangoproject.com/en/2.2/ref/contrib/messages/) instead. But for now, alpha users may just get not-the-best messages.

groovecoder:
Read for another review.

PR #284, opened 2020-05-07
Changed error text color

Eliotnand:
Fixed text color


groovecoder:

PR #285, opened 2020-05-07
Alert for 5 addresses

Eliotnand:
Created simple alert if user hit 5 addresses.


groovecoder:

PR #286, opened 2020-05-07
add debug toolbar for local debugging

groovecoder:


privaterelay/settings.py
```diff
@@ -31,6 +31,10 @@
 SECRET_KEY = config('SECRET_KEY', None, cast=str)
 
 DEBUG = config('DEBUG', False, cast=bool)
+if DEBUG:
+    INTERNAL_IPS = [
+        '127.0.0.1',
```
skim1102, line 36: Doesn't this IP value change based on the environment? If so we should add something like`DEBUG_INTERAL_IP` var to the config instead of adding a static value here. Would be good to update the `.env-dist` as well.

groovecoder: Updated.

PR #288, opened 2020-05-08
Added detecting non-firefox browsers, after redirect to download page.

Eliotnand:
Identify non-Firefox browsers and direct them to Firefox download page.


groovecoder:

PR #294, opened 2020-05-08
fix #211: keep metrics for deleted addresses

### LINKED ISSUE #211: Consider adding num_forwarded, num_blocked, and num_spam to deleted_address table
We've started some analytics queries based on the `num_forwarded`, `num_blocked`, `num_spam` values in `emails_relayaddress` column.

But, we lose those values when the records are deleted and moved to `emails_deletedaddress`.

So, we should add `num_forwarded`, `num_blocked`, `num_spam` columns to `emails_deletedaddress` to keep those aggregate metrics.

groovecoder:


emails/models.py
```diff
@@ -50,7 +53,9 @@ def make_relay_address(user, num_tries=0):
         if num_tries >= 5:
             raise CannotMakeAddressException
         relay_address = RelayAddress.objects.create(user=user)
-        address_hash = sha256(relay_address.address.encode('utf-8')).hexdigest()
+        address_hash = sha256(
+            relay_address.address.encode('utf-8')
+        ).hexdigest()
```
groovecoder, line 58: Just a `flake8` fix.

groovecoder:
Some day I'll get better at naming migrations ... but it is not this day. 😐

PR #300, opened 2020-05-11
fix datetime/timezone code

groovecoder:
Thought this was already fixed/included in Friday's merge of https://github.com/mozilla/fx-private-relay/pull/279 but apparently not? So this is a follow-up to fix the *NEXT* category of errors we're seeing in the `fxa-rp-events` callback endpoint.

### DIFF
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -1,4 +1,4 @@
-from datetime import timezone, timedelta
+from datetime import datetime, timedelta, timezone
 from hashlib import sha256
 import json
 import logging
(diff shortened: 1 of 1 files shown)


groovecoder:
Requesting review, but I will likely merge this to fix dev+stage+prod errors now, and get a post-deploy review whenever.

privaterelay/views.py
```diff
@@ -224,7 +225,9 @@ def _token_updater(new_token):
         'client_secret': client_secret,
     }
 
-    expires_in = (social_token.expires_at - timezone.now()).total_seconds()
+    expires_in = (
+        social_token.expires_at - datetime.now(timezone.utc)
```
skim1102, line 229: Just wanted to make sure that the `social_token.expires_at` is also set to UTC timezone?

Issue #156, opened 2020-04-14
Add "invitations" list


groovecoder:
To invite users on the wait-list, and enable a rolling invitation flow, we need to:

1. Add an `invitations` table:
   * `id`, `email`, `date_added`, `active`, `date_sent`, `date_redeemed`
2. Copy the email addresses from Monitor subscribers' ["waitlists joined"](https://github.com/mozilla/blurts-server/blob/master/db/DB.js#L338-L346) field
   * Set `date_added` to `now()`
3. Write a script to:
   * Select N invitations ordered by `date_added` where `active` is `False`
   * Set `active` to `True`
   * Send invitation email to `email`
   * Set `date_sent` to `now()`
4. Update `mozillians_only` signal handler to `invitations_only`:
   * Check that `sociallogin.account.extra_data['email']` is in `invitations` table with `active=True`
   * Set `date_redeemed` to `now()`
   * Catch `PermissionDenied` and/or redirect other visitors to join the `invitations` with `active` set to `False` and `date_added` set to `now()`

skim1102:
What does the `active` do other then checking to see that the email has been sent? If it only checks that the email has been sent maybe we should just use the `date_sent` column `IS NOT NULL`. 

I am worried about the 4th instruction to update the `mozillians_only` signal handler to `invitations_only`. This could exclude the users with Mozilla email address (which we would always want to allow, right?) and only allow emails in the `invitations` table. Perhaps we want to create a signal like`beta_testers_only` that checks to see if we are testing for invitations as well? Pseudocode looks like:
```
def beta_testers_only(sender, **kwargs):
    sociallogin = kwargs['sociallogin']
    email = sociallogin.account.extra_data['email']
    if waffle.switch_is_active('invitations_allowed')):
        # return True if email in invitations table AND date_sent not null
    if email.split('@')[-1] in ALLOWED_SIGNUP_DOMAINS:
        return True
    raise PermissionDenied
```
What do you think?

P.S. not sure what you normally do to allow additional beta testers from the invitations but may I recommend using `django-waffle`'s [Swtich](https://waffle.readthedocs.io/en/stable/types/switch.html#switches)? That is, if we are using the django-admin in prod.

groovecoder:
Hmm ... `date_sent IS NOT NULL` could effectively be the same as `active=True` - just seems a bit like overloading the field with hidden meaning? No biggie though.

[...]

PR #209, opened 2020-04-24
wip: start wrapped_email.html

groovecoder:

PR #303, opened 2020-05-12
for #156: invite_monitor_waitlist sends email and updates monitor rec…

groovecoder:
…ords


privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,88 @@
...
+                response = socketlabs_send(sl_client, sl_message)
+
+                if not response.result.name == 'Success':
+                    logger.error('socketlabs_error', extra=response.to_json())
+
+                invitation.date_sent = datetime.now()
+                invitation.save(update_fields=['date_sent'])
+
+                invitee.waitlists_joined['email_relay']['notified'] = True
+                invitee.save(update_fields=['waitlists_joined'])
```
skim1102, line 88:
```suggestion
                invitation.date_sent = datetime.now()
                invitee.waitlists_joined['email_relay']['notified'] = True
                invitee.save(update_fields=['date_sent', 'waitlists_joined'])
```

privaterelay/templates/emails/beta_invite_html_email.html
```diff
@@ -0,0 +1 @@
+<a href="http://127.0.0.1:8000/">Join the beta</a>
```
skim1102, line 1:
```suggestion
<a href={{ current_domain }}>Join the beta</a>
```

privaterelay/templates/emails/beta_invite_text_email.txt
```diff
@@ -0,0 +1,3 @@
+Join the beta at:
+
+http://127.0.0.1:8000/
```
skim1102, line 3:
```suggestion
{{ current_domain }}
```

privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,88 @@
 from datetime import datetime
+from email.utils import parseaddr
+import logging
 
+from socketlabs.injectionapi.message.basicmessage import BasicMessage
+from socketlabs.injectionapi.message.emailaddress import EmailAddress
+
+from django.conf import settings
```
skim1102, line 8:
```suggestion
from django.conf import settings
from django.contrib.sites.models import Site
```

privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,88 @@
...
                 print("adding %s to invitations" % invitee.primary_email)
-                Invitations.objects.create(
+                invitation = Invitations.objects.create(
                     email=invitee.primary_email,
                     active=True,
                     date_added=datetime.now(),
                     date_redeemed=None
                 )
+
+                # Send email invite
```
skim1102, line 53:
```suggestion
                # Send email invite
                current_site = Site.objects.get_current()
```

privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,88 @@
...
+                sl_message = BasicMessage()
+                sl_message.subject = (
+                    "You're Invited to the Private Relay Beta!"
+                )
+
```
skim1102, line 58:
```suggestion

                context = {
                    'email': invitee.primary_email,
                    'current_domain': current_site.domain
                }
```

privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,88 @@
...
+                sl_message.html_body = render_to_string(
+                    'emails/beta_invite_html_email.html',
+                    {'email': invitee.primary_email}
+                )
+                sl_message.plain_text_body = render_to_string(
+                    'emails/beta_invite_text_email.txt',
+                    {'email': invitee.primary_email}
+                )
```
skim1102, line 66:
```suggestion
                sl_message.html_body = render_to_string(
                    'emails/beta_invite_html_email.html',
                    context
                )
                sl_message.plain_text_body = render_to_string(
                    'emails/beta_invite_text_email.txt',
                    context
                )
```

privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,88 @@
...
+
+                relay_display_name, relay_from_address = parseaddr(
+                    settings.RELAY_FROM_ADDRESS
+                )
```
skim1102, line 70: We can use the `current_site.domain` and create the config value `RELAY_FROM_ADDRESS` in `settings.py` or use this as is but update the .env for each environment. I think either method works and I don't have a strong preference.

Side note: if you decide to configure in `settings.py` instead of `.env`, may I suggest creating a config value to store the `current_domain` to be used in the `context` variable I suggested?

privaterelay/management/commands/invite_monitor_waitlist.py
groovecoder: This loses the `save()` call on the `invitation` object?

privaterelay/management/commands/invite_monitor_waitlist.py
groovecoder: We've disabled the admin site in stage & prod environments, so using the Django `Site` objects may be harder than using an environment variable.

groovecoder: I'll use an env var that is also put into `settings`.

privaterelay/management/commands/invite_monitor_waitlist.py
skim1102: oh my 👀... You are right. Sorry about that.

groovecoder:
Updated to pull site link from env var. Didn't change anything about the `From:` address yet ... I think @lesleyjanenorton might be working on that with UX on the regular relay emails too.

privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,88 @@
...
             try:
                 Invitations.objects.get(
                     email=invitee.primary_email,
                     active=True,
                 )
-                print("%s already has an active invitation" % invitee.primary_email)
+                print(
+                    "%s already has an active invitation" %
+                    invitee.primary_email
+                )
```
say-yawn, line 43: @groovecoder is there any plan to aggregate the "already active invitation" data and update the `MonitorSubscriber` table so that `waitlists_joined__email_relay__notified` is `True` if not I recommend updating the `invitee` here:
```suggestion
                )
                invitee.waitlists_joined['email_relay']['notified'] = True
                invitee.save(update_fields=['waitlists_joined'])
```

privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,93 @@
...
+                    context,
+                )
+                sl_message.plain_text_body = render_to_string(
+                    'emails/beta_invite_text_email.txt',
+                    context,
+                )
+
+                relay_display_name, relay_from_address = parseaddr(
+                    settings.RELAY_FROM_ADDRESS
+                )
```
say-yawn, line 75: Consider using the function `_generate_relay_From()` from the `emails/views.py` as seen [here](https://github.com/mozilla/fx-private-relay/pull/303/files#diff-2c3a9c773eb79ee1e4b1858c23620e74R201-R207). We may want to move the method to `emails/utils.py`, what do you think?

privaterelay/management/commands/invite_monitor_waitlist.py
groovecoder: Hmm ... good question. I'm trying to think if there's a situation where someone would have an active `Invitation` in Relay, but wouldn't be notified. I don't think there is, so yeah - this could set `notified` to `True` if it finds one.

privaterelay/management/commands/invite_monitor_waitlist.py
groovecoder: `_generate_relay_From()` affixes `via Firefox Private Relay` to a relay email's `From:`. For the invitations, we just want our own `From:`.

groovecoder:
Which changes for `RELAY_FROM_ADDRESS` ?

groovecoder:
Note: we may be able to merge this so it doesn't bit-rot, but we will probably want/need to change this from SocketLabs to SES after https://github.com/mozilla/fx-private-relay/pull/312 lands.

say-yawn:
@groovecoder, `RELAY_FROM_ADDRESS` changes you mentioned that Lesley will be working on [here](https://github.com/mozilla/fx-private-relay/pull/303#issuecomment-628246744) and that was the reason why you didn't change the `From` address's domain to be configured in settings as you mentioned [here](https://github.com/mozilla/fx-private-relay/pull/303#discussion_r424694657). Perhaps I misunderstood something?

Also while reviewing changes for #312 I saw a conflicting change to the `RELAY_FROM_ADDRESS` [here](https://github.com/mozilla/fx-private-relay/pull/312/files#diff-ff34a0bc194457969fa97762e6b0d3ebR21) where the `relay_display_name` is no longer included. Should we just create a separate env variable for the display name?

privaterelay/settings.py
```diff
@@ -80,6 +80,8 @@
 )
 REFERRER_POLICY = 'strict-origin-when-cross-origin'
 
+SITE_URL = config('SITE_URL', None)
```
say-yawn, line 83: Seems like there is an env var for this now, from PR #209. See [here](https://github.com/mozilla/fx-private-relay/blob/master/privaterelay/settings.py#L103), there's a `SITE_ORIGIN` env var which got added to settings config.

groovecoder: Oh yeah, good catch.

privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,95 @@
...
+                sl_message.to_email_address.append(
+                    EmailAddress(invitee.primary_email)
+                )
+
+                sl_client = get_socketlabs_client()
+                response = socketlabs_send(sl_client, sl_message)
+
+                if not response.result.name == 'Success':
+                    logger.error('socketlabs_error', extra=response.to_json())
```
say-yawn, line 89:
```suggestion
                if not response.result.name == 'Success':
                    logger.error('socketlabs_error', extra=response.to_json())
                                    invitee.waitlists_joined['email_relay']['notified'] = False
                invitee.save(update_fields=['waitlists_joined'])
```
Remember the [past](https://github.com/mozilla/fx-private-relay/pull/303/files#diff-b16e3646d7b8f3831528d24b7d04327aR44) you and I who discussed when will we ever have a case where the invitee's email was not notified? We didn't think there was any case, but there is:
If the response for sending an email is not `'Success'` that means they were NOT notified while they were added to the invitation table (can join testing even without the email).

privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,95 @@
...
+
+                invitation.date_sent = datetime.now()
+                invitation.save(update_fields=['date_sent'])
+
+                invitee.waitlists_joined['email_relay']['notified'] = True
+                invitee.save(update_fields=['waitlists_joined'])
```
say-yawn, line 95:
```suggestion
                else:
                    invitation.date_sent = datetime.now()
                    invitation.save(update_fields=['date_sent'])

                    invitee.waitlists_joined['email_relay']['notified'] = True
                    invitee.save(update_fields=['waitlists_joined'])
```
values should be iff email was successfully delivered

privaterelay/management/commands/invite_monitor_waitlist.py
```diff
@@ -1,30 +1,95 @@
...
                     email=invitee.primary_email,
                     active=True,
                 )
-                print("%s already has an active invitation" % invitee.primary_email)
+                print(
+                    "%s already has an active invitation" %
+                    invitee.primary_email
+                )
+                invitee.waitlists_joined['email_relay']['notified'] = True
+                invitee.save(update_fields=['waitlists_joined'])
```
say-yawn, line 45:
```suggestion
```
When the email has failed to successfully be sent to the recipient as noted here, the `notified` field may not be `True` and as a result we do not want to change the said value.

say-yawn:
I created a new branch to add additional logic to handle user with invitation created but email not delivered. Diff between this branch and the new branch is [here](https://github.com/mozilla/fx-private-relay/compare/invite-monitor-waitlist-156...invite-monitor-waitlist-additional-logic-156). Following practical tests were done:
- [ ] New subscriber fail to send email about invitations being active but failed to send invitation email (SocketLabs configurations not set)
1. Run `./manage.py invite_monitor_waitlist 1`
2. Check that you get “adding secondary-email@mailinator.com to invitations” message printed on the console
3. Check that the invitation was created for the given email (try to login to private relay, check DB, etc) but `date_redeemed` field is null
- [ ] User with active invitation but email not sent but failed to send invitation email
1. Run `./manage.py invite_monitor_waitlist 1`
2. Check that you get “secondary-email@mailinator.com has an invitation but was not emailed” message printed on the console
3. Check that the invitation was created for the given email (try to login to private relay, check DB, etc) 
4. Check that if the email was NOT successfully sent the `date_redeemed` remains null

Still need to do this test:
- [ ] User with active invitation but email not sent successfully send invitation email
1. Setup [SocketLabs configuration](https://github.com/mozilla/fx-private-relay#enable-socketlabs-inbound-api)
2. Run `./manage.py invite_monitor_waitlist 1`
3. Check that `redeemed_date` is set for the invtiation

privaterelay/management/commands/invite_monitor_waitlist.py
groovecoder: Good catch!

groovecoder:
Updated your branch with additional logic. Let's close this PR and open one from that branch.

Issue #309, opened 2020-05-14
Add code to receive email via Amazon SES


groovecoder:
https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email.html

PR #312, opened 2020-05-15
for #309: add code to use SES in addition to SocketLabs

groovecoder:
## To spot-check locally
I've been doing this to check the functionality:

1. Save [this gist](https://gist.github.com/groovecoder/bb884b2db993642f9086d0e53bd28fb5) as `ses_NotificationEmail.json` in the base directory of the django project.
2.
```python
In [2]: from emails import views

In [3]: import json

In [4]: json_body = json.load(open('ses_NotificationEmail.json', 'r'))

In [5]: class MockRequest(object):
   ...:     headers = {
   ...:         'X-Amz-Sns-Message-Type': 'Notification',
   ...:         'X-Amz-Sns-Topic-Arn': 'arn:aws:sns:us-east-1:168781634622:ses-inbound'
   ...:     }
   ...:     body = json.dumps(json_body)
   ...:

In [6]: mr = MockRequest()

In [8]: views.sns_inbound(mr)
```

Note: this will return a 404 because the `To:` address in the example `ses_NotificationEmail.json` won't exist on your local host. But the 404 indicates that it got past the SNS verification.

## To spot-check in a cloud!!
To do this on my own app server, I went thru the "[Getting Started Receiving Email](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-getting-started.html)" guide in the AWS docs for my own privaterelay.groovecoder.com domain. 

Overview get things flowing:

1. Verified my privaterelay.groovecoder.com domain with a TXT record
2. Set up SES with a Receipt Rule to publish every email received to an SNS topic
3. Set up an HTTPS endpoint (https://privaterelay.groovecoder.com/emails/sns-inbound) subscription to that SNS topic
4. Confirm the SNS subscription by visiting the confirmation URL logged by the endpoint
5. Set up an SES ConfigSet


emails/sns.py
```diff
@@ -0,0 +1,95 @@
+import base64
+import logging
+import pem
+import six
```
groovecoder, line 4: we may not need `six` because we don't intend to run on anything except python3?

emails/sns.py
```diff
@@ -0,0 +1,95 @@
...
+from urllib.request import urlopen
```
groovecoder, line 5: Should try to refactor to use `requests` library like we do elsewhere in the app code?

emails/sns.py
```diff
@@ -0,0 +1,95 @@
...
+
+from OpenSSL import crypto
```
groovecoder, line 7: do we really want to use `pyOpenSSL` directly? is there a Django wrapper on top? or another nicer library?

emails/views.py
```diff
@@ -118,6 +123,159 @@ def _index_DELETE(request_data, user_profile):
...
+            'SNS message for unsupported type',
+            extra={
+                'supported_sns_types': SUPPORTED_SNS_TYPES,
+                'message_type': message_type,
+            }
+        )
+        return HttpResponse(
+            'Received SNS message for unsupported Type: %s' % message_type,
+            status=400
+        )
```
say-yawn, line 164: I recommend creating a function `validate_sns_header(topic_arn, message_type)` which does this check.

emails/views.py
```diff
@@ -118,6 +123,159 @@ def _index_DELETE(request_data, user_profile):
...
+            extra={'SubscribeURL': json_body['SubscribeURL']}
+        )
+        return HttpResponse('Logged SubscribeURL', status=200)
+    if message_type == 'Notification':
+        logger.info(
+            'SNS Notification',
+            extra={'json_body': json_body},
+        )
+        return _sns_notification(json_body)
+
```
say-yawn, line 198:
```suggestion
        return _sns_notification(json_body)
    logger.error(
        'SNS message type did not fall under the SNS inbound logic',
        extra={'message_type': message_type}
    )
    return HttpResponse('Received SNS message with type not handled in inbound log', status=400)
```
We should have a default behavior.

emails/views.py
```diff
@@ -118,6 +123,159 @@ def _index_DELETE(request_data, user_profile):
...
+    logger.info('email_relay', extra={
+        'fxa_uid': (
+            relay_address.user.socialaccount_set.first().uid
+        ),
+        'relay_address_id': relay_address.id,
+        'relay_address': sha256(local_portion.encode('utf-8')).hexdigest(),
+        'real_address': sha256(
+            relay_address.user.email.encode('utf-8')
+        ).hexdigest(),
```
say-yawn, line 242: I understand that it's encrypted but is there a need to log the `real_address` then?

emails/sns.py
say-yawn: This would be nice and seem do-able. According to the [third comment](https://stackoverflow.com/questions/36516183/what-should-i-use-to-open-a-url-instead-of-urlopen-in-urllib3) on the question you can simply
```
response = requests.get(cert_url)
pemfile = response.raw.read()
```
`requests.get(cert_url)` returns a  `requests.models.Response` and doing `response.raw` returns a `urllib3.response.HTTPResponse` which you can `read()`.

say-yawn: `urlopen` returns a `http.client.HttrpResponse` and its `read()` [returns the response body](https://docs.python.org/3/library/http.client.html#http.client.HTTPResponse.read) and `requests.get` returns a `requests.models.Response` and its `content` [does the same thing.](https://2.python-requests.org/en/master/user/quickstart/#binary-response-content)

emails/views.py
groovecoder: This is for the fraud & abuse pipeline.

emails/views.py
groovecoder: Gonna merge early to test dev env. Can you file an issue for this?

emails/views.py
groovecoder: Gonna merge early to test dev env. Can you file an issue for this?

emails/sns.py
groovecoder: We may be able to drop this entirely if we change to basic auth.

groovecoder: I filed https://github.com/mozilla/fx-private-relay/issues/363 for this.

emails/views.py
say-yawn: Filted #364 for this.

emails/views.py
say-yawn: Filed #364 for this.

PR #345, opened 2020-05-22
Invite monitor waitlist additional logic 156

groovecoder:

Issue #311, opened 2020-05-15
Changing primary email address prevents invited user to access Private Relay


say-yawn:
# About this Issue
Since the `invitations_only` signal checks the email in `Invitations` model, changing primary email address on FX account prevents invited user to access Private Relay.

# Potential Solution
Get FX Account user ID and check that instead.

say-yawn:
@groovecoder just realized we probably want to add a event handler from fxa to get primary email address changes and update the user/invitations accordingly. What do you think?

PR #347, opened 2020-05-22
fix #314: delete invitation if someone deletes their fxa

### LINKED ISSUE #314: If a user deletes their FXA while on the waitlist, delete their data from the invitations table too

groovecoder:
I still need to spot-check this locally.

This now stores a users FXA uid in the invitations table, so we can delete their invitation record (with their email address) if they delete their FXA.


say-yawn:
#311 should use the fxa uid in this field.

privaterelay/views.py
```diff
@@ -140,15 +141,21 @@ def waitlist(request):
...
     try:
         social_account = _get_account_from_jwt(authentic_jwt)
     except SocialAccount.DoesNotExist as e:
         # we received an FXA event for an FXA not in our DB;
-        # capture the exception in sentry, but don't error, or FXA will retry
+        # If the user joined the waitlist, we didn't create a social_account,
+        # but we did create an Invitations object. Delete it if necessary.
+        for event_key in event_keys:
+            if (event_key == FXA_DELETE_EVENT):
+                _delete_invitation(authentic_jwt)
```
say-yawn, line 153: Doesn't social_account get created if the user logs in to our app successfully (can access the private-relay create address page)? If so we should also delete the social account.

groovecoder: Yup, but this call is inside `except SocialAccount.DoesNotExist`. If the user has a social account object, we process the delete event in `_handle_fxa_delete` below here.

privaterelay/templates/socialaccount/authentication_error.html
```diff
@@ -27,6 +27,7 @@
                 <form class="join-waitlist-form flx jst-cntr" id="join-waitlist-form" method="post" action="{% url 'waitlist' %}">
                   {% csrf_token %}
                   <input type="hidden" name="email" type="hidden" value="{{ request.session.waitlist_email }}">
+                  <input type="hidden" name="fxa_uid" type="hidden" value="{{ request.session.waitlist_fxa_uid }}">
```
say-yawn, line 30: Need to update the JS [here](https://github.com/mozilla/fx-private-relay/blob/master/static/js/app.js#L164) to grab the `fxa_uid` field and send the post.

privaterelay/views.py
```diff
@@ -126,11 +126,12 @@ def waitlist(request):
         raise PermissionDenied
     json_body = json.loads(request.body)
     email = json_body['email']
-    if not email:
+    fxa_uid = json_body['fxa_uid']
+    if not email or not fxa_uid:
         return JsonResponse({}, status=400)
 
     _, created = Invitations.objects.get_or_create(
```
say-yawn, line 133: Can't use `get_or_create` since this results into new `Invitations` record being created when the `fxa_uid` does not exist but have the same email address. We can `get_or_create` Invitations using `email` and `active=False`, check for `fxa_uid` value and then add the uid as needed?

privaterelay/views.py
```diff
@@ -209,8 +216,8 @@ def _handle_fxa_profile_change(authentic_jwt, social_account, event_key):
 
 
 def _handle_fxa_delete(authentic_jwt, social_account, event_key):
-    # TODO?: loop over the user's relay addresses and create hard bounce
-    # receipt rules for them? Note: can't just do this in RelayAddress.delete
+    # TODO: Loop over the user's relay addresses and manually call delete()
+    # to create hard bounce receipt rules in SES,
     # because cascade deletes like this don't necessarily call delete()
     deleted_objects = social_account.user.delete()
```
say-yawn, line 222: Maybe we should call `_delete_invitation` here as well, or add `_delete_invitation` on the non-exception case for iterating through the FX events [here](https://github.com/mozilla/fx-private-relay/pull/347/files#diff-f6ad49a624db6b808c53bc30d2ca6713R159-R161).

privaterelay/migrations/0002_auto_20200522_1722.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 2.2.10 on 2020-05-22 17:22
```
say-yawn, line 1: Would be VERY nice if we can use descriptive names like `0002_add_fxa_uid_field_in_monitorsubscriber.py` BEFORE renaming you have to revert the migration and then rename and apply. Otherwise your DB will become 💩

groovecoder: Renamed to `0002_add_fxa_uid_to_invitation.py`

privaterelay/views.py
groovecoder: Added.

privaterelay/templates/socialaccount/authentication_error.html
groovecoder: Fixed.

privaterelay/views.py
groovecoder: I'd like to merge this PR and https://github.com/mozilla/fx-private-relay/pull/345 before tackling this fix. I think this situation will come up if/when someone is on the Monitor wait-list AND directly joined the wait-list.

say-yawn: Got it. Issues created here https://github.com/mozilla/fx-private-relay/issues/369

PR #356, opened 2020-05-26
Use darker icon in extension

lesleyjanenorton:


groovecoder:
Or maybe it's just that the image file doesn't have as much padding/margin?

PR #362, opened 2020-05-27
Fix #361 Display helpful message to stranded users

### LINKED ISSUE #361: Page with no message when ALPHA_INVITE_TOKEN not set
# About this Issue
While testing as a new user and did not setup the `ALPHA_INVITE_TOKEN` and `WAITLIST_OPEN`, I get an empty page without any messages. In other words, if we remove the `ALPHA_INVITE_TOKEN` and `WAITLIST_OPEN` users will get no direction/message. This happens because when this [if check](https://github.com/mozilla/fx-private-relay/blob/invite-monitor-waitlist-additional-logic-156/privaterelay/signals.py#L51) fails, it raises `PermissionDenied` without any message.

# Acceptance Criteria
- [ ] Message is displayed when the waitlist and alphe invites are not open

say-yawn:
# About this PR
When the alpha testing and waitlist is not turned on the users who try to login get an empty page. We want to display a "helpful" message to user to check back later.

# Acceptance Criteria
- [x] When the alpha token and waitlist is not set, user is displayed with a message to come back later

# Practical Tests
## Check that the user gets no message when alpha token and waitlist is not set
1. Do not get the changes from this PR and go to .env file and set the `ALPHA_INVITE_TOKEN` AND `WAITLIST_OPEN` to empty string
2. Run server and login with FXA
3. Check that you are redirected to `http://127.0.0.1:8000/accounts/fxa/login/callback/?code=...` link and the page looks like the following:
Screen Shot 2020-05-27 at 12 01 10 PM

## Check that the user gets a message when alpha token and waitlist is not set
1. Get the changes from this PR and leave the .env file as it was on previous test
2. Run server and login with FXA
3. Check that you are redirected to `http://127.0.0.1:8000/accounts/fxa/login/callback/?code=...` link and the page looks the same but is now displaying a message.

#
Fix #361

### DIFF
--- a/privaterelay/signals.py
+++ b/privaterelay/signals.py
@@ -74,4 +74,5 @@ def invitations_only(sender, **kwargs):
                 return True
 
     # Deny-by-default in case the logic above missed anything
-    raise PermissionDenied
+    raise PermissionDenied("Please check again later to be invited on "
+                           "Private Relay.")
--- a/privaterelay/signals.py
+++ b/privaterelay/signals.py
@@ -59,8 +59,10 @@ def invitations_only(sender, **kwargs):
             # their session
             if 'alpha_token' not in kwargs['request'].session:
                 raise PermissionDenied(
-                    "You must visit the invitation link in your invite email "
-                    "before signing up for Relay.")
+                    "Private Relay is currently invite-only. "
+                    "If you received an invitation email, "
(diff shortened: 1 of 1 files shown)


groovecoder:
I tweaked the language a bit.

PR #370, opened 2020-06-01
Add fxa id on invitation create fix 311

say-yawn:
# About this PR
Fix #311

PR #372, opened 2020-06-01
Get existing user in the waitlist

### LINKED ISSUE #369: New Invitations created when existing inviation's FXA ID is null
# About this Issue
#347 introduced FXA ID column in `Invitations` model. Any existing invitation without the FXA id may get a new invitation created when the user tries to join the waitlist. See comments [here](https://github.com/mozilla/fx-private-relay/pull/347/files#r432071225)

say-yawn:
#About this PR
Fix #369

### DIFF
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -130,10 +130,20 @@ def waitlist(request):
     if not email or not fxa_uid:
         return JsonResponse({}, status=400)
 
-    _, created = Invitations.objects.get_or_create(
-        fxa_uid=fxa_uid, email=email, active=False
-    )
-    status = 201 if created else 200
+    try:
+        invitation =  Invitations.objects.get(
+            Q(fxa_uid=fxa_uid) | Q(email=email), active=False
+        )
+        if not invitation.fxa_uid:
+            invitation.fxa_uid = fxa_uid
+            invitation.save()
+        if invitation.email != email:
+            invitation.email = email
+            invitation.save()
+        status = 200
+    except Invitations.DoesNotExist:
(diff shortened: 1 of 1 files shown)


privaterelay/views.py
```diff
@@ -130,10 +130,20 @@ def waitlist(request):
...
+    try:
+        invitation =  Invitations.objects.get(
+            Q(fxa_uid=fxa_uid) | Q(email=email), active=False
+        )
+        if not invitation.fxa_uid:
+            invitation.fxa_uid = fxa_uid
+            invitation.save()
+        if invitation.email != email:
+            invitation.email = email
+            invitation.save()
```
say-yawn, line 142: Not sure if we need this. If we already have a email change event from Firefox Accounts, we should have a separate event listener for that?

groovecoder:
I mentioned in I mentioned in https://github.com/mozilla/fx-private-relay/pull/370#issuecomment-637707856 a couple lines that could prevent a bunch of these missing fxa_uid values from happening when we import Monitor waitlist. Can we make that change here?

privaterelay/views.py
groovecoder: Interesting. Shouldn't happen, but this logic makes sense too and will fix anything that might be messed up, right?

say-yawn: We talked about keeping this in here as is and adding a TODO to get events from Firefox Accounts and update emails on the invitation as the profile update is already being observed for users.

say-yawn: Opened https://github.com/mozilla/fx-private-relay/issues/378 for this.

Issue #371, opened 2020-06-01
Email HTML broken on Dev


say-yawn:
# About this Issue
After merging #312 the emails being delivered shows broken HTML on dev.
Before:
<img width="1606" alt="socketlab" src="https://user-images.githubusercontent.com/25109943/83455079-e40b2380-a422-11ea-9b01-6938e4b384b3.png">

After:
<img width="1587" alt="ses" src="https://user-images.githubusercontent.com/25109943/83455088-e7061400-a422-11ea-8d6c-8fe807146cd2.png">

The email delivered from SES also do not have the header that indicates that the email was delivered via Private Relay.

groovecoder:
Do you mind sharing the "View Original" email contents? It should show us what the relevant email headers were. (e.g., `Content-Type` with charset, and/or `Content-Transfer-Encoding` could cause this.)

groovecoder:
Fixed in #373.

PR #373, opened 2020-06-02
for #371: check each subpart for content-transfer-encoding

groovecoder:
Some email senders (e.g., StackOverflow) have specific transfer encodings in each subpart of a multi-part message.

```
Content-Type: text/plain; charset="UTF-8"
Content-Transfer-Encoding: quoted-printable

...

Content-Type: text/html; charset="UTF-8"
Content-Transfer-Encoding: base64
```

So, we need to check each part to see if we should decode it differently than default.

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -1,3 +1,4 @@
+from base64 import b64decode
 from datetime import datetime
 from email import message_from_string, policy
 from email.utils import parseaddr
@@ -5,6 +6,7 @@ from hashlib import sha256
 import json
 import logging
 import markus
+import quopri
 import re
 
 import boto3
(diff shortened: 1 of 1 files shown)


emails/views.py
```diff
@@ -249,15 +249,11 @@ def _sns_message(message_json):
...
-        # E.g., b64decode(message_payload.get_payload()).decode('utf-8')
-        if message_payload.get_content_type() == 'text/plain':
-            text_content = message_payload.get_payload()
-        if message_payload.get_content_type() == 'text/html':
-            html_content = message_payload.get_payload()
+    email_message = message_from_string(
+        message_json['content'], policy=policy.default
+    )
+
+    text_content, html_content = _get_text_and_html_content(email_message)
```
groovecoder, line 256: Note: we should probably check `email_message.is_multipart()` before assuming the email has both text and html content.

groovecoder: If the email message is NOT multipart, we should probably use `email_message.get_body()` and relay that as whatever content_type it is.

groovecoder:
StackOverflow email registration before these changes

image

After these changes ...

image

groovecoder:
And the plain text version ...

Original:
image

These changes:

image

say-yawn:
Testing on dev. Looks good but the header is not being shown. Created https://github.com/mozilla/fx-private-relay/issues/375 as a follow-up.

emails/views.py
groovecoder: Filed https://github.com/mozilla/fx-private-relay/issues/377 for this.

PR #374, opened 2020-06-03
wip: add-on metrics events via GMP

groovecoder:


extension/js/add_input_icon.js
```diff
@@ -63,6 +63,14 @@ function addRelayIconToInput(emailInput) {
   };
 
   buttonEl.addEventListener("click", async (e) => {
+    browser.runtime.sendMessage({
+      method: "sendMetricsEvent",
+      eventData: {
+        category: "Addon Input Icon",
+        action: "click",
+      },
+    })
```
groovecoder, line 72: @lesleyjanenorton so, this would be the call to send event pings to the back-end from the add-on. Look good?

groovecoder:
TODO: finish instrumenting the add-on, and add the pings to metrics.md

privaterelay/views.py
```diff
@@ -157,6 +158,38 @@ def waitlist(request):
     return JsonResponse({'email': email}, status=status)
 
 
+# TODO: refactor to consolidate with emails.views
+@csrf_exempt
+def metrics_event(request):
+    request_data = json.loads(request.body)
+    # TODO: set and get ga_uuid in/from add-on, instead of looking it up
+    if (
+        not request.user.is_authenticated and
```
lesleyjanenorton, line 167: Does this means users need to be signed in to the website in order for events to fire?

groovecoder: To send them along to GA, yes. Because we need to send them with a `ga_uuid`.

So in the add-on, I suggest we don't send any pings unless/until the user signs in.

mdboom:
@groovecoder and I are talking later today about Mozilla-built options for this.  Regardless of where that goes, we should make sure this gets a [data review](https://wiki.mozilla.org/Firefox/Data_Collection) prior to merging.

privaterelay/views.py
```diff
@@ -164,6 +165,38 @@ def waitlist(request):
...
+        not request_data.get("api_token", False)
+       ):
+        raise PermissionDenied
+    api_token = request_data.get('api_token')
+    user_profile = _get_user_profile(request, api_token)
+    if not user_profile.ga_uuid:
```
lesleyjanenorton, line 180: Do we want to move this `user_profile` check down to  L188 so that we can still send GA event data for new users who have never signed in? Thinking specifically about panel opens, "Join the Waitlist" and "Sign In" clicks, etc.

privaterelay/settings.py
```diff
@@ -110,6 +110,7 @@
 
 RELAY_FROM_ADDRESS = config('RELAY_FROM_ADDRESS', None)
 SITE_ORIGIN = config('SITE_ORIGIN', None)
+GOOGLE_ANALYTICS_ID = config('GOOGLE_ANALYTICS_ID', None)
```
lesleyjanenorton, line 113: Do we need `GOOGLE_ANALYTICS_ID` to add this to `.env-dist` ?

groovecoder: Added.

extension/js/add_input_icon.js
```diff
@@ -210,15 +212,18 @@ async function addRelayIconToInput(emailInput) {
     const maxNumAliasesReached = numAliasesRemaining === 0;
     if (maxNumAliasesReached) {
       generateAliasBtn.disabled = true;
+      sendInPageEvent("viewed-menu", "input-menu-max-aliases-message")
```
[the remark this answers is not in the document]
lesleyjanenorton: I also wondered about this. If we change the `click` actions to `clicked-something`, it makes for easier GA scanning, but then the `eventLabel` and `eventAction` have basically the same value.

extension/js/popup.js
```diff
@@ -145,6 +158,9 @@ async function popup() {
   document.querySelectorAll(".close-popup-after-click").forEach(el => {
     el.addEventListener("click", async (e) => {
       e.preventDefault();
+      if (e.target.dataset.eventLabel && e.target.dataset.eventAction) {
+        sendRelayEvent("Panel", e.target.dataset.eventAction, e.target.dataset.eventLabel);
+      }
```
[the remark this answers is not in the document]
lesleyjanenorton: Nope, these are the only events attached to the "outbound" CTAs.

extension/js/add_input_icon.js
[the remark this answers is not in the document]
groovecoder: True - it could be that we just don't know the best values to use here because we haven't tried to report on them yet. Fine to leave as-is to avoid paralysis-by-analysis.

privaterelay/views.py
groovecoder: We fixed this by generating the GA uuid in the add-on and sending it with the metrics ping.

extension/js/add_input_icon.js
```diff
@@ -270,6 +277,7 @@ async function addRelayIconToInput(emailInput) {
 
   divEl.appendChild(relayIconBtn);
   emailInputWrapper.appendChild(divEl);
+  sendInPageEvent("input-icon-injected", "input-icon");
```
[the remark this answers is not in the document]
lesleyjanenorton: Not reliably unfortunately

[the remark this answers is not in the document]
groovecoder: no worries then.

PR #384, opened 2020-06-08
Concatenate and minify CSS and JS

lesleyjanenorton:


requirements.txt
```diff
@@ -1,22 +1,59 @@
+attrs==19.3.0
 boto3==1.13.9
-Django==2.2.10
+botocore==1.16.18
+certifi==2019.11.28
+cffi==1.14.0
+chardet==3.0.4
+coverage==5.0.4
+cryptography==2.9
+datadog==0.35.0
```
pdehaan, line 9: Did you mean to check in this file? Seems like a lot of unrelated packages for concatenation and minification.

lesleyjanenorton: Definitely not - @groovecoder and I were just talking about this in a Slack thread. I have some virtual environment sorting out to do. Running `pip freeze > requirements.txt` adds any and every package I've ever installed in a virtual environment and `pip freeze | grep django_compressor >> requirements.txt` produces what you see here.

PR #394, opened 2020-06-09
fix #325: redirect signed-out profile views to fxa_login

### LINKED ISSUE #325: Clicking "Visit Dashboard" in email header sends user to home page
If the user is not already signed into the site, clicking "Visit Dashboard" sends the user to the home page.

It should redirect to the FXA sign-in flow.

groovecoder:
To test, visit http://127.0.0.1:8000/accounts/profile/ while signed-out. You should be redirected to FXA sign-in, and then land back on the dashboard.

### DIFF
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -15,6 +15,7 @@ from django.core.exceptions import PermissionDenied
 from django.db import connections
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import render, redirect
+from django.urls import reverse
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_http_methods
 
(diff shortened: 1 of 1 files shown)


privaterelay/views.py
```diff
@@ -40,7 +41,7 @@
 
 def home(request):
     if (request.user and not request.user.is_anonymous):
-        return redirect('/accounts/profile/')
+        return redirect(reverse('fxa_login'))
```
say-yawn, line 44: I'm confused as to why a logged-in user would be redirected to login?

groovecoder: Oh whoops. Good catch. Not sure why I put this in there.

groovecoder: Updated.

PR #396, opened 2020-06-09
Log FXA profile update event

say-yawn:
# About this PR
Get jwt values on log so we have a sample jwt value to work with.


privaterelay/views.py
```diff
@@ -179,6 +179,10 @@ def fxa_rp_events(request):
 
     for event_key in event_keys:
         if (event_key == FXA_PROFILE_CHANGE_EVENT):
+            logger.info('fxa_profile_update', extra={
```
groovecoder, line 182: Can we wrap this in `if settings.DEBUG` so we can leave this on in dev?

PR #398, opened 2020-06-10
Update invitation email fields and additional user logic

### LINKED ISSUE #378: Update invitation email address when FXA primary email updates
# About this Issue
As pointed out in this [comment](https://github.com/mozilla/fx-private-relay/pull/372/files#r433519642) we should update the invitation emails when the FXA sends event for primary email update. Not doing this cna lead to emails not being delivered to people who wait-listed on Relay.

say-yawn:
# About this PR
If a previously authenticated and verified FXA user changes their email address, the user cannot access Relay due to the login in `signals.py`. We should update this logic to allow any users previously allowed in Relay by checking a value that does not change (like FXA id). Similarly, a user who is listed under waitlist may not get an email if the primary email is changed but not reflected in `Invitiations`
Fix #378

### DIFF
--- a/privaterelay/management/commands/invite_monitor_waitlist.py
+++ b/privaterelay/management/commands/invite_monitor_waitlist.py
@@ -97,6 +97,7 @@ class Command(BaseCommand):
                 print("Creating invitation for %s" % invitee.primary_email)
                 invitation = Invitations.objects.create(
                     email=invitee.primary_email,
+                    fxa_uid=invitee.fxa_uid,
                     active=True,
                     date_added=datetime.now(),
                     date_redeemed=None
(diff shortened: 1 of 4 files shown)


say-yawn:
It's a little fuzzy to me on whether or not `SocialAccount` or `User` should be used to verify users on Private Relay app's `signals.py` and `views.py`. On the `signals.py` (see [here](https://github.com/mozilla/fx-private-relay/blob/master/privaterelay/signals.py#L67)) the `User` object is used while in `views.py` ([here](https://github.com/mozilla/fx-private-relay/blob/master/privaterelay/views.py#L224)) `SocialAccount` object is checked/updated.

PR #402, opened 2020-06-11
separate RELAY_FROM_ADDRESS and RELAY_ALIAS_DOMAIN

groovecoder:
During our stage switch to AWS, we changed the stage "From" address to a domain (@mozilla.com) that's different than the domain we use for the alias addresses (@stage.fxprivaterelay.nonprod.cloudops.mozgcp.net).

But, our alias address code for both the dashboard and the add-on assume the "From" address domain is the same as the alias address domain.

This code changes it so we can (but don't have to) have separate domains for the "From" address and the alias addresses.


say-yawn:
Closing as we found a work-around.

Issue #364, opened 2020-05-28
Refactor and clean SES code


say-yawn:
# About this Issue
After SES code was meged in PR #312 and tested, we should refactor code to make email handling using SES more robust. This issue tracks some refactoring needed.

# Acceptance Criteria
- [x] `sns_inbound` should call a separate function `validate_sns_header(topic_arn, message_type)` that validates the sns headers in [these lines](https://github.com/mozilla/fx-private-relay/blob/master/emails/views.py#L129-L164) of code.
- [ ] `_sns_inbound_logic` should have a default return like:
```
        return _sns_notification(json_body)
    logger.error(
        'SNS message type did not fall under the SNS inbound logic',
        extra={'message_type': message_type}
    )
    return HttpResponse('Received SNS message with type not handled in inbound log', status=400)
```

# 
Related to #312

EMMLynch:
Please reopen if this is still needed.

PR #409, opened 2020-06-15
Refactors SES code for better email handling

commanderHR1:
Suggested in Issue #364:
-Creates separate call in sns_inbound to validate_sns_header
-Added default return logic for _sns_inbound_logic

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -125,7 +125,7 @@ def _index_DELETE(request_data, user_profile):
 @csrf_exempt
 def sns_inbound(request):
     # We can check for some invalid values in headers before processing body
-    #Grabs message information for validation
+    # Grabs message information for validation
     topic_arn = request.headers.get('X-Amz-Sns-Topic-Arn', None)
     message_type = request.headers.get('X-Amz-Sns-Message-Type', None)
     
--- a/emails/views.py
+++ b/emails/views.py
@@ -125,43 +125,13 @@ def _index_DELETE(request_data, user_profile):
 @csrf_exempt
 def sns_inbound(request):
     # We can check for some invalid values in headers before processing body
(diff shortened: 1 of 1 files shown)


emails/views.py
```diff
@@ -125,7 +125,33 @@ def _index_DELETE(request_data, user_profile):
 @csrf_exempt
 def sns_inbound(request):
     # We can check for some invalid values in headers before processing body
+    #Grabs message information for validation
```
say-yawn, line 128:
```suggestion
    # Grabs message information for validation
```

emails/views.py
```diff
@@ -125,7 +125,33 @@ def _index_DELETE(request_data, user_profile):
...
     topic_arn = request.headers.get('X-Amz-Sns-Topic-Arn', None)
+    message_type = request.headers.get('X-Amz-Sns-Message-Type', None)
+    
+    #Validates header
```
say-yawn, line 132:
```suggestion
    # Validates header
```

emails/views.py
```diff
@@ -194,6 +201,12 @@ def _sns_inbound_logic(topic_arn, message_type, json_body):
             extra={'json_body': json_body},
         )
         return _sns_notification(json_body)
+    
+    logger.error(
+        'SNS message type did not fall under the SNS inbound logic',
+        extra={'message_type': message_type}
+    )
+    return HttpResponse('Received SNS message with type not handled in inbound log', status=400)
```
say-yawn, line 209: I think we should also surface this error on Sentry, if possible. Otherwise we can create a follow-up issue for getting the errors to show up Sentry.

commanderHR1: Sounds good, I'll see what I can do!

PR #421, opened 2020-06-16
fix alias labels with addon 1.3.1

groovecoder:


static/js/app.js
```diff
@@ -279,26 +279,56 @@ function addEventListeners() {
 	});
 }
 
+function hasParent(el, selector) {
```
groovecoder, line 282: nit: rename this function parameter from `selector` to `class` or something more accurate?

PR #437, opened 2020-06-17
Add dashboard banners

lesleyjanenorton:
These need final content but are otherwise ready to go.


groovecoder:
Need a conflict resolution before we can merge.

groovecoder:
As soon as we resolve conflict here we can merge.

privaterelay/templates/includes/banners.html
```diff
@@ -0,0 +1,20 @@
...
+        <p class="banner-sub">The Relay extension for Firefox Browser makes creating aliases even easier.</p>
+        <a class="banner-link ff-Met" href="https://www.mozilla.org/firefox/new/?utm_source=fx-relay&utm_medium=banner&utm_campaign=download-fx"" target="_blank" rel="noopener noreferrer">Get Firefox</a>
+      </div>
+    </div>
+    <div class="install-addon-banner banner-content flx flx-row hidden">
+      <div class="banner-relay-logo banner-logo"></div>
+      <div class="banner-copy flx flx-col">
+        <p class="banner-hl ff-Met">Looks like you haven't installed the add-on... yet!</p>
+        <p class="banner-sub">The Relay extension for Firefox Browser makes creating aliases even easier.</p>
+        <a class="banner-link ff-Met" href="https://addons.mozilla.org/firefox/addon/private-relay/?utm_source=fx-relay&utm_medium=banner&utm_campaign=install-addon" target="_blank" rel="noopener noreferrer">Install the Firefox Relay addon</a>
```
pdehaan, line 16: nit: you say "addon" here, but "add-on" above on L14. We should be consistent, but I think the preference is hyphenated (per https://addons.mozilla.org/en-US/firefox/ labels).

static/images/logos/fx-logo.svg
```diff
@@ -0,0 +1,79 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512">
```
pdehaan, line 1: Q: Should this be minified?
There doesnt seem to be a specified width/height, so I can't view the SVG in GitHub's Rich Diff tools.

static/js/app.js
```diff
@@ -350,6 +350,43 @@ function watchForInstalledAddon() {
 }
 
 
+function showBannersIfNecessary() {
+  if (window.clientWidth < 750) {
+    return;
+  }
+  const dashboardBanners = document.querySelector(".dashboard-banners");
+  const bg = dashboardBanners.querySelector(".banner-gradient-bg");
```
pdehaan, line 358: nit: Not sure if we should move this down closer to L367 to save an unnecessary DOM lookup if here are no `dashboardBanners` found or if the user isn't using Firefox or doesn't have the add-on installed.

PR #439, opened 2020-06-17
Concatenate and minify JS and CSS

lesleyjanenorton:


privaterelay/templates/base.html
```diff
@@ -24,9 +27,11 @@
     {% endblock %}
 
     {% include "includes/footer.html" %}
-    <script type="text/javascript" src="{% static 'js/clipboard.min.js' %}" charset="utf-8"></script>
-    <script src="{% static 'js/fx-bento.js' %}"></script>
-    <script src="{% static 'js/app.js' %}"></script>
-    <script src="{% static 'js/analytics.js' %}"></script>
+    {% compress js %}
+      <script type="text/javascript" src="{% static 'js/clipboard.min.js' %}" charset="utf-8"></script>
```
pdehaan, line 31: Not sure if this is a legit concern, but...
Should we _re_-minify the clipboard.min.js file?
Not sure if this should be removed and we just load two .js files.

groovecoder:
IMO we **should** re-minify `clipboard.min.js` file with our scripts, because it eliminates an extra HTTP request which is usually the bigger performance problem than the **size** of scripts.

groovecoder (approved):

groovecoder:
I reverted this branch to the commit BEFORE we moved clipboard.min.js out of the compressed js asset.

pdehaan:
> IMO we **should** re-minify `clipboard.min.js` file with our scripts

Sure, I have no objections if you think it will work. I was afraid of what gateways into the abyss we might create by reminification, but after some light search-engining(tm), I can't find a lot of general warnings about code folding into itself or gaining sentience. Let's try it out on stage and see if we find any sharp edges.

pdehaan:
FWIW, here's what I'm currently seeing on Production re: Network Requests (pre-concat-minification):

Firefox_Relay

Issue #447, opened 2020-06-18
Add more telemetry to SES code


groovecoder:
We have had timers on the SocketLabs sending operation. We should add them for SES too.

groovecoder:
Also using this issue to make sure we have 1:1 telemetry parity with the SocketLabs code.

groovecoder:
Added a number of new metrics calls. Need to add them to the dashboard.

groovecoder:
Added to dashboard: https://earthangel-b40313e5.influxcloud.net/d/qiwPC76Zk/fx-private-relay?orgId=1&refresh=1m&from=now-7d&to=now

PR #451, opened 2020-06-18
for #447: email counters and timers

groovecoder:


emails/views.py
```diff
@@ -274,29 +281,12 @@ def _sns_message(message_json):
     if text_content:
         message_body['Text'] = {'Charset': 'UTF-8', 'Data': text_content}
 
-    relay_from_address, relay_from_display = _generate_relay_From(from_address)
-    formatted_from_address = str(
-        Address(relay_from_display, addr_spec=relay_from_address)
+    ses_client = get_ses_client()
```
say-yawn, line 284: `get_ses_client()` could return `None`, should we check that ses_client is NOT `None` before calling `ses_send_email`?

emails/utils.py
```diff
@@ -22,6 +28,11 @@ def func_wrapper(*args, **kwargs):
     return timing_decorator
 
 
+def incr_if_enabled(name, value):
```
say-yawn, line 31:
```suggestion
def incr_if_enabled(name, value=1):
```
Default value of `1`? It seems like everytime the `incr_if_enabled` increments by 1 only.

emails/views.py
```diff
@@ -40,13 +39,15 @@ def _get_data_from_request(request):
...
 def index(request):
+    incr_if_enabled('emails_index', 1)
     request_data = _get_data_from_request(request)
     if (not request.user.is_authenticated and
         not request_data.get("api_token", False)
        ):
         raise PermissionDenied
     if request.method == 'POST':
         return _index_POST(request)
+    incr_if_enabled('emails_index_GET', 1)
```
say-yawn, line 50:
```suggestion
    incr_if_enabled('emails_index_get', 1)
```
It seems that metric name uses lower cases. Like this [post](https://github.com/mozilla/fx-private-relay/pull/451/files#diff-2c3a9c773eb79ee1e4b1858c23620e74R61), [put](https://github.com/mozilla/fx-private-relay/pull/451/files#diff-2c3a9c773eb79ee1e4b1858c23620e74R106), and [delete](https://github.com/mozilla/fx-private-relay/pull/451/files#diff-2c3a9c773eb79ee1e4b1858c23620e74R123).

emails/views.py
```diff
@@ -57,6 +58,7 @@ def _get_user_profile(request, api_token):
 
 
 def _index_POST(request):
+    incr_if_enabled('emails_index_post', 1)
```
say-yawn, line 61: Should this increment be closer to the `201` response [here](https://github.com/mozilla/fx-private-relay/blob/master/emails/views.py#L88) since this could become a `PUT` or `DELETE` as seen https://github.com/mozilla/fx-private-relay/blob/master/emails/views.py#L69-L72?

emails/views.py
```diff
@@ -96,11 +98,12 @@ def _get_relay_address_from_id(request_data, user_profile):
             user=user_profile.user
         )
         return relay_address
-    except RelayAddress.DoesNotExist as e:
+    except RelayAddress.DoesNotExist:
         return HttpResponse("Address does not exist")
 
 
 def _index_PUT(request_data, user_profile):
+    incr_if_enabled('emails_index_put', 1)
```
say-yawn, line 106: Same question here, should this increment be closer to the successful `JsonResponse`?

emails/views.py
groovecoder: Hmm ... this actually makes me think of a bigger refactor. Should we move the ses client connection "up" into `EmailsConfig.ready()` ? That should hold the client in application memory so we don't have to connect each time.

emails/views.py
groovecoder: Hmm ... I was thinking these would record counters of just the requests made - not just the successful ones.

emails/views.py
groovecoder: I moved this under the PUT and DELETE blocks so we don't double-count those as POSTs too.

say-yawn:
_ping_ just wanted to make sure this PR is not getting old :)

emails/views.py
groovecoder: I think we can go with treating these counters as overall "raw" counts of operations performed. We have other ways to count successful email calls (DB queries) and failed email calls (sentry errors).

groovecoder:
Updated with more review comments, and merged master into this branch to resolve conflicts.

emails/utils.py
```diff
@@ -40,6 +51,33 @@ def socketlabs_send(sl_client, sl_message):
...
+            ConfigurationSetName=settings.AWS_SES_CONFIGSET,
+        )
+        logger.debug('ses_sent_response', extra=ses_response['MessageId'])
+        relay_address.num_forwarded += 1
+        relay_address.last_used_at = datetime.now()
+        relay_address.save(update_fields=['num_forwarded', 'last_used_at'])
+        incr_if_enabled('ses_send_email', 1)
+    except ClientError as e:
+        logger.error('ses_client_error', extra=e.response['Error'])
+        return HttpResponse("SES client error", status=400)
```
say-yawn, line 78:
```suggestion
    except ClientError as e:
        logger.error('ses_client_error', extra=e.response['Error'])
        return HttpResponse("SES client error", status=400)
    return HttpResponse("Sent email to final recipient.", status=200)
```

emails/views.py
```diff
@@ -303,29 +311,10 @@ def _sns_message(message_json):
...
-        relay_address.num_forwarded += 1
-        relay_address.last_used_at = datetime.now()
-        relay_address.save(update_fields=['num_forwarded', 'last_used_at'])
-    except ClientError as e:
-        logger.error('ses_client_error', extra=e.response['Error'])
-        return HttpResponse("SES client error", status=400)
+    if type(result) == HttpResponse:
+        return result
 
     return HttpResponse("Sent email to final recipient.", status=200)
```
say-yawn, line 319:
```suggestion
```

emails/views.py
```diff
@@ -303,29 +311,10 @@ def _sns_message(message_json):
         wrapped_text = relay_header_text + text_content
         message_body['Text'] = {'Charset': 'UTF-8', 'Data': wrapped_text}
 
-    relay_from_address, relay_from_display = _generate_relay_From(from_address)
-    formatted_from_address = str(
-        Address(relay_from_display, addr_spec=relay_from_address)
-    )
+    result = ses_send_email(from_address, relay_address, subject, message_body)
```
say-yawn, line 314:
```suggestion
    return ses_send_email(from_address, relay_address, subject, message_body)
```

PR #454, opened 2020-06-18
Handle text emails

### LINKED ISSUE #452: Handle text emails
# About this is Issue
If the email has the content-type `text/plain` we should wrap the `text/plain` content with a Relay header with texts similar to the header on the `text/html`. We should also duplicate emails that only have `text/plain` content into `text/html` content-type.

# Acceptance Criteria
- [ ] `text/plain` content-type has Relay header-like text
- [ ] email with `text/plain` only should be duplicate the content into `text/html`

say-yawn:
# About this PR
Fix #452

### DIFF
--- a/emails/templates/emails/wrapped_email.html
+++ b/emails/templates/emails/wrapped_email.html
@@ -69,6 +69,6 @@
         </td>
       </tr>
     </table>
-    {{ original_html|safe }}
+    {{ original_html|linebreaks|safe }}
   </body>
 </html>
--- a/emails/views.py
+++ b/emails/views.py
@@ -333,6 +333,7 @@ def _get_text_and_html_content(email_message):
     else:
         if email_message.get_content_type() == 'text/plain':
             text_content = email_message.get_content()
+            html_content = email_message.get_content()
         if email_message.get_content_type() == 'text/html':
             html_content = email_message.get_content()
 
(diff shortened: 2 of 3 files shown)


emails/views.py
```diff
@@ -329,6 +333,7 @@ def _get_text_and_html_content(email_message):
     else:
         if email_message.get_content_type() == 'text/plain':
             text_content = email_message.get_content()
+            html_content = email_message.get_content()
```
groovecoder, line 336: Dang. Setting the content to `text/html` breaks the `\n` line breaks in the plain text part. 😢 (from a reddit signup email)

image

I'm not sure if we should try to go thru the text content and replace `\n` with `<br>` or not? There might be other things that break when plain text is rendered as html.

emails/templates/emails/wrapped_email.html
```diff
@@ -69,6 +69,6 @@
         </td>
       </tr>
     </table>
-    {{ original_html|safe }}
+    {{ original_html|linebreaks|safe }}
```
groovecoder, line 72: Should we apply the `linebreaks` filter here, or should we apply it to `text_content` in the view?

If we apply this here, how does it affect HTML emails? Does it convert all the linebreaks in their markup into more `<p>` and `<br>` tags?

say-yawn: Good catch I'll look into both!

Issue #472, opened 2020-06-23
`css/relay.css` injected into pages conflict with web page styles


ExE-Boss:
Because the extension injects `css/relay.css` into all pages:
<https://github.com/mozilla/fx-private-relay/blob/85607c693764a59d57ecf49905318e5b71905ae2/extension/manifest.json#L45-L52>

And `css/relay.css` contains:
<https://github.com/mozilla/fx-private-relay/blob/85607c693764a59d57ecf49905318e5b71905ae2/extension/css/relay.css#L70-L89>

The injected stylesheet ends up conflicting with web page styles.

## Proposed solutions:
- Separate `css/relay.css` into a stylesheet that is safe to inject into web pages and a stylesheet that is used in extension pages.
- Use **Shadow DOM** to enclose the **Firefox Relay** menu.

dgdavid:
I can confirm this issue. In my case, I cannot distinguish links from the rest of text in Github issues / comments / descriptions.

dgdavid:
Probably related to / duplicated of https://github.com/mozilla/fx-private-relay/issues/341

hoosteeno:
I am seeing an apparent relay-styles-related issue in many other sites (Google Docs, Google Analytics...). The easiest test of my issue is at www.google.com. STR:

1. Visit www.google.com in vanilla profile. See the big rainbow colored logomark above search input box.
2. Install relay extension https://addons.mozilla.org/en-US/firefox/addon/private-relay/
3. Visit www.google.com again. See no big rainbow colored logomark.

![Screen Shot 2020-06-23 at 11 47 44 AM](https://user-images.githubusercontent.com/94519/85437539-f707a400-b547-11ea-9084-e20d9db429b2.png)

AnishRV-NYT:
Having issues on gmail as well. 
**With Private Relay enabled:**
<img width="74" alt="Screen Shot 2020-06-23 at 10 48 27 AM" src="https://user-images.githubusercontent.com/66200765/85437204-13eba980-b53f-11ea-866c-0cdd9797fb81.png"> 
**Without:**
<img width="109" alt="Screen Shot 2020-06-23 at 10 48 02 AM" src="https://user-images.githubusercontent.com/66200765/85437182-07ffe780-b53f-11ea-9321-bae43b68818e.png">

Termuellinator:
[...]

PR #480, opened 2020-06-23
Bandaid fix for #472

lesleyjanenorton:


ExE-Boss:
`css/relay.css` is also depended upon by `panel.html` and other extension pages to set:
```css
* {
box-sizing: border-box;
}
```

groovecoder:
I didn't see any elements in the extension panel or pages that broke without the box-sizing: border-box. Going to make a release with this fix as it is breaking live websites for everyone. We can fix more things as follow-ups.

PR #483, opened 2020-06-23
Update name and description in manifest.json

lesleyjanenorton:


extension/manifest.json
```diff
@@ -1,9 +1,9 @@
 {
   "manifest_version": 2,
-  "name": "Private Relay",
+  "name": "Firefox Relay",
```
groovecoder, line 3: This makes the `.zip` file come out as `firefox_relay-1.4.2.zip` which is fine, but we also need to update the `build` scripts in `package.json` from `private_relay-*.zip` to `firefox_relay-*.zip`

groovecoder: Still needs this update to `package.json` so that `npm run build:prod` will work as before.

PR #487, opened 2020-06-24
First run updates

lesleyjanenorton:
- Prevents first-run.html from opening after automatic updates
- Updates the "Join the waitlist" link copy, and sends users directly into the FxA OAuth flow to hopefully mitigate waitlist flow confusion.


extension/js/background.js
```diff
@@ -4,14 +4,19 @@ browser.storage.local.set({ "maxNumAliases": 5 });
 browser.storage.local.set({ "showInputIcons": "show-input-icons" });
 browser.storage.local.set({ "relaySiteOrigin": RELAY_SITE_ORIGIN });
 browser.storage.local.set({ "fxaOauthFlow": `${RELAY_SITE_ORIGIN}/accounts/fxa/login/?process=login` });
-
+browser.storage.local.set({ "firstRunShown" : false });
```
groovecoder, line 7: This doesn't fix the issue for users who aren't able to sign in, because this line over-writes the `true` value set by the `onInstalled` listener. To verify:

1. Run `npm run build:prod` on this branch
2. Open Nightly with a new profile
3. Set `xpinstall.signatures.required` to `false`
4. In `about:addons`, Install add-on from file and choose the .zip
5a. Do *NOT* sign in at relay.firefox.com, or
5b. Sign in at relay.firefox.com with an FXA that isn't invited
6. In `about:addons`, Install add-on from file and choose the .zip

You'll see the first run page again, because `firstRunShown` is false and `userApiToken` is also false.

PR #499, opened 2020-06-25
Updated README.md

uvinduperera:
Added some code lines for centos users


groovecoder:
Thanks for the PR, but we won't put any platform-specific instructions in the README to keep it as concise as possible.

PR #502, opened 2020-06-25
Update copy in addon

lesleyjanenorton:


extension/js/background.js
```diff
@@ -83,7 +83,7 @@ async function makeRelayAddressForTargetElement(info, tab) {
 if (browser.menus) {
   browser.menus.create({
     id: "fx-private-relay-generate-alias",
-    title: "Generate Email Alias",
+    title: "Generate New Alias",
```
pdehaan, line 86: Q: Why is this "Generate New Alias" (uppercases), but "Generate new alias" above on extension/js/add_input_icon.js#L133 (and L190)?

extension/js/popup.js
```diff
@@ -11,7 +11,7 @@ function getOnboardingPanels() {
     "panel2": {
       "imgSrc": "tip2-icon.svg",
       "tipHeadline": "Don't see the icon?",
-      "tipBody":"Right click (Windows) or control+click (Mac) to open the context menu easily generate a new relay from there",
+      "tipBody":"Right-click (Windows) or Control-click (macOS) to open the context menu easily generate a new relay from there.",
```
pdehaan, line 14: nit: This feels odd here since we've been removing instances of "relay" elsewhere above.

extension/popup.html
```diff
@@ -40,13 +40,13 @@ <h1 class="settings-hl">Settings</h1>
...
         <div class="join-waitlist-wrapper">
-          <span class="">Curious about Private Relay?</span>
+          <span class="">Curious about Relay?</span>
           <a href="" class="join-waitlist-link login-link">Sign up for the waitlist with your Firefox account.</a>
         </div>
       </sign-up-panel>
 
       <main-panel class="signed-in-panel hidden">
-          <p class="ff-Met aliases-remaining text-center remaining-plural">You have <span class="num-aliases-remaining"></span><span class="ff-Met max-num-aliases"></span></span> relay addresses remaining</p>
+          <p class="ff-Met aliases-remaining text-center remaining-plural">You have <span class="num-aliases-remaining"></span><span class="ff-Met max-num-aliases"></span></span> aliases remaining</p>
```
pdehaan, line 49: nit: I think this has a bonus `</span>` in here.

extension/js/fill_relay_address.js
```diff
@@ -44,12 +44,12 @@ async function showModal(modalType, newAlias=null) {
...
-    modalMessage.textContent = "You have reached the maximum number of aliases allowed during the beta phase of Private Relay.";
+    modalMessage.textContent = "You've used all of your beta aliases.";
     modalMessage.classList = ["relay-modal-message relay-modal-headline"];
     modalContent.appendChild(modalMessage);
 
     const manageAliasesLink = document.createElement("a");
-    manageAliasesLink.textContent = "Manage Relay Addresses";
+    manageAliasesLink.textContent = "Manage All Addresses";
     manageAliasesLink.classList = ["new-tab"];
     manageAliasesLink["href"] = `${relaySiteOrigin}?utm_source=fx-relay-addon&utm_medium=context-menu-modal&utm_campaign=beta&utm_content=manage-relay-addresses`;
```
pdehaan, line 54: unrelated, but why is this using object notation for `["href"]` here, but dot notation elsewhere for `.textContent` and `.classList`?

extension/js/background.js
lesleyjanenorton: Wasn't my decision but I think because other context menu items are also capitalized.

extension/js/popup.js
lesleyjanenorton: I will bring this up with UX

extension/js/fill_relay_address.js
lesleyjanenorton: If I had a good reason, I don't remember it. Updated.

extension/js/popup.js
groovecoder: Also says "open the context menu easily generate" when it should say "open the context menu *to* easily generate" ?

lesleyjanenorton: Or should it be "..._to_ open the context menu _and_ easily generate a new..."

Issue #491, opened 2020-06-24
Problem for joining the waiting list on firefox


Sim5912:
Hello,

I tried to get on the waiting list on firefox, but it's not working.
I click on the button to join the waiting list, I log in and then I click on "join the waiting list".

Then, I always get the same error message : 

"Oops, we're sorry but we hit a snag adding your email to the waitlist.

Please try again later or file an issue."

I tried several times without success.

However, when I tried edge (the new one which is based on chromium), everything worked the first time.

What can I do ?

Regard,

![bug](https://user-images.githubusercontent.com/51054004/85600085-6f0bc200-b64d-11ea-8d5f-4d0c694ee91d.JPG)

groovecoder:
Looks like this might be caused by https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/8862456/?query=is%3Aunresolved

Sim5912:
I can't access the link because it asks me for credentials :(

say-yawn:
@Sim5912 apologies for the confusion, that link was for me. I'll look into this. Thanks for filing!

Sim5912:
Sorry !

No problem :) 
I remain available if needed

groovecoder:
I was able to reproduce this with NoScript installed.

Sim5912:
I have NoScript but he's deactivated

PR #503, opened 2020-06-25
Problem joining waitlist 

say-yawn:
# About this PR
There are some cases where the JS is not properly packaging the post data to add the user to waitlist resulting in the error message on #491. This PR handles the post data as either a JSON or queryparam that causes [this](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/8862456/?query=is%3Aunresolved) issue on Sentry.

# Practical Test
### Check that error happens when the user joins the waitlist while script is not loaded
1. Get a no script add-on
2. Go to http://127.0.0.1:8000/ and try to add yourself to waitlist. You may have to "trust"/"allow" the script for FXA.
3. Click the join waitlist button on the /accounts/fxa/... page
4. Check that you get an error: a "oops something went wrong" displayed on /account/..., (prod?), 500 server error on `/waitlist/` page(stage), or stack trace of the error on `/waitlist/` page (dev)

### Check that error DOES NOT happen when the user joins the waitlist while script is not loaded
1. Get a no script add-on
2. Go to http://127.0.0.1:8000/ and try to add yourself to waitlist. You may have to "trust"/"allow" the script for FXA.
3. Click the join waitlist button on the /accounts/fxa/... page
4. Check that you are successfully added to the waitlist

### DIFF
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -155,11 +155,13 @@ def waitlist(request):
         if invitation.email != email:
             invitation.email = email
             invitation.save()
+        message = 'You were already on our waitlist!'
         status = 200
     except Invitations.DoesNotExist:
         Invitations.objects.create(fxa_uid=fxa_uid, email=email, active=False)
         status = 201
-    return JsonResponse({'email': email}, status=status)
+        message = 'You are added on our waitlist!'
+    return JsonResponse({'email': email, 'message': message}, status=status)
 
 
 @csrf_exempt
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -26,6 +26,7 @@ from allauth.socialaccount.providers.fxa.views import (
 )
 
 from emails.models import RelayAddress
+from emails.utils import get_post_data_from_request
 from .models import Invitations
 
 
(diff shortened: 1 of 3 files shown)


privaterelay/views.py
```diff
@@ -138,9 +139,9 @@ def invitation(request):
...
     if not settings.WAITLIST_OPEN:
         raise PermissionDenied
-    json_body = json.loads(request.body)
-    email = json_body['email']
-    fxa_uid = json_body['fxa_uid']
+    request_data = get_post_data_from_request(request)
+    email = request_data['email']
+    fxa_uid = request_data['fxa_uid']
     if not email or not fxa_uid:
         return JsonResponse({}, status=400)
```
groovecoder, line 146: The changes fix joining the wait-list, but this makes the browser window show the JSON response. Should we detect a request that is not `application/json` and redirect them ... somewhere? Or just return a nicer response message?

image

say-yawn: Good point, I think either a successful message or redirect to the home page will do but I am leaning towards nicer response message since the user may not know whether they were added to the waitlist if we just redirect the user.

groovecoder: Agree just a nicer message is an improvement over what happens now (i.e., completely broken!)

PR #514, opened 2020-06-26
fix #494: use python3 decodebytes instead of six.b

### LINKED ISSUE #494: SNS messages with invalid signatures
May be causing #478.

See https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/8801352/?query=is%3Aunresolved ...

groovecoder:


emails/sns.py
```diff
@@ -4,7 +4,6 @@
 import base64
 import logging
 import pem
-import six
```
say-yawn, line 7: I think we can remove `six` from the requirements.txt too.

emails/sns.py
```diff
@@ -55,13 +54,11 @@
...
     cert = crypto.load_certificate(crypto.FILETYPE_PEM, pemfile)
-    signature = base64.decodestring(six.b(json_body['Signature']))
+    signature = base64.decodebytes(json_body['Signature'].encode('utf-8'))
 
     hash_format = _get_hash_format(json_body['Type'])
 
-    crypto.verify(
-        cert, signature, six.b(hash_format.format(**json_body)), 'sha1'
-    )
+    crypto.verify(cert, signature, hash_format.format(**json_body), 'sha1')
```
say-yawn, line 61:
```suggestion
    crypto.verify(cert, signature, hash_format.format(**json_body).encode('utf-8), 'sha1')
```
`crypto.verify()` does [handle](https://github.com/pyca/pyopenssl/blob/master/src/OpenSSL/crypto.py#L2849) the text and convert to a byte object but it returns a warning. Might as well just pass a byte object?

groovecoder: Done. But I re-formatted the line to make it fit.

groovecoder:
Updated with comments.

PR #518, opened 2020-06-29
fix #498: use walk() and try each part of messages

### LINKED ISSUE #498: Support emails with attachments
UPDATED DESCRIPTION:

Changing this from a bug to an enhancement. To scope the work we should first:

* [ ] Start recording the size of attachments so we can predict how much bandwidth it may take
* [ ] Start recording the file types of attachments so we know what we may need to support
* [ ] Start recording the number of attachments so we know what we may need to support

OLD DESCRIPTION:

Sentry error: https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/8811759/activity/?query=is%3Aunresolved

SES charges $0.12/GB of data in messages sent, so (large) attachments could increase costs significantly. So we need to decide how we might support attachments.

At the very least, we should relay the email content without an attachment, and possibly include a note that attachments aren't supported?

groovecoder:


groovecoder:
Tested on dev - this fixes emails with attachments. The attachments still don't go thru, but the email at least parses and delivers.

say-yawn:
Did we want to go ahead and try to get the attachment to work, or fast follow? https://docs.aws.amazon.com/ses/latest/DeveloperGuide/send-email-raw.html

groovecoder:
Fast follow. I'm a bit concerned that attachment traffic could cost more than we've planned. I'll change the commit message to "for" so it doesn't close the issue. We'll leave that issue up as our issue to decide what to do with attachments.

groovecoder:
Added some email counter metrics.

ohiogeek8215:
Please help I  have be attacked digitally

PR #519, opened 2020-06-29
Prevent the "install the add-on" banner from appearing to users who have already installed the add-on

lesleyjanenorton:
…ve already installed the addon


groovecoder:
This looks good. Do we have a follow-up issue to remove this local storage flag if/when the add-on is NOT found?

PR #522, opened 2020-06-29
Inform users attachment not supported for 498

say-yawn:
# About this PR
Warn users when Relay forwards emails with attachment that attachment is not currently supported and fix flake8 errors on emails/views.py

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -42,9 +42,11 @@ metrics = markus.get_metrics('fx-private-relay')
 def index(request):
     incr_if_enabled('emails_index', 1)
     request_data = get_post_data_from_request(request)
-    if (not request.user.is_authenticated and
+    is_not_validated_user = (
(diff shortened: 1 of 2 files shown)


emails/views.py
```diff
@@ -322,6 +341,7 @@ def _get_text_and_html_content(email_message):
                 if part.get_content_type() == 'text/html':
                     html_content = content
                 if part.is_attachment():
+                    has_attachment = True
                     incr_if_enabled('email_with_attachment', 1)
```
say-yawn, line 345: I would like it if we can pass how big the attachment is but not sure how to best do that. Ideally we are not just incrementing the metric but having a record for each attachment size. What do you think @groovecoder?

groovecoder: Yeah, that'd be good. In theory the email could include a `Content-Length` header in addition to its `Content-Disposition` header?

PR #536, opened 2020-07-06
Adds on hover description for "Forwarding"

lesleyjanenorton:
Adds descriptive text which appears when the user hovers over the "Forwarded" column.
Screen Shot 2020-07-06 at 11 48 03 AM

Screen Shot 2020-07-06 at 11 48 29 AM

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1887,7 +1887,7 @@ input.input-has-error {
   margin: auto 3px;
 }
 
-.column-blocked {
+.relay-stat {
   position: relative;
 }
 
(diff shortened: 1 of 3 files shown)


static/css/style.css
```diff
@@ -1887,37 +1887,60 @@ input.input-has-error {
...
+}
+
+.stat-description.show-message {
+  transform: translateY(3px);
+  visibility: visible;
   opacity: 1;
   transition: all 0.3s ease;
 }
 
+.stat-description.blocked-description {
```
pdehaan, line 1924: Would this need to apply to `.stat-description.forwarding-description` as well?

lesleyjanenorton: Nope, just `.blocked-description`.

PR #552, opened 2020-07-15
use save() to trigger auto_now_add on RelayAddress

groovecoder:


groovecoder:
https://github.com/mozilla/fx-private-relay/pull/546 is merged & deployed, but using `update_fields` means the `auto_now_add` functionality doesn't trigger.

groovecoder:
Yeah, I checked the others, but I think the only place we want to update the `last_modified_at` field is during enable/disable.

PR #553, opened 2020-07-15
sort monitor invite waitlist by breaches_last_shown

groovecoder:


groovecoder:
Dang. On dev server, I realized that `NULL` value is "before" other timestamp values. So this will sort the monitor invite waitlist, but it will put any records with `NULL` in the `breaches_last_shown` column on top - which are the LEAST engaged Monitor users. :/

Technically, this shouldn't happen - because someone would have had to view their dashboard to click to join the waitlist. I'm asking @Micheletto to check the Monitor prod db to see if there are any records on the waitlist that somehow have `NULL` in the `breaches_last_shown` column.

groovecoder:
From @Micheletto:
```
=> select count(id) from subscribers WHERE waitlists_joined is not null and breaches_last_shown is null;
 count
-------
     0
(1 row)
```

say-yawn:
Oh, nice catch. I didn't know that was a behavior of order-by (is this the Django behavior or psql?). TIL. Since we have no intersection of users who are on the waitlist and also have `NULL` on the ` breaches_last_show` we can just run the script as is? But to be safe we could consider:
```
        monitor_waitlist = (MonitorSubscriber.objects.using('monitor')
            .filter(waitlists_joined__email_relay__notified=False)
            .exclude(breaches_last_shown__isnull=True)
            .order_by('-breaches_last_shown')
            [:limit]
        )
```

groovecoder:
Looks like a feature of SQL.

I'm fine to either run as-is (no new invites are being added) or make a PR to add the `exclude`.

PR #565, opened 2020-07-22
ENH: significantly reduce size of Docker image using multi-stage builds.

kaczmarj:
Dependencies are compiled and installed in the first stage into a virtual environment, made with python's standard venv. In the second stage, a slim python docker image is used (e.g., no compilers), and the virtual environment is copied into the second stage. The result is a docker image with a size of 341 MB (uncompressed) vs 1.13 GB before this commit.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,41 @@
-FROM python:3.7.6
-
-RUN apt-get update && apt-get -y install libpq-dev
-RUN pip install --upgrade pip
-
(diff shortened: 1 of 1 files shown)


groovecoder:
Hi, thanks for the pull request. We're currently in the midst of some pretty active release schedules, so I don't think we'll make any changes to our infrastructure for the time being. I'm going to close this, but we'll re-visit this when we do an infrastructure sprint in the future.

PR #567, opened 2020-07-22
start invite_relay_waitlist command

groovecoder:
Need to push this branch to the dev server to make sure that the refactoring of the SES code is still working for relay emails too.

### DIFF
--- a/emails/utils.py
+++ b/emails/utils.py
@@ -52,31 +52,45 @@ def socketlabs_send(sl_client, sl_message):
 
 
 @time_if_enabled('ses_send_email')
-def ses_send_email(from_address, relay_address, subject, message_body):
+def ses_send_email(from_address, to_address, subject, message_body):
     emails_config = apps.get_app_config('emails')
-    relay_from_address, relay_from_display = generate_relay_From(from_address)
-    formatted_from_address = str(
-        Address(relay_from_display, addr_spec=relay_from_address)
-    )
     try:
         ses_response = emails_config.ses_client.send_email(
-            Destination={'ToAddresses': [relay_address.user.email]},
+            Destination={'ToAddresses': [to_address]},
             Message={
                 'Body': message_body,
                 'Subject': {'Charset': 'UTF-8', 'Data': subject},
             },
-            Source=formatted_from_address,
+            Source=from_address,
             ConfigurationSetName=settings.AWS_SES_CONFIGSET,
         )
         logger.debug('ses_sent_response', extra=ses_response['MessageId'])
-        relay_address.num_forwarded += 1
-        relay_address.last_used_at = datetime.now()
-        relay_address.save(update_fields=['num_forwarded', 'last_used_at'])
         incr_if_enabled('ses_send_email', 1)
     except ClientError as e:
         logger.error('ses_client_error', extra=e.response['Error'])
         return HttpResponse("SES client error", status=400)
     return HttpResponse("Sent email to final recipient.", status=200)
+
+
+def ses_relay_email(from_address, relay_address, subject, message_body):
+    relay_from_address, relay_from_display = generate_relay_From(from_address)
+    formatted_from_address = str(
+        Address(relay_from_display, addr_spec=relay_from_address)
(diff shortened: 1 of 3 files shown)


emails/utils.py
```diff
@@ -52,31 +52,45 @@ def socketlabs_send(sl_client, sl_message):
...
+        response = ses_send_email(
+            formatted_from_address,
+            relay_address.user.email,
+            subject,
+            message_body
+        )
         relay_address.num_forwarded += 1
         relay_address.last_used_at = datetime.now()
         relay_address.save(update_fields=['num_forwarded', 'last_used_at'])
-        incr_if_enabled('ses_send_email', 1)
```
say-yawn, line 75: I may have misunderstood how this metric was used for but shouldn't we keep the `ses_send_email` metric here since this is the metric used to count how many emails were sent via Relay? If we remove this and instead add it where `ses_send_email` I think we will be getting the number of emails sent for the waitlisted users as well.

privaterelay/management/commands/invite_relay_waitlist.py
```diff
@@ -0,0 +1,79 @@
...
+    message_body = {}
+
+    message_body['Html'] = {'Charset': 'UTF-8', 'Data': render_to_string(
+        'emails/beta_invite_html_email.html',
+        context,
+    )}
+    message_body['Text'] = {'Charset': 'UTF-8', 'Data': render_to_string(
+        'emails/beta_invite_text_email.txt',
+        context,
+    )}
```
say-yawn, line 33:
```suggestion
    message_body['Html'] = {
        'Charset': 'UTF-8',
        'Data': render_to_string(
            'emails/beta_invite_html_email.html',
            context
        )
    }
    message_body['Text'] = {
        'Charset': 'UTF-8', 
        'Data': render_to_string(
            'emails/beta_invite_text_email.txt',
            context
        )
    }
```

privaterelay/management/commands/invite_relay_waitlist.py
```diff
@@ -0,0 +1,79 @@
...
+    help = 'Connects to Monitor waitlist and creates invitations.'
+
+    def add_arguments(self, parser):
+        parser.add_argument('limit', nargs=1, type=int)
+
+    def handle(self, *args, **options):
+        limit = options['limit'][0]
+
+        relay_waitlist_invitees = (
+            Invitations.objects.filter(active=False)
```
say-yawn, line 64:
```suggestion
            Invitations.objects.filter(active=True, date_redeemed=None)
```
I am assuming that this will be ran after `invite_monitor_waitlist` is ran? If so to get the Monitor waitlist the `active` is set to `True` as seen [here](https://github.com/mozilla/fx-private-relay/blob/master/privaterelay/management/commands/invite_monitor_waitlist.py#L104).
Users who were added to waitlist via Relay has `active=False` as seen [here](https://github.com/mozilla/fx-private-relay/blob/master/privaterelay/views.py#L162).

emails/utils.py
groovecoder: Hmm ... good point. I think we'll keep the increment in the `ses_send_email` function though.

This is an op metric we're not actually showing on our dashboard yet. The "Emails forwarded" number on the dashboard actually comes from adding up the sum of all `num_forwarded` records from `emails_relayaddress` and `emails_deletedaddress` tables.

When the wait-list invitations are sent thru SES, we should count them in the op metric.

privaterelay/management/commands/invite_relay_waitlist.py
groovecoder: I planned to run this independently of the monitor wait-list script. So, we will use the Monitor wait-list script to invite the rest of the monitor wait-list, and then use this script to invite the relay wait-list. So we don't want to get the Monitor wait-list in this script. Does that make sense?

groovecoder:
We could try to combine the scripts into one, but there was enough difference that I felt "safer" just keeping them separate. We'll throw these scripts away once we're done using them anyway.

PR #574, opened 2020-07-29
multipart's are just container parts; not errors

groovecoder:


emails/views.py
```diff
@@ -343,11 +343,13 @@ def _get_text_and_html_content(email_message):
     if email_message.is_multipart():
         for part in email_message.walk():
             try:
-                content = part.get_content()
+                # multipart/* are just containers
+                if part.get_content_maintype() == 'multipart':
+                    continue
```
say-yawn, line 348: I would remove this for now since the `multipart/*` would just get ignored in the `if` statements. I am also worried about the multipart having the attachment which will not get checked with this `continue`
```suggestion
```

groovecoder: Oh good point!

groovecoder:
Pushing straight to dev server for mail delivery spot-checks before merging.

groovecoder:
Spot-checks with and without attachments work.

PR #595, opened 2020-08-10
check that profile fetch result is OK before updating

groovecoder:
See https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/8296155/

### DIFF
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -247,6 +247,7 @@ def _get_event_keys_from_jwt(authentic_jwt):
 
 def _handle_fxa_profile_change(authentic_jwt, social_account, event_key):
     client = _get_oauth2_session(social_account)
+    # TODO: more graceful handling of profile fetch failures
     try:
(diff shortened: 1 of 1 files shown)


privaterelay/views.py
```diff
@@ -254,6 +254,15 @@ def _handle_fxa_profile_change(authentic_jwt, social_account, event_key):
         return HttpResponse('202 Accepted', status=202)
 
     extra_data = resp.json()
+
+    if resp.status_code != 200:
+        sentry_sdk.capture_message(
+            'FXA Profile fetch error.',
+            level="error",
+            stack=True
```
say-yawn, line 262: FYI this `stack=True` does not actually return the stack because this is a `capture_message`. If you want the stack it's best to use `logger.error` to capture the message or let the `extra_data['email']` throw an exception and capture the exception using `sentry_sdk.capture_exception` which sends stack trace by default.

groovecoder: Oh good point. Updated to catch and capture the exception and return the 202 to FXA to stop re-trying.

PR #631, opened 2020-08-14
Addon fixes

lesleyjanenorton:
- Updates button text on post-install page
- Changes sign in links to use `/accounts/profile` so that users who are already signed in aren't taken back to the oauth flow.
- Fixes the "Manage All Aliases" link in the modal
- Prevents input icons from getting re-enabled on browser restarts


groovecoder:
Looks good. I removed another last mention of the wait-list from the `add_input_icon.js` script.

PR #632, opened 2020-08-16
No need to change z-index

### LINKED ISSUE #245: High z-index value for email field wrapper causes stacking issues
The `relay-email-input-wrapper` element that gets added around email input fields has the following CSS z-index value:

```css
z-index: 99999999999999999;
```

I'm sure this property is likely needed for a reason, but its value is way too high. This results in the email field appearing visually stacked on top of pretty much any other element that may be on a website.

For example, sticky navigation:

image

humpfhumpf:
resolves #245


groovecoder:
Sorry this PR is outdated. We're looking at more ways to improve the button display.

PR #639, opened 2020-08-20
Support utf8 in headers 523

### LINKED ISSUE #523: problem with "umlauts" in sender's name
I have sent an e-mail from my Google Mail to my Yahoo Mail account via Firefox Relay. This is how Thunderbird showed the sender's name:

Bildschirmfoto 2020-06-29 um 23 37 06

The name should be "Sören" not "S?ren", it's probably a problem with all "umlauts" (ä, ö, ü - very common in German language).

The From mail header looks like:

`From: "\"Sï¿½ren Hentzschel\"  <email@example.org>  [via Relay]`

I compared with other e-mails that did not have this problem. In all other e-mails the From mail header looks like:

`From: =?UTF-8?Q?S=C3=B6ren_Hentzschel?=  <email@example.org>`

say-yawn:
# About this PR
On #523 a user pointed out that UTF-8 encoding is not properly handled on the From display name. This PR ensures that the utf-8 is used on the From display name.

# Acceptance Criteria
- [ ] From display name properly handles UTF-8 encoding
- [ ] Fix #523

# Practical Tests
## Check that umlauts are supported on From, Subject, and Body of the relayed email
1. Send an email to a relay address with umlauts on the From, Subject, and Body
2. Check that you get an email without the umlauts being broken (UTF-8 is properly encoded on the string)
3. Repeat this test with different email clients as the sender with umlauts on their username, in the email subject, and in the body:
- [ ] Yahoo Mail
- [ ] Gmail
- [ ] ProtonMail 
- [ ] AOL Mail
- [ ] Outlook
- [ ] Thunderbird

While testing, I would like information on the following headers:
```
Content-Type
Content-Transfer-Encoding
```
Often times you can get this information from a given email client by clicking for more details/see original/view headers on the email that was relayed.
For example, if you are an FXA user with ProtonMail as your primary email, then when you receive the email via Relay, when you open the email click on the "More" button. Then from the drop down of options click on the "View headers" and search for `Content-Type` and `Content-Transfer-Encoding`

### DIFF
--- a/emails/tests/utils_tests.py
+++ b/emails/tests/utils_tests.py
@@ -10,7 +10,8 @@ class FormattingToolsTest(TestCase):
     def test_generate_relay_From_with_umlaut(self):
         with self.settings(RELAY_FROM_ADDRESS='relay@relay.firefox.com'):
             relay_from_address, relay_from_display = generate_relay_From(
-                self.original_from_address)
+                self.original_from_address
+            )
 
         expected_encoded_display_name = (
             '=?utf-8?b?ImZvw7YgYsOkciIgPGZvb0BiYXIuY29tPiBbdmlhIFJlbGF5XQ==?='
--- a/emails/tests/utils_tests.py
+++ b/emails/tests/utils_tests.py
@@ -1,0 +1,19 @@
+from django.test import TestCase
+
+from emails.utils import generate_relay_From
+
+
+class FormattingToolsTest(TestCase):
+    def setUp(self):
+        self.original_from_address = '"foö bär" <foo@bar.com>'
+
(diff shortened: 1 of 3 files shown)


say-yawn:
@groovecoder as we previously discussed I removed the code for 8-bit Content Transfer Encoding support to only include changes for supporting UTF-8 on the display name. I also left instructions for QA-reproduce for this fix on the PR description above.

groovecoder:
Code changes LGTM. Is this branch on the dev server for testing?

say-yawn:
Awesome, @groovecoder not on dev server right now.

PR #664, opened 2020-09-09
Messaging for attachment support and limits 596

### LINKED ISSUE #596: Add messaging that Relay does not support emails > 150K
At least one user has lost an important attachment thru Relay already. They didn't know Relay stripped attachments until they got their first email.

We should add messaging up front so that more users know that Relay does not support attachments or emails larger than 150K, until we finish https://github.com/mozilla/fx-private-relay/issues/498.

say-yawn:
# About this PR
Update texts on forwarded email headers and FAQ page about attachment support and email size limit. This PR did not make changes to the existing hover text for "Forwarded" text on the dashboard as it was succinct enough to imply that the attachment may be supported. Current hover message reads:
Screen Shot 2020-09-09 at 3 41 21 PM

# Acceptance Criteria
- [ ] When attachment is delivered, we include a line in an header about supporting only up to 150KB email size
- [ ] FAQ has attachment support and relaying email only up to 150KB (bigger than this will get lost) message
- [ ] Fix #596

### DIFF
--- a/emails/templates/emails/wrapped_email.html
+++ b/emails/templates/emails/wrapped_email.html
@@ -39,6 +39,9 @@
               <td width="100%" align="center" style="line-height: 150%;">
                 <p style="color: #363959; font-size: 13px; font-family: sans-serif; margin-top: 0; margin-bottom: 0;">
                   This message was forwarded from <span style="font-family: sans-serif; font-weight: bolder; color: #20123a; text-decoration: none; font-size: 13px;">{{ display_email|safe }}</span> by <a href="{{ SITE_ORIGIN }}" target="_blank" style="font-family: sans-serif; color: #20123a; text-decoration: underline; font-weight: bolder; font-size: 13px;">Firefox Relay</a>.
+                  {% if has_attachment %}
+                    <br>Emails with a maximum size of <b>150KB</b> have attachments support. To learn more check our <a href="{{ faq_page }}" target="_blank" style="font-family: sans-serif; color: #20123a; text-decoration: underline; font-weight: bolder; font-size: 13px;">FAQ page</a>.
+                  {% endif %}
                 </p>
               </td>
             </tr>
--- a/emails/templates/emails/wrapped_email.html
+++ b/emails/templates/emails/wrapped_email.html
@@ -40,7 +40,7 @@
                 <p style="color: #363959; font-size: 13px; font-family: sans-serif; margin-top: 0; margin-bottom: 0;">
                   This message was forwarded from <span style="font-family: sans-serif; font-weight: bolder; color: #20123a; text-decoration: none; font-size: 13px;">{{ display_email|safe }}</span> by <a href="{{ SITE_ORIGIN }}" target="_blank" style="font-family: sans-serif; color: #20123a; text-decoration: underline; font-weight: bolder; font-size: 13px;">Firefox Relay</a>.
                   {% if has_attachment %}
-                    <br>Emails with a maximum size of <b>150KB</b> have attachments support. To learn more check our <a href="{{ faq_page }}" target="_blank" style="font-family: sans-serif; color: #20123a; text-decoration: underline; font-weight: bolder; font-size: 13px;">FAQ page</a>.
+                    <br>Firefox Relay supports email forwarding (including attachments) of email up to 150KB in size. To learn more check our <a href="{{ faq_page }}" target="_blank" style="font-family: sans-serif; color: #20123a; text-decoration: underline; font-weight: bolder; font-size: 13px;">FAQ page</a>.
                   {% endif %}
                 </p>
               </td>
--- a/emails/views.py
+++ b/emails/views.py
@@ -315,6 +315,8 @@ def _sns_message(message_json):
             'email_to': to_address,
             'display_email': display_email,
             'SITE_ORIGIN': settings.SITE_ORIGIN,
+            'has_attachment': bool(attachments),
+            'faq_page': settings.SITE_ORIGIN + reverse('faq')
         })
         message_body['Html'] = {'Charset': 'UTF-8', 'Data': wrapped_html}
 
--- a/emails/views.py
+++ b/emails/views.py
@@ -17,6 +17,7 @@ from django.core.exceptions import PermissionDenied
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import redirect
 from django.template.loader import render_to_string
+from django.urls import reverse
 from django.views.decorators.csrf import csrf_exempt
 
 from .context_processors import relay_from_domain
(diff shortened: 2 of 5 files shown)


emails/views.py
```diff
@@ -314,19 +315,24 @@ def _sns_message(message_json):
...
         relay_header_text = (
             'This email was sent to your alias '
             '{relay_address}. To stop receiving emails sent to this alias, '
             'update the forwarding settings in your dashboard.\n'
             '{extra_msg}---Begin Email---\n'
         ).format(
-            relay_address=display_email, extra_msg=attachment_not_supported
+            relay_address=display_email, extra_msg=attachment_msg
         )
         wrapped_text = relay_header_text + text_content
```
groovecoder, line 337: It just occurred to me that we could use django templates for the plain text email too. I.e., we could have `emails/templates/emails/wrapped_email.txt`, render it with Django, and send it in `message_body['Text'] ?

groovecoder (approved):
Code LGTM. @lesleyjanenorton and/or @mcroud to also check this out.

emails/views.py
mcroud: Forgive me, I'm not entirely sure what I am checking here. One query I have given the tooltip text, when we say " Attachments and emails larger than 150KB are not currently supported and will not be forwarded." Do we mean either:
- Attachments are not supported
or
- Attachments larger than 150KB are not supported

say-yawn: @mcroud no worries and great questions. What we wanted to say was that any emails larger than 150KB will not be forwarded to the user due to the limit.

mcroud:
Hi @say-yawn, Looking at the original issue again, I do think a more "upfront" method of communicating this limit is warranted in addition to what we have, an upfront message will only be useful to certain people however and possibly confusing to others.
In the future it may be desirable to signal that an email was dropped as in most instances it is perhaps out of the recipients hands whether they are to receive a large email or not and whether they were expecting it, maybe a "dropped" or "undelivered" list (future thoughts!).

Do you have an example of what the new header looks like? And the new header with the 150kb message is present on every email regardless of whether there was an attachment or not? 

Some quick thoughts. I first thought about a closable (with a top right "X" button) notification message at the top of the page. This will allow people to acknowledge the information and then hide it. I altered the text to one of the info panels on Monitor's website as a super rough example:



Thinking more of general audiences however, A notification like this may draw too much attention to information they couldn't act on or understand the significance. How big is your average email? How big is a kilobyte? etc.

So perhaps we sample something that is both present but not too alarmist, I quickly cooked up the below to capture my initial thoughts, I can't stress enough this isn't an official design suggestion :) 



I think a message like this will still provoke either questions or a "I have no idea what that means" thought. I feel the main problem is being unaware of when a failed delivery occurs. A visible message like this is probably more helpful than information a person needs to find or click to reveal though.

mcroud:
Hi @say-yawn, how do you feel about the last image I shared showing the line of text near the footer? 
Is there an explanation to this we can link to with a "Learn more" link? 

@groovecoder  I would be curious of your thoughts on the footer messaging in the quick visual above (the whole dashboard screenshot with the footer text). My post outlines my thought stream but very keen on your opinion!

emails/templates/emails/wrapped_email.html
```diff
@@ -39,6 +39,9 @@
               <td width="100%" align="center" style="line-height: 150%;">
                 <p style="color: #363959; font-size: 13px; font-family: sans-serif; margin-top: 0; margin-bottom: 0;">
                   This message was forwarded from <span style="font-family: sans-serif; font-weight: bolder; color: #20123a; text-decoration: none; font-size: 13px;">{{ display_email|safe }}</span> by <a href="{{ SITE_ORIGIN }}" target="_blank" style="font-family: sans-serif; color: #20123a; text-decoration: underline; font-weight: bolder; font-size: 13px;">Firefox Relay</a>.
+                  {% if has_attachment %}
+                    <br>Emails with a maximum size of <b>150KB</b> have attachments support. To learn more check our <a href="{{ faq_page }}" target="_blank" style="font-family: sans-serif; color: #20123a; text-decoration: underline; font-weight: bolder; font-size: 13px;">FAQ page</a>.
```
groovecoder, line 43: Suggest changing to match what's in the dashboard footer (and removing extra `<br>` tag ...
```suggestion
                    Firefox Relay supports email forwarding (including attachments) of email up to 150KB in size. To learn more check our <a href="{{ faq_page }}" target="_blank" style="font-family: sans-serif; color: #20123a; text-decoration: underline; font-weight: bolder; font-size: 13px;">FAQ page</a>.
```

emails/views.py
```diff
@@ -314,19 +315,24 @@ def _sns_message(message_json):
...
+            'faq_page': settings.SITE_ORIGIN + reverse('faq')
         })
         message_body['Html'] = {'Charset': 'UTF-8', 'Data': wrapped_html}
 
     if text_content:
         incr_if_enabled('email_with_text_content', 1)
-        attachment_not_supported = ''
+        attachment_msg = (
+            'Emails with a maximum size (including headers) of 150KB '
+            'have attachments support. To learn more visit {site}{faq}\n'
```
groovecoder, line 327: Suggest changing to match what's in dashboard footer.

```suggestion
            'Firefox Relay supports email forwarding (including attachments) '
            'of email up to 150KB in size. To learn more visit {site}{faq}\n'
```

groovecoder (changes requested):

mcroud:
Having discussed with Gloria about research, I will be taking a selection of the prototypes created and redeveloping them to be more resilient demos for user testing. A testing plan is in the works.

emails/templates/emails/wrapped_email.html
say-yawn: The `<br>` was to ensure that this statement was in a new line as we did [previously](https://github.com/mozilla/fx-private-relay/pull/613/files#diff-8d6893341f0957c36ae14bd59cd9825dL42-L44). Is there a better way to handle this?

groovecoder: Ah, good call. Nvm me.

groovecoder (approved):

groovecoder:
Thanks @mcroud. In the mean-time, we'll merge these first changes to help people who are already asking about undelivered mail.

mcroud:
Three prototypes have been produced for usability testing which include the following variants:
1. Message above footer
2. Stats bar
3. Info icon

Will circle back with the research team to facilitate usability testing with these prototypes.

malqinneh:
@mcroud are next steps related to the conversation scheduled for Friday with Gloria?

mcroud:
@malqinneh Following on from yesterday's Relay Triage meeting, the team will implement the "Footnote" proposal immediately so that the issue is addressed. Once usability testing is complete we may change the way we present the message based on participant feedback and test results. From a research perspective, yes, more info will come on Friday.

groovecoder (approved):

PR #668, opened 2020-09-11
Add ESLint

pdehaan:


lesleyjanenorton:
When I run this locally, it cycles through all of the Django library files as well as the Relay-specific files so the output is huge. Can we limit this to just the files generated by us, or am I just doing something incorrectly?

.eslintignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+static/js/clipboard.min.js
+venv
```
lesleyjanenorton, line 3: Let's add "env" to this list

groovecoder:
Superseding with https://github.com/mozilla/fx-private-relay/pull/716 which adds `env` directory too.

PR #669, opened 2020-09-11
Add stylelint

pdehaan:


static/css/style.css
```diff
@@ -683,10 +682,6 @@ table {
   margin: auto 0 auto auto;
 }
 
-.relay-stat {
-  flex-direction: column-reverse;
```
pdehaan, line 687: See L1670 below.

static/css/style.css
```diff
@@ -1383,10 +1379,6 @@ input.input-has-error {
   opacity: 0.5;
 }
 
-.num-remaining-aliases {
-  margin-right: 4px;
```
pdehaan, line 1387: See L755 above.

groovecoder:
Needs a conflict resolution but then we could merge this?

Issue #683, opened 2020-09-21
Mail delivery problem | 'From' header not respecting RFC5322


groovecoder:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1665441#c0 ...

---

User Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:80.0) Gecko/20100101 Firefox/80.0

Steps to reproduce:

Sent an email to a relay address. For example sent an email from testNDM1ZmE2@gmail.com to ugfv8obfd@relay.firefox.com, that should be forwarded to cyril.schreven@protonmail.com

Actual results:

The mails sent by Firefox relay was not properly delivered.
The reason for this is that the headers were not successfully parsed at reception. More specifically the 'From' header does not contain the quotes as defined by the rfc:
"""
From: john NDM1ZmE2 <testNDM1ZmE2@gmail.com> [via Relay] <noreply@relay.firefox.com>
"""

Expected results:

The 'From' header should include the quotes and be:
"""
From: "john NDM1ZmE2 <testNDM1ZmE2@gmail.com> [via Relay]" <noreply@relay.firefox.com>
"""
The correct format is specified in rfc5322, see for example Appendix A.1.2

pdehaan:
I wonder if this is related to #678 (seemingly all ProtonMail users as well). Possibly related to recent utf-8 From handling work?

schreven:
Thanks for moving it here. The issue likely started the 11th of september.

schreven:
Looked further into it and found this is discussed more precisely at <https://tools.ietf.org/html/rfc2047#section-5>.
The From header sent by the relay always contains `[` `]` `<` and `>` in its display name. For this reason I believe it should indeed be a quoted-string, before it gets encoded.

Before the support-utf8-in-headers commit, the quotes would always get added by [this piece of code](https://github.com/python/cpython/blob/3.8/Lib/email/headerregistry.py#L96).

I wrote a commit on a fork that might be along the line of what you want to do, [here](https://github.com/schreven/fx-private-relay/commit/4ef190d32ae9e10f014884894d6278e8bc9a2fa5).

groovecoder:
Your patch seems like a good one. Would you like to submit a pull request for it?

schreven:
All right I submitted one

say-yawn:
Related to #685

groovecoder:
This should be fixed now!

PR #688, opened 2020-09-23
Set display-name of From header to quoted-string

schreven:
Follow up on: #683 

> Looked further into it and found this is discussed more precisely at <https://tools.ietf.org/html/rfc2047#section-5>.
> The From header sent by the relay always contains `[` `]` `<` and `>` in its display name. For this reason I believe it should indeed be a quoted-string, before it gets encoded.
> 
> Before the support-utf8-in-headers commit, the quotes would always get added by [this piece of code](https://github.com/python/cpython/blob/3.8/Lib/email/headerregistry.py#L96).

This MR offers to always transform the display name of the From header into a quoted-string before being word-encoded.

### DIFF
--- a/emails/tests/utils_tests.py
+++ b/emails/tests/utils_tests.py
@@ -14,7 +14,7 @@ class FormattingToolsTest(TestCase):
             )
 
         expected_encoded_display_name = (
-            '=?utf-8?b?ImZvw7YgYsOkciIgPGZvb0BiYXIuY29tPiBbdmlhIFJlbGF5XQ==?='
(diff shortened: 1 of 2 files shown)


groovecoder:
I couldn't figure out exactly which `quote()` function your code was using. The most popular one in python is from `urllib.parse`, but that created all kinds of `%20` chars in the `From`. Turns out we just needed to restore the `"` characters at the beginning and end of the string value in the `Header`. I submitted https://github.com/mozilla/fx-private-relay/pull/689 based on this commit. Closing this PR.

schreven:
Sorry for the mess up. The `quote()` I intended was [this one here.](https://github.com/python/cpython/blob/3.8/Lib/email/_parseaddr.py#L193)

You probably still need its functionality: avoid double quotes inside the quoted-string.

PR #689, opened 2020-09-24
fix ProtonMail delivery

groovecoder:
Turns out we just needed to restore `"` around the whole `From` value. 

This branch is already on [the dev site](https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/), if you want to test it out there.

### DIFF
--- a/emails/tests/utils_tests.py
+++ b/emails/tests/utils_tests.py
@@ -14,7 +14,7 @@ class FormattingToolsTest(TestCase):
             )
 
         expected_encoded_display_name = (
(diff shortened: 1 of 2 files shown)


say-yawn (commented):
LGTM and testing on dev.

say-yawn:
Seems like there is a weird behavior from Protonmail to strip out the email address inside the `<>` brackets. But the email is received as expected.

Protonmail:
Screen Shot 2020-09-24 at 3 48 53 PM
Gmail (expected email address of the original sender in `<>`):
Screen Shot 2020-09-24 at 3 48 42 PM

say-yawn (approved):

Issue #655, opened 2020-08-26
Support 8-bit Content-Transfer-Encoding


say-yawn:
# About this Issue
Related to this add-on [review](https://addons.mozilla.org/en-US/firefox/addon/private-relay/reviews/1585861/), when we get an email with 8-bit as the Content Transfer Encoding the `message_from_string` does not properly handle the encoding. This may be related to bug filed on Python [here](https://bugs.python.org/issue41307). 

# STR
## Check that email **relayed** from https://nextgame.net/ contains broken encoding on the email body
1. Go to https://nextgame.net/ and create an account with a Relay address(create account [link](https://nextgame.net/auth/registration/?register=yes&backurl=/auth/))
2. Check that you get an email relayed via Relay with broken body that looks like:
```
\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd \ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd \ufffd\ufffd\ufffd\ufffd\ufffd NextGame
------------------------------------------

\ufffd\ufffd \ufffd\ufffd\ufffd\ufffd\ufffd nextgame.net \ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd \ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd \ufffd\ufffd\ufffd\ufffd\ufffd \ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd.

\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd \ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd:
ID \ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd: 311456

\ufffd\ufffd\ufffd: \ufffd\ufffd\ufffd
\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd: \ufffd\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd
E-Mail: <your relay address>
[...]

PR #709, opened 2020-10-12
Support 8 bit content transfer encoding

say-yawn:
# About this PR
Support 8-bit Content Transfer Encoding by using `message_from_bytes` instead of `message_from_string`.

# Acceptance Criteria
- [ ] Email sent with 8-bit Content Transfer Encoding(CTE) no longer has broken body message
- [ ] Email sent with 8-bit CTE is converted to base64
- [ ] Follow-up issue created to document that emails coming in may not have UTF-8 as Content-Type's charset which can cause breakage


emails/views.py
```diff
@@ -299,8 +299,33 @@ def _sns_message(message_json):
...
+        bytes_email_message
+    )
+    strip_texts = []
+    for item in mail['headers']:
+        for k, v in item.items():
+            strip_texts.append(': '.join([k, v]))
+    stripped_content = message_json['content']
+    for item in strip_texts:
+        stipped_content = stripped_content.replace(item, '')
+    logger.error('body_message', extra={
```
groovecoder, line 317: Still need this `logger.error` line?

emails/views.py
```diff
@@ -310,7 +335,7 @@ def _sns_message(message_json):
     if html_content:
```
groovecoder, line 335:
```suggestion
    if b_html_content:
```

emails/views.py
```diff
@@ -299,8 +299,33 @@ def _sns_message(message_json):
     email_message = message_from_string(
         message_json['content'], policy=policy.default
     )
+    bytes_email_message = message_from_bytes(
+        message_json['content'].encode('utf-8'), policy=policy.default
+    )
 
     text_content, html_content, attachments = _get_all_contents(email_message)
```
groovecoder, line 306: Can we remove this now if we're only using the `b_` variables?

groovecoder (changes requested):

groovecoder:
Code and spot-checks on dev look good.

groovecoder (approved):

sv-hyacoub:
We followed the steps from [#655](https://github.com/mozilla/fx-private-relay/issues/655) to verify the first test case of this PR, but the body of the email is still broken in every email service provider and the header of the email is displayed as following:
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: base64

For the second test case we verified that special characters are not correctly displayed in email body section when sending emails from thunderbird to every email service provider via Firefox Relay. We confirm that the email body is displayed correctly now but the header of the email is displayed as following:
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: quoted-printable

@say-yawn could you please take a look at out testing results and let's know your thoughts?

PR #712, opened 2020-10-16
[WIP] Fix collectstatic

birdsarah:
This isn't finished as I'm not sure what we want to do in the case of a non 200 error code as I don't have enough context, but I believe this solves the problem which I was also able to reproduce locally (presumably because I don't have a full setup yet).

### DIFF
--- a/privaterelay/apps.py
+++ b/privaterelay/apps.py
@@ -19,5 +19,3 @@ class PrivateRelayConfig(AppConfig):
         if resp.status_code == 200:
             resp_json = resp.json()
             self.fxa_verifying_keys = resp_json['keys']
-        else:
-            pass  # what's the right thing to do here?
--- a/privaterelay/apps.py
+++ b/privaterelay/apps.py
@@ -16,5 +16,8 @@ class PrivateRelayConfig(AppConfig):
             '%s/jwks' %
             settings.SOCIALACCOUNT_PROVIDERS['fxa']['OAUTH_ENDPOINT']
         )
-        resp_json = resp.json()
-        self.fxa_verifying_keys = resp_json['keys']
+        if resp.status_code == 200:
+            resp_json = resp.json()
+            self.fxa_verifying_keys = resp_json['keys']
+        else:
+            pass  # what's the right thing to do here?
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -168,6 +168,8 @@ def _authenticate_fxa_jwt(req_jwt):
 
 
 def _verify_jwt_with_fxa_key(req_jwt, private_relay_config):
+    if not private_relay_config.fxa_verifying_keys:
+        raise Exception("FXA verifying keys are not available.")
     social_app = SocialApp.objects.get(provider='fxa')
     for verifying_key in private_relay_config.fxa_verifying_keys:
         if verifying_key['alg'] == 'RS256':


privaterelay/apps.py
```diff
@@ -16,5 +16,8 @@ def ready(self):
             '%s/jwks' %
             settings.SOCIALACCOUNT_PROVIDERS['fxa']['OAUTH_ENDPOINT']
         )
-        resp_json = resp.json()
-        self.fxa_verifying_keys = resp_json['keys']
+        if resp.status_code == 200:
+            resp_json = resp.json()
+            self.fxa_verifying_keys = resp_json['keys']
+        else:
+            pass  # what's the right thing to do here?
```
groovecoder, line 23:
```suggestion
        if resp.status_code == 200:
            resp_json = resp.json()
            self.fxa_verifying_keys = resp_json['keys']
```
If we let this code finish and leave `fxa_verifying_keys` as `[]`, then we'll throw `Could not authenticate JWT with FXA key.` exceptions later when we try to use the `fxa_verifying_keys` value that doesn't exist. That's probably the right behavior ...

1. The exceptions will be logged into Sentry, 
2. FXA will retry the callbacks until the callback is successful, 
3. The code in `_authenticate_fxa_jwt` already tries to call this `ready` code again, so:
4. When the FXA verifying keys are available again (because FXA fixes the server), this code will populate `fxa_verifying_keys`,
5. The callback will process successfully,
6. FXA will stop retrying, and finally:
7. We'll have a nice fat spike of FXA callback errors in Sentry so we know about it.


But, the best way to tighten this up is probably to add a check in `_verify_jwt_with_fxa_key` ... ?

```python
  def _verify_jwt_with_fxa_key(req_jwt, private_relay_config):
      if private_relay_config.fxa_verifying_keys == None:
        raise Exception("FXA verifying keys are unavailable.")
      ...
```

groovecoder (changes requested):

birdsarah:
```
    The exceptions will be logged into Sentry,
    FXA will retry the callbacks until the callback is successful,
    The code in _authenticate_fxa_jwt already tries to call this ready code again, so:
    When the FXA verifying keys are available again (because FXA fixes the server), this code will populate fxa_verifying_keys,
    The callback will process successfully,
    FXA will stop retrying, and finally:
    We'll have a nice fat spike of FXA callback errors in Sentry so we know about it.
```

This all makes sense. Is there a need to serve a helpful error message back in some way? Maybe that's handled elsewhere I haven't reviewed the code extensively.

groovecoder:
In this case, the error message shows up in our logs + Sentry interface, and is returned back to FXA calling code (so probably shows up in their logs + Sentry interface too). In both of those places, I think "FXA verifying keys are unavailable." is probably descriptive enough.

birdsarah:
Your suggestion:

```
  def _verify_jwt_with_fxa_key(req_jwt, private_relay_config):
      if private_relay_config.fxa_verifying_keys == None:
        raise Exception("FXA verifying keys are unavailable.")
```

seems great. It can cover a variety of settings where we don't have the `verifying_keys` as opposed to just this json request

Python nits:
- `is None` is more idiomatic
- should it be checking for an empty list?

Can you give me more context about the verifying keys and handling the checks in App setup rather than at runtime. Please explain to me like I don't know, cos I don't. I'm just thinking through where you have multiple instances running as we run a large web service and then some don't spin up because a check fails, but some are already running. What does this mean? What are the implications?

groovecoder:
We load the FXA keys into the app at startup so we don't have to make an additional HTTP request to FXA every time we need them. It's an optimization, though maybe not a big one - we only use these keys when FXA sends us a callback request, and that only happens when a Relay user changes something in their FXA profile.

The code here that handles the FXA callback request *SHOULD* try to fetch new keys once - if it wasn't able to load the keys at startup, or if the signature verification fails because FXA has rotated their signing key. If the app is still missing keys or the signature doesn't verify a second time, then the app will throw the exception.

If 2/5 of the app instances were missing keys, they would throw the "keys are unavailable" errors back up/at FXA, which would retry its request. The FXA retry request should be routed back thru the load-balancer, where it could reach one of the 3/5 of the app instances that DO have the keys to verify the request, which would succeed and stop the FXA retries.

So basically, the current code design is trying to be robust AND performant:

1. Load FXA key once at app startup, to eliminate an HTTP fetch later
2. If the key is unavailable at app startup, don't kill the whole app (this PR)
3. If the key is unavailable when it's needed, try to fetch it in real-time
4. If the key is still unavailable, throw the error
5. If the key is out-dated, try to fetch a new one in real-time
6. If the key is still out-dated, throw the error

groovecoder:
Added commit to check `if not private_relay_config.fxa_verifying_keys` when we need them.

groovecoder (approved):

PR #715, opened 2020-10-21
Lock user profile when alias create

say-yawn:
# About this PR
Prevent race condition where the limit is not respected when creating a new alias.

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -81,7 +81,7 @@ def _index_POST(request):
         locked_profile = Profile.objects.select_for_update().get(
             user=user_profile.user
         )
-        if profile.num_active_address >= settings.MAX_NUM_BETA_ALIASES:
+        if locked_profile.num_active_address >= settings.MAX_NUM_BETA_ALIASES:
             if 'moz-extension' in request.headers.get('Origin', ''):
                 return HttpResponse('Payment Required', status=402)
             messages.error(
--- a/emails/views.py
+++ b/emails/views.py
@@ -14,6 +14,7 @@ from markus.utils import generate_tag
 from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import PermissionDenied
+from django.db import transaction
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import redirect
 from django.template.loader import render_to_string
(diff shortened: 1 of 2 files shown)


emails/views.py
```diff
@@ -76,15 +77,20 @@ def _index_POST(request):
 
     incr_if_enabled('emails_index_post', 1)
     existing_addresses = RelayAddress.objects.filter(user=user_profile.user)
```
groovecoder, line 79: Don't need this line or query anymore.
```suggestion
```

emails/views.py
```diff
@@ -76,15 +77,20 @@ def _index_POST(request):
...
-    if existing_addresses.count() >= settings.MAX_NUM_BETA_ALIASES:
-        if 'moz-extension' in request.headers.get('Origin', ''):
-            return HttpResponse('Payment Required', status=402)
-        messages.error(
-            request, "You already have 5 email addresses. Please upgrade."
+
+    with transaction.atomic():
+        profile = Profile.objects.select_for_update().get(
+            user=user_profile.user
         )
```
groovecoder, line 84: The user profile is already available from line 72's assignment of `user_profile`.
```suggestion
```

emails/views.py
```diff
@@ -76,15 +77,20 @@ def _index_POST(request):
...
-        return redirect('profile')
+        if profile.num_active_address >= settings.MAX_NUM_BETA_ALIASES:
```
groovecoder, line 85:
```suggestion
        if user_profile.num_active_address >= settings.MAX_NUM_BETA_ALIASES:
```

emails/views.py
```diff
@@ -76,15 +77,20 @@ def _index_POST(request):
...
+            if 'moz-extension' in request.headers.get('Origin', ''):
+                return HttpResponse('Payment Required', status=402)
+            messages.error(
+                request, "You already have 5 email addresses. Please upgrade."
+            )
+            return redirect('profile')
+        relay_address = RelayAddress.make_relay_address(profile.user)
```
groovecoder, line 92:
```suggestion
        relay_address = RelayAddress.make_relay_address(user_profile.user)
```

groovecoder (changes requested):

groovecoder:
Ah, I realize your `profile` var is from `select_for_update`, so my suggestions might not be good. But maybe we can re-name that variable to `updating_profile` or something to make it more clear and explicit?

groovecoder:
Spot-checked the attack against this branch on both my local and the dev server and this mitigation takes care of it. So my `user_profile` suggestion is just that - doesn't need to block. (We *should* delete the un-used `existing_addresses` line though)

groovecoder (approved):

PR #720, opened 2020-10-21
also mkdir the CACHE directory

groovecoder:


Micheletto (approved):
r+

Dockerfile
```diff
@@ -12,12 +12,13 @@ EXPOSE 8000
 
 USER app
 
+COPY ./requirements.txt /app/requirements.txt
+RUN pip install -r requirements.txt
+
```
groovecoder, line 17: This helps `docker build` cache the results of the install to speed up builds.

Dockerfile
```diff
@@ -12,12 +12,13 @@ EXPOSE 8000
...
 COPY . /app
 COPY .env-dist /app/.env
 
-RUN pip install -r requirements.txt
-
-RUN mkdir -p /app/static
+RUN mkdir -p /app/static/CACHE
```
groovecoder, line 21: Hmm ... we may not need this at all, if we `COPY . /app`, then `static/CACHE` should now copy over.

Micheletto (approved):
r+

PR #724, opened 2020-10-26
Support Chrome, Opera, and Edge

groovecoder:
Uses `browser-polyfill.js` and generic `-extension://` urls to support all browsers that support Web Extensions.

### DIFF
--- a/extension/js/browser-polyfill.js
+++ b/extension/js/browser-polyfill.js
@@ -1,0 +1,1224 @@
+(function (global, factory) {
+  if (typeof define === "function" && define.amd) {
+    define("webextension-polyfill", ["module"], factory);
+  } else if (typeof exports !== "undefined") {
+    factory(module);
+  } else {
+    var mod = {
+      exports: {}
+    };
+    factory(mod);
+    global.browser = mod.exports;
+  }
+})(typeof globalThis !== "undefined" ? globalThis : typeof self !== "undefined" ? self : this, function (module) {
+  /* webextension-polyfill - v0.6.0 - Mon Dec 23 2019 12:32:53 */
+
+  /* -*- Mode: indent-tabs-mode: nil; js-indent-level: 2 -*- */
+
+  /* vim: set sts=2 sw=2 et tw=80: */
+
+  /* This Source Code Form is subject to the terms of the Mozilla Public
+   * License, v. 2.0. If a copy of the MPL was not distributed with this
+   * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+  "use strict";
+
+  if (typeof browser === "undefined" || Object.getPrototypeOf(browser) !== Object.prototype) {
+    const CHROME_SEND_MESSAGE_CALLBACK_NO_RESPONSE_MESSAGE = "The message port closed before a response was received.";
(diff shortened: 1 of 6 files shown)


extension/popup.html
```diff
@@ -5,7 +5,9 @@
     <title>Firefox Relay</title>
     <link rel="stylesheet" href="css/relay.css">
     <link rel="stylesheet" href="css/popup.css">
-    <script src="/js/data-opt-out-toggle.js"></script>
+    <script src="js/browser-polyfill.js"></script>
+    <script src="js/data-opt-out-toggle.js"></script>
+    <script src="js/add_input_icon.js"></script>
```
pdehaan, line 10: Is this script included twice now (here and L11 below)?

extension/js/browser-polyfill.js
```diff
@@ -0,0 +1,1224 @@
+(function (global, factory) {
```
pdehaan, line 1: Where did this file come from? It looks generated.
Wondering if we can keep it in sync from its upstream source (if this is in npm or somewhere more definitive).

groovecoder: Yeah, it's from [the npm `webextension-polyfill` module](https://www.npmjs.com/package/webextension-polyfill). AFAIK, npm `postinstall` scripts have to be included with the module itself? Is there a way to have our own custom post-install script run that will copy the `browser-polyfill.js` file from `node_modules/` into the `src/` dir?

pdehaan: Nope, we can add our own "postinstall" npm script in the extension/package.json file that would automatically copy over the ./node_module/'ed file into ./extension/js/browser-polyfill.js (and then we can add that file to .gitignore).

Alternatively, we can add some "prebuild" hook or tweak how the extension does <kbd>npm run build</kbd> so the file gets copied over from ./node_modules/ at build time.

pdehaan: In fact, here's what that might look like:

```js
  "dependencies": {
    "copy": "^0.3.2",
    "web-ext": "^5.3.0",
    "webextension-polyfill": "^0.6.0"
  },
  "scripts": {
    "build:clean": "...",
    "postinstall": "copy './node_modules/webextension-polyfill/dist/browser-polyfill.min.*' ./js/",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
```

I use the npm **copy** module to copy the 2 minified polyfill assets (JavaScript and linked sourcemap) from ./node_modules/webextension-polyfill/dist/ to the ./extension/js/ folder after you run <kbd>npm install</kbd>.
We could also add it to the `build:clean` script, but that might not trigger if you run <kbd>npm run build</kbd> directly, and sometimes <kbd>npm run build:clean</kbd> isn't run until after the add-on is built, so the timing can be a bit tricky.

pdehaan: Also looks like we can shave about 26-30KB of bandwidth by using the .min.js versions instead of the non-minified counterparts:

```sh
ls -lh ./node_modules/webextension-polyfill/dist 

-rw-r--r--  1 pdehaan  staff    36K 26 Oct  1985 browser-polyfill.js
-rw-r--r--  1 pdehaan  staff    56K 26 Oct  1985 browser-polyfill.js.map
-rw-r--r--  1 pdehaan  staff   9.9K 26 Oct  1985 browser-polyfill.min.js
-rw-r--r--  1 pdehaan  staff    53K 26 Oct  1985 browser-polyfill.min.js.map
```

PR #735, opened 2020-11-20
bump version number to 1.4.6

groovecoder:
Bumping the version number so we can make a new release to AMO and get our label changed from "Experimental" to "By Firefox"


pdehaan (approved):
R+.
Can confirm that 1.4.6 is greater than 1.4.5.

Issue #104, opened 2020-04-03
Exclude non-extension files during packing


wagnerand:
While packing the add-on files, we might want to exclude some files which aren't necessary for the extension to work, for example:

```
web-ext-artifacts/
package-lock.json
package.json
README.html
README.md
sequence-diagram.mmd
sequence-diagram.svg
```

wagnerand:
Also `.DS_Store` and other macOS-specific files.

groovecoder:
This has apparently regressed. Re-opening ...

pdehaan:
I'm confused, how did we build the add-on?
I ran <kbd>npm run build</kbd> from the ./extension/ dir and don't see any unexpected files:

```sh
npm run build

> fx-private-relay@1.4.6 build /Volumes/Dev/github/mozilla/fx-private-relay/extension
> web-ext build --ignore-files .DS_Store README.* package.json package-lock.json sequence-diagram.* --overwrite-dest -s . -a ../static/downloads/addon/latest/ && mv ../static/downloads/addon/latest/firefox_relay-*.zip ../static/downloads/addon/latest/firefox_relay.zip

Applying config file: ./package.json
Building web extension from /Volumes/Dev/github/mozilla/fx-private-relay/extension
Destination exists, overwriting: ../static/downloads/addon/latest/firefox_relay-1.4.6.zip
Your web extension is ready: ../static/downloads/addon/latest/firefox_relay-1.4.6.zip


zipinfo -1 ../static/downloads/addon/latest/firefox_relay.zip | sort | grep -Ev "/$"
css/first-run.css
css/popup.css
css/relay.css
first-run.html
fonts/Inter/Inter-UI-Bold.woff2
fonts/Inter/Inter-UI-Medium.woff2
fonts/Inter/Inter-UI-Regular.woff2
fonts/Metropolis/Metropolis-Bold.woff2
fonts/Metropolis/Metropolis-Light.woff2
fonts/Metropolis/Metropolis-Medium.woff2
icons/border-48.png
icons/copy-to-clipboard.svg
icons/make-new-alias.png
icons/open-new-tab.svg
icons/placeholder-logo-beta.png
icons/x-close.svg
images/arrowhead-left.svg
images/email-tout-bg.svg
images/first-run-export.svg
images/loader.svg
images/logomark.svg
images/logotype.svg
images/panel-images/high-five.svg
images/panel-images/tip1-icon.svg
images/panel-images/tip2-icon.svg
[...]

PR #736, opened 2020-11-20
bump version to 1.4.7 to re-submit to AMO

groovecoder:


groovecoder:
I had to re-submit because I forgot to run `npm run build:prod`, so I accidentally submitted 1.4.6 with `127.0.0.1:8000` instead of `relay.firefox.com`.

pdehaan:
[AMO already has a 1.4.7](https://addons.mozilla.org/en-US/firefox/addon/private-relay/versions/), I'm confused.
Unless you built locally and pushed to AMO before this PR, in which case makes total sense.

I'd also like to dig into what's the deal with https://github.com/mozilla/fx-private-relay/issues/104 before we do too many more pushes to AMO if we think we're bundling a bunch of dead files.
In fact, we might want to retool the build process a bit to remove some duplication and maybe build for prod by default to avoid future footguns.

pdehaan (approved):
1.4.7 is indeed higher than 1.4.6, so the numbers are trending correctly.

R+ 👍  :shipit: 🚢  🚀

groovecoder:
I did build 1.4.7 locally and publish to AMO before merging this - so I could catch TheOne before they left for the weekend.

In my hurry to build 1.4.7 with the correct domain, I bad-pasted the commands from `npm run build:prod` into my CLI. In doing that, I accidentally changed `--ignore-files` to `--      ignore-files` which is why 1.4.7 got the hidden files this time.

Next time, `npm run build:prod` should correctly ignore the files.

PR #737, opened 2020-11-20
Move web-ext settings into config file

pdehaan:
Moves a bunch of duplicated <kbd>web-ext</kbd> configs into a separate config file to avoid duplicating in 3-4 places.
If you object to yet another new file, I think we can add it as a nested key in the extension/package.json.

Can't think of a better way yet to avoid using `sed` for string substitutions, unless we use Nunjucks/Liquid/lodash for simple templates (and then a combination of gitignore+web-ext ignoreFiles).

### DIFF
--- a/extension/package.json
+++ b/extension/package.json
@@ -8,7 +8,7 @@
   },
   "devDependencies": {},
   "scripts": {
-    "build": "web-ext build && mv ../static/downloads/addon/latest/firefox_relay-*.zip ../static/downloads/addon/latest/firefox_relay.zip && npm run build:clean",
+    "build": "web-ext build && npm run build:clean",
     "build:clean": "git checkout manifest.json js/background.js popup.html",
     "build:dev": "npm run build:clean && sed -i '' 's/:8000//g'  manifest.json js/background.js popup.html && sed -i '' 's@http://127.0.0.1@https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net@g' manifest.json js/background.js popup.html && npm run build",
     "build:stage": "npm run build:clean && sed -i '' 's/:8000//g'  manifest.json js/background.js popup.html && sed -i '' 's@http://127.0.0.1@https://stage.fxprivaterelay.nonprod.cloudops.mozgcp.net@g' manifest.json js/background.js popup.html && npm run build",
--- a/extension/package.json
+++ b/extension/package.json
@@ -8,13 +8,15 @@
   },
   "devDependencies": {},
   "scripts": {
+    "build": "web-ext build && mv ../static/downloads/addon/latest/firefox_relay-*.zip ../static/downloads/addon/latest/firefox_relay.zip && npm run build:clean",
     "build:clean": "git checkout manifest.json js/background.js popup.html",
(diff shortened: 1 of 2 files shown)


extension/web-ext-config.js
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+ module.exports = {
+  // Global options:
+  sourceDir: ".",
+  artifactsDir: "../static/downloads/addon/latest/",
```
groovecoder, line 8: I actually want to stop building the artifacts into the static downloads folder. We never ended up using it.
```suggestion
```

extension/package.json
```diff
@@ -8,13 +8,15 @@
   },
   "devDependencies": {},
   "scripts": {
+    "build": "web-ext build && mv ../static/downloads/addon/latest/firefox_relay-*.zip ../static/downloads/addon/latest/firefox_relay.zip && npm run build:clean",
```
groovecoder, line 11: Trying to get away from `static/downloads`
```suggestion
    "build": "web-ext build && npm run build:clean",
```

groovecoder (changes requested):

pdehaan:
> Can't think of a better way yet to avoid using `sed` for string substitutions, unless we use Nunjucks/Liquid/lodash for simple templates (and then a combination of gitignore+web-ext ignoreFiles).

Yeah, I must be missing something, because this seems to actually _work_ (and is somewhat legible still):

```js
  "dependencies": {
    "liquidjs": "^9.16.1",
    "web-ext": "^5.2.0"
  },
  "devDependencies": {},
  "scripts": {
    "build": "node compile && web-ext build",
    "build:localhost": "PORT=:8000 npm run build",
    "build:dev": "HOST=https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net npm run build",
    "build:stage": "HOST=https://stage.fxprivaterelay.nonprod.cloudops.mozgcp.net npm run build",
    "build:prod": "HOST=https://relay.firefox.com npm run build",
    "lint": "web-ext lint",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
```

Where my ./extension/compile.js file is:

```js
const fs = require("fs");
const { Liquid } = require("liquidjs");

const engine = new Liquid({
  globals: {
    host: process.env.HOST || "http://127.0.0.1",
    port: process.env.PORT || "",
  },
});
const files = ["./manifest.json", "./popup.html", "./js/background.js"];

for (const file of files) {
  // Ie: load "./manifest.json.liquid", inject global vars, and write output to "./manifest.json".
  const output = engine.renderFileSync(`${file}.liquid`);
  fs.writeFileSync(file, output);
}
```

Still, a bit of back end juggling needed still to make sure .liquid templates don't get added to the add-on during <kbd>web-ext build</kbd> and we'd want to .gitignore the generated files so people don't try editing+committing those directly. But a possible future tweak (and fun Saturday hackin').

groovecoder (approved):

PR #739, opened 2020-11-23
Build extension files from liquidjs templates

pdehaan:
Submitted for discussion...

Ref: https://github.com/mozilla/fx-private-relay/pull/737#issuecomment-731644817

PROS:
- Simpler?
- More cross platform
- Less git reverting via `build:clean`
- manifest.json `version` is automatically kept in sync w/ package.json version.

CONS:
- Build step
- Have to remember to edit the .liquid files instead of the generated files (which should be currently .gitignore'd &mdash; but if you edit the generated files then run <kbd>npm run build:prod</kbd>, it'll overwrite all your changes).


extension/package.json
```diff
@@ -31,5 +32,20 @@
...
+    },
+    "ignoreFiles": [
+      ".DS_Store",
+      "**/*.liquid",
+      "README.*",
+      "package.json",
+      "package-lock.json",
+      "sequence-diagram.*"
+    ]
+  }
```
pdehaan, line 50: Moved the web-ext config stuff from its own config file into package.json. 🤷

extension/package.json
```diff
@@ -31,5 +32,20 @@
...
+  "homepage": "https://github.com/mozilla/fx-private-relay#readme",
+  "webExt": {
+    "sourceDir": ".",
+    "verbose": false,
+    "build": {
+      "overwriteDest": true
+    },
+    "ignoreFiles": [
+      ".DS_Store",
+      "**/*.liquid",
```
pdehaan, line 44: Make sure our source .liquid files don't get bundled into the add-on.

extension/package.json
```diff
@@ -8,11 +8,12 @@
   },
   "devDependencies": {},
   "scripts": {
-    "build": "web-ext build && npm run build:clean",
-    "build:clean": "git checkout manifest.json js/background.js popup.html",
```
pdehaan, line 12: Don't need the "build:clean" target anymore.

extension/compile.js
```diff
@@ -0,0 +1,22 @@
...
+const FILES = [
+  "./manifest.json",
+  "./popup.html",
+  "./js/background.js"
+];
+
+const engine = new Liquid({
+  globals: {
+    host: process.env.HOST || "http://127.0.0.1",
+    port: process.env.PORT || "",
```
pdehaan, line 14: I figured these were easier as ENV vars. But with minimal fiddling I could switch to CLI args if you prefer. It's look more like this down in package.json scripts below `node compile --host=https://relay.firefox.com` or `node compile --port=":8000"` (versus `HOST=https://relay.firefox.com node compile`).

PR #761, opened 2021-01-07
wip

groovecoder:
This is a draft pseudo-code outline of how we could support replies thru Relay. I'm sure there's things I'm overlooking or not considering. @say-yawn @birdsarah double-check me on this logic?

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -36,6 +36,17 @@ class CannotMakeAddressException(Exception):
     pass
 
 
+'''
+class ReplyAddress(models.Model):
+    address = models.CharField(
+        max_length=64, default=address_default, unique=True
+    )
+    sender = models.CharField(
+        max_length=64, default=address_default, unique=True
+    )
+    relay_address = models.ForeignKey(RelayAddress, on_delete=models.CASCADE)
+'''
+
 class RelayAddress(models.Model):
     user = models.ForeignKey(User, on_delete=models.CASCADE)
     address = models.CharField(
--- a/emails/views.py
+++ b/emails/views.py
@@ -257,6 +257,7 @@ def _sns_message(message_json):
             status=400
         )
 
+    from_address = parseaddr(mail['commonHeaders']['from'])[1]
     to_address = parseaddr(mail['commonHeaders']['to'][0])[1]
     local_portion = to_address.split('@')[0]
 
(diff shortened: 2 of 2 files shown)


emails/models.py
```diff
@@ -36,6 +36,17 @@ class CannotMakeAddressException(Exception):
     pass
 
 
+'''
+class ReplyAddress(models.Model):
+    address = models.CharField(
+        max_length=64, default=address_default, unique=True
+    )
+    sender = models.CharField(
+        max_length=64, default=address_default, unique=True
```
groovecoder, line 45: we should also add a `validators=[ensure_sha256_hash]` here. to make sure we never store plaintext values.

```python
def ensure_hash(value):
    if (not all(c in string.hexdigits for c in value) or 
        not len(value) == 64):
        raise ValidationError()
```

emails/views.py
```diff
@@ -280,13 +289,21 @@ def _sns_message(message_json):
             )
             incr_if_enabled('email_for_deleted_address', 1)
             # TODO: create a hard bounce receipt rule in SES
+            return HttpResponse("Address deleted.", status=404)
```
birdsarah, line 292: Assuming the user's in good standing (not abusing the system) then I think that:
- if the user is attempting to email to a defunct email address, we send them an email response saying address not available - an email 404 if you will.
- if it's not the user of a defunct email address, then we just eat the email quietly - so as to not alert the people that they've been spam blocked.

this will be based on from header which of course people could fake, but i think that's an acceptable edge case.

emails/views.py
```diff
@@ -280,13 +289,21 @@ def _sns_message(message_json):
...
             logger.error(
                 'Received email for unknown address.',
                 extra={'to_address': to_address}
             )
-        return HttpResponse("Address does not exist", status=404)
+            '''
+            try:
+                reply_address = ReplyAddress.objects.get(address=local_portion)
+                replace From: and Reply-To: with reply_address.relay_address.address
+                incr_if_enabled('email_for_reply_address', 1)
```
birdsarah, line 303: curious what this line is about - i'm just unfamiliar with codebase

birdsarah:
I've left a couple of comments. I don't have enough experience with the codebase to give an intelligent review on whether this covers the bases of how to implement.

emails/views.py
groovecoder: Yeah let's make sure we consider those cases. In this particular code, these http responses are sent back to Amazon SNS, so I don't think they get any further than that. But maybe SNS could do something different based on the status code?

emails/views.py
groovecoder: oh this is [a util for telemetry/metrics](https://github.com/mozilla/fx-private-relay/blob/641f691f0ad921caffcf9c2d8b0e1738e7b0ede9/emails/utils.py#L37).

Issue #79, opened 2020-03-17
When a user deletes an address, hard bounce it and prevent future users from getting the same address


groovecoder:
When a user deletes an address, we still need to:

* [x] Prevent any future addresses from using the same localpart
* [ ] Hard-bounce any messages sent to the address

To do this AND still delete our record of the plaintext email address:

1. [x] Insert the deleted address into a new `emails.deleted_addresses` table
2. [x] Delete the address from the `emails.relay_address` table
3. [ ] Hard bounce all emails sent to unknown addresses
   * Call [`CreateReceiptRule`](https://docs.aws.amazon.com/ses/latest/APIReference/API_CreateReceiptRule.html) with `BounceAction` for the recipient.

groovecoder:
SocketLabs may not have an API to bounce emails sent to unknown/deleted addresses.

Amazon SES has a [CreateReceiptRule API](https://docs.aws.amazon.com/ses/latest/APIReference/API_CreateReceiptRule.html) so we could create bounce rules for addresses that are deleted.

EMMLynch:
Please reopen if this is still needed.

PR #769, opened 2021-01-13
 fix #395: pause email forwarding for bounced addresses 

### LINKED ISSUE #395: Handle SES bounce notifications
When we deliver emails via SES, we should also open an endpoint for SES to send bounce notifications back, so we can disable those aliases. (And/or eventually delete the FXA?)

Example Monitor code:
https://github.com/mozilla/blurts-server/blob/649d27a8ce94041ee3d5ea71707bfa6bf66f6a05/controllers/ses.js#L64-L89

SES + SNS docs:
https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html#bounce-object

groovecoder:
This introduces 2 new fields on the `Profile` model:

* `last_soft_bounce`
* `last_hard_bounce` 

These are DateTime values of the last time we received a soft bounce when we sent to the user's final delivery email address.

This also adds 2 new env vars:

* `SOFT_BOUNCE_ALLOWED_DAYS`
* `HARD_BOUNCE_ALLOWED_DAYS`

These are the number of days after a bounce that we will resume sending emails to the user's delivery email address. The default values are `1` and `30`, respectively, to match FXA's behavior.


emails/views.py
```diff
@@ -358,6 +383,35 @@ def _sns_message(message_json):
     )
```
say-yawn, line 383: Technically, while relaying the email and we get a bounce response does that mean we response to the original sender about the bounce reply with the receiver's real address instead of the relay address? If so, we should probably check for the bounce email and replace that information?

privaterelay/settings.py
```diff
@@ -99,10 +99,6 @@
 AWS_SNS_TOPIC = config('AWS_SNS_TOPIC', None)
 AWS_SNS_KEY_CACHE = config('AWS_SNS_KEY_CACHE', 'default')
 AWS_SES_CONFIGSET = config('AWS_SES_CONFIGSET', None)
-SOCKETLABS_SERVER_ID = config('SOCKETLABS_SERVER_ID', 0, cast=int)
-SOCKETLABS_API_KEY = config('SOCKETLABS_API_KEY', None)
-SOCKETLABS_SECRET_KEY = config('SOCKETLABS_SECRET_KEY', None)
-SOCKETLABS_VALIDATION_KEY = config('SOCKETLABS_VALIDATION_KEY', None)
```
say-yawn, line 105: CircleCI error:
```
```

say-yawn:
https://medium.com/@serbanmihai/how-to-handle-aws-ses-bounces-and-complaints-53d6e7455443 Cool diagrams on how SES/SNS handles bounces.

groovecoder:
TODO: Audit more email headers like `Return-Path` to make sure nothing in the bounces may potentially leak the receiver's real email address.

emails/views.py
```diff
@@ -288,6 +288,31 @@ def _sns_message(message_json):
...
+        incr_if_enabled('email_suppressed_for_soft_bounce', 1)
+        return HttpResponse("Address is temporarily disabled.")
+    last_hard_bounce = user_profile.last_soft_bounce
+    last_hard_bounce_allowed = (
+        datetime.now(timezone.utc) -
+        timedelta(days=settings.HARD_BOUNCE_ALLOWED_DAYS)
+    )
+    if last_hard_bounce and last_hard_bounce > last_hard_bounce_allowed:
+        incr_if_enabled('email_suppressed_for_hard_bounce', 1)
+        return HttpResponse("Address is temporarily disabled.")
```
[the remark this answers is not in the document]
groovecoder: Uplifting from our conversation ...

These HTTP responses go back to SNS, which doesn't have a way to pass back to the original sender. We have [another issue](https://github.com/mozilla/fx-private-relay/issues/79) to use SES receipt rules that will send bounces back to the original senders.

privaterelay/settings.py
groovecoder: Fixed.

emails/views.py
groovecoder: I thought about this a bit more. In https://github.com/mozilla/fx-private-relay/issues/79 we plan to create a receipt rule to permanently hard bounce to the original sender for **_deleted aliases_**, which will inform original senders with appropriate bounces.

So the question is what we do when the **_real recipient's email address_** is bounce ...

If it's a soft bounce, we **_don't_** want to "relay" that bounce back to the original sender, as that would leak temporary state information about the real recipient's email address.

If it's a hard bounce (for >X days), we may want to set their aliases to disabled, and create a receipt rule to soft bounce emails sent to their aliases. 

But I think we don't ever want to **_permanently delete_** aliases (and hard bounce emails sent to them) based on some heuristics. We should only permanently delete aliases when the user explicitly permanently deletes them.

groovecoder:
@say-yawn - can you review this again? I made some tweaks along the way to fixing the tests on Circle CI.

emails/views.py
say-yawn: @groovecoder can we create a separate issue to document this? I am happy to just copy paste what you wrote into a separate issue and reference this PR.

emails/views.py
```diff
@@ -358,6 +383,35 @@ def _sns_message(message_json):
...
+        incr_if_enabled(
+            'email_bounce_%s_%s' % (
+                bounce.get('bounceType'), bounce.get('bounceSubType')
+            ),
+            1
+        )
+        if bounce.get('bounceType') == 'Permanent':
+            profile.last_hard_bounce = now
+        if bounce.get('bounceType') == 'Transient':
+            profile.last_soft_bounce = now
```
say-yawn, line 409:
```suggestion
        if bounce.get('bounceType') == 'Permanent':
            profile.last_hard_bounce = now
            incr_if_enabled('email_hard_bounce', 1)
        if bounce.get('bounceType') == 'Transient':
            profile.last_soft_bounce = now
            incr_if_enabled('email_soft_bounce', 1)
```
More specific metrics?

say-yawn (approved):
Code LGTM. Some minor changes requested. Does not need to be fixed if the author thinks it not necessary.

groovecoder:
Okay I updated this to remove the secret-ish test fixture files. They are available in a private repository of my own here:

https://github.com/groovecoder/private-private-relay/tree/handle-bounces/emails/tests/fixtures

PR #770, opened 2021-01-14
wip for reply support

groovecoder:
Re submitting after https://github.com/mozilla/fx-private-relay/pull/761 was closed with the default branch change to `main`

tech design doc: https://docs.google.com/document/d/19ltX9r9yIEpnVZzG85KYYi_cIRRXSceQ_XIqGA682ss/edit#


emails/views.py
```diff
@@ -268,6 +269,14 @@ def _sns_message(message_json):
 
     try:
         relay_address = RelayAddress.objects.get(address=local_portion)
+
+        '''
+        reply_address = ReplyAddress.objects.get_or_create(
+            relay_address=relay_address, sender=sha256(from_address).hexdigest()
+        )
+        replace Reply-To: with reply_address.address
```
groovecoder, line 277: We also need to replace `Return-Path` header.

groovecoder: We should also look at the `Received` header value, which could leak the mail *provider* of the person sending the reply.

groovecoder: `Authentication-Results`, `Received-SPF`, `DKIM-Signature` could also leak the *provider* of the person sending a reply.

emails/models.py
```diff
@@ -36,6 +36,17 @@ class CannotMakeAddressException(Exception):
     pass
 
 
+'''
+class ReplyAddress(models.Model):
+    address = models.CharField(
+        max_length=64, default=address_default, unique=True
+    )
+    sender = models.CharField(
```
groovecoder, line 44: note: this will capture the sender's `Reply-To` so maybe this column should be `sender_reply_to` ?

emails/views.py
```diff
@@ -280,13 +289,21 @@ def _sns_message(message_json):
...
             incr_if_enabled('email_for_unknown_address', 1)
             logger.error(
                 'Received email for unknown address.',
                 extra={'to_address': to_address}
             )
-        return HttpResponse("Address does not exist", status=404)
+            '''
+            try:
+                reply_address = ReplyAddress.objects.get(address=local_portion)
+                replace From: and Reply-To: with reply_address.relay_address.address
```
groovecoder, line 302: also, replace To: with `reply_address.sender_reply_to`

PR #775, opened 2021-01-27
fix stage DeletedAddress.MultipleObjectsReturned

groovecoder:
Only seeing this on stage. Maybe bad or corrupted data there from before we did proper data integrity checks? https://sentry.prod.mozaws.net/operations/fx-private-relay-stage/issues/10374815/?query=is%3Aunresolved

Or maybe we've created multiple records for `relay@stage.fxprivaterelay.nonprod.cloudops.mozgcp.net` at different times in the past?

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -286,6 +286,9 @@ def _sns_message(message_json):
                 'Received email for unknown address.',
                 extra={'to_address': to_address}
             )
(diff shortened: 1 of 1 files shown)


say-yawn (approved):
LGTM. Yeah this is weird but hopefully we can measure how frequently this happens. Can you please make this metric be surfaced on Grafana as well?

PR #776, opened 2021-01-27
fix address header encoding for python 3.7.9

groovecoder:
What does all this mean?

We are now truncating the `From:` address when it is longer than 36 characters. So something like:

`For quality discount Gamestop stock, be sure to visit evil.com <scam@evil.com>`

to:

`For quality discount Gamestop st [via Relay] <relay@relay.firefox.com>`

The gross details ...

python 3.7.6 has a problem in the way that it handles email address headers. (https://bugs.python.org/issue39073)

it’s fixed in 3.7.8+. but the fix is to disallow `\n` or `\r` characters in email address headers. ok, fair enough …

when we use the `Header` `encode()` method, it follows RFC 2822 and auto-wraps headers at 78 chars …. wait for it … BY ADDING `\n` CHARS! so our encoding is busted in 3.7.9.

so, after trying a bunch of values in the test, I truncated the original From address we receive to 32 chars so we don’t end up with a `\n` char in there.

### DIFF
--- a/emails/tests/utils_tests.py
+++ b/emails/tests/utils_tests.py
@@ -1,3 +1,6 @@
+from email.utils import parseaddr
+
+from django.conf import settings
 from django.test import TestCase
 
 from emails.utils import generate_relay_From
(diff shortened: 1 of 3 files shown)


say-yawn (approved):
Code LGTM though *later* I think we should figure out what to do if the From address is truncated with Design's help?

groovecoder:
Updated the description here to loop in Design.

PR #780, opened 2021-01-29
fix #778: strip line breaks in From

### LINKED ISSUE #778: ValueError: invalid arguments; address parts cannot contain CR or LF in /emails/sns-inbound
https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/10500353/?query=is%3Aunresolved

If Relay receives an email address with a `From:` value that contains a unicode character, e.g., `"Terry ... ct\u2028" <info@a...t.org>`, our `Header` `encode()` call transforms the `\u2028` into a `\n` which is not allowed in python 3.7.9.

groovecoder:


say-yawn (approved):
LGTM

groovecoder:
Updated to also remove `\r` and `\n` chars we see causing sentry errors.

PR #784, opened 2021-02-03
fix #772: popup needs areInputIconsEnabled

### LINKED ISSUE #772: Setting "Show Relay icon in email fields on websites" does not toggle
## Repro steps

1. Install extension
2. Open extension's pop-up (click on extension icon)
3. Open "settings menu" by clicking "cog" icon in the top right corner
4. Click on toggle "Show Relay icon in email fields on websites"

## Expected result

The toggle changes state (goes from "ON" to "OFF")

## Actual result

Nothing happens. The toggle just gets a highlight, but does not change state.

Note that a similar toggle called "Allow Mozilla to collect interaction data" on the same page works fine.

groovecoder:


extension/js/utils.js
```diff
@@ -0,0 +1,13 @@
+/* exported areInputIconsEnabled */
+
+// eslint-disable-next-line no-redeclare
```
lesleyjanenorton, line 3: Do we still  need the `eslint-disable`?

lesleyjanenorton (approved):

extension/js/utils.js
groovecoder: Yeah without it eslint says `'areInputIconsEnabled' is already defined as a built-in global variable. (no-redeclare)`

PR #790, opened 2021-02-09
fix #789: show users when they are "bounce paused"

### LINKED ISSUE #789: Inform users when we receive bounces for their email inbox
To help users who don't receive email thru their aliases because of bounces, on the dashboard show them:

* if we have received a bounce for their email address
* when we received the bounce
* when we will automatically try sending emails to them again
* a link to check their email address again

groovecoder:
This adds a new banner on the user dashboard to show them when they are "bounce paused":

- When we received a bounce from their address
- What type of bounce it was
- When we will try to resume delivering email to their address

For examples ...

image
image

Also refactors a lot of the bounce logic into the `Profile` model; tests incoming.

Note: need to check delivery and bounce recording thru the dev server again since this refactors that code which is hard to test.

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -1,15 +1,20 @@
-from datetime import datetime
+from collections import namedtuple
+from datetime import datetime, timedelta, timezone
 from hashlib import sha256
 import random
 import string
 import uuid
 
 from django.apps import apps
+from django.conf import settings
 from django.contrib.auth.models import User
 from django.db import models
 
 
 emails_config = apps.get_app_config('emails')
+
+
+BounceStatus = namedtuple('BounceStatus', 'paused type')
 
 
 class Profile(models.Model):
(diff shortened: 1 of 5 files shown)


emails/models.py
```diff
@@ -33,6 +34,51 @@ def __str__(self):
...
         return RelayAddress.objects.filter(user=self.user).count()
 
+    def check_bounce_pause(self):
+        if self.last_hard_bounce:
+            last_hard_bounce_allowed = (
+                datetime.now(timezone.utc) -
+                timedelta(days=settings.HARD_BOUNCE_ALLOWED_DAYS)
+            )
+            if self.last_hard_bounce > last_hard_bounce_allowed:
+                return True, 'hard'
```
birdsarah, line 44: This is fine but a named tuple might be cleaner for the code that's checking these answers.

privaterelay/templates/includes/banners.html
```diff
@@ -1,3 +1,16 @@
+    {% if last_bounce_type %}
+<div class="dashboard-banners flx hide-750">
+  <div class="banner-gradient-bg">
+    <div class="bounced-email-banner banner-content flx flx-row">
+      <div class="banner-bounce-warning banner-logo"></div>
+      <div class="banner-copy flx flx-col">
+        <p class="banner-hl ff-Met">Relay couldn't deliver your email.</p>
+        <p class="banner-sub">On {{ last_bounce_date|date:"SHORT_DATETIME_FORMAT" }}, Relay received a <em>{{ last_bounce_type }}</em> "bounce" from your email provider when trying to deliver your email. This can happen if Relay couldn't connect to your email provider, or if your mailbox was full. Relay will try to resume delivering email to you at {{ next_email_try|date:"SHORT_DATETIME_FORMAT" }}</p>
+      </div>
```
birdsarah, line 9: I find this copy confusing even though I know what you're trying to say. The main confusion is which way things aren't working. I would suggest something like:

```
We are currently unable to send email to your forwarding email address myrealaddress@gmail.com. We have received a <em>{{ last_bounce_type }}</em> "bounce" from your email provider @gmail.com when trying to forward relay emails to you. This can happen if we couldn't connect to your email provider, or your mailbox was full. We will try again on {{ next_email_try|date:"SHORT_DATETIME_FORMAT" }}.
```

privaterelay/views.py
```diff
@@ -64,10 +64,16 @@ def profile(request):
...
 
-    return render(request, 'profile.html', {
-        'relay_addresses': relay_addresses, 'avatar': avatar
-    })
+    profile = request.user.profile_set.first()
+    bounce_paused, bounce_type = profile.check_bounce_pause()
+    if bounce_paused:
+        context.update({'last_bounce_type': bounce_type})
+        context.update({'last_bounce_date': profile.last_bounce_date})
+        context.update({'next_email_try': profile.next_email_try})
```
birdsarah, line 74: you can do one context update with all three properties.

privaterelay/views.py
```diff
@@ -64,10 +64,16 @@ def profile(request):
...
+
+    return render(request, 'profile.html', context)
```
birdsarah, line 76: if you ever let me at this code base, we will be using class-based views :)

birdsarah (commented):
Made some comments. I don't have anything intelligent to say about code logic other than it looks reasonable.

emails/models.py
groovecoder: TIL. ❤️

privaterelay/views.py
groovecoder: you're welcome to the code-base! we can have the implicit code flow argument anytime.

groovecoder:
Updated to address comments.

emails/models.py
```diff
@@ -1,17 +1,22 @@
...
 from django.apps import apps
+from django.conf import settings
 from django.contrib.auth.models import User
 from django.db import models
 
 
 emails_config = apps.get_app_config('emails')
 
 
+BounceStatus = namedtuple('BounceStatus', 'paused type')
```
birdsarah, line 17: It's merged now, but for future ref you can leverage the named tuple more to make the code more readable and less error prone (always used keyword args to help reduce errors)

```python
BounceStatus = namedtuple('active', 'type')

def check_bounce_pause(self):
    .....
    # example return 
    return BounceStatus(active=True, type='hard')
 
def next_email_try(self):
    ...
    # example consumption
    bounce_status = self.check_bounce_pause()
    if not bounce_status.active:
         ...
    if bounce_status.type == 'soft':
        return ...
```

groovecoder: But that doesn't fit in a regex replace statement in vim. 😛

PR #802, opened 2021-03-23
fix #750: remove buggy unnecessary log output

### LINKED ISSUE #750: deleted_user_objects log causing bigquery export errors
The following code is causing bigquery export errors, because it's an array that contains two different data types. The first element of the array is a number, the second is a dictionary.

https://github.com/mozilla/fx-private-relay/blob/60e5d9191056140f8e39f5e2802691be3b39518d/privaterelay/views.py#L233-L236

Can we change this to output two elements of the same type or change the whole dataset to a dictionary?

image

groovecoder:


Micheletto (approved):
Fine with me.

say-yawn (approved):
LGTM.

PR #814, opened 2021-04-20
Fix #806 - Fade out copied success message after 5 seconds

### LINKED ISSUE #806: UI: Make email copied tag go after a fixed time
See the email copied tag

This tag `Copied!` is permanent right now. What we can do is, make it go away (after some fixed time, let's say 15 seconds) after the link has been copied. Recently I faced this small issue, where I cannot copy the link again just because the copy icon was not there anymore. Although, I could have just clicked on the email and copy it. I think this fix will improve the user experience :100: 

So, please let me know if I can work on this and make a PR :)

Browser: Mozilla Firefox 87.0

maxxcrawford:
## Testing

1. Go to http://127.0.0.1:8000/accounts/profile/
2. Create multiple aliases
3. Click copy.
4. **Expected**: Message should fade out and reset to copy symbol after 5 seconds. 
5. **Expected**: Clicking copy on multiple messages should reset the other email _(matches current functionality) _

Screenshot: 
image

Fading away: 
image

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -653,15 +653,33 @@ table {
 }
 
 .relay-address.alias-copied {
-  opacity: 1 !important;
   text-decoration: none !important;
 }
 
-
-.relay-address.alias-copied.alias-copied-fadeout .alias-copied-message {
-  opacity: 0 !important;
+.alias-copied-fadeout .alias-copied-message {
+  animation-duration: 4s;
+  animation-name: fadeOut;
+  animation-fill-mode: forwards;
 }
 
+.alias-copied-fadeout .alias-copied-icon {
+  opacity: 0;
+  animation-delay: 3s;
+  animation-duration: 0.1s;
+  animation-name: fadeOut;
+  animation-direction: reverse;
+  animation-fill-mode: forwards;
+}
+
+@keyframes fadeOut {
+  from {
+    opacity: 1;
+  }
+
+  to {
+    opacity: 0;
+  }
+}
 
 .alias-copied-icon {
   width: 18px;
@@ -683,12 +701,7 @@ table {
   pointer-events: none;
 }
 
-.alias-copied .alias-copied-icon {
-  display: none;
-}
-
-.alias-copied-message,
-.alias-copied .alias-copied-icon {
+.alias-copied-message {
   visibility: hidden;
 }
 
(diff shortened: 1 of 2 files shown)


static/js/app.js
```diff
@@ -91,12 +91,28 @@ async function sendForm(formAction, formData) {
...
-    previouslyCopiedAlias.classList.remove("alias-copied");
+    previouslyCopiedAlias.classList.remove("alias-copied", "alias-copied-fadeout");
     previouslyCopiedAlias.title = "Copy alias to clipboard";
   }
+  
   aliasCopyBtn.classList.add("alias-copied");
   aliasCopyBtn.title="Alias copied to clipboard";
+
+  // Countdown timer to fade out the "Copied" message after 5 seconds of being displayed
+  let copyMessageCountdown = 0;
```
maxxcrawford, line 104: Using `let` here bc it's being updated by the `setInterval`

static/css/style.css
```diff
@@ -657,6 +657,12 @@ table {
   text-decoration: none !important;
 }
 
+
+.relay-address.alias-copied.alias-copied-fadeout .alias-copied-message {
+  opacity: 0 !important;
```
maxxcrawford, line 662: `The !important` may not be required, but I wanted to match previous convention.

static/js/app.js
groovecoder: Could we do this with a CSS animation instead? Something like:

```
.alias-copied-fadeout {
  animation-duration: 5s;
  animation-name: fadeOut;
}

@keyframes fadeOut {
  from: {
    opacity: 100%;
  }

  to {
    opacity: 0%;
  }
}
```
?

static/js/app.js
```diff
@@ -91,12 +91,20 @@ async function sendForm(formAction, formData) {
...
-    previouslyCopiedAlias.classList.remove("alias-copied");
+
+  if (previouslyCopiedAlias || previouslyCopiedAlias == triggeringEl) {
+    previouslyCopiedAlias.classList.remove("alias-copied", "alias-copied-fadeout");
     previouslyCopiedAlias.title = "Copy alias to clipboard";
   }
-  aliasCopyBtn.classList.add("alias-copied");
-  aliasCopyBtn.title="Alias copied to clipboard";
+
+  setTimeout(() => {
```
maxxcrawford, line 100: This `setTimeout` is critical to reset the copy button if a user clicks on the same email address button twice in a row.

groovecoder (approved):

PR #815, opened 2021-04-21
Add Fluent library to the website project

maxxcrawford:
## Summary

This PR adds Fluent (via [django-ftl](https://github.com/django-ftl/django-ftl) to the Django app

TODO: 

- [x] Header
- [x] Footer
- [x] Homepage
- [x] FAQ
- [x] Profile (logged in)
- [x] Survey Banner
- [x] ~Add Bento box strings [See Monitor bento.ftl implementation](https://github.com/mozilla/blurts-server/blob/main/locales/en/bento.ftl)~ Filed a follow-up issue here:  #822
- [x] ~Update hero images (Text is embedded)~ The strings are added to `app.ftl`, but we'll need to bust up the art to make this work fully. Filed a follow-up issue here:  #821

## Testing 

Verify English:
- Set browser language as a `en*` variant. 
- Navigate to homepage: http://localhost:8000/
- Expected: Headline should have `[EN]` string visible (See screenshot)

Verify German (non-English): 
_Note - To verify German, please checkout this commit: 2cc98dc0efac18306bb8f22e5cb6ae82f71f9f8d_
- Set browser language as a `de*` variant. 
- Navigate to homepage: http://localhost:8000/
- Expected: Headline should have `[DE]` string visible

## Screenshots

image


privaterelay/locales/de/privaterelay/brands.ftl
```diff
@@ -0,0 +1,171 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### Please read the trademarks localization guide
+### https://mozilla-l10n.github.io/styleguides/mozilla_general/index.html#brands-copyright-and-trademark
```
maxxcrawford, line 6: This directory (`privaterelay/locales/de/`) is temporary to verify Fluent works as expected. These files will be removed before merging.

privaterelay/locales/de/privaterelay/home.ftl
```diff
@@ -0,0 +1,10 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
maxxcrawford, line 1: For testing purposes only.

privaterelay/locales/en/privaterelay/brands.ftl
```diff
@@ -0,0 +1,171 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
maxxcrawford, line 3: Taken from [Bedrock](https://github.com/mozilla/bedrock/blob/master/l10n/en/brands.ftl)

Question: Should we par the list down to only strings that are used? I pared the list down considerable from Bedrock, but there are still strings that are not used.

privaterelay/templates/home.html
```diff
@@ -1,15 +1,18 @@
 {% extends "base.html" %}
 
+{% load ftl %}
 {% load socialaccount %}
 
 {% block content %}
 
 {% include "includes/messages.html" %}
 
+{% withftl bundle='privaterelay.ftl_bundles.main' %}
```
maxxcrawford, line 10: I used the [`withftl`](https://django-ftl.readthedocs.io/en/latest/usage.html#withftl) method, rather than including the file via [`ftlconf`](https://django-ftl.readthedocs.io/en/latest/usage.html#ftlconf). This can be revised as needed.

privaterelay/templates/includes/remaining-aliases.html
```diff
@@ -1,17 +1,24 @@
 {% load static %}
 {% load relay_tags %}
+{% load ftl %}
+{% ftlconf bundle='privaterelay.ftl_bundles.main' %}
 
+{% remaining_free_aliases relay_addresses as remaining_free_aliases_count %}
```
maxxcrawford, line 6: This variable is set to capture the `simple_tag` function call and the output is passed as an argument to the translation string.

static/js/analytics.js
```diff
@@ -153,6 +153,7 @@ function analyticsSurveyLogic() {
   switch (surveyType) {
     case "nps": {
       const notLikely = document.createElement("li");
+      // TODO: Translate these strings
```
maxxcrawford, line 156: We need to add Fluent  support to the JavaScript for this function.

groovecoder: FWIW, my approach in Monitor for translating the JS-powered micro-survey responses was to add data attributes for the strings (e.g., `data-{{ msg_id }}-translated="{% ftlmsg msg_id %}"`) and then have the JS put that value into the `textContent` of the proper element.

groovecoder (changes requested):
Code looks good and works, though I haven't finished reading the django FTL docs yet. Could we move all the `ftlconf` tags up right under the `{% load ftl %}` tag? To keep things consistent and have our concerns all grouped together.

Is this ready for i18n/l10n review?

privaterelay/templates/faq.html
```diff
@@ -1,61 +1,45 @@
 {% extends "base.html" %}
 {% load static %}
 {% load socialaccount %}
+{% load ftl %}
+
+
 
 {% block content %}
+{% ftlconf bundle='privaterelay.ftl_bundles.main' %}
```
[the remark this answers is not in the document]
maxxcrawford: @groovecoder Let me confirm, but I think I needed it to be after the `{% block content %}` opening.

privaterelay/locales/config/pontoon.toml
```diff
@@ -0,0 +1,15 @@
...
+# be covered by the [[paths]] in this file.
+
+basepath = ".."
+
+locales = [
+    "de",
+    "en-CA",
+    "en-GB",
+    "fr",
+]
```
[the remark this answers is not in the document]
maxxcrawford: No – This is where I think we set which languages we want to support. We need @peiying2 to take a look to see how we need to actually configure this before merging. 

@peiying2 Could you take a quick look to make sure we've set this up correctly? Additionally, if this needs to stay OUT of Pontoon for now, please advise if we need to remove anything.

privaterelay/views.py
```diff
@@ -55,7 +56,9 @@
 def home(request):
     if (request.user and not request.user.is_anonymous):
         return redirect(reverse('profile'))
-    return render(request, 'home.html')
+    return render(request, 'home.html', {
+        'ftl_bundle': main_bundle,
+    })
```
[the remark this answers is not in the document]
maxxcrawford: This is some copy-pasta left over from the docs when I first installed it. I think this can be removed (as the other pages don't need them to work). I'll confirm and revise as necessary.

privaterelay/templates/faq.html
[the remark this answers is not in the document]
maxxcrawford: Removed!

privaterelay/views.py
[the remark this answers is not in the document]
maxxcrawford: Removed!

privaterelay/locales/config/pontoon.toml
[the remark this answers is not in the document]
peiying2: If this project is not ready for Pontoon, then can we exclude the file? Or since the project is not set up in Pontoon,  so nothing will be picked up there, and you can keep your work here.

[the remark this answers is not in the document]
peiying2: A side question, is this considered marketing  content, thus Anika and the like would be working on this once enabled in Pontoon. Or is it for the communities?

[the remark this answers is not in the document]
maxxcrawford: > A side question, is this considered marketing content, thus Anika and the like would be working on this once enabled in Pontoon. Or is it for the communities?

I would assume it would be treated the same as [Monitor](https://pontoon.mozilla.org/projects/firefox-monitor-website/) or [Facebook Container](https://pontoon.mozilla.org/projects/facebook-container/)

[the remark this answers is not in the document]
groovecoder: @maxxcrawford - I would say that we should pull this list from an env var like we do for Monitor, but that just shifts the responsibility from an app code change to an ops code change, so I guess 🤷‍♂️ . Maybe best to keep it in our own code.

privaterelay/views.py
```diff
@@ -37,6 +37,7 @@
 )
 from emails.utils import get_post_data_from_request
 
+from .ftl_bundles import main as main_bundle
```
groovecoder, line 40: Don't need this line anymore.

privaterelay/locales/l10n-pontoon.toml
```diff
@@ -0,0 +1,7 @@
+# Entrypoint for community localization.
+# Edits to add content for l10n should be made in configs/pontoon.toml
+
+basepath = "."
+
+[[includes]]
+    path = "configs/pontoon.toml"
```
groovecoder, line 7: This path `configs/pontoon.toml` doesn't match the directory structure `config/pontoon.toml`?

groovecoder (changes requested):

privaterelay/views.py
maxxcrawford: Fixed!

privaterelay/locales/l10n-pontoon.toml
maxxcrawford: Fixed!

groovecoder (approved):

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+fx-makes-tech = { -brand-name-firefox } is tech that fights for your online privacy.
+made-by-mozilla = Made by { -brand-name-mozilla }
+fx-desktop = { -brand-name-firefox-browser } for Desktop
+fx-lockwise = { -brand-name-firefox-lockwise }
+fx-mobile = { -brand-name-firefox-browser } for Mobile
+fx-monitor = { -brand-name-firefox-monitor }
+fx-pocket = { -brand-name-pocket }
+mobileCloseBentoButtonTitle = Close menu
+
+## Home Page - https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/
```
peiying2, line 55: This is not needed. The product staging info will provided in Pontoon on project dashboard.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+how-it-works-step-2-headline = Create a new alias
+how-it-works-step-2-copy = As you browse, the { -brand-name-relay } icon will appear in form fields where sites ask for your email address. Select it to generate a new, random address that ends in @relay.firefox.com. { -brand-name-relay } will forward messages to the primary email address associated with your account.
+how-it-works-step-3-headline = Manage your account
+how-it-works-step-3-copy = Sign in to the { -brand-name-relay } website to keep track of the aliases you’ve created.  If you find that one receives spam or unwanted messages, you can block all messages or even delete the alias, right from the management page.
+
+hero-image-copy-trust = Can you even trust this company with your personal email?
+hero-image-copy-unique-html = <strong>Use a unique relay address</strong> for every new account…
+hero-image-copy-protect-html = …so you can <strong>protect your real email</strong> from tracking and spam.
+hero-image-copy-control-html = Now <em>you're in control</em> of what's arriving to your inbox!
+
```
peiying2, line 78: Remove extra line space (nit pick)

peiying2 (changes requested):
Structure of  the files:
- This app.ftl source file is for `en` or `en-US`. Make sure to use the correct locale code. The  toml file refers to `en`. The difference between en and en-US is that  en-US is treated like a locale, and it allows you to add US specific content while en is the source for all other languages.
- File path/directory structure: Recommend to use: {productname}/locales/{locale}/app.ftl. Right now, locaels/{locale}/**privatereplay**/app.ftl. If you have more  than one file than app.ftl, you can list them all under locales/{locael}/...
- Are we going to have an l10n repo or is localized content going to be part of this product code.
- If there is no interest in using vendor for the localization, we can do away toml files which configures which locales goes to where (pontoon or vendor platform). If this is for community to  localize, we can remove the pontoon.toml file.

Here is the summary of the common issues identified in the app.ftl file itself:
- Brands section, 1). add parameter for Firefox Account entry depending on the function of the term. 2). Consider adding Relay Premium and Premium as terms and reference them in the file
- Use curly quote throughout: single and double
- Use `extension` instead of `add-on`.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+-brand-name-firefox-account = Firefox Account
+-brand-name-firefox-browser = Firefox Browser
+-brand-name-mozilla = Mozilla
+-brand-name-firefox-lockwise = Firefox Lockwise
+-brand-name-firefox-monitor = Firefox Monitor
+-brand-name-pocket = Pocket
+
+## Meta Data
+
+meta-title = { -brand-name-firefox-relay }
```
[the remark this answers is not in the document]
flodolo: Note: this string is OK, there was some misunderstanding between Peiying and I.

privaterelay/locales/config/pontoon.toml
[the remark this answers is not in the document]
maxxcrawford: @peiying2 Excellent. Thanks for the insight on this.  

@groovecoder I recommend we remove the file all together and let it supported locales be set in Pontoon.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+how-it-works-step-2-headline = Create a new alias
+how-it-works-step-2-copy = As you browse, the { -brand-name-relay } icon will appear in form fields where sites ask for your email address. Select it to generate a new, random address that ends in @relay.firefox.com. { -brand-name-relay } will forward messages to the primary email address associated with your account.
+how-it-works-step-3-headline = Manage your account
+how-it-works-step-3-copy = Sign in to the { -brand-name-relay } website to keep track of the aliases you’ve created.  If you find that one receives spam or unwanted messages, you can block all messages or even delete the alias, right from the management page.
+
+hero-image-copy-trust = Can you even trust this company with your personal email?
+hero-image-copy-unique-html = <strong>Use a unique relay address</strong> for every new account…
+hero-image-copy-protect-html = …so you can <strong>protect your real email</strong> from tracking and spam.
+hero-image-copy-control-html = Now <em>you're in control</em> of what's arriving to your inbox!
```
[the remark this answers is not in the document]
maxxcrawford: It's [a convention used](https://django-ftl.readthedocs.io/en/latest/usage.html#html-escaping) from our Fluent python lib to escape HTML into strings. If there is a better way to handle this, I'm happy to revisit.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+
+## FAQ Page - https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/faq
+
+faq-headline = Frequently Asked Questions
+faq-question-1-question = What about spam?
+
+# Variables:
+#   $url (url) - https://addons.mozilla.org/firefox/addon/private-relay/
+#   $attrs (string) - specific attributes added to external links
+faq-question-1-answer-html = While { -brand-name-relay } does not filter for spam, our email partner Amazon SES does block spam and malware. If { -brand-name-relay } forwards messages you don’t want, you can update your { -brand-name-relay } settings to block messages from the alias forwarding them. <br/><br/> If you see a broader problem of unwanted email from all of your aliases, please <a  href="{ $url }" { $attrs }>report this to us</a> so we can consider adjusting the SES spam thresholds for this service. <br/><br/> If you report these as spam, your email provider will see { -brand-name-relay } as the source of spam, not the original sender.
```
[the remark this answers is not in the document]
maxxcrawford: Broken up. I also broke up other lines as well.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+
+# Variables:
+#   $email (string) - User email address
+profile-label-welcome = Welcome, { $email }!
+
+profile-headline-manage-relay = Manage your { -brand-name-relay } aliases
+profile-headline-manage-domain = Manage your Domain aliases
+profile-supports-email-forwarding = { -brand-name-firefox-relay } supports email forwarding (including attachments) of email up to 150KB in size
+profile-promo-premium-headline = Go premium for more features.
+profile-promo-premium-copy = With { -brand-name-relay } Premium, you get your own email domain.
```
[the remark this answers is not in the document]
maxxcrawford: I don't think it is – or at least, it isn't yet. Let me confirm. 

@groovecoder Should we use "Firefox Relay Premium" here?

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+profile-promo-premium-cta = Buy Premium
+profile-label-saved = Label saved!
+profile-label-generate-new-alias = Generate New Alias
+profile-label-delete = Delete
+profile-label-forward-emails = Forward emails to
```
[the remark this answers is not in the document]
maxxcrawford: Whatever the users' email address is will populate here. There was not a colon originally, so I kept it the same. Should we add on, or should I add a comment for context?

image

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+how-it-works-step-1-headline = Install the extension
+
+# Variables:
+#   $url (url) - https://addons.mozilla.org/firefox/addon/private-relay/
+#   $attrs (string) - specific attributes added to external links
+how-it-works-step-1-copy-html = <a href="{ $url }" { $attrs }>Download the { -brand-name-relay } extension for { -brand-name-firefox }</a>. Select the icon that appears on your { -brand-name-firefox } toolbar to access the sign in page. Sign in with your { -brand-name-firefox-account } to get started.
+how-it-works-step-2-headline = Create a new alias
+how-it-works-step-2-copy = As you browse, the { -brand-name-relay } icon will appear in form fields where sites ask for your email address. Select it to generate a new, random address that ends in @relay.firefox.com. { -brand-name-relay } will forward messages to the primary email address associated with your account.
+how-it-works-step-3-headline = Manage your account
+how-it-works-step-3-copy = Sign in to the { -brand-name-relay } website to keep track of the aliases you’ve created.  If you find that one receives spam or unwanted messages, you can block all messages or even delete the alias, right from the management page.
```
peiying2, line 72: Single space two sentences `...you’ve created. If you find...`

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
peiying2: Add a comment will be good, `This is followed by an email address`

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+fx-mobile = { -brand-name-firefox-browser } for Mobile
+fx-monitor = { -brand-name-firefox-monitor }
+fx-pocket = { -brand-name-pocket }
+mobileCloseBentoButtonTitle = Close menu
+
+## Home Page - https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/
+
+home-hero-headline = Hide your real email address to help protect your identity
+home-hero-copy = { -brand-name-firefox-relay } makes it easy to create aliases, randomly generated email addresses that forward to your real inbox. Use it to protect your online accounts - and your identity - from hackers. Sign in with your { -brand-name-firefox-account } to get started.
+home-hero-cta = { nav-profile-sign-in }
```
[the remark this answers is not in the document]
flodolo: Note: this comment (and others) is marked as resolved but it really isn't? My guess is that you didn't push your changes?

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
flodolo: I think you're talking about two different things? This is about consistently using typographic apostrophes instead of straight quotes, i.e. not `what's`, but `what’s`. Same applies in the rest of the file, and for `""` vs `“”` (**not** when used in markup).

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+survey-option-strongly-disagree = Strongly disagree
+survey-option-disagree = Disagree
+survey-option-unsure = Unsure
+survey-option-agree = Agree
+survey-option-strongly-agree = Strongly agree
+survey-option-i-wouldnt-care = I wouldn't care
+survey-option-somewhat-disappointed = Somewhat disappointed
+survey-option-very-disappointed = Very disappointed
+survey-option-very-likely = Very likely
+survey-option-not-likely = Not likely
```
flodolo, line 208: Please keep an empty line at the end of the file

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: @flodolo Yup. Just pushed it up. I'll refrain from marking resolved until the commit is up officially.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+how-it-works-step-2-headline = Create a new alias
+how-it-works-step-2-copy = As you browse, the { -brand-name-relay } icon will appear in form fields where sites ask for your email address. Select it to generate a new, random address that ends in @relay.firefox.com. { -brand-name-relay } will forward messages to the primary email address associated with your account.
+how-it-works-step-3-headline = Manage your account
+how-it-works-step-3-copy = Sign in to the { -brand-name-relay } website to keep track of the aliases you’ve created.  If you find that one receives spam or unwanted messages, you can block all messages or even delete the alias, right from the management page.
+
+hero-image-copy-trust = Can you even trust this company with your personal email?
+hero-image-copy-unique-html = <strong>Use a unique relay address</strong> for every new account…
+hero-image-copy-protect-html = …so you can <strong>protect your real email</strong> from tracking and spam.
```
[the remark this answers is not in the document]
maxxcrawford: These strings are part of the image gallery on the homepage (currently, this text is embedded in the images - #821). It will be resolved in a follow-up PR.

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: @flodolo Understood about the single/double quotes. 

I was answering this question with my reply: 
> why these last three string IDs contain -html. Just curious.

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: (and removed the double space)

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: Also, I'll add the `:`.  We should aim for consistency, and `Created:` has it.

privaterelay/locales/en-us/privaterelay/app.ftl
maxxcrawford: I broke this line up into separate lines, and removed the extra space. Please let me know if my new line breaks are causing any issues.

privaterelay/locales/en-us/privaterelay/app.ftl
maxxcrawford: Removed!  I removed some additional extra lines throughout the file too.

privaterelay/locales/en-us/privaterelay/app.ftl
maxxcrawford: Cool. I pulled this convention from the Bedrock project. Should I remove all of the URLs?

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+# Variables:
+#   $attrs (string) - specific attributes added to HTML element
+modal-delete-warning-recovery-html = Once you delete this alias, it cannot be recovered. { -brand-name-firefox-relay } will no longer forward messages sent to <span { $attrs }></span>, including messages that allow you to reset lost passwords.
+
+modal-delete-warning-upgrade = If you use this alias to log in to sites you care about, you should update your login with a different email address before you delete this one.
+modal-delete-confirmation = Yes, I want to delete this alias
+
+## Evergreen Survey (displayed on the profile page) - https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/accounts/profile/
+
+survey-question-1 = On a scale from 1-10, how likely are you to recommend Relay to a friend or colleague?
```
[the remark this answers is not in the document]
maxxcrawford: Fixed on all the survey entries!

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+profile-supports-email-forwarding = { -brand-name-firefox-relay } supports email forwarding (including attachments) of email up to 150KB in size
+profile-promo-premium-headline = Go premium for more features.
+profile-promo-premium-copy = With { -brand-name-relay } Premium, you get your own email domain.
+profile-promo-premium-cta = Buy Premium
+profile-label-saved = Label saved!
+profile-label-generate-new-alias = Generate New Alias
+profile-label-delete = Delete
+profile-label-forward-emails = Forward emails to
+profile-label-first-emailed = First emailed:
+profile-label-created = Created:
```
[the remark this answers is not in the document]
maxxcrawford: Context added.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+profile-label-details-show = Show Details
+profile-label-details-hide = Hide Details
+profile-label-forwarding = forwarding
```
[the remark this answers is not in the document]
maxxcrawford: It's a label for a switch to turn off/on. It's lowercase per the design. 

image

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+profile-headline-manage-domain = Manage your Domain aliases
+profile-supports-email-forwarding = { -brand-name-firefox-relay } supports email forwarding (including attachments) of email up to 150KB in size
+profile-promo-premium-headline = Go premium for more features.
+profile-promo-premium-copy = With { -brand-name-relay } Premium, you get your own email domain.
+profile-promo-premium-cta = Buy Premium
+profile-label-saved = Label saved!
+profile-label-generate-new-alias = Generate New Alias
+profile-label-delete = Delete
+profile-label-forward-emails = Forward emails to
+profile-label-first-emailed = First emailed:
```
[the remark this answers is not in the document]
maxxcrawford: Context added.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+profile-label-details-show = Show Details
+profile-label-details-hide = Hide Details
+profile-label-forwarding = forwarding
+profile-label-blocking = blocking
+profile-label-copied = Copied!
+profile-label-blocked = Blocked
+profile-label-forwarded = Forwarded
+profile-label-cancel = Cancel
+profile-blocked-copy = { -brand-name-firefox-relay } will delete messages before they can reach your inbox when you select blocking for this alias.
+profile-forwarded-copy-html = { -brand-name-firefox-relay } will send messages to your inbox when you select forwarding for this alias. <br /><span class="fwd-note"><span class="bold ff-Met">Note:</span> Email (including attachments) larger than 150KB are not currently supported and will not be forwarded.
```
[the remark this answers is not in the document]
maxxcrawford: I can break up this HTML into three separate strings: 
```
profile-forwarded-copy: { -brand-name-firefox-relay } will send messages…
profile-forwarded-note: Note: 
profile-forwarded-note-copy: Email (including attachments) larger…

File: privaterelay/locales/en-us/privaterelay/app.ftl [hunk 28]
```diff
@@ -0,0 +1,208 @@
...
+banner-download-firefox-cta = Get { -brand-name-firefox }
+banner-download-install-add-on-headline = Looks like you haven't installed the add-on... yet!
+banner-download-install-add-on-copy = The { -brand-name-relay } extension for { -brand-name-firefox-browser } makes creating aliases even easier.
+banner-download-install-add-on-cta = Get the add-on for { -brand-name-firefox }
+banner-premium-headline = Go premium for more features.
+banner-premium-copy = With { -brand-name-relay } Premium, you get unlimited email addresses and your own email domain.
+banner-premium-cta = Buy Premium
+banner-choose-subdomain-headline = Choose your own domain
+banner-choose-subdomain-copy = You can choose your own domain for "wildcard" aliases. NOTE: YOU CANNOT CHANGE THIS LATER!
+banner-choose-subdomain-label = Your domain is:
```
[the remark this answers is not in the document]
maxxcrawford: Yes. Context added.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,226 @@
...
+profile-label-details-hide = Hide Details
+profile-label-forwarding = forwarding
+profile-label-blocking = blocking
+profile-label-copied = Copied!
+profile-label-blocked = Blocked
+profile-label-forwarded = Forwarded
+profile-label-cancel = Cancel
+profile-blocked-copy = { -brand-name-firefox-relay } will delete messages before they can reach your inbox when you select blocking for this alias.
+profile-forwarded-copy-html = { -brand-name-firefox-relay } will send messages to your inbox when you select forwarding for this alias. <br />
+    <span class="fwd-note"><span class="bold ff-Met">Note:</span> Email (including attachments) larger than 150KB are not currently supported and will not be forwarded.
```
pdehaan, line 154: Can you please confirm these HTML tags?
I see two opening `<span` tags but only one closing `</span>` tag.
My guts are saying this should be a combined `<span class="fwd-note bold ff-Met">` tag.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,226 @@
...
+#   $username (string) - Username
+#   $bounce_type (string) - Type of bounced email
+#   $date (string) - A date for the next time the services tries to resend the email
+banner-bounced-copy = We are currently unable to send email to { $username }. 
+    We received a <em>{ $bounce_type }</em> “bounce” from your email provider when trying to forward emails to you. 
+    This can happen if { -brand-name-relay } couldn’t connect to your email provider, or if your mailbox was full. We will try again on { $date }.
+banner-download-firefox-headline = { -brand-name-relay } is even better in { -brand-name-firefox }
+banner-download-firefox-copy = The { -brand-name-relay } extension for { -brand-name-firefox-browser } makes creating aliases even easier.
+banner-download-firefox-cta = Get { -brand-name-firefox }
+banner-download-install-add-on-headline = Looks like you haven’t installed the add-on…yet!
```
pdehaan, line 170: Q: Should there be a space between the "&hellip;" and "yet!"?

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+#   $bounce_type (string) - Type of bounced email
+#   $date (string) - A date for the next time the services tries to resend the email
+banner-bounced-copy = We are currently unable to send email to { $username }. We received a <em>{ $bounce_type }</em> "bounce" from your email provider when trying to forward emails to you. This can happen if { -brand-name-relay } couldn't connect to your email provider, or if your mailbox was full. We will try again on { $date }.
+
+banner-download-firefox-headline = { -brand-name-relay } is even better in { -brand-name-firefox }
+banner-download-firefox-copy = The { -brand-name-relay } extension for { -brand-name-firefox-browser } makes creating aliases even easier.
+banner-download-firefox-cta = Get { -brand-name-firefox }
+banner-download-install-add-on-headline = Looks like you haven't installed the add-on... yet!
+banner-download-install-add-on-copy = The { -brand-name-relay } extension for { -brand-name-firefox-browser } makes creating aliases even easier.
+banner-download-install-add-on-cta = Get the add-on for { -brand-name-firefox }
```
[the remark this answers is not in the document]
pdehaan: Excellent question! I see "extension" used in strings 4 times, and "add-on" twice (here and L170 above).  Probably best to standardize on a single term.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,226 @@
...
+
+bento-button-title = { -brand-name-firefox } apps and services
+fx-makes-tech = { -brand-name-firefox } is tech that fights for your online privacy.
+made-by-mozilla = Made by { -brand-name-mozilla }
+fx-desktop = { -brand-name-firefox-browser } for Desktop
+fx-lockwise = { -brand-name-firefox-lockwise }
+fx-mobile = { -brand-name-firefox-browser } for Mobile
+fx-monitor = { -brand-name-firefox-monitor }
+fx-pocket = { -brand-name-pocket }
+mobileCloseBentoButtonTitle = Close menu
```
pdehaan, line 54: This looks non-kebab-case, whereas everything else looks kebaby. 🤷

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,226 @@
...
+#   $url (url) - https://github.com/mozilla/fx-private-relay/issues/99
+#   $attrs (string) - specific attributes added to external links
+faq-question-4-answer-html = { -brand-name-relay } does not yet offer the ability to reply using an alias. 
+    If you try, nothing will happen. We are planning an additional feature to let you <a href="{ $url }" {$attrs}>reply anonymously to senders</a>.
+faq-question-5-question = Can I make up my own { -brand-name-relay } alias using the @relay.firefox.com domain?
+faq-question-5-answer = Not currently, but we are considering new features including letting you create your own alias with a designated domain.
+faq-question-6-question = What happens if { -brand-name-mozilla } shuts down the { -brand-name-firefox-relay } service?
+faq-question-6-answer = We will give you advance notice that you need to change the email address of any accounts that are using { -brand-name-relay } aliases.
+faq-question-7-question = What if an email sent to my alias contains an attachment?
+faq-question-7-answer = We now support attachment forwarding. However, there is a 150KB limit for email forwarding using { -brand-name-relay }.
```
pdehaan, line 118: Possibly a Q for @flodolo and @peiying2, but should the `150KB` be a variable? I think I found 4&times; instances in this .ftl file (here, L128, and L154), meaning that if we change that limit to 300KB at some point, we'd invalidate 4 strings per locale. Perhaps possibly using a `{ -email-size-limit }` variable and change it in one place per locale.

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: I know that the official term is "Add-on" based on AMO, Bedrock — but over in about:addons, it says "Extensions".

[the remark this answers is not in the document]
flodolo: "Add-on" is an umbrella term that includes themes, extensions, and plugins. It's also being deprecated in Firefox in favor of "extensions", with a few exception.

privaterelay/locales/en-us/privaterelay/app.ftl
flodolo: Definitely not a bad idea.

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: Extension it is! :)

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -55,8 +56,9 @@ mobileCloseBentoButtonTitle = Close menu
 ## Home Page - https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/
 
 home-hero-headline = Hide your real email address to help protect your identity
-home-hero-copy = { -brand-name-firefox-relay } makes it easy to create aliases, randomly generated email addresses that forward to your real inbox. Use it to protect your online accounts - and your identity - from hackers. Sign in with your { -brand-name-firefox-account } to get started.
-home-hero-cta = { nav-profile-sign-in }
+home-hero-copy = { -brand-name-firefox-relay } makes it easy to create aliases, randomly generated email addresses that forward to your real inbox.
+    Use it to protect your online accounts - and your identity - from hackers. Sign in with your { -brand-name-firefox-account } to get started.
```
peiying2, line 60: This needs to have a string ID, maybe make the string IDs as 
- `home-hero-copy-1 =`
- `home-hero-copy-2 =`

Can you split it up like this? I think it is ok to combine this one. The ones with line breaks can be broken up.

privaterelay/locales/en-us/privaterelay/app.ftl
peiying2: Agreed. Don't want to update the string when the number changes.

privaterelay/locales/en-us/privaterelay/app.ftl
peiying2: Can we do away the CSS code in the string? If not, yeah, if we can consolidate them, that's better/cleaner. so it won't risk being broken up.

privaterelay/locales/en-us/privaterelay/app.ftl
peiying2: I missed this one.

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: Updated.

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: Done.

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: Added the comment:  `# This string is a label for a toggle (on/off) switch`

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
groovecoder: No - neither "Firefox Relay Premium" nor "Relay Premium" will be a product name. "premium" is just a description that the particular feature requires "premium" access, which will be paid.

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: Broken up (and FWIW re-wrote the logic to show/hide this to not be JS dependent)

privaterelay/locales/en-us/privaterelay/app.ftl
maxxcrawford: Removed!

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
flodolo: Let me rephrase the question: is the expectation that Premium will stay in English, or it can be translated without limitations?

privaterelay/locales/en-us/privaterelay/app.ftl
maxxcrawford: I updated this entire string structure. It's now a standard `<strong>` tag, and no HTML is included in the string.

privaterelay/locales/en-us/privaterelay/app.ftl
maxxcrawford: Context for the string. I think we should remove it too. 

image

maxxcrawford: Revised: 
image

privaterelay/locales/en-us/privaterelay/app.ftl
maxxcrawford: Awesome. I'll add it.

maxxcrawford: Added a comment for additional context too. 

```
# The file size limit for emails forwarded through the Relay service. KB is an abbreviation for kilobyte.  
-email-size-limit = 150KB
```

privaterelay/locales/config/pontoon.toml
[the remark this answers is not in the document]
maxxcrawford: This file has been removed.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Brands
+# Dev Note: When adding to this section, use this file for naming conventions: https://github.com/mozilla/bedrock/blob/master/l10n/en/brands.ftl
+
+-brand-name-firefox = Firefox
+-brand-name-firefox-relay = Firefox Relay
+-brand-name-relay = Relay
+-brand-name-firefox-account = Firefox Account
```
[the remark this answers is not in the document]
maxxcrawford: @peiying2 What does `{ -fxaccount-brand-name(nbroken: "frase") }` mean in your example?

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: Translated without limitations. It's to let them know it's extra features.

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
flodolo: https://mozilla-l10n.github.io/localizer-documentation/tools/fluent/basic_syntax.html#terms-and-parameterized-terms

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: Quick update – I reached out to our copywriter to address the entire "Relay Premium" issue. It'll avoid any sort of branded name so the string can be fully translated (besides "Relay"). I'll updated strings tomorrow.

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
peiying2: If used in an image, make sure there is space between the punctuation and the first word after it so it looks like `… so you can`

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,233 @@
...
+how-it-works-step-2-headline = Create a new alias
+how-it-works-step-2-copy = As you browse, the { -brand-name-relay } icon will appear in form fields where sites ask for your email address. 
+    Select it to generate a new, random address that ends in @relay.firefox.com.
+    { -brand-name-relay } will forward messages to the primary email address associated with your account.
+how-it-works-step-3-headline = Manage your account
+how-it-works-step-3-copy = Sign in to the { -brand-name-relay } website to keep track of the aliases you’ve created.
+    If you find that one receives spam or unwanted messages, you can block all messages or even delete the alias, 
+    right from the management page.
+hero-image-copy-trust = Can you even trust this company with your personal email?
+hero-image-copy-unique-html = <strong>Use a unique relay address</strong> for every new account…
```
peiying2, line 84: Should have space between the last word and the punctuation so it looks like `every new account …`

privaterelay/locales/en-us/privaterelay/app.ftl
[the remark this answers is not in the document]
maxxcrawford: Revised! I made the lowercase version the default. There were only 3 places (all within the profile component) with it was supposed to be capitalized.

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+banner-choose-subdomain-copy = You can choose your own domain for "wildcard" aliases. NOTE: YOU CANNOT CHANGE THIS LATER!
+banner-choose-subdomain-label = Your domain is:
+
+# Variables:
+# $subdomain (url) - User-set subdomain
+banner-choose-subdomain-label = You can make up any address @{ $subdomain }
+
+# Variables:
+# $number (number) - User-set subdomain
+remaining-aliases-promo-label = { $number } aliases. You can make unlimited!
```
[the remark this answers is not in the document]
maxxcrawford: Revised (and confirmed locally). Thanks!

privaterelay/locales/en-us/privaterelay/app.ftl
```diff
@@ -0,0 +1,208 @@
...
+
+remaining-aliases-label-html = <span>1</span> remaining alias
+
+# Variables:
+# $number (number) - Number of remaining aliases. This will always be more than 1. 
+remaining-aliases-label-plural-html = <span>{ $number }</span> remaining aliases
```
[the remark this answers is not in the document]
maxxcrawford: Revised/removed!

privaterelay/locales/en-us/privaterelay/app.ftl
maxxcrawford: @peiying2 Can you confirm this is the change you want to see? 
```suggestion
hero-image-copy-unique-html = <strong>Use a unique relay address</strong> for every new account …
```

maxxcrawford:
@peiying2 

- This app.ftl source file is for en or en-US. Make sure to use the correct locale code. The toml file refers to en. The difference between en and en-US is that en-US is treated like a locale, and it allows you to add US specific content while en is the source for all other languages. **I removed the TOML file, and will leave it as en-US. (The website breaks on other languages when I have only a /en/ file.**
- File path/directory structure: Recommend to use: {productname}/locales/{locale}/app.ftl. Right now, locaels/{locale}/privatereplay/app.ftl. If you have more than one file than app.ftl, you can list them all under locales/{locael}/... **I removed the extra {productname} directory. It's now `/privaterelay/locales/en-us/app.ftl`**
- Are we going to have an l10n repo or is localized content going to be part of this product code. **I do not know the answer to this. Do you have a recommendation? @groovecoder Do you have an opinion? @moz-astults?**
- If there is no interest in using vendor for the localization, we can do away toml files which configures which locales goes to where (pontoon or vendor platform). This is a product so it will go to Pontoon, to be localized by the communities. **Copy that. Files removed!**

peiying2:
@maxxcrawford Thanks for the update. I realize the comments in this PR is getting very long, so hopefully your questions are answered and additional comments/suggestion for change is updated too. 

One more thing regarding the changes you made after our suggestions to breaking up sentences that contained <br> markups. After the breakup, each string needs to have a string ID. If you want them to group together, you can add an extra element like same-paragraph-copy-a, same-paragraph-copy-b, same-paragraph-copy-c, for example, so you can tell these strings are grouped together.

maxxcrawford:
> @maxxcrawford Thanks for the update. I realize the comments in this PR is getting very long, so hopefully your questions are answered and additional comments/suggestion for change is updated too.
> 
> One more thing regarding the changes you made after our suggestions to breaking up sentences that contained
> markups. After the breakup, each string needs to have a string ID. If you want them to group together, you can add an extra element like same-paragraph-copy-a, same-paragraph-copy-b, same-paragraph-copy-c, for example, so you can tell these strings are grouped together.

I revised the string you pointed to in this commit: https://github.com/mozilla/fx-private-relay/pull/815/commits/87cfcd9c18ab1534889e95febc706b0cd765c397

maxxcrawford:
> ID `b` contains `html` but ID `a` doesn't. Let's standardize the naming convention.

The second string contains HTML – but the first doesn't. I'll update them both to use `-html`.

privaterelay/locales/en-us/app.ftl
```diff
@@ -0,0 +1,239 @@
...
+how-it-works-step-2-copy = As you browse, the { -brand-name-relay } icon will appear in form fields where sites ask for your email address. 
+    Select it to generate a new, random address that ends in @relay.firefox.com.
+    { -brand-name-relay } will forward messages to the primary email address associated with your account.
+how-it-works-step-3-headline = Manage your account
+how-it-works-step-3-copy = Sign in to the { -brand-name-relay } website to keep track of the aliases you’ve created.
+    If you find that one receives spam or unwanted messages, you can block all messages or even delete the alias, 
+    right from the management page.
+hero-image-copy-trust = Can you even trust this company with your personal email?
+hero-image-copy-unique-html = <strong>Use a unique relay address</strong> for every new account …
+hero-image-copy-protect-html = …so you can <strong>protect your real email</strong> from tracking and spam.
```
peiying2, line 90: should there a space after the ellipsis

peiying2:
1). `en-us` should be `en-US`: lower case is locale code, all cap is country code. 
2). Ellipsis usage: space before it when it ends a sentence, and a space after it when starting a sentence as in [row90](https://github.com/mozilla/fx-private-relay/pull/815/files#diff-e14e064050f9c7c508129a37ef6c96054af00b08bb421ba06ca0adf2378f16d5R90).
3). I assume the decision the repo for l10n is yet to be decided.

privaterelay/locales/en-us/app.ftl
```diff
@@ -0,0 +1,239 @@
...
+-brand-name-firefox-relay = Firefox Relay
+-brand-name-relay = Relay
+-brand-name-firefox-browser = Firefox Browser
+-brand-name-mozilla = Mozilla
+-brand-name-firefox-lockwise = Firefox Lockwise
+-brand-name-firefox-monitor = Firefox Monitor
+-brand-name-pocket = Pocket
+-brand-name-firefox-account = 
+    { $capitalization ->
+        *[lowercase] Firefox account
```
flodolo, line 19: nit: align the parenthesis, not the * sign

```
-brand-name-firefox-account = 
    { $capitalization ->
       *[lowercase] Firefox account
        [uppercase] Firefox Account
    }
```

privaterelay/locales/en-us/app.ftl
```diff
@@ -0,0 +1,239 @@
...
+        [uppercase] Firefox Account
+    }
+    
+# The file size limit for emails forwarded through the Relay service. KB is an abbreviation for kilobyte.  
+-email-size-limit = 150KB
```
flodolo, line 24: This doesn't really scale.

One issue is that terms should be used for things like brand/product names, where you might need private attributes. For this you can use a normal message (standard messages can be referenced in other messages, but also used stand-alone in code).

Second issue: what happens if you change the value? You need to update this string, which requires a new ID, and breaks all existing translations (which will require new IDs).

This should be

```
# Variables:
#   $size (number): maximum size for attachments
email-size-limit = { $size } KB
```

Even better, so we don't break if we switch to some MB.

```
# String used to display the attachment limit, e.g. 150 KB
# Variables:
#  $size (number): maximum size for attachments
#  $unit (string): unit of measurement (e.g. KB)
email-size-limit = { $size } { $unit }
unit-kb = KB
```

Questions on the English:
* Shouldn't there be a space between number and unit?
* `kB` or `KB`?

This should also move closer to the string that uses it.

privaterelay/locales/en-us/app.ftl
```diff
@@ -0,0 +1,239 @@
...
+# Variables:
+#   $url (url) - https://addons.mozilla.org/firefox/addon/private-relay/
+#   $attrs (string) - specific attributes added to external links
+faq-question-1-answer-b-html = If you see a broader problem of unwanted email from all of your aliases, please <a href="{ $url }" { $attrs }>report this to us</a> so we can consider adjusting the SES spam thresholds for this service. If you report these as spam, your email provider will see { -brand-name-relay } as the source of spam, not the original sender.
+faq-question-2-question = Why won’t a site accept my { -brand-name-relay } alias?
+
+# Variables:
+#   $url (url) - https://addons.mozilla.org/firefox/addon/private-relay/
+#   $attrs (string) - specific attributes added to external links
+faq-question-2-answer-html = Some sites may not accept an email address that includes a subdomain (ie, the “relay” portion of @relay.firefox.com) and others have stopped accepting all addresses except those from Gmail, Hotmail, or Yahoo accounts.
```
flodolo, line 108: ie or i.e.?

privaterelay/locales/en-us/app.ftl
maxxcrawford: > Shouldn't there be a space between number and unit? 
Yes. Adding. 

> `kB` or `KB`?

KB because this file size limit is in kilobytes, rather than bits. (@groovecoder – Can you confirm?) 

---

I want to use the two-argument variable string. How do I reference `unit-kb`? In the example below I set the unit value in the `HTML`. 

**In the FTL:** 
```
faq-question-7-answer = We now support attachment forwarding. However, there is a { -email-size-limit } limit for email forwarding using { -brand-name-relay }. Any emails larger than { -email-size-limit } will not be forwarded.
```

**In the HTML:** 
```
<p class="faq-answer">{% ftlmsg 'faq-question-7-answer' %}</p>
```

```
<p class="faq-answer">{% ftlmsg 'faq-question-7-answer' size='150' unit='KB' %}</p>
```

privaterelay/locales/en-us/app.ftl
maxxcrawford: "i.e." is the correct usage from what I can gather.

maxxcrawford: (Updated to i.e.)

privaterelay/locales/en-us/app.ftl
maxxcrawford: I passed arguments everywhere the string is referenced. I have the `unit-kb` variable defined in the FTL file, but not used anywhere. Please advise.

flodolo: > I want to use the two-argument variable string. How do I reference `unit-kb`? In the example below I set the unit value in the `HTML`.

I'm afraid I'm out of my depth when it comes to the actual code, and I'm not sure who else is familiar with this library at this point.

Worst case scenario, I would `KB` to `unit`, because having the message reference (i.e. `email-size-limit = { $size } { mb-unit }`) would have the same scalability/maintainability problem.

privaterelay/locales/en-us/app.ftl
maxxcrawford: I added a space. Happy to match the usage you defined: 

> Ellipsis usage: space before it when it ends a sentence, and a space after it when starting a sentence as in row90.

maxxcrawford:
> 1). `en-us` should be `en-US`: lower case is locale code, all cap is country code.
> 2). Ellipsis usage: space before it when it ends a sentence, and a space after it when starting a sentence as in [row90](https://github.com/mozilla/fx-private-relay/pull/815/files#diff-e14e064050f9c7c508129a37ef6c96054af00b08bb421ba06ca0adf2378f16d5R90).
> 3). I assume the decision the repo for l10n is yet to be decided.

1. Updated the folder to the `en-US` correct capitalization 
2. Updated ellipsis spacing. Quick reminder that this is part of an outstanding issue (#821) with the strings being embedded in the images currently. 
3. No decision yet. As this isn't immediately being localized, I think we can defer this decision immediately.

privaterelay/locales/en-us/app.ftl
flodolo: Not sure if [this](https://django-ftl.readthedocs.io/en/latest/usage.html#lazy-translations) helps (can you store the unit-kb translation in variable, and pass that as the argument? I have no clue…)

maxxcrawford: Okay – Talking with the team, this is okay as is. We are not planning on changing the limit any time in the near/far future. It is not on our roadmap.

maxxcrawford:
@flodolo @peiying2 I think this is good to merge in, unless y'all wanna review anything further!

peiying2:
> @flodolo @peiying2 I think this is good to merge in, unless y'all wanna review anything further!

I don't  have any additional comment. I want to make sure the locale name is fixed. It should be: en-US.

privaterelay/locales/en-us/app.ftl
flodolo: Not the first time we've been bitten by this approach…

Out of curiosity, the suggestion I linked to in the previous comment doesn't work? I'd still prefer to have that fully localizable and future-proof, if possible. If not, you should remove the "KB" string, since nothing is using it.

maxxcrawford:
> > @flodolo @peiying2 I think this is good to merge in, unless y'all wanna review anything further!
> 
> I don't have any additional comment. I want to make sure the locale name is fixed. It should be: en-US.

Great catch, @peiying2. I pushed an update to revise it. 

https://github.com/mozilla/fx-private-relay/blob/813-add-fluent/privaterelay/locales/en-US/app.ftl

privaterelay/locales/en-us/app.ftl
maxxcrawford: Oh! Yeah – the previous comment didn't work because this is visible without any user interaction. The lazy function is for modals and things that only show when the user does an action to display them. However, this could be used for other strings in this application (Namely – all the strings used in `/includes/modal.html` and some in the `profile.html`). 

I'll file an issue to revist this. 

(also I'm removing that unused variable for now)

maxxcrawford:
@groovecoder I think we're ready to merge – could you please re-review one more time? (I know we're already approved by you, but wanted to double confirm)

privaterelay/views.py
```diff
@@ -37,7 +37,6 @@
 )
 from emails.utils import get_post_data_from_request
 
-
```
groovecoder, line 40: F this line in particular.

groovecoder:
Code and spot-check looks good. I found an un-related bug I'll file separately.

Issue #808, opened 2021-04-13
Add create and delete functionality for subdomain aliases


say-yawn:
# About This Issue
For subdomain alias (email aliasing with a subdomain) allow alias to be created and deleted on the dashboard.

# Acceptance Criteria
- [ ] Domain Address impromptu creation (e.g. alias generated because email was received with the given alias) with `first_email_at` field populated
- [ ] Domain Address random creation (like Relay Address)
- [ ] Domain Address named creation (e.g. alias generated with specific name on add-on or dashboard)
- [ ] Deleted Domain Address becomes a DeletedAddress where the entire address is hashed for the `address` field

PR #827, opened 2021-05-04
Handle errors properly on Emails

say-yawn:
# About this PR
Fast follow to fix some dev tested errors on #808. Some fixes include:
- adding helpful messages on the disabled button
Screen Shot 2021-05-04 at 14 23 36

- making the errors be more helpful
- formatting errors on the profile page (has been broken since this [change](https://github.com/mozilla/fx-private-relay/commit/0b534d452255aadf23fa03a4b6e3cfabfdc4dbce#diff-a5ea33c888430601a659bcbef2da1944097953737f2606282efc13ca6e5fbaba))
image (16)

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -16,7 +16,7 @@ emails_config = apps.get_app_config('emails')
 
 BounceStatus = namedtuple('BounceStatus', 'paused type')
 
-NOT_PREMIUM_USER_ERR_MSG = 'You must be a premium subscriber to set a subdomain.'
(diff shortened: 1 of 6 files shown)


emails/models.py
```diff
@@ -262,7 +267,7 @@ def make_domain_address(user_profile, address=None, made_via_email=False):
             )
 
         address_contains_badword = False
-        if address is None:
+        if not address:
```
say-yawn, line 270: Doing `not address` because address value can be an empty string which would fail this check.

say-yawn:
Merge now. Fix later.

PR #828, opened 2021-05-06
Handle heroku mx record issue

say-yawn:
# About this PR
While testing on dev environment we found that emails were no longer being relayed due to `mail` added on the email domain if settings was set to DEBUG see [here](https://github.com/mozilla/fx-private-relay/blame/main/privaterelay/templatetags/relay_tags.py#L18-L19).
This caused the failure [here](https://github.com/mozilla/fx-private-relay/blob/main/emails/views.py#L399) since the `urlparse(settings.SITE_ORIGIN).netloc` does not contain the `mail` but the `domain_portion` resulting in [this Sentry error](https://sentry.prod.mozaws.net/operations/fx-private-relay-dev/issues/11253393/?query=is%3Aunresolved). Fix is to add `mail` only when the service is running on Heroku because we can’t publish MX records on Heroku and blocks the ability to test subdomain feature on dev environment.

# Acceptance Criteria
- [ ] Email is relayed via RelayAddress
- [ ] `mail` on email domain if the service is running on Heroku
- [ ] Email is relayed via impromptu DomainAddress generation

# STR
## Check that RelayAddress works with premium features turned off
1. On the .env file set the `PREMIUM_ENABLED` to `False`
2. Run the service and send an email to existing RelayAddress
3. Check that the email is relayed via the RelayAddress and that the number of emails forwarded is incremented
## Check that RelayAddress works with premium features turned on
1. On the .env file set the `PREMIUM_ENABLED` to `True`
2. Run the service and send an email to existing RelayAddress
3. Check that the email is relayed via the RelayAddress and that the number of emails forwarded is incremented
## Check that email DomainAddress works with premium features turned on
1. Leaving the `PREMIUM_ENABLED` to `True` send an email to any DomainAddress (impromptu generation)
2. Check that the email is relayed via the DomainAddress, DomainAddress with the impromptu generation is created on the Relay dashboard, and that the number of emails forwarded is incremented

#
Related to #808


emails/tests/utils_tests.py
```diff
@@ -59,3 +65,13 @@ def test_generate_relay_From_with_linebreak_chars(self):
...
+
+    @override_settings(ON_HEROKU=True, SITE_ORIGIN='https://test.com')
+    def test_get_email_domain_from_settings_on_heroku(self):
+        email_network_locality = get_email_domain_from_settings()
+        assert 'mail.test.com' == email_network_locality
+
+    @override_settings(ON_HEROKU=False, SITE_ORIGIN='https://test.com')
+    def test_get_email_domain_from_settings_not_on_heroku(self):
+        email_network_locality = get_email_domain_from_settings()
+        assert 'test.com' == email_network_locality
```
groovecoder, line 77: `locality` is a bit confusing here? Can it be just `email_domain`

emails/views.py
```diff
@@ -396,7 +396,8 @@ def _get_domain_address(to_address, local_portion, domain_portion):
 def _get_address(to_address, local_portion, domain_portion):
```
groovecoder, line 396: Could we include some tests for this `_get_address` function while we're in here?

groovecoder (changes requested):

emails/views.py
say-yawn: Done 8b54661

emails/tests/utils_tests.py
say-yawn: Done 2ad82db

PR #840, opened 2021-05-25
Fix #817 - Remove map function, set URLs directly

### LINKED ISSUE #817: Please update CSP with new avatar URL
Reported on Slack:

Relay is blocking the avatar URL from FxA:

> Content Security Policy: The page’s settings blocked the loading of a resource at https://profile.accounts.firefox.com/v1/avatar/w (“img-src”).

Resulting in mis-rendering of the account menu in the top left. 
image

The fix is to add `profile.accounts.firefox.com` to the img-src CSP rule

maxxcrawford:
Testing steps:
- Open Relay stage website (https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/)
- Login with FXA
- Expected: Profile image should display correctly

Note – Be sure to NOT have a profile image set.

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -55,17 +55,13 @@ SESSION_COOKIE_SECURE = config(
 )
 # maps fxa profile hosts to respective avatar hosts for CSP
 AVATAR_IMG_SRC_MAP = {
-    'https://stable.dev.lcip.org/profile/v1':      [
-        'stable.dev.lcip.org',
-        'https://stable.dev.lcip.org/profile/v1/avatar/w',
-    ],
     'https://profile.stage.mozaws.net/v1':      [
         'mozillausercontent.com',
-        'https://profile.stage.mozaws.net/v1/avatar/w',
+        'https://profile.stage.mozaws.net',
     ],
     'https://profile.accounts.firefox.com/v1':  [
         'firefoxusercontent.com',
-        'https://profile.accounts.firefox.com/v1/avatar/w',
+        'https://profile.accounts.firefox.com',
     ],
 }
 AVATAR_IMG_SRC = AVATAR_IMG_SRC_MAP[config(
@@ -83,7 +79,7 @@ CSP_SCRIPT_SRC = (
 )
 CSP_STYLE_SRC = ("'self'",)
 CSP_IMG_SRC = ["'self'"] + AVATAR_IMG_SRC
-REFERRER_POLICY = 'strict-origin-when-cross-origin' \
+REFERRER_POLICY = 'strict-origin-when-cross-origin'
 
 ALLOWED_HOSTS = []
 DJANGO_ALLOWED_HOST = config('DJANGO_ALLOWED_HOST', None)
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -55,9 +55,18 @@ SESSION_COOKIE_SECURE = config(
 )
 # maps fxa profile hosts to respective avatar hosts for CSP
 AVATAR_IMG_SRC_MAP = {
-    'https://stable.dev.lcip.org/profile/v1':   'stable.dev.lcip.org',
-    'https://profile.stage.mozaws.net/v1':      'mozillausercontent.com',
-    'https://profile.accounts.firefox.com/v1':  'firefoxusercontent.com',
+    'https://stable.dev.lcip.org/profile/v1':      [
+        'stable.dev.lcip.org',
+        'https://stable.dev.lcip.org/profile/v1/avatar/w',
+    ],
+    'https://profile.stage.mozaws.net/v1':      [
+        'mozillausercontent.com',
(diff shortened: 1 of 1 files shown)


privaterelay/settings.py
```diff
@@ -53,15 +53,6 @@
 SESSION_COOKIE_SECURE = config(
     'DJANGO_SESSION_COOKIE_SECURE', False, cast=bool
 )
-# maps fxa profile hosts to respective avatar hosts for CSP
-AVATAR_IMG_SRC_MAP = {
-    'https://stable.dev.lcip.org/profile/v1':   'stable.dev.lcip.org',
-    'https://profile.stage.mozaws.net/v1':      'mozillausercontent.com',
-    'https://profile.accounts.firefox.com/v1':  'firefoxusercontent.com',
```
maxxcrawford, line 60: If it makes more sense to nest another array here, we can try that? 

```
'https://profile.accounts.firefox.com/v1':  [
   'firefoxusercontent.com',
   'https://profile.accounts.firefox.com/v1/avatar/w',
]
```

privaterelay/settings.py
```diff
@@ -75,7 +66,10 @@
 CSP_STYLE_SRC = ("'self'",)
 CSP_IMG_SRC = (
     "'self'",
-    AVATAR_IMG_SRC,
+    'mozillausercontent.com',
+    'https://profile.stage.mozaws.net/v1/avatar/w',
+    'firefoxusercontent.com',
+    'https://profile.accounts.firefox.com/v1/avatar/w',
```
maxxcrawford, line 72: The security risk here is allowing the stage URLs to operate on production.

privaterelay/settings.py
```diff
@@ -55,9 +55,18 @@
 )
 # maps fxa profile hosts to respective avatar hosts for CSP
 AVATAR_IMG_SRC_MAP = {
-    'https://stable.dev.lcip.org/profile/v1':   'stable.dev.lcip.org',
-    'https://profile.stage.mozaws.net/v1':      'mozillausercontent.com',
-    'https://profile.accounts.firefox.com/v1':  'firefoxusercontent.com',
+    'https://stable.dev.lcip.org/profile/v1':      [
+        'stable.dev.lcip.org',
+        'https://stable.dev.lcip.org/profile/v1/avatar/w',
+    ],
```
groovecoder, line 61: We should be able to remove all mentions of `stable` now that we've converted all environments to stage.
```suggestion
```

privaterelay/settings.py
```diff
@@ -73,11 +82,8 @@
     'https://www.google-analytics.com/',
 )
 CSP_STYLE_SRC = ("'self'",)
-CSP_IMG_SRC = (
-    "'self'",
-    AVATAR_IMG_SRC,
-)
-REFERRER_POLICY = 'strict-origin-when-cross-origin'
+CSP_IMG_SRC = ["'self'"] + AVATAR_IMG_SRC
+REFERRER_POLICY = 'strict-origin-when-cross-origin' \
```
groovecoder, line 86:
```suggestion
REFERRER_POLICY = 'strict-origin-when-cross-origin'
```

privaterelay/settings.py
```diff
@@ -55,9 +55,18 @@
...
-    'https://stable.dev.lcip.org/profile/v1':   'stable.dev.lcip.org',
-    'https://profile.stage.mozaws.net/v1':      'mozillausercontent.com',
-    'https://profile.accounts.firefox.com/v1':  'firefoxusercontent.com',
+    'https://stable.dev.lcip.org/profile/v1':      [
+        'stable.dev.lcip.org',
+        'https://stable.dev.lcip.org/profile/v1/avatar/w',
+    ],
+    'https://profile.stage.mozaws.net/v1':      [
+        'mozillausercontent.com',
+        'https://profile.stage.mozaws.net/v1/avatar/w',
```
groovecoder, line 64: When I spot-checked on local, the avatar image was `https://profile.stage.mozaws.net/v1/avatar/l` because my stage fxa name starts with `L`. So the path in these directive values needs to be at *most* `/v1/avatar`, but even more flexible (and still safe, assuming we trust ALL images from `profile.stage.mozaws.net`) is to omit the path completely and allow images from the entire origin:
```suggestion
        'https://profile.stage.mozaws.net',
```

privaterelay/settings.py
```diff
@@ -55,9 +55,18 @@
...
+    ],
+    'https://profile.accounts.firefox.com/v1':  [
+        'firefoxusercontent.com',
+        'https://profile.accounts.firefox.com/v1/avatar/w',
```
groovecoder, line 68:
```suggestion
        'https://profile.accounts.firefox.com',
```

groovecoder (changes requested):

privaterelay/settings.py
maxxcrawford: Great call out. Tested with a different letter. 
image

PR #859, opened 2021-06-09
Update dockerfile for npm installer

maxxcrawford:


moz-astults:
@maxxcrawford  Do we want to pin a specific version of npm/node? Or is it ok to be installing the latest every time? I mean, `npm`  dependencies can be kinda crazy, but other than that is there a reason we shouldn't use this?

Dockerfile
```diff
@@ -1,6 +1,8 @@
 FROM python:3.7.9
 
 RUN apt-get update && apt-get -y install libpq-dev
+RUN apt-get install -y nodejs
+RUN apt-get install -y npm
```
maxxcrawford, line 5: @moz-astults Do you have an opinion on which version? I have the following locally. 

```
node: v10.20.1
npm: 6.14.4
```

maxxcrawford:
```suggestion
RUN apt-get install -y nodejs=v10.20.1
RUN apt-get install -y npm=6.14.4
```

maxxcrawford, line 5:
```suggestion
RUN apt-get install -y nodejs=v10.20.1
RUN apt-get install -y npm=6.14.4
```

moz-astults: @maxxcrawford I don't - I would generally say "latest" so we avoid security vulnerability issues, but I don't know if there are any version specific dependencies in the application? Sorry I don't really have enough context to say one way or the other. Are there any concerns you have with using `npm`?

Just curious, are we only using the output of `npm install` for use with the django app? If so, maybe we could use a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds) and do the `npm` bits in a separate container to keep the final image as small as possible?

Dockerfile
```diff
@@ -15,12 +17,20 @@ USER app
 COPY --chown=app ./requirements.txt /app/requirements.txt
 RUN pip install -r requirements.txt
 
+RUN echo "Node: " && node -v
+RUN echo "NPM: " && npm -v
+
+COPY package*.json ./
+RUN npm install
```
groovecoder, line 24: Is this enough? Does the `django-gulp` logic hook right into the existing `collectstatic` step?

maxxcrawford: cc/ @moz-astults 

Quick context for what this patch resolves. We were getting [this error](https://app.circleci.com/pipelines/github/mozilla/fx-private-relay/1640/workflows/62add57c-67ef-42ab-a409-60d71c191da8/jobs/3102) from Circle CI: 

```bash
/bin/sh: 1: gulp: not found
CommandError: Command 'gulp build --cwd /app --production' returned non-zero exit status 127.
The command '/bin/sh -c python manage.py collectstatic --no-input -v 2' returned a non-zero code: 1

Exited with code exit status 1
```

This lead me down the rabbit hole of making sure node/npm is included in the Container file.

maxxcrawford: Also we can totally remove the `echo` lines. That was just for debugging.

maxxcrawford: > Does the django-gulp logic hook right into the existing collectstatic step?

@groovecoder Yeah. This resolves the original error linked above.

Dockerfile
moz-astults: Also, if you don't care about the minor version being pinned, you could start your multi-stage build with `node:12` (or whatever major version you'd like) which will make sure you're pulling the latest for that major version while still keeping it pinned.

maxxcrawford: (Catching up from offline discussion with @moz-astults) 

We tried the multi-stage build, but node/gulp is required in the same container as the core app. We reverted back to this method, however – instead of `apt-get update`, we use [nodesource](https://github.com/nodesource/distributions/blob/master/README.md#debinstall) to install Node (includes `npm`). We're using `14.x` to get the latest stable in that major version.

moz-astults (approved):
LGTM - we were close with the multi-stage build, but unless we break the dependency of django gulp this is probably a good approach for now.

groovecoder (approved):

PR #882, opened 2021-06-15
DRAFT/WIP: django-hosts to serve different views for different hosts

groovecoder:
This is very raw code to simply show how we might use `django-hosts` to serve both relay.firefox.com and privacy.mozilla.org from the same app code.

To see this locally:

1. Check out this branch
2. Copy the necessary new lines from `.env-dist` to your local `.env`
3. Go to http://127.0.0.1:8000/
   * Everything should work as before
4. Go to http://privacy.localtest.me:8000/
   * Should see a different page with just "privacy.mozilla.org" in the content

Still to do/check:

Can we ... 
1. Set up a separate FXA OAuth client for http://privacy.localtest.me:8000/
2. Add that client to the same django-allauth db
3. Bind that client to a new http://privacy.localtest.me:8000/ site in the django admin
4. Have all that work

Also still to decide:
Is this worth the trouble? Or should we create a completely new Django app for privacy.mozilla.org?


groovecoder:
Closing this as I'm 90% sure we'll go another way and stand up a separate Django code-base, project, and app for privacy.mozilla.org.

PR #885, opened 2021-06-16
 fix #874: run gulp in docker; not during collectstatic 

### LINKED ISSUE #874: Update Docker containers to be multi-part
Currently we have one big container with Python AND Node. We should switch this out to multi-part per @moz-astults recommendation.  

To do this, we need to move `django-gulp` to be `DEV`mode only.  

In the first container (`node`), we should install npm and run `npm install` and THEN run `gulp` to compile all the CSS. Copy over the compiled filed and then move on to the Python. 

Note – @moz-astults made great progress down this path and we had to abandon it. See this commit for how it should work: https://github.com/mozilla/fx-private-relay/pull/859/commits/2429d53627fafcf8a3ecc7c8358afa9a8b85321f

groovecoder:


Dockerfile
```diff
@@ -1,8 +1,14 @@
+FROM node:14 AS builder
+WORKDIR /app
+COPY package*.json ./
+COPY gulpfile.js ./
+COPY static ./static/
+RUN npm install
+RUN ./node_modules/.bin/gulp build
+
 FROM python:3.7.9
```
moz-astults, line 9: Is there a reason we need to specify `python:3.7.9` or can we just specify `python:3.7` (currently pulls `3.7.10`) so we keep up to date on the patch version? 

FWIW the image still builds successfully when specifying `python:3.7`

Otherwise, LGTM. I'm not sure how to test the docker image built locally though since it doesn't seem to respond without environment variables specified, so will we need to push this to stage to validate?

groovecoder: No specific reason against 3.7.10. But as a general tactic I like to pin to very specific versions of dependencies to lower the risk of an un-planned dependency update breaking something. In theory, all 3.7.x should be compatible, but we already [ran into an issue from 3.7.6 to 3.7.9](https://github.com/mozilla/fx-private-relay/pull/776#issue-562866493) in Relay. 😢 

I know dependabot watches our `requirements.txt` and `package.json` dependency versions for sec updates ... I wonder if it also watches `Dockerfile` `FROM` dependencies?

moz-astults: Oh wow, that's a huge bummer. Thanks for clarifying! Not totally sure on dependabot looking at `Dockerfile` but its something I could look into!

moz-astults (approved):

PR #889, opened 2021-06-17
Add UI: Unlimited Aliases 

maxxcrawford:
This PR includes: 

- Filter Bar
- Revised Alias Feed

TODO: 
- [x] Hide unused filter UI (Categories, Date Picker)


static/js/filter.js
```diff
@@ -0,0 +1,7 @@
+(function() {
+	"use strict";
+
+
```
groovecoder, line 4: [nice](https://www.tate.org.uk/art/artworks/li-monochrome-white-painting-t11871).

privaterelay/templates/includes/dashboard-filter.html
```diff
@@ -0,0 +1,59 @@
+{% load static %}
+{% load ftl %}
+{% ftlconf bundle='privaterelay.ftl_bundles.main' %}
+
+<div class="c-filter">
+    <div class="c-filter-search">
+        <div class="c-filter-icon">
+            <img src="/static/images/icon-search.svg" alt="">
```
groovecoder, line 8: Just remember to change all these to `{% static %}` tags.

privaterelay/templates/includes/dashboard-filter.html
```diff
@@ -0,0 +1,59 @@
...
+
+    </div>
+    <div class="c-filter-cateogry">
+        <div class="c-filter-icon">
+            <img src="/static/images/icon-filter.svg" alt="">
+        </div>
+    </div>
+    <div class="c-filter-action">
+        <!-- Button -->
+        <form action="/emails/" class="dash-create" method="POST">
```
groovecoder, line 27: Ooo, and let's fix this action to `{% url  'emails-index' %}` tag.

privaterelay/templates/profile.html
```diff
@@ -67,226 +72,229 @@
...
+                    <input data-label="" type="text" maxlength="50" value="" aria-label="Edit the label for this alias" class="relay-email-address-label ff-Met">
+                    <span class="saved-confirmation">{% ftlmsg 'profile-label-saved' %}</span>
+                    <span class="input-error"></span>
                   </div>
-                  <button class="toggle-details-visibility card-small-text">
-                    <span class="show-details-label detail-label flx">{% ftlmsg 'profile-label-details-show' %}</span>
-                    <span class="hide-details-label detail-label flx">{% ftlmsg 'profile-label-details-hide' %}</span>
-                    <span class="detail-label-icon"></span>
+
+                  <button title="Click to copy"
```
groovecoder, line 111: String needs to be internationalized.

privaterelay/templates/profile.html
```diff
@@ -67,226 +72,229 @@
...
+
+
+              </div>
+              <!--end enable/disable toggle -->
+
+              <div class="c-alias-name xflx xflx-col">
+                <div class="relay-email-row card-top-row xflx xflx-row inset">
+                  <!-- alias label -->
+                  <div class="additional-notes">
+                    <input data-label="" type="text" maxlength="50" value="" aria-label="Edit the label for this alias" class="relay-email-address-label ff-Met">
```
groovecoder, line 216: i18n this `aria-label`

privaterelay/templates/profile.html
```diff
@@ -67,226 +72,229 @@
...
+                </div>
+
+
+                <!-- delete email address -->
+                <div class="column-delete">
+                  <form action="{% url 'emails-index' %}" method="POST" class="delete-email-form xflx jst-cntr al-cntr" data-delete-relay="{{ relay_address.address }}@{{ RELAY_DOMAIN }}">
+                    <input type="hidden" name="method_override" value="DELETE">
+                    <input type="hidden" name="api_token" value="{{ user_profile.api_token }}">
+                    <input type="hidden" name="relay_address_id" value="{{ relay_address.id }}">
+                    <button type="submit" name="delete" value="" class="dashboard-delete-button delete-relay-email-button" aria-label="Delete this alias">
```
groovecoder, line 288: i18n this `aria-label`

static/js/filter.js
maxxcrawford: 👁️ 👄 👁️  
_Write clean code™_

groovecoder (commented):
I know this is still draft/WIP but I left a few comments of things to make sure we do before we finalize it ... mostly:

* Make sure we use `{% static %}` everywhere we can instead of hard-coding static urls
* Likewise, make sure we use `{% url %}` everywhere we can instead of hard-coding urls
* Use `{% ftlmsg %}` tags everywhere we can

In addition, as we embrace i18n in the code-base more, we should consider if we want to switch from `top`, `left`, etc. to css logical properties and values to help support more RTL languages.

privaterelay/templates/profile.html
```diff
@@ -67,226 +72,229 @@
...
+                      <span class="c-alias-data-value">{{ request.user.email }}</span>
+                    </div>
+                  </div>
+
+                  <!-- created by date -->
+                  <div class="c-alias-data created-by">
+                    <img class="" src="/static/images/icon-calendar-gray.svg" alt="">
+                    <div>
+                      <span class="c-alias-data-label">{% ftlmsg 'profile-label-created' %}</label>
+                      <span class="c-alias-data-value">{{ relay_address.created_at|date:"M j, Y" }}</span>
```
pdehaan, line 276: Not sure if we can localize the date formats to work internationally.

static/js/app.js
```diff
@@ -53,12 +53,12 @@ async function updateEmailForwardingPrefs(submitEvent) {
       if (prefToggle.value === "Enable") {
         prefToggle.title = "Disable email forwarding for this alias";
         toggleLabel.textContent = "forwarding";
-        wrappingEmailCard.classList.add("card-enabled");
+        wrappingEmailCard.classList.add("is-enabled");
         return prefToggle.value = "Disable";
       } else if (prefToggle.value === "Disable") {
         prefToggle.title="Enable email forwarding to this alias";
```
pdehaan, line 59: Should these be localized?

static/js/filter.js
```diff
@@ -1,7 +1,95 @@
...
-
     
-	
-	
-})();
\ No newline at end of file
+	const filterEmailAddresses = [];
+	const filterAliasLabels = [];
+	const aliasesWithLabelsCollection = [];
+    const aliasCollection = [];
```
groovecoder, line 7: preceding lines are indented with 2 spaces, these are 4?

static/js/filter.js
```diff
@@ -1,7 +1,95 @@
...
+    const aliases = document.querySelectorAll(".c-alias");
+
+    const filterLabelTotalCases = document.querySelector(".js-filter-case-total");
+    const filterLabelVisibleCases = document.querySelector(".js-filter-case-visible");
+
+    const filterResetButton = document.querySelector(".js-filter-reset")
+
+    const filterInput = document.querySelector(".js-filter-search-input");
```
groovecoder, line 15: Looks like these are all single elements, so these could change from class names to element IDs and the JS could be `document.getElementById` ?

static/js/filter.js
```diff
@@ -1,7 +1,95 @@
...
+
+	function filterInputWatcher(e) {
+		const query = e.target.value.toLowerCase();
+
+        // if (query === "") {
+        //     resetFilter();
+        //     return;
+        // }
```
groovecoder, line 23: be sure to delete commented code before merge.

static/js/filter.js
```diff
@@ -1,7 +1,95 @@
...
+            }
+        }
+
+        // TODO: Map the entire list of aliases better so we dont have two seperate DOM collection arrays to loop through.
+        for (const alias of matchListAliasLabels) {
+            let index = filterAliasLabels.indexOf(alias);
+            if (index >= 0) {
+                aliasesWithLabelsCollection[index].style.display = "block";
+            }
+        }
```
groovecoder, line 53: Could try to refactor these 2 functions into a single function that accepts `matchListEmailAddresses, aliasCollection` and then `matchListAliasLabels, aliasesWithLabelsCollection` as arguments?

But that might end up being more complex/confusing to read?

static/js/filter.js
```diff
@@ -1,7 +1,95 @@
...
+            if (aliasLabel) {
+                aliasesWithLabelsCollection.push(alias);
+                filterAliasLabels.push( aliasLabel.dataset.label.toString().toLowerCase() );
+            }
+
+		});
+
+		// // Set ##/## in filter input field to show how many aliases have been filtered.
+        filterLabelVisibleCases.textContent = aliases.length;
+        filterLabelTotalCases.textContent = aliases.length;
```
groovecoder, line 77: Is this part done? Should these be different numbers?

groovecoder (changes requested):

privaterelay/templates/includes/dashboard-filter.html
```diff
@@ -0,0 +1,66 @@
...
+                class="btn btn-blue--ghost"
+                data-event-label="Upgrade"
+            >
+              <span class="generate-new-alias-text">
+                {% ftlmsg 'profile-label-upgrade' %}
+              </span>
+            </a>
+
+            {% else %}
+            <button
```
groovecoder, line 50: Because this is in `dashboard-filter.html` and `profile.html` has this:

```
        {% if relay_addresses|length == 0 %}
          {% include "includes/dashboard_onboarding.html" %}
        {% else %}
          {% include "includes/dashboard-filter.html" %}
        {% endif %}
```

It means that there's no "Generate new alias" button on the dashboard when someone has 0 aliases. 😢 

image

groovecoder (changes requested):

privaterelay/templates/includes/dashboard-filter.html
maxxcrawford: Ended up moving these to be displayed via CSS to account for hover-state colors. We also could import the actual SVG, rather than the image tag. Not sure the best way to manage that. (Would at least only require one SVG file per icon instead of two)

privaterelay/templates/includes/dashboard-filter.html
```diff
@@ -0,0 +1,66 @@
...
+    </div>
+    <div class="c-filter-action">
+        <!-- Button -->
+        <form action="/emails/" class="dash-create" method="POST">
+            <input type="hidden" name="api_token" value="{{ user_profile.api_token }}">
+            {% if user_profile.at_max_free_aliases and not user_profile.has_unlimited %}
+            <a
+                href="{{ settings.FXA_SUBSCRIPTIONS_URL }}/products/{{ settings.PREMIUM_PROD_ID }}?plan={{ settings.PREMIUM_PRICE_ID }}"
+                target="_blank"
+                rel="noopener noreferrer""
```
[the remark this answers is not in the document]
maxxcrawford: ⚡ removed!

privaterelay/templates/includes/dashboard-filter.html
maxxcrawford: There's an entire new on-boarding flow I need to layout. We need to fix this pronto.

privaterelay/templates/includes/dashboard-filter.html
```diff
@@ -0,0 +1,66 @@
...
+            >
+              <span class="generate-new-alias-text">
+                {% ftlmsg 'profile-label-upgrade' %}
+              </span>
+            </a>
+
+            {% else %}
+            <button
+                class="blue-btn-states dash-create-new-relay flx al-cntr jst-cntr"
+                title="Generate new alias"
```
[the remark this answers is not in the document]
maxxcrawford: Yup. We even have the string ready to go. `profile-label-generate-new-alias`

privaterelay/templates/includes/dashboard-filter.html
```diff
@@ -0,0 +1,66 @@
...
+                type="submit"
+                value="Generate new alias"
```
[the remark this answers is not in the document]
maxxcrawford: See above! 🛠️

static/js/app.js
maxxcrawford: Yes. Filing this as a seperate issue.

maxxcrawford: Filed! https://github.com/mozilla/fx-private-relay/issues/906

static/js/filter.js
maxxcrawford: Actual answer: We need a linter™.

static/js/filter.js
maxxcrawford: I'm wanting to use the `js-` naming convention to make it easy to identify JavaScript enabled items in the code. This may need a larger conversation, bc I agree with you on leveraging `getElementById` over `querySelector`.

static/js/filter.js
maxxcrawford: My TODO really states it – we're actually looping through two different arrays. There's an entirely better way to do this. I would recommend leaving it as is for the moment. 

Note that I did find an edge case in the logic where if you name two aliases the same thing, it will only count/display the first one. That alone is worth fixing this up into a more flexible system. (Maybe a `Map`)?

static/js/filter.js
maxxcrawford: It's only on init – it should display `7/7` if you have 7 aliases.

maxxcrawford:
@groovecoder Fixed everything but the on-boarding. Should that be blocking (probably?) or file a separate issue?

privaterelay/templates/profile.html
```diff
@@ -67,226 +72,229 @@
...
+                <div class="c-alias-data-collection">
                   <!-- forwarding message -->
-                  <section class="relay-email-column hide-mobile details-column flx flx-col">
-                    <span class="relay-email-label email-forwarding-label card-small-text">{% ftlmsg 'profile-label-forward-emails' %}</span>
-                    <span class="text-dark demi primary-email-address card-small-text">{{ request.user.email }}</span>
-                  </section>
+                  <div class="c-alias-data forward ">
+                    <img class="" src="/static/images/icon-email.svg" alt="">
+                    <div>
+                      <span class="c-alias-data-label">{% ftlmsg 'profile-label-forward-emails' %}</label>
```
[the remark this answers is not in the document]
maxxcrawford: Great catch. Resolved!

privaterelay/templates/profile.html
gregtatum: Providing a pattern here is not localizable, you'll need to use something that will generate the pattern for you correctly. For example, see the Fluent DATETIME function, which lets you choose the components you need, and it will generate the localized pattern.

https://projectfluent.org/fluent/guide/functions.html

gregtatum: In JS-land this is the [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) API.

groovecoder:
Yeah, that needs to block. I'm sure QA will try to go thru this with brand new accounts.

static/js/filter.js
maxxcrawford: _(This was removed when I added the feature back)_

privaterelay/templates/includes/dashboard-filter.html
maxxcrawford: There's an entirely new on-boarding comp, but this fix works for now:

image

privaterelay/templates/profile.html
maxxcrawford: Issue filed https://github.com/mozilla/fx-private-relay/issues/907

privaterelay/templates/profile.html
maxxcrawford: Updated!

privaterelay/templates/profile.html
maxxcrawford: Updated!

.eslintrc.js
```diff
@@ -3,6 +3,7 @@
 module.exports = {
   env: {
     es6: true,
+    node: true,
```
groovecoder, line 6: TODO: clean up this file to remove the `extension/**` directives.

privaterelay/templates/includes/dashboard-filter.html
```diff
@@ -0,0 +1,60 @@
...
+    </div>
+    <div class="c-filter-date hidden">
+        <div class="c-filter-icon c-filter-icon-calendar"></div>
+    </div>
+    <div class="c-filter-cateogry hidden">
+        <div class="c-filter-icon c-filter-icon-filter"></div>
+    </div>
+    <div class="c-filter-action">
+        <!-- Button -->
+        <form action="/emails/" class="dash-create" method="POST">
```
groovecoder, line 28:
```suggestion
        <form action="{$ url 'emails-index' %}" class="dash-create" method="POST">
```

groovecoder (changes requested):

.eslintrc.js
maxxcrawford: Cleaned! 🧹

privaterelay/templates/includes/dashboard-filter.html
maxxcrawford: Fixed!

maxxcrawford: Fixed!

package.json
```diff
@@ -32,7 +32,7 @@
     "url": "git+https://github.com/mozilla/fx-private-relay.git"
   },
   "scripts": {
-    "lint": "npm run lint:js",
+    "lint": "npm run lint:js && npm run lint:css",
```
groovecoder, line 35: woohoo!

groovecoder (approved):

PR #893, opened 2021-06-17
non-random addresses in DomainAddressTest

groovecoder:
We see a lot of `CannotMakeAddressException`s during Circle test runs. It seems like the `has_bad_words` check is a bit over-aggressive? Changing tests so they don't accidentally make bad-words addresses during tests.


say-yawn (approved):
LGTM

Issue #872, opened 2021-06-11
Display name in Firefox account is not showing in Firefox Relay page


Bianca-O:
**Build:**
Dev

**Affected platforms:**
Windows 10 64-bit
macOS 10.15

**Browsers tested on:**
Firefox

**Steps to reproduce:**

1. Navigate to https://accounts.stage.mozaws.net/ and login using an existing account.
2. Go to Display name section and add a display name.
3. Navigate to https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/ and sign in with the same account.
4. Observe the Sign in successful message.
5. Observe the Welcome message.
6. Click the Avatar icon and observe the username.

**Actual result:** 
The display name is not showing. The email address is displayed.

**Expected result:** 
The display name is displayed instead of the email address.

![display name not showing](https://user-images.githubusercontent.com/79697848/121706996-d2049c00-cade-11eb-8df3-52fa6f86d9e2.PNG)

maxxcrawford:
@say-yawn I need some help exposing the correct information from the FxA API

maxxcrawford:
https://github.com/mozilla/fx-private-relay/projects/3

say-yawn:
https://github.com/mozilla/fxa/blob/main/packages/fxa-profile-server/docs/API.md#get-v1profile this information is available from the `/profile` request we make and I just need to make sure this information is available in the view.

PR #894, opened 2021-06-17
Get display name from profile

say-yawn:
# About this PR
According to FxA document on keys and values returned for a given profile [here](https://github.com/mozilla/fxa/blob/main/packages/fxa-profile-server/docs/API.md#get-v1profile), the display name is under the extra data of a Social Account authorized from FxA. However, if display name is not set, the `displayName` key will not show up on the `extra_data`.

#
Related to #872

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -101,6 +101,12 @@ class Profile(models.Model):
         return self.user.socialaccount_set.filter(provider='fxa').first()
 
     @property
+    def display_name(self):
+        # if display name is not set on FxA the
+        # displayName key will not exist on the extra_data
+        return self.fxa.extra_data.get('displayName')
+
+    @property
     def has_unlimited(self):
         # FIXME: as we don't have all the tiers defined we are over-defining
         # this to mark the user as a premium user as well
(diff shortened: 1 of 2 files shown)


emails/models.py
```diff
@@ -100,6 +100,12 @@ def at_max_free_aliases(self):
     def fxa(self):
         return self.user.socialaccount_set.filter(provider='fxa').first()
 
+    @property
+    def display_name(self):
+        # if display name is not set on FxA the
+        # displayName key will not exist on the extra_data
+        return self.fxa.extra_data.get('displayName')
```
groovecoder, line 107: should we make this return the email address if `displayName` doesn't exist?

```suggestion
        return self.fxa.extra_data.get('displayName', self.fxa.extra_data.email)
```

say-yawn: @maxxcrawford, do you want the backend to return the email address for the display name var in the context if the display name doesn't exist or just return `None`?

maxxcrawford: @say-yawn Great questions – Lemme check with `@efeo`

groovecoder: Any update here?

maxxcrawford: Update: If they don't have a display name set, we're just going to say "Welcome"…so `none` should work. As long as we can wrap up the logic there. I'm happy to impliment that portion after this lands with the name.

Issue #711, opened 2020-10-13
Have Circle sign into Docker


groovecoder:
Just like https://github.com/mozilla/ichnaea/pull/1369. (See https://github.com/mozilla/ichnaea/compare/6655b9b...f8da96dc28 for Circle code changes.)

groovecoder:
We can either do this, or if we have a different comparable mechanism or technique we can do that instead.

PR #896, opened 2021-06-21
Updating circleci config to user dockerhub auth

moz-astults:
Regarding: https://support.circleci.com/hc/en-us/articles/360050623311-Docker-Hub-rate-limiting-FAQ

Before this change we were getting warning messages in CircleCI mentioned in the support FAQ:
```
Warning: No authentication provided, using CircleCI credentials for pulls from Docker Hub.
```

After this change, the CircleCI run no longer shows this warning. Taken from the example here: https://github.com/mozilla-services/Dockerflow/blob/8aa28daed38aa55de42e9248e1e0ac4bac33b0c8/.circleci/config.yml

Closes https://github.com/mozilla/fx-private-relay/issues/711

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,9 @@ jobs:
   build:
     docker:
       - image: docker:stable-git
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
     working_directory: /dockerflow
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -105,4 +114,4 @@ workflows:
             tags:
               ignore: /addon-.*/
             branches:
-              only: master
+              only: main
```
moz-astults, line 117: I'm not sure how this was working before this change, but I figured we should update this since we are using `main` as the default branch. Perhaps CircleCI aliases `master` with whatever your default branch is, but I haven't found any documentation to support that.

groovecoder (approved):

PR #937, opened 2021-07-20
Fix #926 - Remove search icon when no alias

### LINKED ISSUE #926: Magnifying glass (search) icon is displayed on mobile even if no alias is added
**Build:**
Stage

**Affected platforms:**
Android 11
iOS 14.6

**Browsers tested on:**
Firefox

**Preconditions:**
The user doesn’t have any aliases added.

**Steps to reproduce:**

2. Observe the Magnifying glass (search) icon and tap on it.

**Actual result:**
Magnifying glass (search) icon is displayed even if no alias is added.

**Expected result:** 
Magnifying glass (search) icon is not displayed when no alias is added.

https://user-images.githubusercontent.com/79697848/126285305-e4dadb36-fb2a-4005-b0ec-dcb9b5a0c599.mp4

codemist:
Added template tag to disable search icon when there are no aliases listed.

### DIFF
--- a/privaterelay/templates/includes/dashboard-filter.html
+++ b/privaterelay/templates/includes/dashboard-filter.html
@@ -4,7 +4,7 @@
 
 <div class="c-filter js-filter-container">
     <div class="c-filter-search">
-        {% if relay_addresses|length >= 1 %}
+        {% if relay_addresses %}
             <div class="c-filter-icon c-filter-icon-search js-filter-mobile-search-toggle"></div>
             <form class="js-filter-search-form c-filter-search-form" action="">
                 <input class="c-filter-search-input js-filter-search-input " type="search" name="search" id="search">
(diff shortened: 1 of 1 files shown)


privaterelay/templates/includes/dashboard-filter.html
```diff
@@ -4,18 +4,20 @@
...
-            <div class="c-filter-case-count">
-                <span class="js-filter-case-visible"></span>
-                /
-                <span class="js-filter-case-total"></span>
-            </div>
-            <button class="c-filter-reset js-filter-reset">
-                <img alt="" src="{% static '/images/x-close.svg' %}"/>
-            </button>
-        </form>
+        {% if relay_addresses|length >= 1 %}
```
groovecoder, line 7: I think a python list is evaluated as `False` if it has no elements, so you could change this to:
```suggestion
        {% if relay_addresses %}
```

groovecoder (changes requested):

codemist:
I just selected _commit changes_ to your review - am I allowed to do that? Does that change my commit?

maxxcrawford (approved):
Clean logic. Ship it!

PR #955, opened 2021-07-23
Fix #949 - Truncate email address to prevent overlap

### LINKED ISSUE #949: Long email address overlaps text on desktop and creates layout issues on mobile
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit
macOS 10.14
Android 11
iOS 14.6

**Browsers tested on:**
Firefox

**Steps to reproduce:**

1. Navigate to https://stage.fxprivaterelay.nonprod.cloudops.mozgcp.net and create an account using a long email address.
2. Upgrade to Mozilla Privacy Pack.
3. Observe the username displayed after "Welcome" message.
4. Only for mobile - Zoom out the page.

**Actual result:**
The username overlaps the "Email aliases used" text.
On mobile there is a layout issue.

**Expected result:**
No overlapping. No layout issue.

**Desktop 1:**

long alias text overlaps

**Desktop 2:**

long alias text overlaps2
**Mobile:**

mobile layout issue

codemist:
Preview:
image

### DIFF
--- a/static/scss/partials/dashboard.scss
+++ b/static/scss/partials/dashboard.scss
@@ -5,8 +5,10 @@
 
 .mpp-dashbaord-header-name {
     font-weight: 700;
-    width:100%;
-    word-wrap:break-word;
+    width: 100%;
+    word-wrap: break-word;
+    padding-right:5%;
+
     span {
         font-weight: 400;
         display: block;
@@ -140,7 +142,7 @@
 @media screen and (min-width: 768px) {
     .mpp-dashbaord-header-container {
         display: flex;
-        align-items: flex-end;
+        align-items:center;
         justify-content: space-between;
     }    
     .mpp-dashbaord-header-title {
--- a/static/scss/partials/dashboard.scss
+++ b/static/scss/partials/dashboard.scss
@@ -5,6 +5,8 @@
 
 .mpp-dashbaord-header-name {
     font-weight: 700;
+    width:100%;
+    word-wrap:break-word;
     span {
         font-weight: 400;
         display: block;
--- a/static/scss/partials/dashboard.scss
+++ b/static/scss/partials/dashboard.scss
@@ -7,7 +7,7 @@
     font-weight: 700;
     width: 100%;
     word-wrap: break-word;
-    padding-right:5%;
+    padding-right: $spacing-sm;
 
     span {
         font-weight: 400;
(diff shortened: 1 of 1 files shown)


static/scss/partials/dashboard.scss
```diff
@@ -5,6 +5,8 @@
 
 .mpp-dashbaord-header-name {
     font-weight: 700;
+    width:100%;
```
maxxcrawford, line 8: Is the `width` here necessary?

maxxcrawford (changes requested):

static/scss/partials/dashboard.scss
codemist: Yes, it helps keep the word-wrap property valid on mobile.
https://user-images.githubusercontent.com/13066134/127021103-30bc7aff-5b4b-4c35-9917-b5d950b56b26.mov

codemist:
I changed the container to `align-items: center;` so that when the email breaks off into two lines the dashboard stats remains centered. I also added a `padding-right` property to increase the gutter space between the header title and the stats (visually nicer in my opinion).

Preview:
image

static/scss/partials/dashboard.scss
```diff
@@ -5,6 +5,10 @@
 
 .mpp-dashbaord-header-name {
     font-weight: 700;
+    width: 100%;
+    word-wrap: break-word;
+    padding-right:5%;
```
maxxcrawford, line 10: nit: whitespace

And also – percentage padding feels funky here. Does it work with a token? 

```suggestion
    padding-right: 5%;
```

static/scss/partials/dashboard.scss
```diff
@@ -138,7 +142,7 @@
 @media screen and (min-width: 768px) {
     .mpp-dashbaord-header-container {
         display: flex;
-        align-items: flex-end;
+        align-items:center;
```
maxxcrawford, line 145: With centering, does it look the same when only on one line?  Same question for three/four lines? I want to understand how the different length levels impact this layout. 

nit: whitespace

```suggestion
        align-items: center;
```

maxxcrawford (changes requested):

codemist:
Without `align-items: center;` :
image


With it:

One line:
image

Two lines:
image

Three lines:
image

Four lines:
image

static/scss/partials/dashboard.scss
codemist: Switched out the percentage for `$spacing-sm` instead! 😄

maxxcrawford (approved):
Nice work!

PR #976, opened 2021-07-28
Fix #822 - Set bento strings from FTL file

### LINKED ISSUE #822: Revise bento box menu to use translated strings
- Bento box strings [See Monitor bento.ftl implementation](https://github.com/mozilla/blurts-server/blob/main/locales/en/bento.ftl)

Blocked by #815

maxxcrawford:


static/js/fx-bento.js
```diff
@@ -20,10 +20,44 @@ function createAndAppendEl(wrapper, tagName, className = null) {
...
+      "bentoButtonTitle":"Firefox apps and services",
+      "bentoHeadline":"Firefox is tech that fights for your online privacy.",
+      "bentoBottomLink":"Made by Mozilla",
+      "fxDesktop":"Firefox Browser for Desktop",
+      "fxLockwise":"Firefox Lockwise",
+      "fxMobile":"Firefox Browser for Mobile",
+      "fxMonitor":"Firefox Monitor",
+      "pocket":"Pocket",
+      "fxSend":"Firefox Send",
+      "mobileCloseBentoButtonTitle":"Close menu",
```
groovecoder, line 36: Should this be `bentoButtonCloseLabel` like the other references?

groovecoder (changes requested):

static/js/fx-bento.js
maxxcrawford: Yup. Pocket too. Good catch. This was the old block of copy. I'll update the string refs to match (and test it failing)

maxxcrawford: Updated all of them. Good to go!

groovecoder (approved):

PR #978, opened 2021-07-29
Fix #942 - Truncate Alias Label

### LINKED ISSUE #942: Longer alias labels are cut off
#### Build:
- Stage;

#### Affected Platforms:
- Windows 10;

#### Browsers tested on:
- Firefox, latest version;

#### Prerequisites:
- Have a few aliases created;
- Have the stage version of the Relay add-on installed;

#### Steps to reproduce:
2. Click on the “Add account name” bar above any created alias;
3. Write a long name;

#### Expected result:
- The label is displayed in its entirety;

#### Actual result:
- The label is cut off;

#### Notes:
- N/A;

labelcut1 1
labelcut1 2
labelcut1 3

codemist:
**Preview:**
image
image

### DIFF
--- a/static/scss/partials/main.scss
+++ b/static/scss/partials/main.scss
@@ -940,17 +940,18 @@ input.relay-email-address-label {
   font-size: 1em;
   padding: 6px 24px 6px 6px;
   width: 188px;
+  border-radius: 4px;
   text-overflow: ellipsis;
   white-space: nowrap;
   overflow: hidden;
   transition: all .2s ease;
   // Reset styles from protocol
   background-color: none;
-  min-width: 0;
   border-radius: 0;
   border: 0;
   margin: 0;
   list-style: inherit;
+  min-width: 0;
 }
 
 .additional-notes {
@@ -960,6 +961,7 @@ input.relay-email-address-label {
   pointer-events: none;
   margin: auto auto 0 0;
   display: none;
+  width: 100%;
 }
 
 .legacy-addon-show-label {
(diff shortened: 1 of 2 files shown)


static/scss/partials/main.scss
```diff
@@ -1080,7 +1080,7 @@ input:focus::placeholder {
 
 .relay-email-address-label:focus,
 input.input-has-error {
-  width: 200px !important;
+ // width: 200px !important;
```
groovecoder, line 1083: nit: let's go ahead and remove the line rather than comment.

codemist: Done!

static/scss/partials/main.scss
```diff
@@ -1032,6 +1032,7 @@ input::placeholder,
 .relay-email-address-label:placeholder-shown {
   color: var(--grey50);
   opacity: 1;
+  width: 192px !important;
```
maxxcrawford, line 1035: Is the `!important` tag necessary here? 
```suggestion
  width: 192px !important;
```

maxxcrawford: Also, there is a token, `$layout-2xl` that is `192px` wide. We might use that instead? (That may not be the most semantically correct use of the token)

maxxcrawford (changes requested):

static/scss/partials/main.scss
codemist: Hmm, the `192px` is more of a Magic Number fix for the length of the placeholder, and from my understanding `$layout-2xl` should be reserved for spacing between elements.

maxxcrawford: Well reasoned. Good to go!

maxxcrawford (changes requested):
Good to go as soon as you resolve the conflicts! 🙈

maxxcrawford (approved):

PR #994, opened 2021-08-05
Fix #693 (NEW) - Add Sign up Btn in Header

### LINKED ISSUE #693: Private Relay home page is ambigous on sign-in sign-up
# About this Issue
As mentioned by @birdsarah currently the page only has `sign in` which can confuse the user to expecting that they must have an account with Relay before they can access Relay.

codemist:
**Home page preview:**

Desktop - 
image

Mobile - 
image

**Profile page stays the same**
image

**Updated breakpoint to `660px` demo:**

https://user-images.githubusercontent.com/13066134/128380846-f316a9e1-f5b3-4622-b83b-41b1219b0c8b.mov

### DIFF
--- a/privaterelay/templates/includes/login.html
+++ b/privaterelay/templates/includes/login.html
@@ -6,8 +6,15 @@
 {% block content %}
 
 <div class="user-area">
-	<firefox-apps></firefox-apps>
 	{% if request.user.is_anonymous %}
+	<a
+	 data-entrypoint="relay-sign-in-header"
+	 data-form_type="button"
+	 data-ga="send-ga-funnel-pings"
+	 data-event-category="Sign In Buttons"
+	 data-event-label="sign-in-btn-header"
+	 class="sign-in sign-up button"
+	 href="{% provider_login_url 'fxa' process='login' %}">{% ftlmsg 'nav-profile-sign-up' %}</a>
     <a
       data-entrypoint="relay-sign-in-header"
       data-form_type="button"
@@ -16,8 +23,10 @@
       data-event-label="sign-in-btn-header"
       class="sign-in button btn-transparent"
       href="{% provider_login_url 'fxa' process='login' %}">{% ftlmsg 'nav-profile-sign-in' %}</a>
+	<firefox-apps class="home-page"></firefox-apps>
 	{% else %}
 		{% get_social_accounts request.user as accounts %}
+		<firefox-apps class="profile-page"></firefox-apps>
 		<glocal-menu class="glocal-menu">
 			<button class="avatar-wrapper" aria-label="{% ftlmsg 'label-open-menu' %}">
 				<img class="avatar" src="{{ avatar }}" alt="{% ftlmsg 'nav-profile-image-alt' %}" />
--- a/privaterelay/templates/includes/login.html
+++ b/privaterelay/templates/includes/login.html
@@ -13,7 +13,7 @@
 	 data-ga="send-ga-funnel-pings"
 	 data-event-category="Sign In Buttons"
 	 data-event-label="sign-in-btn-header"
-	 class="sign-up button"
+	 class="sign-up button header-link"
 	 href="{% provider_login_url 'fxa' process='login' %}">{% ftlmsg 'nav-profile-sign-up' %}</a>
     <a
       data-entrypoint="relay-sign-in-header"
@@ -21,12 +21,12 @@
       data-ga="send-ga-funnel-pings"
       data-event-category="Sign In Buttons"
       data-event-label="sign-in-btn-header"
-      class="mzp-c-button mzp-t-secondary mzp-t-dark home-sign-in"
+      class="mzp-c-button mzp-t-secondary mzp-t-dark u-font-metropolis u-font-weight-normal"
       href="{% provider_login_url 'fxa' process='login' %}">{% ftlmsg 'nav-profile-sign-in' %}</a>
-	<firefox-apps class="home-page"></firefox-apps>
+	<firefox-apps class="fx-bento-home-page"></firefox-apps>
 	{% else %}
 		{% get_social_accounts request.user as accounts %}
-		<firefox-apps class="profile-page"></firefox-apps>
+		<firefox-apps class="fx-bento-profile-page"></firefox-apps>
 		<glocal-menu class="glocal-menu">
 			<button class="avatar-wrapper" aria-label="{% ftlmsg 'label-open-menu' %}">
 				<img class="avatar" src="{{ avatar }}" alt="{% ftlmsg 'nav-profile-image-alt' %}" />
--- a/privaterelay/templates/includes/login.html
+++ b/privaterelay/templates/includes/login.html
@@ -13,7 +13,7 @@
 	 data-ga="send-ga-funnel-pings"
 	 data-event-category="Sign In Buttons"
 	 data-event-label="sign-in-btn-header"
-	 class="sign-in sign-up button"
+	 class="sign-up button"
 	 href="{% provider_login_url 'fxa' process='login' %}">{% ftlmsg 'nav-profile-sign-up' %}</a>
     <a
       data-entrypoint="relay-sign-in-header"
--- a/privaterelay/templates/includes/login.html
+++ b/privaterelay/templates/includes/login.html
@@ -21,7 +21,7 @@
       data-ga="send-ga-funnel-pings"
       data-event-category="Sign In Buttons"
       data-event-label="sign-in-btn-header"
-      class="sign-in button btn-transparent"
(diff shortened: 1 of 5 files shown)


privaterelay/templates/includes/login.html
```diff
@@ -6,8 +6,15 @@
...
 <div class="user-area">
-	<firefox-apps></firefox-apps>
 	{% if request.user.is_anonymous %}
+	<a
+	 data-entrypoint="relay-sign-in-header"
+	 data-form_type="button"
+	 data-ga="send-ga-funnel-pings"
+	 data-event-category="Sign In Buttons"
+	 data-event-label="sign-in-btn-header"
+	 class="sign-in sign-up button"
```
maxxcrawford, line 16: Suggestion: Use the `.header-link` class to make the button behave like the Home/FAQ buttons. You may need to play with the padding to compensate. 

```suggestion
	 class="header-link"
```

image

static/scss/partials/header.scss
```diff
@@ -39,6 +39,14 @@ header {
     white-space: nowrap;
 }
 
+.sign-up {
```
maxxcrawford, line 42: I recommend showing the sign-in button on the mid-size screens if there's room. 

Caveat: If Eduardo is good with it as is, please disregard this recommendation. 

image

maxxcrawford (changes requested):

codemist:
**Demo:**

https://user-images.githubusercontent.com/13066134/128549045-ebb36e6e-3099-4409-9a31-c3c707ac6e2c.mov

privaterelay/templates/includes/login.html
```diff
@@ -6,18 +6,27 @@
...
+	 class="sign-up button"
+	 href="{% provider_login_url 'fxa' process='login' %}">{% ftlmsg 'nav-profile-sign-up' %}</a>
     <a
       data-entrypoint="relay-sign-in-header"
       data-form_type="button"
       data-ga="send-ga-funnel-pings"
       data-event-category="Sign In Buttons"
       data-event-label="sign-in-btn-header"
-      class="sign-in button btn-transparent"
+      class="mzp-c-button mzp-t-secondary mzp-t-dark home-sign-in"
```
codemist, line 24: Added Protocol button classes here

static/scss/components/buttons.scss
```diff
@@ -41,3 +41,9 @@
     align-items: center;
     padding-left: $spacing-md;
 }
+
+.mzp-c-button.home-sign-in {
+    font-family: $font-metropolis;
+    font-weight: normal;
+    font-size: 1.1rem;
```
codemist, line 48: I overwrote some of the Protocol properties to fit the Figma design. One thing I'm not sure about is increasing the `font-size` to `1.1rem`, because for some reason the button looks subtly smaller by default.

https://user-images.githubusercontent.com/13066134/128549873-38b04aea-bbe4-4ab3-abbd-bdb6afea6bd6.mov

static/scss/partials/header.scss
```diff
@@ -194,6 +209,14 @@ header {
     align-items: center;
 }
 
+.mobile-home-link:last-child {
+    display: none;
+
+    @media screen and (max-width: 576px) {
+        display: inline;
```
codemist, line 216: Added this class so that the Sign Up button does not repeat in specific breakpoints (if its in the header, disable it in the dropdown).

static/scss/partials/header.scss
```diff
@@ -349,7 +372,7 @@ header {
     }
 }
 
-@media screen and (max-width: 550px) {
+@media screen and (max-width: 690px) {
```
codemist, line 375: Had to increase the media query by a `10px` magic number to prevent line breaks in the buttons.

static/scss/components/buttons.scss
maxxcrawford: Good call out about the `1.1rem`. I would actually recommend updating the `.header-link` size to match the font size of the default button. That keeps it cleaner.  Great job switching to a Protocol button here!

privaterelay/templates/includes/login.html
maxxcrawford: nit: Recommend using a different class name than `home-sign-in` that closer matches the [class naming convention](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md#class-naming-convention).

static/scss/partials/fx-bento.scss
```diff
@@ -255,8 +255,12 @@ body.hide-bento firefox-apps > button {
...
 
 @media screen and (max-width: 900px) {
-  firefox-apps {
-    margin: 0 var(--bentoPadding);
+  firefox-apps.home-page {
+    margin: 0 0 0 $spacing-lg;
+  }
+
+  firefox-apps.profile-page {
+    margin: 0 $spacing-lg;
```
maxxcrawford, line 263: nit: This specificity may not be necessary. 
```suggestion
  .home-page {
    margin: 0 0 0 $spacing-lg;
  }

  .profile-page {
    margin: 0 $spacing-lg;
```

static/scss/partials/header.scss
maxxcrawford: This works but I wanted to call out the fact that l10n will mean line breaks will totally happen (and it's okay!)

maxxcrawford (changes requested):

privaterelay/templates/includes/login.html
```diff
@@ -13,20 +13,20 @@
 	 data-ga="send-ga-funnel-pings"
 	 data-event-category="Sign In Buttons"
 	 data-event-label="sign-in-btn-header"
-	 class="sign-up button"
+	 class="sign-up button header-link"
```
codemist, line 16: Added `header-link` so that this button behaves like Home & FAQ button.

privaterelay/templates/includes/login.html
```diff
@@ -13,20 +13,20 @@
...
 	 href="{% provider_login_url 'fxa' process='login' %}">{% ftlmsg 'nav-profile-sign-up' %}</a>
     <a
       data-entrypoint="relay-sign-in-header"
       data-form_type="button"
       data-ga="send-ga-funnel-pings"
       data-event-category="Sign In Buttons"
       data-event-label="sign-in-btn-header"
-      class="mzp-c-button mzp-t-secondary mzp-t-dark home-sign-in"
+      class="mzp-c-button mzp-t-secondary mzp-t-dark u-font-metropolis u-font-weight-normal"
```
codemist, line 24: Following Protocol naming convention here: https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md#class-naming-convention

Opted for `u-` utility classes since I'm overwriting certain Protocol button styling attributes.

static/scss/partials/header.scss
```diff
@@ -86,6 +86,7 @@ header {
     padding: $spacing-xs $spacing-sm;
     color: $color-white;
     margin-left: $spacing-md;
+    font-size: 1rem;
```
codemist, line 89: Removed `1.1rem` in the Protocol button overwrite, I found that just adding `1rem` standardizes the size for all of the header buttons. Thank you for that suggestion!

codemist (commented):
@maxxcrawford

static/scss/components/buttons.scss
```diff
@@ -41,3 +41,11 @@
...
     padding-left: $spacing-md;
 }
+
+.mzp-c-button.u-font-metropolis {
+    font-family: $font-metropolis;
+}
+
+.mzp-c-button.u-font-weight-normal {
+    font-weight: normal;
+}
```
maxxcrawford, line 51: Since this is a utility class, I think you can move these over to [the `utils.scss` partial](https://github.com/mozilla/fx-private-relay/blob/rearrange-sign-up-btn/static/scss/includes/utils.scss) (and make it apply to other things outside of `.mzp-c-button`.

maxxcrawford: FWIW – I was actually expecting more of a `.c-button-sign-in` instead of `.home-sign-in`. Good call on the utils class.

maxxcrawford (changes requested):

static/scss/includes/utils.scss
```diff
@@ -1,3 +1,11 @@
 .hidden {
     display: none;
 }
+
+.u-font-metropolis {
+    font-family: $font-metropolis !important;
```
codemist, line 6: I know we shouldn't be using `!important` too much, but it didn't work without it. Since it's an overwriting class, is it fine?

maxxcrawford: Understood. Good for now. The linter may get angry. 😅

maxxcrawford (approved):

PR #995, opened 2021-08-05
Fix #966 - Revise alias search feature

### LINKED ISSUE #966: When searching for an alias which shares its name with other aliases, the list doesn’t display all of them
#### Build:
- Stage;

#### Affected Platforms:
- Windows 10;

#### Browsers tested on:
- Firefox, latest version;

#### Prerequisites:
- Be signed in into a Firefox Relay account;
- Have the stage version of the Relay add-on installed;
- Have a few aliases created;

#### Steps to reproduce:
2. Name some of the aliases with the same name (e.g. “google”);
3. In the alias search bar, search for those aliases by typing the name;
4. Observe the results;

#### Expected result:
- The list updates showing all the aliases named like that;

#### Actual result:
- The list updates and shows only the first one named like that;

#### Notes:
- N/A;
search filters1 3

maxxcrawford:
## Summary

This PR refactor how the label search filter function works. It can now account for: 

- Labels named the same name
- Labels that are updated (Ex: If you change a label, and then search for it, it will now appear) 

_Caveat – If you change a label while a search is active, it will not hide itself until a new search is started._ 

## STR: 

- Run a local copy of both the add-on and the app
- Log into FXA
- Create multiple aliases 
- Add labels to some/all of the aliases, including some that are identical. 
- Search for that label
- **Expected:** All items with that label should be visible

### DIFF
--- a/static/js/filter.js
+++ b/static/js/filter.js
@@ -26,6 +26,7 @@
     filterToggleSearchInput.addEventListener("click", toggleAliasSearchBar, false);
 
 	function filterInputWatcher(e) {
+
 		const query = e.target.value.toLowerCase();
 
         // Reset filter if the input is empty, however, do not steal focus to input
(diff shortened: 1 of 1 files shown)


static/js/filter.js
```diff
@@ -36,43 +37,74 @@
             alias.style.display = "none";
         });
 
-        // BUG: If more than one labels are named the same thing (identical), 
-        // the filter only counts/shows the first result.
-		// Loop through both arrays, making that alias visible if it is a match.
+        // Fix GitHub/#966: Create temporary array of objects containing labels and their parent `.js-alias` DOM element
+        const searchIndexWithLabels = [];
+
+        for (const [index, alias] of aliasesWithLabelsCollection.entries()) {
```
maxxcrawford, line 43: This is the crux of the PR: I join the both the array of label entries and each parent object together. When I filter it below, I can access (read: show/hide) the matched items.

static/js/filter.js
```diff
@@ -36,43 +37,74 @@
...
         if (labelArrayIndex > -1) {
-            filterAliasLabels.splice(labelArrayIndex, 1, newLabel);
+            filterAliasLabels.splice(labelArrayIndex, 1, newLabelLowercased);
+            return;
         }
+
+        // Case: User created a label for an alias
+        aliasesWithLabelsCollection.push(alias);
+        filterAliasLabels.push(newLabelLowercased);
+
```
maxxcrawford, line 107: This function got a major update as well – it now can account for all the different cases where a user interacts with the label input form. _(Creating a new label, changing an existing one, or removing an existing one)_

static/js/filter.js
```diff
@@ -99,13 +131,18 @@
             }
 
             const aliasLabel = alias.querySelector(".relay-email-address-label");
+            
 
             if (addOnDetected) {
                 aliasLabel.addEventListener("blur", updateAliasLabel);
+            }
+
+            if (aliasLabel.dataset.label) {
```
maxxcrawford, line 140: This change was necessary because every alias was being added to the `aliasesWithLabelsCollection` array, regardless if they actually had a label value saved.

codemist (approved):
I checked out the branch and ran a bunch of test cases and it looks to be working flawlessly!!

PR #1005, opened 2021-08-09
Redis env var

moz-astults:
In order to work better with Heroku, we should change `REDIS_CACHE_URL` to `REDIS_URL`: https://devcenter.heroku.com/articles/heroku-redis#connecting-in-django

Confirmed this is working by logging into dev, and seeing a new key was created with redis-cli.

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -254,12 +254,12 @@ DATABASES = {
     )
 }
 
-REDIS_CACHE_URL = config('REDIS_CACHE_URL', '', cast=str)
-if REDIS_CACHE_URL:
+REDIS_URL = config('REDIS_URL', '', cast=str)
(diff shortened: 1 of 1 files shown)


moz-astults:
@maxxcrawford good call out - I think for this PR we don't need to worry about the `.env-dist` file because django-redis will only be used as a cache and session storage. If the environment variable for `REDIS_URL` doesn't exist, nothing changes for the application. It just won't use redis. For now, the local dev experience (even after this change) doesn't need redis at all. 

That being said, in the future if we try to implement a rate limiting pattern using redis, this will become a hard dependency and should be included in the `.env-dist`. 

I'd defer to @say-yawn though if it's best to include now, or wait until it's a hard dependency.

groovecoder (approved):

PR #1027, opened 2021-08-19
Domain registration confirmation modal

maxxcrawford:
This PR adds the Domain Registration modal. 

To test: 

- Create/log into PREMIUM account who hasn't yet registered a subdomain
- Fill out form, click "register domain" button
- **Expected:**
  - If the domain is available: The user should see a modal asking them to confirm their selection. 
  - If the domain is not available (because it's either already registered or a disallowed word), the form should NOT show the modal and should submit as normal. This will cause the page to reload with a new error message displayed. 

Note: The confirmation modal submit button should be disabled until the user clicks the checkbox. 

Other UX Notes: 
- Clicking anywhere outside of the modal should dismiss it
- Clicking the ESC key should dismiss the modal 
- Clicking the modal cancel button should dismiss the modal

Completes MPP-406
Related to #1043

Screenshot: https://drive.google.com/file/d/1EcPJBC27kR1bG1m8E_sfKME0BjUoldUV/view?usp=sharing


static/scss/includes/variables.scss
```diff
@@ -17,6 +17,8 @@ $color-mpp-gray-30: #ECECEC;
 
 $box-shadow-mpp-sm: 0px 0px 4px rgba(0, 0, 0, .12);
 
+// Custom width
+$content-sm-md: 546px;
```
maxxcrawford, line 21: _Goldilocks_ width in between Protocol small and medium sizes.

static/scss/partials/modals.scss
```diff
@@ -0,0 +1,151 @@
...
+}
+
+.alias-to-delete {
+    display: inline;
+    word-break: break-all;
+
+    &.demi {
+        display: inline-block;
+        margin: 0 auto $spacing-md;
+        // margin: 0 auto $spacing-md 0;
```
maxxcrawford, line 22: nit: remove comment SCSS

static/scss/partials/modals.scss
```diff
@@ -0,0 +1,151 @@
...
+        overflow: hidden;
+        word-wrap: unset;
+    }
+}
+
+// Modal Protocol Refactor
+.c-modal-wrapper {
+    position: fixed;
+    display: none;
+    background: rgba(0, 0, 0, .4);
```
maxxcrawford, line 35: Should this be a variable?

static/scss/partials/modals.scss
```diff
@@ -0,0 +1,151 @@
...
+}
+
+.c-modal-header-content {
+    @include text-title-3xs;
+    
+    h3 {
+        @include text-title-xs;
+        color: $color-grey-40;
+        margin: $spacing-sm 0;
+        font-weight: 700;
```
maxxcrawford, line 75: Do we use font-weight numbers (`700`) or names (`bold`)?

static/scss/partials/modals.scss
```diff
@@ -0,0 +1,151 @@
...
+        align-items: center;
+        justify-content: center;
+
+        img {
+            margin-right: $spacing-sm;
+        }
+    }
+}
+
+.t-domain-registration-hero {
```
maxxcrawford, line 98: Protocol: Is this a correct use of a `t-` class?

cc/ @craigcook

static/scss/includes/variables.scss
maxxcrawford: TODO: Consider proposing this on Protocol?

static/js/app.js
```diff
@@ -125,25 +125,25 @@ function deleteAliasConfirmation(submitEvent) {
...
     addressEl.textContent = aliasToDelete;
   });
 
   // Show alias email address in delete warning modal
-  const aliasToDeleteLabel = document.querySelector(".modal-message strong");
+  const aliasToDeleteLabel = confirmDeleteModal.querySelector(".modal-message strong");
   aliasToDeleteLabel.textContent = aliasToDelete;
 
-  confirmDeleteModal.classList.add("show-modal");
+  confirmDeleteModal.classList.add("is-visible");
```
maxxcrawford, line 138: Migrate to state based class

static/js/app.js
```diff
@@ -125,25 +125,25 @@ function deleteAliasConfirmation(submitEvent) {
...
   trapFocusInModal("delete-modal", true);
   sendGaPing("Dashboard Alias Settings", "Delete Alias", "Delete Alias");
-  const checkbox = confirmDeleteModal.querySelector(".checkbox");
+  const checkbox = confirmDeleteModal.querySelector(".js-modal-delete-alias-checkbox");
```
maxxcrawford, line 141: Migrate to js based class

static/js/domain-registration.js
```diff
@@ -0,0 +1,136 @@
+(function() {
+	"use strict";
+
+    const domainRegistration = {
+        init: () => {
+            const domainRegistrationForm = document.getElementById("domainRegistration");
+
+            if (domainRegistrationForm) {
+                domainRegistrationForm.addEventListener("submit", domainRegistration.events.onSubmit, false);  
+            }
```
maxxcrawford, line 10: Init check to only fire if the form is visible.

static/js/domain-registration.js
```diff
@@ -0,0 +1,136 @@
...
+            const response = await fetch(requestUrl, {
+                method: "get",
+                mode: "same-origin",
+                credentials: "same-origin",
+            });
+
+            // Catch Redirect Failure
+            if (response.redirected) {
+                return false;
+            }
```
maxxcrawford, line 28: TODO: Add JavaScript explaining the redirect catch

static/js/domain-registration.js
```diff
@@ -0,0 +1,136 @@
...
+                });
+
+                // Close modal if the user clicks the Escape key
+                document.addEventListener("keydown", domainRegistration.modal.close, false);
+            },
+            close: (e) => {
+                if (e && e.key && (e.key !== "Escape")) {
+                    // Do nothing! 
+                    return;
+                }
```
maxxcrawford, line 110: This logic catches all keystrokes while the modal is open/active. If a user clicks the Escape key, it continues on to close the modal. Other wise it drops/ignores it here.

privaterelay/templates/base.html
```diff
@@ -30,7 +30,8 @@
   </head>
   <body class="{% with request.user.profile_set.first as user_profile %}{% if user_profile.has_unlimited %}is-premium{% endif %}{% endwith %}" data-fxa-settings-url="{{ settings.FXA_SETTINGS_URL }}" data-site-origin="{{ settings.SITE_ORIGIN }}" data-google-analytics-id="{{ settings.GOOGLE_ANALYTICS_ID }}" data-debug="{{ settings.DEBUG }}">
     <firefox-private-relay-addon data-addon-installed="false"></firefox-private-relay-addon>
-    {% include "includes/modal.html" %}
+    {% include "includes/modal-delete.html" %}
+    {% include "includes/modal-domain-registration-confirmation.html" %}
```
maxxcrawford, line 34: As there was only ONE modal before (delete), I refactored it into two modals, separated by use-cases.

privaterelay/templates/includes/banners.html
```diff
@@ -1,7 +1,13 @@
+{% load static %}
 {% load relay_tags %}
 {% load ftl %}
 {% ftlconf bundle='privaterelay.ftl_bundles.main' %}
 
+{% if settings.DOMAIN_REGISTRATION_MODAL %}
+  {% block javascript %}
+    <script src="{% static 'js/domain-registration.js' %}"></script>
+  {% endblock %}
+{% endif %}
```
maxxcrawford, line 10: Putting this feature behind a flag. If the ENV isn't set, the form just submits as expected (without any confirmation)

maxxcrawford (commented):
Self-review

static/js/domain-registration.js
maxxcrawford: Added! :)

static/scss/includes/variables.scss
codemist: Could you somehow `calc($content-sm - $content-xs)/2` ? That's 560px. 👀

static/scss/partials/modals.scss
codemist: I've personally been using font-weight numbers since there's usually a CSS property for it in Figma.

maxxcrawford: +1 for numbers. Was curious how you were handling it.

static/scss/partials/modals.scss
codemist: Don't think so, alternatively to incorporate Protocol I think you could use the `::before` selector and then set that to a `$color-black` background and `opacity: 0.4`

codemist (approved):

codemist:
I left some comments but they were mostly nits and suggestions

static/scss/partials/modals.scss
craigcook: "t" is for theming, so it's usually a modifier of some other shared class, and most often presentational. Roughly equivalent to the "M" in BEM. 

If this class is being used to visually differentiate this element from other elements of the same kind, I'd say that's an appropriate use of theming. It says "here is a box that is functionally the same as other boxes, but this particular one has a slightly different appearance." So if you have other "hero" elements but the "domain registration" hero is a different color, that would be a case for a theme class (and would probably be something like `.t-hero-domain-registration`, assuming that's how you want to describe the style and wouldn't use the same style on other heroes that aren't about domain registration).

But this class isn't getting baked into Protocol, you're just using it for this one project so you can really name it whatever makes sense to you. 🙂

static/scss/partials/modals.scss
maxxcrawford: I can also use the SCSS `rgba($base-color, 0.88)` [function](https://sass-lang.com/documentation/variables). I'll do that.

static/scss/includes/variables.scss
maxxcrawford: @codemist Great idea. Dropping that pesky in-between variable.

static/js/domain-registration.js
```diff
@@ -0,0 +1,142 @@
...
+            });
+
+            // Catch Redirect Failure
+            // If the endpoint returns false on a queried subdomain, it means one of two things: 
+            //      1. The domain is already registered
+            //      2. The domain is on the emails/badwords.txt 
+            // 
+            // If we know their request is going to fail, we stop trying to catch the submission and let it fail. 
+            // This will add a messages cookie with the "Domain Not Available" error and reload the page. 
+            if (response.redirected) {
```
groovecoder, line 32: TIL `fetch` response objects have some nice properties!

static/js/domain-registration.js
```diff
@@ -0,0 +1,142 @@
...
+                e.preventDefault();
+
+                const requestedDomain = document.querySelector(".js-subdomain-value").value;
+                const domainCanBeRegistered = await domainRegistration.checkIfDomainIsSafeAndAvailable(requestedDomain);
+                
+                if (domainCanBeRegistered) {
+                    domainRegistration.modal.open();
+                } else {
+                    // If the domain cannot be registered, submit the form to init an error message.
+                    console.log("Domain not available. Please try again");
```
groovecoder, line 68: Remove this?

static/js/domain-registration.js
```diff
@@ -0,0 +1,142 @@
...
+                modal.classList.remove("is-visible");
+
+                document.removeEventListener("keydown", domainRegistration.modal.close, false);
+
+            },
+            formSubmit: async (e) => {
+                const modalConfirmCheckbox = document.querySelector(".js-modal-domain-registration-confirmation-checkbox");
+
+                if (!modalConfirmCheckbox.checked) {
+                    console.log("Not checked");
```
groovecoder, line 128: Remove this?

groovecoder (changes requested):

privaterelay/settings.py
```diff
@@ -243,6 +243,8 @@ def _get_initial_middleware():
     'SUBSCRIPTIONS_WITH_UNLIMITED', default=''
 )
 
+DOMAIN_REGISTRATION_MODAL = config('DOMAIN_REGISTRATION_MODAL', False, cast=bool)
```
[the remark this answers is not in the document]
maxxcrawford: Answer: I'm shipping everything right now with a gate for each feature management.

static/js/domain-registration.js
maxxcrawford: Resolved (and added lint rule to catch future violations!)

groovecoder (approved):

PR #1029, opened 2021-08-20
FixLanguageCodeForDjangoFtl middleware

groovecoder:
Uplifting from code comment:
```
    # Django converts all zh-* codes into either zh-hans or zh-hant
    # django_ftl looks for zh-cn, zh-tw, etc.
```
So, this change adds `FixLanguageCodeForDjangoFtl` middleware that restores the zh- code that django_ftl expects.

To test:

1. In your browser, pick Chinese (Taiwan) [zh-tw] as your top language
2. Visit http://127.0.0.1:8000/faq
   * You should see the FAQ content in `zh-tw`

### DIFF
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -1,3 +1,5 @@
+import functools
+import re
 import time
 
 import markus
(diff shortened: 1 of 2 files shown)


groovecoder:
After some more digging, I saw that the Django code checks `settings.LANGUAGES` for the exact language code before falling back to one of its own languages. So, I think the better fix is to add `zh-tw` and `zh-cn` to `LANGUAGES`. New PR: https://github.com/mozilla/fx-private-relay/pull/1030

PR #1031, opened 2021-08-23
add endpoint for checking subdomain availability

groovecoder:
To test:

1. http://127.0.0.1:8000/accounts/profile/subdomain?subdomain=<subdomain-to-check>
   * Note: if your browser session isn't signed in, it should return an error
2. Also verify that you can still register a new subdomain

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -40,6 +40,12 @@ class Profile(models.Model):
     def __str__(self):
         return '%s Profile' % self.user
 
+    @staticmethod
+    def subdomain_available(subdomain):
+        bad_word = has_bad_words(subdomain)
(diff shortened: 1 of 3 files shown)


maxxcrawford (commented):
I tested the bad word filter with "butt" (http://127.0.0.1:8000/accounts/profile/subdomain?subdomain=butt) and it returned `true`. Should it return false, or is it baked into another query?

groovecoder:
Oh good catch! I didn't include the badwords check in there. Will fix and update.

groovecoder:
Okay I moved the "already taken" AND the "has bad word" check into the method that checks for availability. Should be fixed now. (Also added a couple tests)

maxxcrawford (approved):

PR #1036, opened 2021-08-25
Add add-on data attr if subdomain is set

maxxcrawford:
Expected: 
If you're a user who is premium AND logged in AND have registered a subdomain, then the data-attr `data-premium-subdomain-set` will be set to `true` in the `firefox-private-relay-addon-data` custom element. 

Otherwise, that data-attr will not exist in the `firefox-private-relay-addon-data` element.

### DIFF
--- a/privaterelay/templates/profile.html
+++ b/privaterelay/templates/profile.html
@@ -18,6 +18,13 @@
   data-fxa-subscriptions-url="{{ settings.FXA_SUBSCRIPTIONS_URL }}" 
   data-premium-prod-id="{{ settings.PREMIUM_PROD_ID }}" 
   data-premium-price-id="{{ settings.PREMIUM_PRICE_ID }}" 
+  {% if settings.PREMIUM_ENABLED %}
+    {% with request.user.profile_set.first as user_profile %}
+      {% if user_profile.subdomain %}
+        data-premium-subdomain-set="true"
+      {% endif %}
+    {% endwith %}
+  {% endif %}
 >
 </firefox-private-relay-addon-data>
 


privaterelay/templates/profile.html
```diff
@@ -18,6 +18,13 @@
   data-fxa-subscriptions-url="{{ settings.FXA_SUBSCRIPTIONS_URL }}" 
   data-premium-prod-id="{{ settings.PREMIUM_PROD_ID }}" 
   data-premium-price-id="{{ settings.PREMIUM_PRICE_ID }}" 
+  {% if settings.PREMIUM_ENABLED %}
+    {% with request.user.profile_set.first as user_profile %}
+      {% if user_profile.subdomain %}
```
maxxcrawford, line 23: Is there a better way to chain this than three if/with type statements?

privaterelay/templates/profile.html
```diff
@@ -18,6 +18,13 @@
...
+        data-premium-subdomain-set="true"
```
maxxcrawford, line 24: Rather than just setting `true`, can I get an output from the variable? Similar to `var foo = (x > y)` in JS?

privaterelay/templates/profile.html
groovecoder: Hmm ... we can probably add this markup to the page even if `PREMIUM_ENABLED` is `False`? It won't hurt anything, right?

We should also make our own context processor that always puts a `user_profile` variable in every template context.

privaterelay/templates/profile.html
groovecoder: Yup, the `user_profile.subdomain` value should be a string of the user's subdomain. So it could be:
```suggestion
        data-premium-subdomain-set="{{ user_profile.subdomain }}"
```

groovecoder (approved):

PR #1038, opened 2021-08-25
Relay replies

groovecoder:
Still needs tests and I'm sure there's more code cleanup to do, but this makes replies work in addition to the latest code for mozmail domain.

Let's keep this PR to a code review for now, and I can help you do an end-to-end test if you like. (Or we can push this to the Relay dev server to test.)


emails/views.py
```diff
@@ -374,11 +375,86 @@ def _sns_message(message_json):
...
+        if header['name'].lower() == 'in-reply-to':
+            in_reply_to = header['value']
+            message_id_bytes = get_message_id_bytes(in_reply_to)
+    if not in_reply_to:
+        incr_if_enabled('mail_to_replies_without_in_reply_to', 1)
+        raise Exception("Received mail to reply address without In-Reply-To")
+    (lookup_key, encryption_key) = derive_reply_keys(
+        message_id_bytes
+    )
+    lookup=base64.urlsafe_b64encode(lookup_key).decode('ascii')
```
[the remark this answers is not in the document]
groovecoder: added a new `b64_lookup_key` util.

moz-astults:
Per conversations with @say-yawn and @groovecoder we should also include a time stamp in the model so we can have retention policies on this table if necessary.

emails/models.py
```diff
@@ -10,8 +10,6 @@
 from django.contrib.auth.models import User
 from django.db import models
 
-from emails.utils import get_domains_from_settings
```
groovecoder, line 13: I brought this function into this module to avoid a circular dependency.

emails/models.py
```diff
@@ -120,11 +124,13 @@ def display_name(self):
...
 
     @property
-    def has_unlimited(self):
+    def has_premium(self):
         # FIXME: as we don't have all the tiers defined we are over-defining
         # this to mark the user as a premium user as well
         if not self.fxa:
             return False
+        if self.user.email.endswith('@mozilla.com'):
+            return True
```
groovecoder, line 133: Changed this property to `has_premium` and returning `True` if their FXA ends with `@mozilla.com` so we can fox-food premium features.

emails/models.py
```diff
@@ -363,3 +372,17 @@ def delete(self, *args, **kwargs):
         self.user_profile.num_address_deleted += 1
         self.user_profile.save()
         return super(DomainAddress, self).delete(*args, **kwargs)
+
+    @property
+    def full_address(self):
+        return '%s@%s.%s' % (
+            self.address, self.user_profile.subdomain, DEFAULT_DOMAIN
+        )
```
groovecoder, line 380: The reply code needs a `full_address` to put in the `From:` header.

emails/utils.py
```diff
@@ -53,36 +60,10 @@ def get_email_domain_from_settings():
     return email_network_locality
 
 
-def get_domains_from_settings():
-    return {
-        'RELAY_FIREFOX_DOMAIN': settings.RELAY_FIREFOX_DOMAIN,
-        'MOZMAIL_DOMAIN': settings.MOZMAIL_DOMAIN
-    }
```
groovecoder, line 60: This is moved into `models` module.

emails/utils.py
```diff
@@ -53,36 +60,10 @@ def get_email_domain_from_settings():
...
-                'Subject': {'Charset': 'UTF-8', 'Data': subject},
-            },
-            Source=from_address,
-            ConfigurationSetName=settings.AWS_SES_CONFIGSET,
-        )
-        incr_if_enabled('ses_send_email', 1)
-    except ClientError as e:
-        logger.error('ses_client_error', extra=e.response['Error'])
-        return HttpResponse("SES client error", status=400)
-    return HttpResponse("Sent email to final recipient.", status=200)
```
groovecoder, line 80: Everything is using the `ses_send_raw_email` now.

emails/utils.py
```diff
@@ -53,36 +60,10 @@ def get_email_domain_from_settings():
...
-
-
-@time_if_enabled('ses_send_email')
+@time_if_enabled('ses_send_raw_email')
 def ses_send_raw_email(
-        from_address, to_address, subject, message_body, attachments
+    from_address, to_address, subject, message_body, attachments,
+    reply_address, mail, address
```
groovecoder, line 66: I'm sure there's some more clean-up that can be done here now that this function has the entire `mail` object in it.

emails/utils.py
```diff
@@ -132,7 +114,7 @@ def ses_send_raw_email(
     try:
         # Provide the contents of the email.
         emails_config = apps.get_app_config('emails')
-        emails_config.ses_client.send_raw_email(
+        ses_response = emails_config.ses_client.send_raw_email(
```
groovecoder, line 117: Capturing and using the `ses_response` so we can use its ephemeral `MessageId` value to derive encryption keys.

emails/views.py
```diff
@@ -281,21 +287,27 @@ def _sns_message(message_json):
             status=400
         )
 
+    from_address = parseaddr(mail['commonHeaders']['from'][0])[1]
     to_address = parseaddr(mail['commonHeaders']['to'][0])[1]
-    local_portion = to_address.split('@')[0]
-
-    if local_portion == 'noreply':
```
groovecoder, line 287: We no longer send from `noreply@` as the `Reply-To` value ... we send with `replies@`.

groovecoder:
@say-yawn okay, this is working for both Relay addresses and Domain addresses. I added some inline comments here on the PR to hopefully help make the review easier, but ping me if you want to review in real-time.

emails/migrations/0020_reply_created_at.py
```diff
@@ -0,0 +1,20 @@
...
+    dependencies = [
+        ('emails', '0019_merge_20210825_1737'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='reply',
+            name='created_at',
+            field=models.DateField(auto_now_add=True, default=django.utils.timezone.now),
+            preserve_default=False,
```
say-yawn, line 18: We want to verify that the date of the Reply object created is properly populated.

emails/models.py
say-yawn: We want to use the `self.domain` instead of the `DEFAULT_DOMAIN` when the code for moving the subdomain registration to mozmail.com instead of relay.firefox.com

emails/models.py
```diff
@@ -363,3 +372,17 @@ def delete(self, *args, **kwargs):
...
+
+
+class Reply(models.Model):
+    relay_address = models.ForeignKey(RelayAddress, on_delete=models.CASCADE, blank=True, null=True)
+    domain_address = models.ForeignKey(DomainAddress, on_delete=models.CASCADE, blank=True, null=True)
```
say-yawn, line 385: Not a big problem if we do not have the reply object be editable by API endpoints.

emails/models.py
```diff
@@ -363,3 +372,17 @@ def delete(self, *args, **kwargs):
...
+    lookup = models.CharField(max_length=255, blank=False)
```
say-yawn, line 386: Maybe consider setting this as an index?

emails/utils.py
say-yawn: [Not Required] Some code we can refactor into a separate function:
- attachment
- reply object creation (this can after where the `ses_send_raw_email` is called)

emails/utils.py
```diff
@@ -141,36 +123,48 @@ def ses_send_raw_email(
                 'Data': msg.as_string(),
             },
         )
-        incr_if_enabled('ses_send_email', 1)
+        incr_if_enabled('ses_send_raw_email', 1)
```
say-yawn, line 126: Change the query on Grafana to pull from both `ses_send_raw_email` and `ses_send_email` otherwise we'll no longer have accurate count of emails relayed.

emails/utils.py
```diff
@@ -220,3 +214,50 @@ def generate_relay_From(original_from_address):
         )
     )
     return formatted_from_address
+
+
+def get_message_id_bytes(message_id_str):
+    message_id = message_id_str.split("@", 1)[0].rsplit("<", 1)[-1].strip()
+    return message_id.encode()
```
say-yawn, line 221: This would be a good one to have unittests on.

emails/utils.py
```diff
@@ -220,3 +214,50 @@ def generate_relay_From(original_from_address):
...
+
+
+def b64_lookup_key(lookup_key):
+    return base64.urlsafe_b64encode(lookup_key).decode('ascii')
```
say-yawn, line 225: look into seeing if we can use the `utf-8`. The encode only returns the `ascii` strings only.

emails/views.py
```diff
@@ -279,21 +287,27 @@ def _sns_message(message_json):
...
-        incr_if_enabled('email_for_noreply_address', 1)
-        return HttpResponse('noreply address is not supported.')
-
-    domain_portion = to_address.split('@')[1]
+    [to_local_portion, to_domain_portion] = to_address.split('@')
+    # to_domain_portion = to_address.split('@')[1]
```
say-yawn, line 293:
```suggestion
    [to_local_portion, to_domain_portion] = to_address.split('@')
```

emails/views.py
```diff
@@ -279,21 +287,27 @@ def _sns_message(message_json):
...
         # up a bit.
-        address = _get_address(to_address, local_portion, domain_portion)
+        address = _get_address(to_address, to_local_portion, to_domain_portion)
         user_profile = address.user.profile_set.first()
     except Exception:
+        if to_local_portion == 'replies':
+            user = User.objects.get(email=from_address)
+            profile = user.profile_set.first()
+            if not profile.has_premium:
+                # TODO: send the user an email that replies are a premium feature
```
say-yawn, line 305: Increment here on how many users are trying to reply.

say-yawn (changes requested):
First changes requested.

emails/utils.py
say-yawn: cursory search showed that urlsafe_b64encode uses binary data or `ascii` string so we should continue to use that format. See this [doc](https://docs.python.org/3/library/base64.html) for more info.

emails/views.py
say-yawn: We may still want this for users who are trying to respond to the old relayed emails that was sent with `noreply` cal portion.

emails/views.py
```diff
@@ -279,21 +287,27 @@ def _sns_message(message_json):
...
+                return HttpResponse(
+                    "Rely replies require a premium account", status=403
+                )
+            return _handle_reply(message_json)
+
         return HttpResponse("Address does not exist", status=404)
 
     address_hash = sha256(to_address.encode('utf-8')).hexdigest()
```
say-yawn, line 313: We could potential move the rest of the code snippet here to a separate function like ` _handle_incoming_email` to follow the `_handle_reply`. Wdyt?

say-yawn (changes requested):
2nd changes requested

emails/views.py
```diff
@@ -372,11 +375,87 @@ def _sns_message(message_json):
...
+    decrypted_metadata = json.loads(decrypt_reply_metadata(
+        encryption_key, reply_record.encrypted_metadata
+    ))
+    incr_if_enabled('reply_email', 1)
+    outbound_reply_to_address = outbound_from_address
+    subject = mail['commonHeaders'].get('subject', '')
+    if 'reply-to' in decrypted_metadata:
+        to_address = decrypted_metadata['reply-to']
+    elif 'from' in decrypted_metadata:
+        to_address = decrypted_metadata['from']
```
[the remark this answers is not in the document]
groovecoder: Maybe. But `Sender` is a different email header, so "sender" isn't a strictly accurate term either. 🤷‍♂️

emails/utils.py
groovecoder: We don't have any Grafana queries that use this yet, but I'll make sure we include both when we do.

emails/migrations/0020_reply_created_at.py
groovecoder: I verified the reply records all have the proper (i.e., distinct) values.
image

emails/views.py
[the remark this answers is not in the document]
groovecoder: I'm going to keep it as `from` to match the exact header field.

emails/utils.py
groovecoder: good call ... much nicer function break-down now in https://github.com/mozilla/fx-private-relay/pull/1038/commits/901310c81f8b2a5d1eb2054d28ffa44108a5a83e ... new functions: 
* _start_message_with_headers
* _add_body_to_message
* _add_attachments_to_message
* _store_reply_record

PR #1050, opened 2021-08-30
Hide category filter by default

maxxcrawford:
This PR addresses two issues: 

- Only show alias label/edit input when add-on is detected
- Only show the category filter bar if the user is premium 

Screenshows: 

### Alias labels

With add-on installed: 
image

Without add-on installed:
image

### Category filter

As premium user: 
image

As normal user (no filter bar): 
image

### DIFF
--- a/privaterelay/templates/includes/dashboard-filter.html
+++ b/privaterelay/templates/includes/dashboard-filter.html
@@ -22,80 +22,82 @@
     <div class="c-filter-date hidden">
         <div class="c-filter-icon c-filter-icon-calendar"></div>
     </div>
-    <div class="c-filter-category">
+    <div class="c-filter-category {% if not user_profile.has_unlimited %} hidden {% endif %}">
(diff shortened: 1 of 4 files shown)


privaterelay/templates/includes/dashboard-filter.html
```diff
@@ -22,80 +22,82 @@
     <div class="c-filter-date hidden">
         <div class="c-filter-icon c-filter-icon-calendar"></div>
     </div>
-    <div class="c-filter-category">
+    <div class="c-filter-category {% if not user_profile.has_unlimited %} hidden {% endif %}">
```
codemist, line 25: Should this be `is-hidden`?

maxxcrawford: Yes, but I'm leaving as is for the moment. Hot-fix'in!

groovecoder (approved):

PR #1066, opened 2021-08-31
Decouple Homepage Hero Image Text/Rotator

maxxcrawford:
The homepage rotator now has the strings that were embedded in them outside, so that they can display in the users' requested language.


static/scss/pages/home.scss
```diff
@@ -0,0 +1,60 @@
...
+        @include text-title-3xs;
+        color: $color-purple-90;
+        text-align: center;
+        margin-bottom: $spacing-md;
+    }
+
+    img {
+        display: block;
+        width: $content-xs;
+        margin-left: auto;
```
codemist, line 57: nit: `margin: 0 auto;`?

codemist (approved):
That's some Protocol-ed CSS!

static/scss/pages/home.scss
maxxcrawford: Revised. I had an additional commit that retooled a ton of the "lines" of the site with the `$content-max` width set everything. I ended up bailing on it, but I was constantly only using `margin-left/right` to preserve the Y axis margins from a parent style. 

TLDR; Great catch!

PR #1078, opened 2021-09-07
Wrap alias label input in a form

Vinnl:
This should allow submitting it by pressing Enter, and might better
assist screen readers. Note that this will break versions of the add-on that do not include


maxxcrawford:
Putting up some `blocked` tags on this PR as we will need to cut/release/ship a new release of the add-on before merging this in.

maxxcrawford (approved):
Works as expected in both environments (with and without the form tag around the labels)

maxxcrawford:
@Vinnl Can you re-open a new PR with this change? I approved/merged it accidentally (and then revert it) 🙈

PR #1085, opened 2021-09-08
Fix #1059 - Revise category filter buttons to wrap if necessary

### LINKED ISSUE #1059: Apply button gets outside of filter dropdown when page is in Deutsch
**Build:**
Stage

**Affected platforms:**
Windows 10 64 bit
macOS 10.14

**Browsers tested on:**
Firefox

**Preconditions:**
The browser's language is set to Deutsch.

**Steps to reproduce:**

2. Click the filter button and observe the drop-down elements.

**Actual result:**
The "Apply (Anwenden)" button gets outside of filter dropdown.

**Expected result:**
The "Apply (Anwenden)" button fits in the filter dropdown.

button displayed outside of dropdown

maxxcrawford:
This PR fixes MPP-869

This PR refactors how the padding/wrapping works on the category filter menu. If there is not enough room to fit both buttons, it wraps to the next line, but maintains the expected spacing/margins. 

## Screenshots

German (edit): 
image

English: 
image

### DIFF
--- a/static/scss/components/filter-bar.scss
+++ b/static/scss/components/filter-bar.scss
@@ -241,13 +241,18 @@
 }
 
 .c-filter-category-buttons {
-    padding: $spacing-md $spacing-md 0;
+    padding: $spacing-md $spacing-md 0 0;
     width: 100%;
     display: flex;
     justify-content: space-between;
     flex-wrap: wrap;
+    flex-direction: row;
+    align-items: stretch;
 }
 
 .c-filter-category-buttons button {
     margin-bottom: $spacing-md;
+    margin-left: $spacing-md;
+    width: auto;
+    flex: 1;
 }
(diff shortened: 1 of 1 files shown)


Vinnl (approved):
Does what it says on the tin :) (Slightly related design question: maybe they should be full width when wrapped? Fine as-is too though.)

codemist (approved):
Looks good

maxxcrawford:
> Does what it says on the tin :) (Slightly related design question: maybe they should be full width when wrapped? Fine as-is too though.)

Ahhh. I agreed with this. Let me revise.

maxxcrawford:
@Vinnl Updated: 

German:
image

English: _(Note that these buttons fille the space minus margins)_
image

Vinnl (approved):
Looks great!

PR #1088, opened 2021-09-09
Remove unused vars, enable in lint rule

Vinnl:
Instead of manually checking (e.g.
https://github.com/mozilla/fx-private-relay/pull/1082#discussion_r704475602),
we can just have the computer do it for us :)

PR #1089, opened 2021-09-09
Run linting in CI

Vinnl:
There's already a configuration for CircleCI, so we might want to move this there. However, the add-on's repository does have a GitHub Actions config, and this integrates better with GitHub (plus, I was already familiar with the syntax :) ), so created this as a first proposal.

This complements #1088 to avoid having to manually check for linting errors in PR reviews.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,25 @@
+name: CI
+
+on:
+  push:
+
+env:
+  CI: true
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # Available OS's: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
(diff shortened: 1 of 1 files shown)


maxxcrawford (commented):
Currently – we're catching lint failures in the here: https://github.com/mozilla/fx-private-relay/blob/main/Dockerfile#L1-L10

Is there any specific benefit to do it this way instead? 

cc/ @moz-astults @groovecoder

moz-astults:
It might be more obvious since it is called out in a separate PR check (rather than just the build step), but the main benefit to doing it in the Dockerfile is that it will use the same version that gets built for stage/prod. I'm not sure how much version pinning matters for this level of linting, but that's my 2 cents.

If you want to have this in GHA I'd say maybe call it `Lint` or something more specific than `CI`

Vinnl:
@maxxcrawford Oh, no, I'd just missed that the existing CI job already checked for linting errors :) Not much point to this PR then!

PR #1090, opened 2021-09-10
Add link target to "Get your email domain"

Vinnl:
I added a little bit of styling to hopefully make it a little bit clearer to the user what they need to be looking at, i.e. I added a blue border to the custom domain banner when pointed to it by the link at the top. This might be something for Eduardo to take another look at though.

https://user-images.githubusercontent.com/4251/132840340-6002169e-9497-4ade-bcfc-319dd178e5f8.mp4

### DIFF
--- a/static/scss/components/banners.scss
+++ b/static/scss/components/banners.scss
@@ -54,6 +54,31 @@
     }
 }
 
+#mpp-choose-subdomain {
+    // Add a transparent border by default so its content doesn't jump around when the border appears:
+    border: 4px solid transparent;
+
+    &:target {
+        border-color: $color-informational;
+        // The height of the fixed header can vary depending on e.g. whether .micro-survey-banner
+        // is displayed, so this calculation might not result in #mpp-choose-subdomain to be
+        // anchored exactly to the top of the page, but at least it won't be hidden behind
+        // the fixed headers:
+        scroll-margin-top: calc(
+            /* .micro-survey-banner vertical padding: */
+            #{($spacing-md * 2)} +
+            /* .header-top vertical padding: */
+            #{($spacing-xl * 2)} +
+            /* .logo-wrapper height: */
+            40px +
+            /* #mpp-choose-subdomain top border: */
+            4px +
+            /* .branner-gradient-bg bottom margin: */
+            2rem
+        );
+    }
+}
+
 @media screen and (min-width: 540px) {
 
     .mpp-choose-subdomain form {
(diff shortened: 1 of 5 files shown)


static/scss/partials/main.scss
```diff
@@ -4,6 +4,15 @@
 
 html {
     font-size: 100%;
+    scroll-behavior: smooth;
+
+    @media (prefers-reduced-motion) {
```
maxxcrawford, line 9: Good looking out!

static/scss/components/banners.scss
```diff
@@ -54,6 +54,31 @@
...
+            #{($spacing-xl * 2)} +
+            /* .logo-wrapper height: */
+            40px +
+            /* #mpp-choose-subdomain top border: */
+            4px +
+            /* .branner-gradient-bg bottom margin: */
+            2rem
+        );
+    }
+}
```
maxxcrawford, line 80: This feels really fragile/a magic-number. Is there a more generic way we can set this? Or can we use JavaScript to calculate this value?

Vinnl: It is :/ I might be able to calculate the bounding box of the header elements and set this property in an `onclick` handler, but then I'd need to know for sure what elements all could be part of the header, since I saw at least the regular header and an NPS survey banner - do you know if there's anything else to take into account?

maxxcrawford: We are using `position: sticky` on the header. Does that help targeting here? (We were actually using `position: fixed` before and had to listen via JS to set the main container top-margin to compensate. 

What happens when it just hits it via the `#mpp-choose-subdomain`? (with no scroll-margin-top)?

Vinnl: Hmm, unfortunately I don't think it's possible to use selectors to target specific styles. I could try to iterate over all (or a set decided via some heuristics) DOM elements, inspect all their styles and calculate the heights of everything with `position: fixed`, but I expect that that would add a noticeable and annoying delay to the scroll.

Without `scroll-margin-top` (or if the calculation above results in a number that is too low), the subdomain chooser will be at the top of the viewport, hidden behind the header and, if present, the microsurvey and any other elements with `position: fixed` and `top: 0`. If the calculation above results in a number that is too high, it will be located a bit below the elements with `position: fixed` and `top: 0` (but likely to still be in view).

maxxcrawford (commented):
@Vinnl We received some early feedback from some of the internal testing and wanted to get this merged in. I caught the branch up and added back the ℹ️ icon to the right of the link.

image

Thanks for getting this teed up for approval.

groovecoder (approved):

PR #1098, opened 2021-09-13
Make total stats order consistent with alias stats

### LINKED ISSUE #1094: The overall consistency of the order in which forwarded and blocked emails are displayed could be improved
#### Build:
- Stage;

#### Affected Platforms:
- Across platforms;

#### Browsers tested on:
- Firefox, latest version;

#### Prerequisites:
- Have a few aliases created;
- Have some forwarded and blocked emails through the aliases;
- Have the latest version of the Relay Add-on installed (for broader comparison);

#### Steps to reproduce:
2. Open the Relay Add-on toolbar menu;
3. Observe the layout of all the email counters together on the screen;

#### Expected result:
- The layout is consistent;

#### Actual result:
- The consistency could be improved, as the counters on the alias list are displayed in reverse order than the counters at the top of the dashboard;

#### Notes:
- This is a small design enhancement that could be implemented, for an issue that isn’t really that visible, but could irk some users;
- In my opinion reversing the display order in the counter at the top of the dashboard would fix this in the most easy and efficient way, aligning it with the “columns” that form on the alias list and fitting it with the order the counters are displayed in the Relay Add-on toolbar menu as well - emails blocked first, emails forwarded second;
blockvsfwd

Vinnl:
Fixes #1094.

In other words, Blocked comes before Forwarded in both the stats at
the top of the page as well as in the stats to the right of a
single alias. They're in the same order in the designs.

Before:

image

After:

image

### DIFF
--- a/privaterelay/templates/profile.html
+++ b/privaterelay/templates/profile.html
@@ -60,8 +60,8 @@
       </div>
       <ul class="mpp-dashbaord-header-stats">
         <li><span class="label">{% ftlmsg 'profile-stat-label-aliases-used' %}</span> <span>{{relay_addresses|length}}</span> </li>
-        <li><span class="label">{% ftlmsg 'profile-stat-label-forwarded' %}</span> <span>{{user_profile.emails_forwarded}}</span> </li>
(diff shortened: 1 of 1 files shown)


maxxcrawford (approved):
🔁  Nice one!

PR #1100, opened 2021-09-13
Remove "Copied!" tooltip after it's hidden

### LINKED ISSUE #1099: "Copied" tooltip for an alias is displayed again when closing/deleting the search filter
**Build:**
Stage

**Affected platforms:**
Windows 10 64 bit
macOS 10.14

**Browsers tested on:**
Firefox

**Preconditions:**

**Steps to reproduce:**

2. Select an alias and click the "Copy" button.
3. Search for an alias by entering a text in the Search aliases bar.
4. Delete the text in the Search bar or close the search filter.

**Actual result:**
"Copied" tooltip is displayed again for the previously selected alias.

**Expected result:**
"Copied" tooltip is not displayed anymore for the previously selected alias.

copied tooltip appears again

Vinnl:
Fixes #1099.

Where the "Copied!" tooltip would previously only be removed when
another alias was copied, it is now removed as soon as it's hidden
from view. This prevents the animation from being replayed every
time the alias is hidden from view, then shown again (i.e. when a
filter is applied, then unapplied).

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -67,18 +67,17 @@ async function sendForm(formAction, formData) {
 }
 
 function copyToClipboardAndShowMessage(triggeringEl) {
-  const previouslyCopiedAlias = document.querySelector(".alias-copied");
+  triggeringEl.classList.add("alias-copied", "alias-copied-fadeout");
+  triggeringEl.title="Alias copied to clipboard";
 
-  if (previouslyCopiedAlias || previouslyCopiedAlias == triggeringEl) {
-    previouslyCopiedAlias.classList.remove("alias-copied", "alias-copied-fadeout");
-    previouslyCopiedAlias.title = "Copy alias to clipboard";
-  }
-
-  setTimeout(() => { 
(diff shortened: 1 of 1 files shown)


static/js/app.js
```diff
@@ -67,18 +67,17 @@ async function sendForm(formAction, formData) {
...
+  triggeringEl.title="Alias copied to clipboard";
+
+  setTimeout(() => {
+    // When the fade-out animation is done (it takes four seconds),
+    // reset the copy button. This avoids the animation getting replayed
+    // every time the alias is being hidden and reshown (i.e. when a filter
+    // is applied and removed again).
+    triggeringEl.classList.remove("alias-copied", "alias-copied-fadeout");
+    triggeringEl.title = "Copy alias to clipboard";
+  }, 4 * 1000);
```
maxxcrawford, line 80: Nice readability math here!

maxxcrawford (approved):
This works as expected. Great fix!

PR #1101, opened 2021-09-14
Treat warnings as errors in CI

Vinnl:
While warnings shouldn't interrupt your flow during development, I think it
would be good not to accumulate too many of them by resolving them
before merging to `main`. (But feel free to reject if you don't agree :) )


maxxcrawford (approved):
I love this idea – it's a good balance. We have some other projects where the warning count is…troubling. This feels safe to land anyways since we only had two warnings (from `main`) anyways. 

Lint for the win!

PR #1108, opened 2021-09-16
Only show domain tooltip when there is none yet

### LINKED ISSUE #1105: Tooltip could be displayed before adding an email domain/ tooltip text could be improved
**Build:**
Stage

**Affected platforms:**
Windows 10 64 bit
macOS 10.14

**Browsers tested on:**
Firefox

**Preconditions:**
The user has an active subscription with an email domain added.

**Steps to reproduce:**

2. Hover over the info icon near the added email domain and observe the tooltip.

**Actual result:**
The tooltip text says "Create your unique and custom email domain" when the email domain is already created.

**Expected result:**
The tooltip is displayed before adding an email domain, or
The tooltip text is changed.

tooltip improvement
tooltip place

Vinnl:
Fixes #1105.

The tooltip says "Create your unique and custom email domain", so it is only
relevant when the user doesn't have a domain yet.

Since I didn't know what to put in the tooltip for when the user _does_ have a domain, I just removed it there altogether. It looks like this now:

image

We can of course re-add it when we have appropriate content there later.

### DIFF
--- a/privaterelay/templates/profile.html
+++ b/privaterelay/templates/profile.html
@@ -45,20 +45,19 @@
         </div>
         {% if not user_profile.subdomain %}
           <div class="mpp-dashbaord-header-action">
-            <button class="mpp-action-link"> <img class="margin-right" src="/static/images/icon-check.svg" alt=""> {% ftlmsg 'profile-label-create-domain' %}</button>
-          </div>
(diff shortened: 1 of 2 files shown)


codemist (approved):
Tested this, looks good!

PR #1109, opened 2021-09-16
Make the chosen domain fit in more situations

### LINKED ISSUE #1091: Alias domain not fully visible in the confirmation dialog
**Build:**
Stage

**Affected platforms:**
Windows 10 64 bit
macOS 10.14
Android 11
iOS 14.6

**Browsers tested on:**
Firefox

**Preconditions:**
The user has an active subscription but no email domain added.

**Steps to reproduce:**

2. Fill in the Search domain field and click the "Get Domain" button.
3. Observe the confirmation dialog.

**Actual result:**
The alias domain is not fully visible in the dialog.

**Expected result:**
The alias domain is fully visible in the dialog.

domain not fully visible in confirmationdialog

Vinnl:
Fixes #1091.

If the chosen domain is somewhat long, the suffix will be hidden
on small screens. If the suffix is very long, the suffix will
always be hidden, and the font size will be decreased on small
screens.

https://user-images.githubusercontent.com/4251/133632079-f069417e-6ac9-415c-be0a-467673d9de22.mp4

### DIFF
--- a/static/js/domain-registration.js
+++ b/static/js/domain-registration.js
@@ -76,16 +76,27 @@
                 modal.classList.add("is-visible");
 
                 const requestedDomain = document.querySelector(".js-subdomain-value");
-                const requestedDomainPreview = document.querySelectorAll(".js-modal-domain-registration-confirmation-domain-preview");
+                const requestedDomainPreviews = document.querySelectorAll(".js-modal-domain-registration-confirmation-domain-preview");
                 const requestedDomainPreviewEnding = document.querySelector(".js-modal-domain-registration-confirmation-domain-ending")
                 
                 // Preview the domain the user requested
-                requestedDomainPreview.forEach( preview => {
-                    preview.textContent = requestedDomain.value;
-                })
+                requestedDomainPreviews.forEach(requestedDomainPreview => {
+                    requestedDomainPreview.textContent = requestedDomain.value;
+                });
                 
                 // .mozmail.com
                 requestedDomainPreviewEnding.textContent = `.${requestedDomain.dataset.domain}`;
+
+                requestedDomainPreviewEnding.parentElement.classList.remove("has-long-domain", "has-very-long-domain");
+                if (requestedDomain.value.length > 10) {
+                    // The domain is so long that it won't fit side-by side with the suffix
+                    // on small screens:
+                    requestedDomainPreviewEnding.parentElement.classList.add("has-long-domain");
+                }
+                if (requestedDomain.value.length > 25) {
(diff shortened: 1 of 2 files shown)


static/js/domain-registration.js
```diff
@@ -76,17 +76,26 @@
                 modal.classList.add("is-visible");
 
                 const requestedDomain = document.querySelector(".js-subdomain-value");
-                const requestedDomainPreview = document.querySelectorAll(".js-modal-domain-registration-confirmation-domain-preview");
+                const requestedDomainPreview = document.querySelector(".js-modal-domain-registration-confirmation-domain-preview");
```
Vinnl, line 79: @maxxcrawford Any idea why this was a `querySelectorAll` instead of a `querySelector`? As far as I can see there's only one such element in the entire codebase.

maxxcrawford: It's in two locations, both in the `modal-domain-registration-confirmation.html` file: 
- https://github.com/mozilla/fx-private-relay/blob/main/privaterelay/templates/includes/modal-domain-registration-confirmation.html#L21
- https://github.com/mozilla/fx-private-relay/blob/main/privaterelay/templates/includes/modal-domain-registration-confirmation.html#L11

Per your demo video, it's here and here (the second instance is currently broken bc of the lack of the `querySelectorAll`) 

image

static/js/domain-registration.js
```diff
@@ -76,17 +76,26 @@
...
                 const requestedDomainPreviewEnding = document.querySelector(".js-modal-domain-registration-confirmation-domain-ending")
                 
                 // Preview the domain the user requested
-                requestedDomainPreview.forEach( preview => {
-                    preview.textContent = requestedDomain.value;
-                })
```
maxxcrawford, line 85: Revert this change to restore that funcitonality (as called out in the [other comment](https://github.com/mozilla/fx-private-relay/pull/1109/files#r714451468)).

maxxcrawford (changes requested):
This is a fantastic design solution for this. Thanks for handling all the dynamics here! 

I have a blocking change that I outlined below. Other than that, good to go!

static/js/domain-registration.js
Vinnl: Haaa that was silly; I was looking for element accessors (`[0]`, `[1]`, etc.) and just mentally ignored this `forEach` :joy:  Fixed in c5dfbc1 (which I've rebased already to avoid accidental merges; hope that doesn't interfere with local checkouts.)

maxxcrawford (approved):
Awesome. Let's goooooo.

PR #1110, opened 2021-09-16
Client-side validation of subdomain length

Vinnl:
Validate subdomain length on the client-side as well, to align with the validation from #1107. I didn't add a `pattern` match since that won't generally result in helpful error messages.

@groovecoder I noticed you set a min length of 1 character, so I mirrored that, but I was thinking that maybe we should start with a higher min length? We can always lower the min length later, but otherwise there might be a scramble for the short domain names, and possibly names that are easily confused with others.

### DIFF
--- a/privaterelay/templates/includes/banners.html
+++ b/privaterelay/templates/includes/banners.html
@@ -66,7 +66,16 @@
                 <p>{% ftlmsg 'banner-choose-subdomain-copy' %}</p>
                 <form id="domainRegistration" method="post" action="{% url 'profile_subdomain' %}">
                     {% csrf_token %}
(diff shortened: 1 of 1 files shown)


groovecoder:
But we *want* to cultivate FOMO in the world, right? ;) But yeah, you're probably right. @tcinotto - what should be the minimum length for the sub-domain?

groovecoder:
I went ahead and merged https://github.com/mozilla/fx-private-relay/pull/1107. Can you re-base this branch on that new commit in `main`?

Vinnl:
Done!

groovecoder (approved):

groovecoder:
Merging to set up for deploy + QA. @maxxcrawford let us know if we need to change something post-merge?

PR #1111, opened 2021-09-16
refresh access tokens automatically, or log the user out

groovecoder:
To test:

### Expired access token, VALID refresh token should automatically refresh
1. Sign in as a regular user
2. (In private window) Sign in to [the admin interface for social tokens](http://127.0.0.1:8000/admin/socialaccount/socialtoken/)
3. As the admin, go to the social token for the signed-in user
4. Change the "Expires at" field to "Today" and "Now" values
5. Go back to the regular user window and refresh
   * The page should refresh no problem - the expired access token should automatically be refreshed
6. In the admin interface, refresh the list of social tokens
   * The social token object for the user should have a new token value and a new expires at value

### Expired access token, REVOKED refresh token should sign out
1. In the account menu, click "Manage your Firefox Account" (should take you [here](https://accounts.stage.mozaws.net/settings))
2. Click "Connected Services"
3. Click "Sign out" in the "Firefox Private Relay local dev" entry
   * Note: you may need to refresh the page and click "Sign out" multiple times, as FXA's collapses multiple active sessions into 1 in the UI. Just keep refreshing and clicking "Sign out" until you've signed out of ALL of them.
4. Back on your local Relay, as the admin, go to the social token for the signed-in user
5. Change the "Expires at" field to "Today" and "Now" values
6. Go back to the regular user window and refresh
   * Your session should be signed out (because your access token is expired AND your refresh token is revoked)

### DIFF
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -1,7 +1,19 @@
+from datetime import datetime, timezone
 import time
 
 import markus
+from oauthlib.oauth2.rfc6749.errors import CustomOAuth2Error
 
+from django.contrib.auth import logout
+from django.shortcuts import redirect
+from django.urls import reverse
+
+from allauth.socialaccount.models import SocialToken
+from allauth.socialaccount.providers.fxa.views import (
+    FirefoxAccountsOAuth2Adapter
+)
+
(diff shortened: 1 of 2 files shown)


say-yawn (approved):
LGTM

PR #1112, opened 2021-09-17
Add some easily-abused words to the blocklist

Vinnl:
Just a couple I could come up with off the top of my head — there's
probably a more thorough list we could use somewhere.


groovecoder:
FWIW, I think the ones in these 2 lists may be pretty solid first additions:

https://github.com/wesleyraptor/streamingphish/blob/master/training_data/targeted_brands/initial_brands.txt
https://github.com/wesleyraptor/streamingphish/blob/master/training_data/keywords/initial.txt

groovecoder:
Uplifting from Slack ... https://github.com/michaldudek/subdomain-blacklist/blob/master/blacklist.txt looks good and is MIT. Check https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=130927170 for guidelines on incorporating MIT-licensed code into MPL2 code-base.

Vinnl:
Replaced my manually compiled list with those, and added a few Relay-specific ones myself.

Vinnl:
Ah, the badwords filter just checks whether the bad word is part of the chosen domain or alias, whereas the blocklists match on the whole domain (e.g. it contains `my`).

I've now split it up to match on the whole word for the blocklist. I don't know Python though, so please pay extra close attention when reviewing :)

emails/apps.py
```diff
@@ -30,8 +30,21 @@ def __init__(self, app_name, app_module):
...
+        blocklist_file_path = os.path.join(
+            settings.BASE_DIR, 'emails', 'blocklist.txt'
+        )
+        with open(blocklist_file_path, 'r') as blocklist_file:
+            for word in blocklist_file:
+                if len(word.strip()) > 0 and word.strip()[0] == "#":
+                    continue
+                blocklist.append(word.strip())
+        self.blocklist = blocklist
+
```
groovecoder, line 48: Might be able to reduce some code duplication here (note: I didn't test this, just an idea for it):
```
diff --git a/emails/apps.py b/emails/apps.py
index 2cd9cbc..4eea17e 100644
--- a/emails/apps.py
+++ b/emails/apps.py
@@ -22,29 +22,24 @@ class EmailsConfig(AppConfig):
         except Exception:
             logger.exception("exception during SES connect")
 
-        badwords = []
         # badwords file from:
         # https://www.cs.cmu.edu/~biglou/resources/bad-words.txt
-        badwords_file_path = os.path.join(
-            settings.BASE_DIR, 'emails', 'badwords.txt'
-        )
-        with open(badwords_file_path, 'r') as badwords_file:
-            for word in badwords_file:
-                if len(word.strip()) > 0 and word.strip()[0] == "#":
-                    continue
-                badwords.append(word.strip())
-        self.badwords = badwords
+        self.badwords = self._load_terms('badwords.txt')
+        # blocklist file from:
+        # https://github.com/michaldudek/subdomain-blacklist
+        self.blocklist = self._load_terms('blocklist.txt')
+
 
-        blocklist = []
-        blocklist_file_path = os.path.join(
-            settings.BASE_DIR, 'emails', 'blocklist.txt'
-        )
-        with open(blocklist_file_path, 'r') as blocklist_file:
-            for word in blocklist_file:
-                if len(word.strip()) > 0 and word.strip()[0] == "#":
+    def _load_terms(self, filename):
+        terms = []
+        terms_file_path = os.path.join(settings.BASE_DIR, 'emails', filename)
+        with open(terms_file_path, 'r') as terms_file:
+            for term in terms_file:
+                if len(term.strip()) > 0 and term.strip()[0] == "#":
                     continue
-                blocklist.append(word.strip())
-        self.blocklist = blocklist
+                terms.append(term.strip())
+        return terms
+
 
     def ready(self):
         import emails.signals
```

groovecoder (changes requested):
1 possible cleanup in `emails/apps.py`. Argh, looks like our previous change to verify the subdomain values forgot to include a DB migration that increased the length of the subdomain field in the DB. I'll need to fix that separately.

emails/apps.py
Vinnl: Heh, I was being a bit lazy since it would cost me more time than I'd like to look up how to do this in Python, so thanks for the helpful pointer :) e9b4016

groovecoder (approved):

Vinnl:
Oh oops, meant to rebase before merging, now my fixup commit's still in there :sweat_smile:

PR #1115, opened 2021-09-18
initial api: django-rest-framework

groovecoder:
## New features for backend label storage
* `Profile.server_storage` is a boolean toggle for whether or not the user has enabled/disabled server storage
* `RelayAddress.generated_for` is a 255-character text field to hold the site on which the user generated the alias

## The API for the Relay back-end... 
* It uses [Django REST Framework](https://www.django-rest-framework.org/) library and [drf-yasg](https://drf-yasg.readthedocs.io/en/stable/) for auto-generated docs.
* The API lives at `/api/` and the docs live at `/api/docs/`.
   * `/api/` has a browsable interface if `DEBUG` is set to `True`
   * `/api/docs/` is only available if `DEBUG` is set to `True`
 * The API currently exposes `Profile`, `DomainAddress`, and `RelayAddress` models via HTTP + JSON endpoints.
 * Every Model/endpoint filters its responses to the data associated with `request.user`
    * Note: this leaves the existing auth mechanisms for the add-on and website un-touched.

## TODO
* [x] Verify POST, PUT, and DELETE operations are properly access-controlled by the `IsOwner` permission
* [x] Update Serializers to make some fields read-only
* [x] ~move blocklist and badword checks "up" into model validators for Relay and Domain addresses~
   * Instead, I marked the model fields that have extra/special validation logic to read_only_fields so that clients can't write to them unless/until we refactor that validation logic.

### Browsable API at `/api/`
image
image
### Interactive Swagger Docs at `/api/docs/`
image
image
image


Vinnl:
I'll use this comment to gather my current findings trying to use this:

- According to the Swagger interface I should be able to call `/relayaddresses/{id}/` (I'm particularly interested in `PUT` or `PATCH` to update a label), but I don't see a way of obtaining `id`.
- What does it mean that `/profiles/` can return multiple profiles? I'd assume each user has exactly one?
- Do you know already how we'll be able to query whether the user has opted in to storing labels on the server? Or will we be tracking that client-side?
- Is `description` meant to fit the label? And do we have a place to store the linked origin yet?

groovecoder:
> I'll use this comment to gather my current findings trying to use this:
> 
>     * According to the Swagger interface I should be able to call `/relayaddresses/{id}/` (I'm particularly interested in `PUT` or `PATCH` to update a label), but I don't see a way of obtaining `id`.

Fixed this. Now when you look at `/relayaddresses/` list, it should include the `id` of each.

> 
>     * What does it mean that `/profiles/` can return multiple profiles? I'd assume each user has exactly one?

This is just the way that DRF works out-of-the-box. The endpoint is indeed filtered to only show the "profiles" that belong to the requesting user, and it should only ever return one. If we want, I can look into customizing the endpoint so we have just an `/api/profile` endpoint instead?

>     * Do you know already how we'll be able to query whether the user has opted in to storing labels on the server? Or will we be tracking that client-side?

Added a `server_storage` boolean field to the profile.

> 
>     * Is `description` meant to fit the label? And do we have a place to store the linked origin yet?

Yes, we can use `description` for the label. I added a `generated_for` field to the `RelayAddress` model to store the linked origin.

> 
>     * I guess more of a general back-end/Django question, but do I just make root-relative requests (e.g. to `/api/whatever`), or should/can I get the website root URL in some way?

Start with root-relative requests.

groovecoder:
Oh, and you'll need to run `python manage.py migrate` to create the new db fields on the models.

moz-astults:
@groovecoder We should add a versioning scheme, or at least a `v1` in these paths. That way, if we ever need to make breaking changes, we can still maintain backwards compatibility and avoid tightly coordinated changes for anything that calls this API.

groovecoder:
Added `/v1/` url space for the API.

Also checked that the API properly controls write-access to only the resources that the `request.user` owns, though I think it's actually doing this with the `ModelViewSet.get_queryset` function, because the response is a `404` rather than an permission denied response. The effect is the same, and maybe preferred(?) - because a `404` in both cases (ID doesn't exist, OR ID exists but the client request is not allowed to write to it) helps to hide some info.

I did notice that the API endpoints don't go thru the our checks for blocklist or bad words, so we should probably:

* [ ] move blocklist and badword checks into custom `save()` methods for each `Model`

api/views.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def update(self, request, pk=None, *args, **kwargs):
+        profile = request.user.profile_set.first()
+        if not profile.has_unlimited:
+            raise CannotMakeSubdomainException('error-premium-check-subdomain')
+        subdomain = request.data.get('subdomain', None)
+        available = Profile.subdomain_available(subdomain)
+        if not available:
+            raise CannotMakeSubdomainException('error-subdomain-not-available')
+        return super().update(request, pk, *args, **kwargs)
```
say-yawn, line 60: This does not currently update anything on the `Profile` since the `subdomain_available` only checks to see if the subdomain is available. If this is a call to check to see if the `subdomain` is available we should create a separate routing/path using `action` as seen in example [here](https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing).

privaterelay/settings.py
```diff
@@ -170,12 +172,7 @@
     ]
 
 
-def download_xpis(headers, path, url):
-    if path.endswith('.xpi'):
-        headers['Content-Disposition'] = 'attachment'
-
-WHITENOISE_ADD_HEADERS_FUNCTION = download_xpis
-
```
say-yawn, line 178: I'm guessing this was for [signing Firefox add-ons](https://wiki.mozilla.org/Add-ons/Extension_Signing) which is now in separate repo?

say-yawn (changes requested):
First pass through

api/views.py
say-yawn: 🤦 should've known that the `super` is calling update on the entire object based on request data which are validated against serializer fields. Ignore my comment.

privaterelay/settings.py
groovecoder: Well, it was when we though we would self-distribute the `.xpi` file thru the website. But we're linking over to AMO now.

api/views.py
groovecoder: Updated.

Vinnl:
@groovecoder Is the namespace now `/v1/` or `/api/v1/`? I'd prefer the latter; if we ever want to separate out the front-end from the API, it would be more future-proof if requests to anything under `/api/` could pass through to the back-end server rather than having to manually maintain a list of API versions.

As for `/profiles/`: if it's not very little work I'd just keep it this way - just wanted to make sure I understood the data model :)

Vinnl:
@groovecoder Trying this out now, sending a PUT from a `fetch` (with `credentials: "same-origin"`), I see this in the response:

```
PUT /api/v1/profiles/6/

HTTP 403 Forbidden
Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": "CSRF Failed: CSRF token missing or incorrect."
}
```

Is there a way for me to obtain the CSRF token to send with the request from JS?

groovecoder:
Sadly, I punted on the problem of moving data validation logic "up" into model layer code.

Instead, I marked the model fields that have extra/special validation logic to `read_only_fields` so that clients can't write to them unless/until we refactor that validation logic.

But, with that done, I'm marking this as ready for final review.

say-yawn (approved):
LGTM

moz-astults (approved):
LGTM!

groovecoder:
Cool, thanks! I'm also going to see if I can play with https://github.com/microsoft/restler-fuzzer to do some fuzz-testing of the API endpoints.

PR #1117, opened 2021-09-20
Prevent focused .btn-blue--ghosts being unreadable

### LINKED ISSUE #974: "Get unlimited aliases" button text becomes white when changing tabs
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit
macOS 10.14 

**Browsers tested on:**
Firefox

**Preconditions:**
Relay extension is not installed.
No upgrade to MPP.

**Steps to reproduce:**

2. Click on “Get unlimited aliases” button and then go back to Firefox Relay tab.

**Actual result:**
The "Get unlimited aliases" text is written with white color and it is not that visible.

**Expected result:** 
The "Get unlimited aliases" text is written with blue color like before.

Get unlimited aliases button text white

Vinnl:
Fixes #974.

They would become white-on-transparent when focused/active due to the default styling
of visited links.

### DIFF
--- a/static/scss/components/buttons.scss
+++ b/static/scss/components/buttons.scss
@@ -24,7 +24,17 @@
     color: $color-blue-50;
     background: none;
 
-    &:hover {
+    &:focus,
+    &:active,
+    &:visited:focus,
+    &:visited:active {
+        color: $color-blue-50;
+        outline: none;
+        box-shadow: var(--buttonFocus);
+    }
+
+    &:hover,
+    &:visited:hover {
         color: $color-white;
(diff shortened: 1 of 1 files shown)


Vinnl:
Unless the `tabindex` is explicitly disabled, using proper `<button>`s or `<a>`s should make them available to tab to. We have a number of other accessibility issues with buttons though, such as `cursor: pointer` being explicitly disabled and I suppose more of them not having focus states. I still have some local, uncommitted changes to bring those up for discussion, but they're mostly separate from this specific issue I think :)

PR #1121, opened 2021-09-23
Use mozmail on domain address 708

say-yawn:
# About this PR
Since we are moving away from generating aliases with relay.firefox.com to generating them with mozmail.com, the Domain Aliasing should also use the mozmail.com

# Acceptance Criteria
- [x] Domain Address is generated with mozmail.com equivalent domain
- [x] Alias generated with Domain Address with mozmail.com domain is relayed properly
- [x] Reply on emails with Domain Address works
- [x] Domain Address is displayed with correct address on profile page
- [x] Domain Address is displayed with correct address on header/footer of relayed email

# Need to follow-up
- Currently we have two way of generating the full address of a given alias by using: `relay_address.address`@`relay_address.domain_value` or with `relay_address.full_address`. For DomainAddress changes I used `domain_address.full_address`. We should consider unifying it one way or the other.
- Edited logic on checking for bad words or block listed words to check for all DomainAddress generated rather than just randomly generated ones.

#
Related to #1112 
Related to #993 
Completes [MPP-708](https://mozilla-hub.atlassian.net/browse/MPP-708)
Completes [MPP-713](https://mozilla-hub.atlassian.net/browse/MPP-713)


emails/models.py
```diff
@@ -318,6 +318,7 @@ class DomainAddress(models.Model):
     address = models.CharField(max_length=64)
     enabled = models.BooleanField(default=True)
     description = models.CharField(max_length=64, blank=True)
+    domain = models.PositiveSmallIntegerField(choices=DOMAIN_CHOICES, default=2)
```
groovecoder, line 321: Should we change `RelayAddress.domain` to `default=2` also?

emails/views.py
```diff
@@ -453,14 +453,16 @@ def _handle_reply(message_json):
...
-def _get_domain_address(to_address, local_portion, domain_portion):
+def _get_domain_address(local_portion, domain_portion):
     address_subdomain = domain_portion.split('.')[0]
+    address_domain = domain_portion.split('.')[1]
     try:
         user_profile = Profile.objects.get(subdomain=address_subdomain)
+        domain_numerical = get_domain_numerical(address_domain)
         # filter DomainAddress because it may not exist
-        # which will throw an error with get()
+        # which will throw an error with get()
```
groovecoder, line 463:
```suggestion
        # which will throw an error with get()
```

privaterelay/templates/profile.html
```diff
@@ -183,7 +183,7 @@
 
               <!-- delete email address -->
               <div class="column-delete">
-                <form action="{% url 'emails-index' %}" method="POST" class="delete-email-form xflx jst-cntr al-cntr" data-delete-relay="{{ relay_address.address }}@{{ RELAY_DOMAIN }}">
+                <form action="{% url 'emails-index' %}" method="POST" class="delete-email-form xflx jst-cntr al-cntr" data-delete-relay="{{ relay_address.address }}@{{ relay_address.domain_value }}">
```
groovecoder, line 186: Yeah, I like the idea to use `relay_address.full_address` everywhere instead of using `{{ relay_address.address }}@{{ relay_address.domain_value }}`

emails/models.py
say-yawn: No, because that will make everyone irregardless of them being a premium user create alias with the mozmail.com. We should do this if/when Tony is ready to have everyone generate the alias with the new domain.

privaterelay/templates/profile.html
say-yawn: Done at https://github.com/mozilla/fx-private-relay/pull/1121/commits/f0d1093363f7cc27a7edc039be40f1ce9624f884

emails/views.py
```diff
@@ -453,14 +453,16 @@ def _handle_reply(message_json):
         return HttpResponse("SES client error", status=400)
 
 
-def _get_domain_address(to_address, local_portion, domain_portion):
+def _get_domain_address(local_portion, domain_portion):
```
groovecoder, line 456: This causes `TypeError: _get_domain_address() takes 2 positional arguments but 3 were given` on `views.py(488)_get_address()` which is accidentally covered up by a too-broad `try`/`except Exception` at `views.py(303)`. 😢

Needs this:
```
index 200cc2b..c2f9dcb 100644
--- a/emails/views.py
+++ b/emails/views.py
@@ -484,7 +483,7 @@ def _get_address(to_address, local_portion, domain_portion):
     # it may be for a user's subdomain
     email_domains = get_domains_from_settings().values()
     if domain_portion not in email_domains:
-        return _get_domain_address(to_address, local_portion, domain_portion)
+        return _get_domain_address(local_portion, domain_portion)
```

emails/views.py
```diff
@@ -453,14 +453,16 @@ def _handle_reply(message_json):
...
     address_subdomain = domain_portion.split('.')[0]
+    address_domain = domain_portion.split('.')[1]
```
groovecoder, line 458:
```suggestion
    [address_subdomain, address_domain] = domain_portion.split('.', 1)
```

groovecoder (changes requested):
Couple changes needed to make it work on my local.

emails/views.py
groovecoder: And we may want to change that `except` to more specifically catch just the `Exception`s that it needs to.

emails/views.py
say-yawn: Ooh nice!

groovecoder (approved):
Code and spot-checks look good! It's ready to be not a draft and ready to merge!

PR #1125, opened 2021-09-23
Check user created before premium release

say-yawn:
# About this PR
Provide information on whether or not the user was created before the premium release to inform users on pref changes.

# Needs Follow-up
~Need to make sure that the `joined_date` from the `User` Django object is actually the date user joins the Relay service AND NOT the date user joined FxA.~ Confirmed User's `date_joined` value comes from when the user joins our app and NOT when user created account on FxA by going to db and checking the `date_joined` and comparing it against when I first created the account (confirmation email).

# 
[Subtask 1199](https://mozilla-hub.atlassian.net/browse/MPP-1199) for main [ticket 1059](https://mozilla-hub.atlassian.net/browse/MPP-1059)


emails/models.py
```diff
@@ -157,6 +157,11 @@ def emails_blocked(self):
         ).values('num_blocked')
         return sum(blocked['num_blocked'] for blocked in relay_addresses_blocked)
 
+    @property
+    def created_before_premium_release(self):
+        date_created = self.user.date_joined
+        return date_created < datetime(2021, 10, 27, 18, 00)
```
groovecoder, line 163: Can we make this come from an env var so it's not hard-coded? Like, `2021-10-27` ?

groovecoder (changes requested):

say-yawn:
@groovecoder good point. That's probably a good idea. I created a subtask [here](https://mozilla-hub.atlassian.net/browse/MPP-1199) for pref behavior [ticket 1059](https://mozilla-hub.atlassian.net/browse/MPP-1059) to capture this PR but mentioning this on [ticket 1059](https://mozilla-hub.atlassian.net/browse/MPP-1059) would be a good idea too.

privaterelay/settings.py
```diff
@@ -242,6 +243,7 @@ def _get_initial_middleware():
 SUBSCRIPTIONS_WITH_UNLIMITED = config(
     'SUBSCRIPTIONS_WITH_UNLIMITED', default=''
 )
+PREMIUM_RELEASE_DATE = config('PREMIUM_RELEASE_DATE', str(datetime.now(timezone.utc)), cast=str)
```
groovecoder, line 246: can we go ahead and do `datetime.fromisoformat(config('PREMIUM_RELEASE_DATE', str(datetime.now(timezone.utc)), cast=str))` ? so we can just treat `settings.PREMIUM_RELEASE_DATE` as a `datetime` object elsewhere in the code?

Specifically, I'm thinking of how I'll implement "automatically opted in to data collection" after `settings.PREMIUM_RELEASE_DATE` and I think I'll change `Profile.server_storage` field to:

```
def default_server_storage_value():
    return datetime.now(timezone.utc) > settings.PREMIUM_RELEASE_DATE
...
class Profile(models.Model):
    ...
    server_storage = models.BooleanField(default=default_server_storage_value)
```

groovecoder (changes requested):

emails/tests/models_tests.py
```diff
@@ -511,6 +511,24 @@ def test_display_name_does_not_exist(self):
         profile = Profile.objects.get(user=social_account.user)
         assert profile.display_name == None
 
+    @override_settings(PREMIUM_RELEASE_DATE=datetime.fromisoformat('2021-10-27 17:00:00+00:00'))
+    def test_user_joined_before_premium_release_returns_True(self):
+        user = baker.make(
+            User,
+            date_joined=datetime.now(timezone.utc)
```
groovecoder, line 518:
```suggestion
            date_joined=datetime.fromisoformat('2021-10-18 17:00:00+00:00')
```
or this test will start failing on premium release day.

groovecoder (changes requested):

PR #1129, opened 2021-09-24
fix #1126 replies for FXA's with plus addressing

### LINKED ISSUE #1126: Reply to emails through alias issue
**Build:**
Stage

**Affected platforms:**
Windows 10 64 bit

**Browsers tested on:**
Firefox

**Preconditions:**
The user has already added a few aliases.

**Steps to reproduce:**

2. Copy an alias address.
3. Go to inbox and send an email to the original address through the alias.
3. Check the email received and click Reply.
4. Verify your inbox for the Reply email.

**Actual result:**
The Reply email is not received. The Forwarded emails number is not increased.

**Expected result:**
The Reply email is received. The Forwarded emails number is increased.

**Note:** 
I tried it with both existing and new Relay accounts and the result is the same.

reply issue

groovecoder:
strip the localpart "tags" before checking that the From: value
is a valid premium Relay user.

also some new convenient Reply properties, and some refactoring of
_handle_reply code into some sub-functions:

_get_keys_from_headers, _get_reply_record_from_lookup_key, and
_strip_localpart_tag

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -301,16 +301,7 @@ def _sns_message(message_json):
         user_profile = address.user.profile_set.first()
     except Exception:
         if to_local_portion == 'replies':
-            user = User.objects.get(email=from_address)
-            profile = user.profile_set.first()
-            if not profile.has_premium:
-                # TODO: send the user an email
-                # that replies are a premium feature
-                incr_if_enabled('free_user_reply_attempt', 1)
-                return HttpResponse(
-                    "Rely replies require a premium account", status=403
-                )
-            return _handle_reply(message_json)
+            return _handle_reply(from_address, message_json)
 
         return HttpResponse("Address does not exist", status=404)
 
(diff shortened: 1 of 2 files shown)


emails/views.py
```diff
@@ -393,22 +384,50 @@ def _sns_message(message_json):
...
+def _get_reply_record_from_lookup_key(lookup_key):
     lookup = b64_lookup_key(lookup_key)
-    reply_record = Reply.objects.get(lookup=lookup)
-    address = reply_record.relay_address or reply_record.domain_address
+    return Reply.objects.get(lookup=lookup)
+
+
+def _strip_localpart_tag(address):
+    [localpart, domain] = address.split('@')
+    subaddress_parts = localpart.split('+')
```
Vinnl, line 405: As far as I know no spec defines that the `+` part is stripped from email addresses - it's just a convention used by e.g. Gmail. In other words, it's perfectly possible that for some email providers, email addresses with a `+` and the same prefix before the `+` should go to different people.

In practice I suppose it'll probably be fine, but I thought it couldn't harm to call it out :)

groovecoder: Hmm ... good point. And the other complication is that we don't know which domains are "backed" by an email provider like Gmail that supports localpart tags.

After walking thru how this affects this reply code if an email provider `example.com` treats `user123` and `user123+456` as *different* email addresses ...

This will break `user123+456@example.com`'s replies if:
* `user123+456@example.com` IS a premium relay user
* `user123@example.com` is also a relay user
* `user123@example.com` is NOT a premium user

So, I updated the code to check if the `From` matches either exactly OR with the localpart tag stripped off.

Good catch!

say-yawn (approved):
Changes LGTM. Really love that the replies chunk of code was able to be broken into functions. Maybe tests soon?

PR #1130, opened 2021-09-24
Fix #1123 - Disable Upgrade CTA Button when user is already premium

### LINKED ISSUE #1123: “Get unlimited aliases” button is present even if the user has an active subscription
#### Build:
- Stage;

#### Affected Platforms:
- Across platforms;

#### Browsers tested on:
- Firefox, latest version;

#### Prerequisites:
- Have at least 5 aliases generated;

#### Steps to reproduce:
2. Observe the Relay dashboard, specifically the alias list section;
3. Click on the “Get unlimited aliases” button and observe the results;
4. Return to the Relay dashboard by clicking on the “Continue to download” button;
5. Observe the Relay dashboard again;

#### Expected result:
- The premium user sees the “Generate new alias” button;

#### Actual result:
- The premium user sees the “Get unlimited aliases” button instead of the generate one;
- The premium user can click on the ”Get unlimited aliases” which will redirect him to a subscription page that acknowledges the user has already subscribed;
- Returning to the dashboard after the above does not change anything;

#### Notes:
- This is related to #1122  ;
- This issue seems to be present to all accounts, new and old;
- The user can generate new aliases through the context menu or Relay add-on field button;
problem2
problem3
problemvid2

codemist:
This removes the upgrade CTA button along the search container when the user is already premium.

Demo: https://drive.google.com/file/d/1--A9qCgsAKJV-gdpsGLP-ADfp6Q6p-Px/view?usp=sharing

### DIFF
--- a/privaterelay/templates/includes/dashboard-filter.html
+++ b/privaterelay/templates/includes/dashboard-filter.html
@@ -103,7 +103,7 @@
         <!-- Button -->
         <form action="{% url 'emails-index' %}" class="dash-create" method="POST">
             <input type="hidden" name="api_token" value="{{ user_profile.api_token }}">
-            {% if user_profile.at_max_free_aliases and not user_profile.has_unlimited %}
+            {% if user_profile.at_max_free_aliases and not user_profile.has_premium %}
                 {% if settings.PREMIUM_ENABLED %}
                     <a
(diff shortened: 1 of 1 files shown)


Vinnl (changes requested):
Hmm, it seems like the "Generate New Alias" button now is completely gone for users who have fewer than five aliases and don't have a Premium subscription?

codemist:
@Vinnl added a demo at the top, thanks for the catch :)

Vinnl (approved):
Works as expected!

PR #1137, opened 2021-09-28
Align on toolchain versions

Vinnl:
@codemist @maxxcrawford if you both also install [Volta](https://volta.sh/), we can be sure that we're all on the same versions of Node and npm, and hence that we won't have endless back-and-forth changes to `package-lock.json` when different people are on different versions :)

(I set it to Node 12 because that's what the README says, although the Dockerfile uses Node 14. On which version shall we align?)


maxxcrawford (approved):
I'm down! I uninstalled `nodenv`. Let's do this. :)

PR #1142, opened 2021-09-30
:red_circle: DNM: New settings page

Vinnl:
It's reachable, but not referred to yet. Also, even if the "Store
alias labels" preference is turned on, they're not actually stored
yet.

The copy is not final yet, and will still need to be added to the l10n repository.


privaterelay/templates/settings.html
```diff
@@ -0,0 +1,121 @@
...
+  data-premium-price-id="{{ settings.PREMIUM_PRICE_ID }}" 
+  data-aliases-used-val="{{ relay_addresses|length }}" 
+  data-emails-forwarded-val="{{ user_profile.emails_forwarded }}"
+  data-emails-blocked-val="{{ user_profile.emails_blocked }}"
+  {% with request.user.profile_set.first as user_profile %}
+    data-premium-subdomain-set="{{ user_profile.subdomain }}"
+  {% endwith %}
+  data-premium-enabled="{{ settings.PREMIUM_ENABLED }}"
+>
+</firefox-private-relay-addon-data>
```
Vinnl, line 29: Not actually sure if this is necessary? Copied over from `profile.html`.

privaterelay/templates/settings.html
```diff
@@ -0,0 +1,121 @@
...
+
+<main id="settings-main" class=" bg-light dashboard-container " data-api-token="{{ user_profile.api_token }}">
+  <div class="c-meta">
+    <h1 class="mzp-u-title-md c-settings-headline">{% ftlmsg 'settings-headline' %}</h1>
+    <ul class="c-meta--links">
+      <li>
```
Vinnl, line 35: This is still missing the "Contact us" link, because I didn't know the link target yet.

privaterelay/views.py
```diff
@@ -97,6 +97,20 @@ def profile(request):
     return render(request, 'profile.html', context)
 
 
+def settings(request):
+    if (not request.user or request.user.is_anonymous):
+        return redirect(reverse('fxa_login'))
+    if request.GET.get('fxa_refresh') == 1:
+        fxa = _get_fxa(request)
+        _handle_fxa_profile_change(fxa)
```
Vinnl, line 105: I'm not sure what this does, so I left it in?

static/js/settings.js
```diff
@@ -0,0 +1,61 @@
...
+        const response = await apiRequest(
+            `/profiles/${profileId}/`,
+            {
+                method: "PUT",
+                body: JSON.stringify(settings),
+            },
+        );
+        if (response.ok) {
+            // Re-render the page on the server, to make sure all {% if %} statements in the template are rendered correctly:
+            document.location.reload();
```
Vinnl, line 47: So this is terrible. I wanted to avoid trying to keep the front-end in sync with what the server would render (i.e. manually making the warning at the top appear or disappear or not when the form is submitted depending on the value of the setting), so I just reloaded the entire page to get the server to render it. The disadvantage is that there's no "Saved!" confirmation message (in addition to the page just getting a superfluous reload.)

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,116 @@
...
+            display: block;
+            padding: 0 $spacing-lg $spacing-md;
+
+            a {
+                @include text-body-lg;
+
+                img {
+                    margin-right: $spacing-sm;
+                    stroke: #5953F3;
+                    width: 15px;
```
Vinnl, line 95: Don't know if there's an appropriate token for this somewhere?

privaterelay/urls.py
```diff
@@ -36,6 +36,7 @@
         views.profile_subdomain,
         name='profile_subdomain'
     ),
+    path('accounts/settings/', views.settings, name='settings'),
```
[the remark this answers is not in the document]
tcinotto: works for me!

static/js/settings.js
```diff
@@ -0,0 +1,63 @@
...
+        const profileId = settingsForm.dataset.profileId;
+
+        const settings = {
+            server_storage: labelCollectionCheckbox.checked,
+        };
+
+        const response = await apiRequest(
+            `/profiles/${profileId}/`,
+            {
+                method: "PUT",
```
groovecoder, line 43: [DRF wants clients to use `PATCH` HTTP request method to do "partial_update" actions](https://www.django-rest-framework.org/api-guide/routers/#api-guide) - i.e., where the client only sends the fields that it wants to update.

This doesn't matter yet, because the `PUT /profiles/{id}/` endpoint only accepts the single `server_storage` field right now anyway, but, if we keep this `PUT` request with this object like ...

```
{"server_storage": true}
```
... and then later we add the `subdomain` field to the API endpoint ... then this code will break because it's sending a `PUT` request without a value for `subdomain`.

static/js/settings.js
groovecoder: I've ~seen~ written worse. 😉

groovecoder (commented):
Did a spot-check but couldn't quite check everything because of all the "???" text. Can you send a branch of the new messages (even if they're not final) up to the l10n repo so I can pull that down into my local to see things more realistically?

Also did quick code check of the python, html, and JS. I'll let @maxxcrawford and @codemist check the the css.

privaterelay/templates/settings.html
```diff
@@ -0,0 +1,121 @@
...
+  {% with request.user.profile_set.first as user_profile %}
+    data-premium-subdomain-set="{{ user_profile.subdomain }}"
+  {% endwith %}
+  data-premium-enabled="{{ settings.PREMIUM_ENABLED }}"
+>
+</firefox-private-relay-addon-data>
+
+<main id="settings-main" class=" bg-light dashboard-container " data-api-token="{{ user_profile.api_token }}">
+  <div class="c-meta">
+    <h1 class="mzp-u-title-md c-settings-headline">{% ftlmsg 'settings-headline' %}</h1>
```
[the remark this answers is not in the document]
Vinnl: Done, same branch name! https://github.com/mozilla-l10n/fx-private-relay-l10n/tree/MPP-1049-settings-page

static/js/settings.js
```diff
@@ -0,0 +1,63 @@
+(function() {
+	"use strict";
```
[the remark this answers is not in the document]
Vinnl: Admittedly I've been using ES modules for so long that it's been a long time I've manually written `"use strict";`, so I didn't give this much thought and just copied it from `profile.js`, but... [Yes!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#strict_mode_for_functions)

static/js/settings.js
```diff
@@ -0,0 +1,63 @@
...
+
+    // TODO Find a way to make this reusable across the app?
```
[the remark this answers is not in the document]
Vinnl: Would like to get input from @maxxcrawford and @codemist on that too - I'm kinda hesitant about adding globals and separating the import from usage, though I'm not sure whether e.g. module imports are supported widely enough already to be able to use those, or if there's a better solution.

privaterelay/templates/settings.html
Vinnl: 76f17e7

(Btw, I'm also not sure what the `get_social_accounts` and `load` statements above do...)

privaterelay/views.py
Vinnl: 85b3c35

static/js/settings.js
```diff
@@ -0,0 +1,63 @@
...
+            `/api/v1${path}`,
+            {
+                mode: "same-origin",
+                ...options,
+                headers: headers,
+            },
+        );
+    }
+
+    const settingsForm = document.querySelector("#settings-form");
```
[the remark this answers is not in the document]
Vinnl: Ha, so that's why I defaulted to first, but then I saw the user of `querySelectorAll` elsewhere in the codebase instead of `getElementsByTagName` (for example) and I looked into it a bit. Turns out the latter returns an `HTMLCollection` (which stays up to date with DOM changes) whereas the former just returns a static `NodeList`, so I figured that that was probably for performance reasons too.

But of course, that doesn't carry over to `getElementById` (which just returns a single element in the first place).

All of which is a long-winded way of saying: 652f38ead0ddf7dc1fa2ac140208fdc0bf243059.

static/js/settings.js
Vinnl: That's good to know! (/cc @maxxcrawford for visibility.) And I guess makes sense, in hindsight. 4ffa807424d26a6436718d1d184311ff3ad7daf8

Vinnl:
@groovecoder I realised that after toggling this off, we'll also need to remove the stored labels and origins. I've added 55df351 to do this in the front-end, but it might make more sense to do this in the back-end so it can be done in a transaction and with just a single HTTP request?

privaterelay/templates/settings.html
groovecoder: `load` is how Django templates import [custom tags and filters](https://docs.djangoproject.com/en/3.2/howto/custom-template-tags/). [`get_social_accounts` is a custom tag from `django-allauth`](https://django-allauth.readthedocs.io/en/latest/templates.html#social-account-tags) to get easier access to the user's social account objects. I don't see this template using `accounts` at all, though, so it looks like it's not needed. (And that would mean the template doesn't need to `load socialaccount` either.

privaterelay/templates/settings.html
groovecoder: This is for free users too, so I'm not sure we can link them to a "contact us"? @tcinotto - where do we want 200K+ free users to send their contact?

groovecoder:
Yes, I can add to this PR to remove the data from the back-end. Do you want to, for example, clean up the `fixup` commits *before* I add other commits, or can I just add to this?

Vinnl:
@groovecoder You can add to it and I'll clean it up before merging this in - just in case someone else has already checked out this branch locally. I've reverted the commit in which I remove the labels from the front-end.

static/scss/pages/settings.scss
maxxcrawford: Not really. I'm fan of `1rem` so it'll get larger to match their text size. That might work?

static/js/settings.js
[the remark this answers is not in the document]
maxxcrawford: As a stop gap, I'd recommend moving shared stuff to some sort of `utils.js` that we can toss on different pages (profile vs settings) and then keeping individual pages to call/interact with them. 

Not sure if that's as static as individual script tags or if we wanna go the module route. Are there any downsides to the static approach?

privaterelay/templates/settings.html
```diff
@@ -0,0 +1,104 @@
...
+    <ul class="c-meta--links">
+      <li>
+        <a
+          href="https://support.mozilla.org/"
+          target="_blank"
+          rel="noopener noreferrer"
```
maxxcrawford, line 22: Question: Do  we want referral traffic here? (Specifically to another page in our ecosystem)

privaterelay/templates/settings.html
```diff
@@ -0,0 +1,104 @@
...
+          class="glocal-fxa-settings-link"
+        >
+        <img src="/static/images/performance.svg" alt="">
+          {% ftlmsg 'settings-meta-status-label' %}
+        </a>
+      </li>
+    </ul>
+  </div>
+  <div class="c-settings">
+    <div id="save-error" class="hidden c-settings-form--notification">
```
maxxcrawford, line 45: Global conversation: The `--` line isn't in line with our current naming convention. I mention it bc I know it's a known pattern in [BEM naming](http://getbem.com/naming/). 

Is this something we should revise to match everything else? I don't have a hard opinion about it, but there's only a [few minor instances](https://github.com/mozilla/fx-private-relay/blob/main/static/scss/components/buttons.scss#L12-L22) (that I wrote 🙈)  of that naming convention elsewhere in the code base. I would like to stick to convention, unless there's a strong reason to deviate.  

_(Flagging this instance here only) _

image

static/images/help.svg
```diff
@@ -0,0 +1,8 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
```
maxxcrawford, line 3: Nice. 😎

static/js/settings.js
```diff
@@ -0,0 +1,64 @@
...
+
+    // TODO Find a way to make this reusable across the app?
+    function apiRequest(path, options) {
+        const cookieString = typeof document.cookie === "string" ? document.cookie : "";
+        const cookieStringArray = cookieString
+            .split(";")
+            .map(individualCookieString => individualCookieString.split("="))
+            .map(([cookieKey, cookieValue]) => [cookieKey.trim(), cookieValue.trim()]);
+        // Looks like the `argsIgnorePattern` option for ESLint doesn't like array destructuring:
+        // eslint-disable-next-line no-unused-vars
```
maxxcrawford, line 12: 👍

static/js/settings.js
```diff
@@ -0,0 +1,64 @@
...
+                ...options,
+                headers: headers,
+            },
+        );
+    }
+
+    const settingsForm = document.getElementById("settings-form");
+    const labelCollectionCheckbox = document.getElementById("label-collection");
+    const labelCollectionOffWarning = document.getElementById("label-collection-off-warning");
+    const saveError = document.getElementById("save-error");
```
maxxcrawford, line 31: nit: I think all of this is specific to the Settings page, but if not, I would recommend migrating from an ID to a `js-` prefix class for more usability.

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,120 @@
+#settings-main {
```
maxxcrawford, line 1: Blocking request: Could you swap this to a `.settings-main` class? (As they can be super difficult to override, as outlined in the [Protocol coding docs](https://protocol.mozilla.org/docs/css-guide.html#simple-selectors)) I think we should refrain from any ID selector name-spacing when possible.

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,120 @@
...
+
+    .c-settings-headline {
+        @include text-title-2xs;
+        padding: $spacing-lg $spacing-lg $spacing-md ;
```
maxxcrawford, line 5: nit: ⚡ this line! (Channeling @groovecoder) 

```suggestion
        padding: $spacing-lg $spacing-lg $spacing-md;
```

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,120 @@
...
+        background-color: $color-white;
+        margin: $spacing-md;
+        padding: $spacing-md;
+        border-radius: $border-radius-md;
+    }
+
+    .c-settings-form--notification-heading {
+        @include text-title-3xs;
+        color: $color-grey-60;
+        border-left-width: 5px;
```
maxxcrawford, line 50: nit: This is an instance where we can make the design work within the design system. I know we don't have a specific token for this, but could you make it a `base 8` unit: `4px`?

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,120 @@
...
+        border-left-style: solid;
+        margin: 0;
+        padding: $spacing-sm $spacing-md;
+
+        &.t-error {
```
maxxcrawford, line 55: `t-` class 👌

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,120 @@
...
+        box-shadow: none;
+        text-align: start;
+        margin-left: 0;
+    }
+
+    .c-settings-form--submit {
+        padding: $spacing-sm $spacing-2xl;
+    }
+
+    .c-settings-form--notification {
```
maxxcrawford, line 39: Info: I wanted to call this out here, but @codemist may have already tagged you. She also created the [same settings-notification](https://github.com/mozilla/fx-private-relay/pull/1154) (visible on the profile page, with some additional functionality). Could you two sync up your code here? That may mean moving it out of this partial entirely.

maxxcrawford (changes requested):
Nice job. I know there's still some TODOs but this checks all the boxes we needed to get the Settings off the ground! 

I only had one blocking change, a few questions and a few nits. 😺

static/scss/pages/settings.scss
codemist: ref: https://github.com/mozilla/fx-private-relay/pull/1154

codemist: @maxxcrawford for my PR, the scss code is in `main.scss` right now. Would it be better fit in `settings.scss`?

groovecoder:
@Vinnl I added https://github.com/mozilla/fx-private-relay/pull/1142/commits/0dc4507022302bf07a86c472148a7cdbba736da2 to automatically remove the server data when a profile's `server_storage` is set to `False`. (@say-yawn can you check out that commit for me?)

privaterelay/templates/settings.html
Vinnl: I'm not sure if it's still necessary today, but it's used more as a way to get Firefox to do what it's supposed to do with `noopener`: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

privaterelay/templates/settings.html
Vinnl: Heh, I was wondering about that - just wasn't sure what the convention then is. Just more `c-` classes? (Maybe you could suggest a name for this specific class - I can then probably deduce appropriate class names for the other instances from that.)

static/images/help.svg
Vinnl: Haha neat, hadn't even seen that - just copied it from the Firefox source code :)

privaterelay/templates/settings.html
maxxcrawford: Ahh – Got'cha. We can leave it. 

@groovecoder Is there any sort of telemetry we want to add here?

static/scss/pages/settings.scss
maxxcrawford: @codemist @Vinnl No – It needs to be outside of both. Most likely in the `components/banners` SCSS file.

Vinnl: Oh nice, and yours is far more complete than mine! I've submitted a PR to extract that into a separate component that we can reuse, maybe we can collaborate on that and reuse it in our respective PRs? Or maybe it's a completely crazy approach and we better just use copy-paste the HTML in the different locations :)

https://github.com/mozilla/fx-private-relay/pull/1161

static/scss/pages/settings.scss
Vinnl: Heh, why doesn't stylelint have a rule we can turn on to catch this :) 3fb3187

privaterelay/templates/settings.html
groovecoder: We could add `utm_source=relay.firefox.com` (or `utm_source={{ settings.SITE_ORIGIN }}`) so that SUMO's analytics will show the traffic came from Relay.

groovecoder: Let's use https://support.mozilla.org/kb/activate-firefox-relay-hide-your-email-address-alias for now.

groovecoder (changes requested):
Local spot-check looks good. 1 code change request, and then if you and @maxxcrawford can align on any remaining changes, I think this is ready to go? (Assuming content/messages/strings get approved?)

privaterelay/templates/settings.html
Vinnl: d2db60d

static/js/settings.js
Vinnl: Whoops yeah, still have to get into that habit (haven't manually targeted elements from JS in a looong time): 753b47b

Feels a bit weird to be using class selectors for unique elements though :/

static/scss/pages/settings.scss
Vinnl: Haha, I was explicitly trying to avoid the styles being overridable/reusable to make sure they don't leak into other parts of the website :P But here: 6925537e48674cb604ccac31773e38473c8651a7

static/scss/pages/settings.scss
Vinnl: Oops, looks like my comment here got lost. So I went ahead and copied the exact text size: 87d83b7aa837ce6da223b9e6ccf67415daf0fac1

static/scss/pages/settings.scss
Vinnl: Let's work on reusing it in that PR, and replace the code here with whatever we land on there after this is merged.

Vinnl:
Strings are now ready, and blocking comments have been incorporated. I rebased it, so when https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/17 is incorporated in this repo I think it's ready to go. (Without a "Contact us" link.)

privaterelay/templates/settings.html
```diff
@@ -0,0 +1,104 @@
...
+
+{% block content %}
+
+<main class="settings-main bg-light dashboard-container " data-api-token="{{ user_profile.api_token }}">
+  <div class="c-meta">
+    <h1 class="mzp-u-title-md c-settings-headline">{% ftlmsg 'settings-headline' %}</h1>
+    <ul class="c-meta--links">
+      <li>
+        <a
+          href="https://support.mozilla.org/en-US/kb/activate-firefox-relay-hide-your-email-address-alias?utm_source={{ settings.SITE_ORIGIN }}"
```
groovecoder, line 20: We should remove the `en-US` part. SUMO has locale middleware that will automatically detect the browser language and redirect the user to the right translation. But if `en-US` is hard-coded into the url, this link will always only go to `en-US` translation of the page.

groovecoder (changes requested):

privaterelay/templates/settings.html
Vinnl: Whoops good catch. As a non-American, I should've caught that really: d4ee27a

Vinnl:
Strings are now merged in! @maxxcrawford could you approve and/or respond to open review questions?

privaterelay/views.py
```diff
@@ -97,6 +97,17 @@ def profile(request):
     return render(request, 'profile.html', context)
 
 
+def settings(request):
```
maxxcrawford, line 91: This is missing the avatar context, which is not setting/rendering the FxA profile image correctly. This could be added to the context processor, or exposed like the FAQ view does: https://github.com/mozilla/fx-private-relay/blob/main/privaterelay/views.py#L61-L65

Broken: 
image

Working: 
image

static/js/settings.js
```diff
@@ -0,0 +1,64 @@
...
+            {
+                method: "PATCH",
+                body: JSON.stringify(settings),
+            },
+        );
+        if (response.ok) {
+            // Re-render the page on the server, to make sure all {% if %} statements in the template are rendered correctly:
+            document.location.reload();
+            return;
+        }
```
maxxcrawford, line 52: Q: Is there any sort of error catching here?

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    .c-meta--links {
+        display: none;
+    }
+
+    .c-settings-form--heading {
+        @include text-title-3xs;
+        color: $color-grey-50;
+        font-weight: normal;
```
maxxcrawford, line 16: nit: We have been using numbered units to set font weighs else where. 
```suggestion
        font-weight: 400;
```

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,125 @@
...
+        display: block;
+    }
+
+    .c-settings-form--field {
+        border-bottom: 1px solid $color-grey-10;
+    }
+
+    .c-settings-form--inline-warning {
+        box-shadow: none;
+        text-align: start;
```
maxxcrawford, line 31: TIL! 

This is great. We need it for @codemist in a comment from @pdehaan.

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,125 @@
...
+            border-color: $color-error;
+        }
+
+        &.t-warning {
+            border-color: $color-warning;
+        }
+    }
+
+    .c-settings-form--notification-description {
+        border-left-width: 5px;
```
maxxcrawford, line 65: Can we use something  `4px` here? We're in a pseudo–base 8 format for CSS units. (With `4` being a `1/2` of [8](url)) 
```suggestion
        border-left-width: 4px;
```

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,125 @@
...
+            li {
+                padding: $spacing-sm 0;
+            }
+
+            a {
+                @include text-body-lg;
+
+                img {
+                    margin-right: $spacing-sm;
+                    stroke: #5953F3;
```
maxxcrawford, line 102: nit: Can we set this a variable color? I'm not sure where this color falls into the design system.

static/scss/pages/settings.scss
```diff
@@ -0,0 +1,125 @@
...
+        .c-settings-form--field {
+            display: flex;
+        }
+
+        .c-settings-form--heading {
+            width: $content-sm;
+        }
+
+        .c-settings-form--footer {
+            text-align: right;
```
maxxcrawford, line 122: Since I just learned this…

```suggestion
            text-align: end;
```

privaterelay/views.py
groovecoder: Yup, fixed.

static/scss/pages/settings.scss
Vinnl: https://github.com/mozilla/fx-private-relay/pull/1180/commits/0174cefcbe4edf8ef5e4f88c9261675a3c91c4f3

static/scss/pages/settings.scss
Vinnl: https://github.com/mozilla/fx-private-relay/pull/1180/commits/0afc499f686fa6c1775ee38d4ef93cdab34bfc4d

PR #1143, opened 2021-09-30
show domain modal for user.has_premium

groovecoder:
even if settings.PREMIUM_ENABLED is false


maxxcrawford (approved):
This is a nice logic fix. Get it! 🏁

PR #1145, opened 2021-10-01
Avoid calling methods on `undefined`

### LINKED ISSUE #1133: Aliases number issue in Search bar
**Build:**
Stage

**Affected platforms:**
Windows 10 64 bit
macOS 10.14
Android 11
iOS 14.8

**Browsers tested on:**
Firefox

**Preconditions:**

**Steps to reproduce:**

2. Click in the Search aliases bar and observe the number of aliases.

**Actual result:**
The number of aliases is not displayed. Only the slash is visible.

**Expected result:**
The number of aliases is correctly displayed.

aliases Number in search bar issue

Vinnl:
Fixes #1133.

### DIFF
--- a/static/js/filter.js
+++ b/static/js/filter.js
@@ -125,30 +125,36 @@
         document.removeEventListener("keydown", categoryMenuEscListener, false);
     }
     // TODO: Add "f" key listener to toggle category filter
-    filterToggleCategoryInput.addEventListener("click", () => {
-        if (filterToggleCategoryInput.classList.contains("is-enabled")) {
-            closeCategoryMenu();
-        } else {
-            openCategoryMenu();
-        }
-    });
+    // This element is only available when the user has Premium:
+    if (filterToggleCategoryInput) {
+        filterToggleCategoryInput.addEventListener("click", () => {
+            if (filterToggleCategoryInput.classList.contains("is-enabled")) {
+                closeCategoryMenu();
+            } else {
+                openCategoryMenu();
+            }
+        });
+    }
 
     function setCategoryFilters(categories) {
         activeCategoryFilters = categories;
         applyFilters();
     }
-    filterCategoryCheckboxes[0].form.addEventListener("submit", (event) => {
-        event.preventDefault();
-        const categoriesToSet = Array.from(filterCategoryCheckboxes)
-            .filter(checkbox => checkbox.checked)
-            .map(checkbox => checkbox.dataset.categoryType);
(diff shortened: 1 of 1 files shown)


static/js/filter.js
```diff
@@ -125,30 +125,36 @@
...
-            .filter(checkbox => checkbox.checked)
-            .map(checkbox => checkbox.dataset.categoryType);
-        setCategoryFilters(categoriesToSet);
-        closeCategoryMenu();
-    });
-    filterCategoryCheckboxes[0].form.addEventListener("reset", () => {
-        setCategoryFilters([]);
-        closeCategoryMenu();
-    });
+    // This element is only available when the user has Premium:
```
maxxcrawford, line 143: This comment put me on a path to test and I discovered a separate issue: https://github.com/mozilla/fx-private-relay/blob/1133-count-in-search-bar/privaterelay/templates/includes/dashboard-filter.html#L25-L26

The `user_profile.has_unlimited` changed to `user_profile.has_premium`. Can you update the logic here? This fix will restore the category filter to be visible to premium users.  

If not fixed in this PR, we can/need to open an issue.

maxxcrawford, line 143: This comment put me on a path to test and I discovered a separate issue: https://github.com/mozilla/fx-private-relay/blob/1133-count-in-search-bar/privaterelay/templates/includes/dashboard-filter.html#L25-L26

The `user_profile.has_unlimited` changed to `user_profile.has_premium`. Can you update the logic here? This fix will restore the category filter to be visible to premium users.  

If not fixed in this PR, we can/need to open an issue.

maxxcrawford (approved):
Great catch! 

See the nit below. It may warrant changing for this PR if possible.

Vinnl:
The catch is all @Bianca-O's, but great work indeed :)

static/js/filter.js
Vinnl: Nice find! Any idea how/why that happened? @codemist already [fixed](https://github.com/mozilla/fx-private-relay/pull/1130/files#diff-08d75e4f3be8657c70a33e3b353b7c6a6fc3ff628ecbe09c6cfdf952aa0fc880R106) a similar occurrence elsewhere, so it's a bit worrying that this was missed in multiple instances. I'll fix it in a separate PR :)

Vinnl: PR here: https://github.com/mozilla/fx-private-relay/pull/1152

PR #1152, opened 2021-10-04
Fix check for whether user has Premium

Vinnl:
Found by @maxxcrawford: https://github.com/mozilla/fx-private-relay/pull/1145#discussion_r721419135


maxxcrawford (approved):
Confirmed visible for premium and hidden for others. Thanks for the follow-up fix!

PR #1159, opened 2021-10-05
fix #1157: use == instead of "is" for checking relay user email

### LINKED ISSUE #1157: A non premium user can reply to emails sent through the relay
#### Build:
- Stage;
- Production;

#### Affected Platforms:
- Windows 10;

#### Browsers tested on:
- Firefox, latest version;

#### Prerequisites:
- Be signed in into a Firefox Relay account without an active subscription, specifically a non-Mozilla account;
- Have at least one alias generated;
- Have another email address from which to send emails on hand;

#### Steps to reproduce:
1. From the other email address send an email to a generated address;
2. From the email address tied to the Relay, reply to the mail received through the Relay;
3. Observe the results;

#### Expected result:
- The non premium user cannot reply through the Relay and the reply is not received;

#### Actual result:
- The reply is received, even though the user does not have premium benefits;

#### Notes:
- This happens on both stage and production environments;
reply1 1
reply1 2

groovecoder:
The reply code checks to see if the reply is coming from an email address that matches an existing Relay user. But it was using `is` which evaluates to `False` even when the values ARE equal. This means that Relay assumes (maybe another bug to dig into and fix?) the sender is the external party sending their reply.

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -417,8 +417,8 @@ def _handle_reply(from_address, message_json):
     stripped_reply_record_address = _strip_localpart_tag(reply_record_email)
 
     if (
-        (from_address is reply_record_email) or
-        (stripped_from_address is stripped_reply_record_address)
+        (from_address == reply_record_email) or
+        (stripped_from_address == stripped_reply_record_address)
     ):
         # This is a Relay user replying to an external sender;
         # verify they are premium


say-yawn (approved):
LGTM. `is` checks the `id` of the object and which means values being the same does not necessary have the same same `id`

PR #1154, opened 2021-10-04
Add data opt in banner

codemist:
Preview: https://drive.google.com/file/d/1SoqK4MwY5lmateXRqhEQ3LpIqxFmJ2WW/view?usp=sharing

Waiting on updated strings to get merged in https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/23

PR #1161, opened 2021-10-05
Add a partial for notifications

Vinnl:
OK, so there are multiple places we need notifications, and they've got a rather elaborate HTML structure. I tried to turn the banner as @codemist made it (in https://github.com/mozilla/fx-private-relay/pull/1154/) into a reusable template, but obviously not knowing Python nor Django it's probably not ideal - hence y'all as reviewers :)

This is how it's used:

```
{% load notification %}

{% notification 'warning' %}

          <p class="data-notification-banner-header">Allow data collection for future features</p>
          <p class="data-notification-banner-sub">This data will allow you to label your aliases with the sites where they're generated and used in a future release. You can allow this data collection in “Settings.”</p>

{% endnotification %}
```

What I don't like is that you still have to know the HTML to use for the header and the body, but I couldn't see how to pass template tags (specifically `{% ftlmsg %}` as arguments to other template tags.

### DIFF
--- a/privaterelay/templatetags/notification.py
+++ b/privaterelay/templatetags/notification.py
@@ -1,0 +1,41 @@
+from django import template
+
+register = template.Library()
+
+def do_notification(parser, token):
+    nodelist = parser.parse(('endnotification',))
+    parser.delete_first_token()
+    _tag_name, level = token.split_contents()
+    return NotificationNode(nodelist, level[1:-1])
+
+class NotificationNode(template.Node):
+    def __init__(self, nodelist, level):
+        self.nodelist = nodelist
+        self.level = 'info'
+        if (level == 'warning'):
+            self.level = 'warning'
+
+    def render(self, context):
+        content = self.nodelist.render(context)
+
+        theme = 't-info'
+        if (self.level == 'warning'):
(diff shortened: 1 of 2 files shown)


groovecoder:
Oh interesting. There's definitely a way to do this, where we would activate the local and call for the translation inside the template tag. (i.e., [use the bundle from python](https://django-ftl.readthedocs.io/en/latest/usage.html#using-bundles-from-python) rather than the django templates)

But we'd have to look into it a bit more.

maxxcrawford (commented):
I think there may be some miscommunication from my initial request. 

I want the SCSS for this feature broken out into a seperate partial from the settings page SCSS. The HTML can be as static as necessary. 

That said, I like this method, but am also in the same boat not knowing exactly what to do next.

@groovecoder Is it okay to have some of the markup passed in the argument?

groovecoder:
Let's avoid passing markup into template tags as arguments until we can dig into this more.

So it sounds like we can close this particular change - what we really want is a new SCSS partial?

maxxcrawford:
@groovecoder Correct. I'm going to close this but leave the branch up in case we need to revisit.

PR #1163, opened 2021-10-05
Increase subdomain length on profile

say-yawn:
# About this PR
Increase the subdomain length on profile as per spec from design


groovecoder (approved):
Code and local spot-check looks good.

PR #1164, opened 2021-10-05
Fix #1136 - Add ellipsis overflow to filter bar input

### LINKED ISSUE #1136: Placeholder text cut off in Search aliases bar when browser's language is set to Portuguese
**Build:**
Stage

**Affected platforms:**
Windows 10 64 bit
macOS 10.14
Android 11
iOS 14.8

**Browsers tested on:**
Firefox

**Preconditions:**
The browser's language is set to Portuguese (Brazil).

**Steps to reproduce:**

2. Observe the Search aliases bar placeholder text.

**Actual result:**
The text is cut off and not fully visible.

**Expected result:**
The text is fully visible or ellipsis is used.

portuguese - text not visible

lennartschoch:
Fixes #1136 by adding ellipsis overflow to the placeholder of the filter bar. Keeps the placeholder from being cut-off.

### DIFF
--- a/static/scss/components/filter-bar.scss
+++ b/static/scss/components/filter-bar.scss
@@ -72,6 +72,10 @@
         &:focus, &:hover {
             background-image: url("/static/images/icon-search-blue.svg");
         }
+
+        &::placeholder {
+            text-overflow: ellipsis;
+        }
     }
 
     .c-filter-search-controls {


Vinnl:
Thanks for your contribution @lennartschoch! Unfortunately, I don't think this is the best approach: the problem is the superfluous whitespace on the right-hand side of the input field, rather than there not being an ellipsis. Hope you don't mind my fixing that instead in #1175. Appreciate you looking into this though!

Vinnl (approved):

PR #1165, opened 2021-10-05
Increase email size limit beyong 150kb

say-yawn:
# Acceptance Criteria
- [ ] Get email message from S3 using the S3 client on boto3 Relay emails back to the user after processing the emails 
- [ ] Wait to respond back to SNS topic until after the email is successfully delivered on via SES
- [ ]  Make sure we are not repeatedly relaying emails to the user because we get multiple sns pings during the processing (race condition) 
- [ ] Delete emails after successfully relaying email to real users’ email
- [ ]  Add metrics on emails size Add metrics on email delivered over 150KB
- [ ]  Update email size limit to 10MB on SES (expandable to 40MB but most email clients do not accept emails sizes that big)


emails/utils.py
```diff
@@ -275,3 +277,13 @@ def decrypt_reply_metadata(key, jwe):
     e.deserialize(jwe)
     e.decrypt(k)
     return e.plaintext
+
+def get_message_content_from_s3(bucket, object_key):
+    emails_config = apps.get_app_config('emails')
+
+    # attachment = SpooledTemporaryFile()
+    with open('temp_file', 'w+b') as f:
```
groovecoder, line 285: Can we try https://stackoverflow.com/a/38910315 so we don't have to touch the file-system? Putting any files-from-the-internet in our app server's file-system makes me nervous.

groovecoder (changes requested):

emails/utils.py
say-yawn: Thanks for bringing this to my attention. I was using BufferedRandom to temporarily store the data and I agree with your concern and mediation on using streamed data instead. Instead of following the answer on StackOverflow, I used the [get_object](https://boto3.amazonaws.com/v1/documentation/api/1.9.42/reference/services/s3.html?highlight=delete_object#S3.Client.get_object) function in the S3 client which does the same thing and less code :)

emails/views.py
```diff
@@ -556,9 +566,19 @@ def _handle_bounce(message_json):
...
+        message_content = message_json['content'].encode('utf-8')
+    elif 'receipt' in message_json and 'action' in message_json['receipt']:
+        if 'S3' in message_json['receipt']['action']['type']: 
+            bucket = message_json['receipt']['action']['bucketName']
+            object_key = message_json['receipt']['action']['objectKey']
+            message_content = get_message_content_from_s3(bucket, object_key)
+            histogram_if_enabled(
+                'relayed_email.size',
+                len(message_content)
+            )
```
say-yawn, line 579: @groovecoder, using a histogram here so we can look at the data as an individual and in aggregate. Let me know if you think the name should be changed or anything?

groovecoder: Nice!

emails/utils.py
```diff
@@ -278,12 +278,33 @@ def decrypt_reply_metadata(key, jwe):
     e.decrypt(k)
     return e.plaintext
 
+
+class S3ClientException(Exception):
+    """Exception raised by Profile due to error on subdomain creation.
```
groovecoder, line 283: This comment looks like it's copy-pasted.

emails/views.py
```diff
@@ -391,12 +382,15 @@ def _sns_message(message_json):
...
         message_body, attachments, mail, address
     )
-    print(f'Response: {response}')
     address.num_forwarded += 1
     address.last_used_at = datetime.now(timezone.utc)
     address.save(
         update_fields=['num_forwarded', 'last_used_at']
     )
+    
+    message_removed = remove_message_from_s3()
```
groovecoder, line 391: This calls `remove_message_from_s3` without any arguments, but the function definition looks like it needs `bucket, object_key`?

emails/utils.py
```diff
@@ -275,3 +275,37 @@ def decrypt_reply_metadata(key, jwe):
...
+    """
+
+    def __init__(self, message=None):
+        self.message = message
+
+
+def get_message_content_from_s3(bucket, object_key):
+    try:
+        emails_config = apps.get_app_config('emails')
+        streamed_s3_object = emails_config.s3_client.get_object(
```
groovecoder, line 294: Not sure if this looks any cleaner to skip the `emails_config` var and go straight to `s3_client`?
```suggestion
        s3_client = apps.get_app_config('emails').s3_client
        streamed_s3_object = s3_client.get_object(
```

emails/utils.py
```diff
@@ -275,3 +275,37 @@ def decrypt_reply_metadata(key, jwe):
...
+        return streamed_s3_object.read()
+    except ClientError as e:
+        logger.error('s3_client_error_get_email', extra=e.response['Error'])
+    raise S3ClientException('Failed to fetch email from S3')
+
+
+def remove_message_from_s3(bucket, object_key):
+    try:
+        emails_config = apps.get_app_config('emails')
+        response = emails_config.s3_client.delete_object(bucket, object_key)
```
groovecoder, line 306: Not sure if this looks any cleaner to skip the `emails_config` var and go straight to `s3_client`?
```suggestion
        s3_client = apps.get_app_config('emails').s3_client
        response = s3_client.delete_object(bucket, object_key)
```

emails/views.py
```diff
@@ -279,16 +281,16 @@ def _sns_message(message_json):
             'Received SNS notification without commonHeaders.',
             status=400
         )
-
-    if 'to' not in mail['commonHeaders']:
+    common_headers = mail['commonHeaders']
```
groovecoder, line 284: nice clean-up.

emails/views.py
```diff
@@ -381,6 +387,14 @@ def _sns_message(message_json):
     address.save(
         update_fields=['num_forwarded', 'last_used_at']
     )
+
+    # only remove message from S3 if the email was stored in S3
+    if 'receipt' in message_json and 'action' in message_json['receipt']:
+        if 'S3' in message_json['receipt']['action']['type']:
+            bucket = message_json['receipt']['action']['bucketName']
+            object_key = message_json['receipt']['action']['objectKey']
+            remove_message_from_s3(bucket, object_key)
```
groovecoder, line 396: I was going to say we should handle failures to remove messages from S3, but it looks like `remove_message_from_s3` does enough error logging that we can detect problems and clean out the s3 bucket if we need to do it manually later?

emails/views.py
```diff
@@ -556,9 +570,19 @@ def _handle_bounce(message_json):
...
 def _get_text_html_attachments(message_json):
-    bytes_email_message = message_from_bytes(
-        message_json['content'].encode('utf-8'), policy=policy.default
-    )
+    if 'content' in message_json:
+        message_content = message_json['content'].encode('utf-8')
+    elif 'receipt' in message_json and 'action' in message_json['receipt']:
+        if 'S3' in message_json['receipt']['action']['type']:
+            bucket = message_json['receipt']['action']['bucketName']
+            object_key = message_json['receipt']['action']['objectKey']
```
groovecoder, line 578: These 3 lines are duplicated up in `_sns_message`. Could we put them in a helper function?
```
bucket, object_key = _get_bucket_and_key_from_s3_json(message_json)
```
?

emails/views.py
say-yawn: Yup. Since `remove_message_from_s3` runs in the background I wanted it to fail more silently. So we are logging it on Sentry AND on the metrics so ops and engineers can be more aware of it when the failure to delete the message object happens.

PR #1087, opened 2021-09-08
Wrap alias label input in a form"

Vinnl:
Reverts mozilla/fx-private-relay#1084 to recreate #1078 as requested in https://github.com/mozilla/fx-private-relay/pull/1078#issuecomment-915302190.

PR #1168, opened 2021-10-06
Display and store server-side labels

Vinnl:
This updates the profile page such that, if the user has server-side storage of labels enabled, they are displayed and, when edited, are stored to the server again.

Note: this builds on #1087, so probably easiest to just review 0ff57d2a704aea58d74fd78f35d31f4263248b65 in isolation.


static/js/labels.js
```diff
@@ -0,0 +1,104 @@
...
+            );
+            if (!response.ok) {
+                throw new Error("Immediately catch'd to show an error message.");
+            }
+            aliasLabelWrapper.classList.add("show-saved-confirmation");
+            labelInputElement.dataset.label = labelInputElement.value;
+            setTimeout(() => {
+                aliasLabelWrapper.classList.remove("show-saved-confirmation");
+            }, 1000);
+        } catch(e) {
```
maxxcrawford, line 74: Nice catch!

PR #1179, opened 2021-10-08
Link to the settings page from the menu

Vinnl:
I'm not sure what `glocal` is, but `.glocal-site-option` did not seem to be used anywhere yet and seemed to be intended for this use.

Depends on https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/20.


maxxcrawford (approved):
Approved!

I did have one note: The vertical alignment seems off. Could you vertically center this? Or maybe top-align it? (In case longer words break over for other items in the future.) Having it aligned on baseline seems off compared to the Log Out button. 

image

Vinnl:
You have a sharp eye! Fixed in 274f8d8188a39087650d365011578b01e78efdad.

PR #1183, opened 2021-10-08
fix mpp-1145: only exact match for short bad words

groovecoder:


emails/tests/models_tests.py
```diff
@@ -41,6 +41,12 @@ def test_has_bad_words_with_bad_words(self):
     def test_has_bad_words_without_bad_words(self):
         assert not has_bad_words('happy')
 
+    def test_has_bad_words_exact_match_on_small_words(self):
+        assert has_bad_words('ho')
+        assert not has_bad_words('horse')
+        assert has_bad_words('ass')
+        assert not has_bad_words('cassandra')
```
pdehaan, line 48: Do we need a couple more tests here for the cases where we have a 4+ letter word from the bad list? If I'm reading the code/rule correctly this should still mean that a domain like "shell" would be invalid, correct?

emails/models.py
```diff
@@ -223,10 +223,12 @@ def address_default():
 
 
 def has_bad_words(value):
-    return any(
-        badword in value
-        for badword in emails_config.badwords
-    )
+    for badword in emails_config.badwords:
+        if len(badword) <= 3 and badword == value:
+            return True
```
pdehaan, line 228: Potentially very irrelevant, but while looking at the /emails/badwords.txt list, I noticed a few of them have trailing whitespace. I don't know django, but just wanted to confirm that somewhere is trimming the bad word list before comparing word lengths. Although I think only one of the results is 3 characters (but 4 with trailing space) and might have a slightly different behavior if trimmed or not.

```sh
cat emails/badwords.txt | grep -e "\s$"
```

emails/models.py
```diff
@@ -223,10 +223,12 @@ def address_default():
...
 def has_bad_words(value):
-    return any(
-        badword in value
-        for badword in emails_config.badwords
-    )
+    for badword in emails_config.badwords:
+        if len(badword) <= 4 and badword == value:
+            return True
+        if len(badword) > 4 and badword in value:
+            return True
```
say-yawn, line 230:
```suggestion
        badword = badword.strip()
        if len(badword) <= 4 and badword == value:
            return True
        if len(badword) > 4 and badword in value:
            return True
```

say-yawn (changes requested):
One recommendation to strip for leading or trailing white spaces.

PR #1184, opened 2021-10-11
Make the word filter wider

### LINKED ISSUE #1136: Placeholder text cut off in Search aliases bar when browser's language is set to Portuguese

Vinnl:
In some languages the placeholder is rather long, so his gives it
more opportunity to be fully visible.

Fixes #1136, for good this time I hope.

### DIFF
--- a/static/scss/components/filter-bar.scss
+++ b/static/scss/components/filter-bar.scss
@@ -60,6 +60,7 @@
         position: relative;
         z-index: 0;
         transition: padding-right .2s ease-out;
+        width: $content-sm;
 
         &.is-filtered,
         &:focus {


static/scss/components/filter-bar.scss
```diff
@@ -60,6 +60,7 @@
         position: relative;
         z-index: 0;
         transition: padding-right .2s ease-out;
+        width: 400px;
```
maxxcrawford, line 63: I recommend using a [placeholder value](https://protocol.mozilla.org/fundamentals/tokens.html) here. Closest I see is `$content-sm`. 

```suggestion
        width: $content-sm;
```

Preview at that length: 
image

Vinnl:
Thanks, good call! And yeah, the other issue seems very much an edge case with very limited impact :)

PR #1191, opened 2021-10-12
Disallow whitespace-only labels

Vinnl:
Fixes https://github.com/mozilla/fx-private-relay-add-on/issues/104.

Using the `pattern` attribute doesn't provide a particularly
helpful error message, but since it's pretty clear what you're
doing wrong if you're only inserting whitespace, and also not
something people often do, I think that's probably fine.

### DIFF
--- a/privaterelay/templates/profile.html
+++ b/privaterelay/templates/profile.html
@@ -135,6 +135,8 @@
                       aria-label="{% ftlmsg 'profile-label-edit' %}"
                       placeholder="{% ftlmsg 'profile-label-placeholder' %}"
                       class="relay-email-address-label js-relay-email-address-label ff-Met"
+                      {# Require at least one non-whitespace character:  #}
+                      pattern=".*\S.*"
                     >
                     <span class="saved-confirmation">{% ftlmsg 'profile-label-saved' %}</span>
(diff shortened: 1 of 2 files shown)


privaterelay/templates/profile.html
```diff
@@ -135,6 +135,8 @@
                       aria-label="{% ftlmsg 'profile-label-edit' %}"
                       placeholder="{% ftlmsg 'profile-label-placeholder' %}"
                       class="relay-email-address-label js-relay-email-address-label ff-Met"
+                      {% comment %} Require at least one non-whitespace character: {% endcomment %}
```
maxxcrawford, line 138: As this isn't a [multi-line comment](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#std:templatetag-comment), I think you can get away with just the [single-line syntax](https://docs.djangoproject.com/en/3.2/topics/templates/#comments): 
```suggestion
                      {# Require at least one non-whitespace character:  #}
```

Vinnl: TIL! That's just what I got after pressing <kbd>Ctrl</kbd>+<kbd>/</kbd> :P

PR #1193, opened 2021-10-12
Add client-side validation of the subdomain name

### LINKED ISSUE #939: White-space can be added as email domain/ "Server error (500)" is displayed when white-space is added in the "Search your new domain" field
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox

**Preconditions:**
The user has already upgraded to Mozilla Privacy Pack.

**Steps to reproduce:**

2. Go to your domain card and fill in white-space in the “Search your new domain” field.
3. Click the “Get it” button.
4. Repeat step 2 and 3 a few times.

**Actual result:**
Sometimes white-space is added as email domain.
Other times a Server error 500 is displayed.

**Expected result:** 
White-space cannot be added as email domain. A warning message is displayed.

white space added as email domain

server error 500 email domain

Vinnl:
Fixes #939.

Although UX-wise, the `pattern` attribute doesn't give great user
feedback in terms of telling them what they're doing wrong, the
status quo is that they'll get a 500 error, so this is somewhat
better.

### DIFF
--- a/privaterelay/templates/includes/banners.html
+++ b/privaterelay/templates/includes/banners.html
@@ -76,6 +76,8 @@
                       name="subdomain"
                       minlength="1"
                       maxlength="63"
+                      {% comment %} Make sure to keep this regex aligned with the one in valid_available_subdomain() in models.py {% endcomment %}
+                      pattern="^(?!-)[A-Za-z0-9-]{1,63}(?<!-)$"
                     >
                     <input class="btn btn--gray" type="submit" value="{% ftlmsg 'banner-choose-subdomain-submit' %}">
                 </form>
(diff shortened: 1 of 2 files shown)


privaterelay/templates/includes/banners.html
```diff
@@ -76,6 +76,8 @@ <h2>{% ftlmsg 'banner-choose-subdomain-headline-aliases' %}</h2>
                       name="subdomain"
                       minlength="1"
                       maxlength="63"
+                      {% comment %} Make sure to keep this regex aligned with the one in valid_available_subdomain() in models.py {% endcomment %}
+                      pattern="^(?!-)[A-Za-z0-9-]{1,63}(?<!-)$"
```
groovecoder, line 80: TIL about [`pattern=`](https://developer.mozilla.org/docs/Web/HTML/Attributes/pattern)

Vinnl:
Agreed, if we can get that error message then that is definitely better (I should've reproduced it myself first - which I did now, and I also get that message). @AOiegas @Bianca-O can you confirm whether #939 still applies?

Issue #1209, opened 2021-10-14
Generating aliases through the Relay Add-on is not working


AOiegas:
#### Build:
- Stage;

#### Affected Platforms:
- Windows 10;

#### Browsers tested on:
- Firefox, latest version;

#### Prerequisites:
- Be signed in into a Firefox Relay account with or without an active subscription;
- Have the latest version of the Relay Add-on installed;

#### Steps to reproduce:
1. Go to https://stage.fxprivaterelay.nonprod.cloudops.mozgcp.net/accounts/profile/ ;
2. In another tab, go to the register page for a site (e.g. YouTube);
3. In the email field of the form, generate a new alias through the context menu and then through the Relay button;
4. Check the Relay dashboard;
5. Observe the results;

#### Expected result:
- The user can generate new aliases through the add-on without issue and they show up in the Relay dashboard;

#### Actual result:
- Aliases do not get generated and do not show up in the Relay dashboard;

#### Notes:
- Happens for both premium and free users;
- The Relay button menus, both the email button and the toolbar button, which have alias counters, do count a new alias after clicking on the “Generate New Alias” button, and they revert back to the actual number of aliases only after refreshing the Relay dashboard;
- Refreshing the register page doesn’t update the Add-on menus;
For a free user:
![addoncreate2](https://user-images.githubusercontent.com/85748693/137296614-086fce44-6f42-4e1b-b853-1d42aa5c35a6.gif)
For a premium user:
![addoncreate](https://user-images.githubusercontent.com/85748693/137296869-30a7e349-35db-4a6b-b5f5-2697abfec2cd.gif)

Vinnl:
![image](https://user-images.githubusercontent.com/4251/137303207-24d2b608-3216-428a-93d3-156579a813e1.png)

I think the issue might be that the API is attempting to prevent CSRF, but in this case it should be callable from outside the Relay website (i.e. from the add-on). It looks like it occurs on stage but not for me locally because stage runs on HTTPS, whereas locally it goes over HTTP. And [supposedly](https://github.com/encode/django-rest-framework/issues/839), the problem is that the request is trying to use session authentication instead of token authentication? @groovecoder and @maxxcrawford worked on that, so they'll probably be able to figure this out in no time :)

AOiegas:
Verified this as working on Stage, Windows 10 with the latest version of the Stage add-on.

PR #1212, opened 2021-10-14
try token auth before session auth

groovecoder:


say-yawn:
Related to: https://github.com/mozilla/fx-private-relay/issues/1209

PR #1214, opened 2021-10-14
Settings Saved Redirect

maxxcrawford:
Depends on https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/26

Add homepage redirect that sets a success message after user saves settings

Preview: 
image

### DIFF
--- a/privaterelay/urls.py
+++ b/privaterelay/urls.py
@@ -38,7 +38,7 @@ urlpatterns = [
     ),
     path('accounts/settings/', views.settings_view, name='settings'),
     # This redirects users back to the homepage after updating settings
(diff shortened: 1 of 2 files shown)


privaterelay/urls.py
```diff
@@ -37,6 +37,8 @@
         name='profile_subdomain'
     ),
     path('accounts/settings/', views.settings_view, name='settings'),
+    # This redirects users back to the homepage after updating settings
+    path('accounts/profile/settings_update', views.settings_update_view, name='settings'),
```
groovecoder, line 41:
```suggestion
    path('accounts/profile/settings_update', views.settings_update_view, name='settings-update'),
```

PR #1218, opened 2021-10-15
Store used subdomains

### LINKED ISSUE #1155: Make sure custom domains don't get recycled if somebody deletes their FxA/Relay account
If I register my subdomain of “pdehaan.mozmail.com” using my LDAP email, and then decide to delete my FxA/Relay account, could somebody else ever get my pdehaan suffix, or is that recorded in some permanant database somewhere so that pdehaan subdomain can never ever be reregistered?

Per @say-yawn:
> Yeah, currently we are not handling the case for when a user deletes their FxA. The profile gets immediately deleted when the user is deleted. A possible fix for this may be storing the subdomain in a separate db altogether or storing the hash of deleted subdomains like we are currently handling aliases on Relay.

Per @groovecoder:
> we should add something to to store a hash of deleted/deactivated (by downgrade) subdomains like we do for aliases. it creates a denial-of-service for others to get the subdomain, but the alternative is that someone else in the future can get a previously-used subdomain and start receiving emails for it. (e.g., account reset/recovery emails)

### LINKED ISSUE #1156: Potential domain squatting if you upgrade to premium, register a custom subdomain, downgrade to free tier
Related to #1155

I see the big red note of “cannot be changed or transferred” but curious how/where you were storing that pref. So I guess I could also get a paid personal Relay account, quickly register “p.mozmail.com” and then cancel my premium subscription and effectively block anybody from ever using my 1 char subdomain in case I ever decide to resume my premium account.

Per @say-yawn:
> Yes, if we implement the above fix for the subdomain (see #1155) this would be the case. I do wonder about how often this abuse will happen and when? We could do a complicated check/logic of seeing if the subdomain is actually used but this may be a risk we accept rather than trying to handle for.

Per @groovecoder:
> we should add something to to store a hash of deleted/deactivated (by downgrade) subdomains like we do for aliases. it creates a denial-of-service for others to get the subdomain, but the alternative is that someone else in the future can get a previously-used subdomain and start receiving emails for it. (e.g., account reset/recovery emails)

I think the squatting is a valid [although unfortunate] case. If somebody downgrades and their custom domain is re-released into general population, that could lead to receiving the previous person's emails that they may have registered elsewhere. So I think I'm pro "denial-of-service".

say-yawn:
# About this PR
When user deletes the account on FxA we also delete all information related to the user including the registered subdomain if they were a premium user. This may result in another user registering a previously registered subdomain resulting in possible abuse like account take over. We want to ensure that subdomain registration is checked against all current and previously registered subdomain.

# Acceptance Criteria
- [ ] Store hashed subdomain in a separate table RegisteredSubdomain
- [ ] When registering new subdomain check if subdomain exists in RegisteredSubdomain, if found fail valid_available_subdomain

#
Close #1155 
Close #1156 
Resolves [Bug 1119](https://mozilla-hub.atlassian.net/browse/MPP-1119)


emails/models.py
```diff
@@ -54,7 +54,9 @@ def valid_available_subdomain(subdomain, *args, **kwargs):
     #   can't have "blocked" words in them
     blocked_word = is_blocklisted(subdomain)
     #   can't be taken by someone else
-    taken = Profile.objects.filter(subdomain=subdomain).count() > 0
+    taken = RegisteredSubdomain.objects.filter(
+        subdomain_hash=hash_subdomain(subdomain)
+    ).count() > 0
```
groovecoder, line 59: We need to make this either/or, since we have existing subdomains that are only in `Profile`.

```suggestion
    registered_before = RegisteredSubdomain.objects.filter(
        subdomain_hash=hash_subdomain(subdomain)
    ).count() > 0
    profile_currently_has = Profile.objects.filter(
        subdomain=subdomain
    ).count() > 0
    taken = registered_before or profile_currently_has
```

say-yawn: Done with migration `emails/migrations/0028_copy_subdomain_to_registeredsubdomain.py`

emails/migrations/0028_copy_subdomain_to_registeredsubdomain.py
```diff
@@ -0,0 +1,22 @@
...
+
+from django.db import migrations, models
+
+from emails.models import hash_subdomain
+
+def copy_already_registered_subdomains_from_profile(apps, schema_editor):
+    Profile = apps.get_model('emails', 'Profile')
+    RegisteredSubdomain = apps.get_model('emails', 'RegisteredSubdomain')
+    for profile in Profile.objects.exclude(subdomain=None):
+        RegisteredSubdomain.objects.create(subdomain_hash=hash_subdomain(profile.subdomain))
```
groovecoder, line 11: Our `0025` migration did `Profile.objects.all()` which triggered out-of-memory error. But looks like this `exclude(subdomain=None)` should make this queryset ONLY get the Profiles of the few users who have gotten a subdomain via fox-fooding. So it should be 🆗

PR #1226, opened 2021-10-18
try/except around GA calls

groovecoder:
https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/11170665/?query=is%3Aunresolved


privaterelay/views.py
```diff
@@ -184,9 +185,17 @@ def metrics_event(request):
...
+        report(
+            settings.GOOGLE_ANALYTICS_ID,
+            request_data.get('ga_uuid'),
+            event_data
+        )
+    except:
+        return JsonResponse(
+            {'msg': 'Unable to report metrics event.'},
+            status=500
+        )
```
say-yawn, line 198: Can we log the error too?

PR #1227, opened 2021-10-18
fix MPP-1030: default_server_storage function

groovecoder:
Update Profile model to use new default_server_storage callable function
to determine its default value. The function returns False before
premium release and True after premium release.


say-yawn (approved):
Lgtm

PR #1228, opened 2021-10-18
fix MPP-1114: autoremove data notification banner

groovecoder:
Builds on the new settings.PREMIUM_RELEASE_DATE env var to dynamically
and automatically remove the data notification banner after the premium
release date.

To spot check:

1. Go to http://127.0.0.1:8000/accounts/profile/
   * You should see the banner (because you are before the default premium release date of 10/27)
2. Change env var to `PREMIUM_RELEASE_DATE="2021-10-17 17:00:00+00:00"`
3. Restart `runserver`
4. Go to http://127.0.0.1:8000/accounts/profile/
   * You should NOT see the banner (because you are now past the release date)

### DIFF
--- a/privaterelay/templates/includes/banners.html
+++ b/privaterelay/templates/includes/banners.html
@@ -21,20 +21,24 @@
     </div>
   </div>
 {% elif show_data_notification_banner %}
-  <div class="c-notification-banner c-data-notification-banner t-warning js-notification">
-    <div class="dismiss-wrapper">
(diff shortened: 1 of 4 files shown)


maxxcrawford (changes requested):
I cannot get this to display. I thought maybe because it was in the `{% if last_bounce_type %}` IF block, so I moved it. No dice. 

Can you include additional steps to reproduce?

PR #1231, opened 2021-10-19
🛑 DNM: Update attachment limit communication

Vinnl:
It's been increased from 150KB to 10MB.


groovecoder:
Closing this so we don't accidentally merge it. We have a blocker on launching 10mb emails next week so we need to wait before we make this change.

PR #1232, opened 2021-10-19
🛑 DNM: Use new string for the email attachment limit

Vinnl:
~Do not merge until this is merged: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/29~

Builds on #1231, so only 5cc08e2 is relevant for this PR.


maxxcrawford:
@Vinnl Could you retarget this PR to merge into #1231? That would make merging this easy to do, and if #1231 gets merged in, this PR automatically re-targets `main`. That would be helpful for these sort of two-part feature lands.

PR #1237, opened 2021-10-19
Add custom domain property on profile

say-yawn:
# About this PR
On the premium dashboard, the user's email dashboard showed only the subdomain instead of the full custom domain. This PR adds a property `custom_domain` on the `Profile` model to show the full custom domain on the premium dashboard.

# STR
Check that full custom domain is displayed on the premium profile dashboard
1. Login to premium Relay user
2. Add subdomain
3. Check that on the premium dashboard, the custom subdomain and the mozmail domain is displayed

# 
Resolves [MPP-1122](https://mozilla-hub.atlassian.net/browse/MPP-1122)

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -180,7 +180,7 @@ class Profile(models.Model):
 
     @property
     def custom_domain(self):
-        return f'{self.subdomain}@{settings.MOZMAIL_DOMAIN}'
+        return f'@{self.subdomain}.{settings.MOZMAIL_DOMAIN}'
 
     @property
     def has_premium(self):


emails/models.py
```diff
@@ -178,6 +178,10 @@ def display_name(self):
         # displayName key will not exist on the extra_data
         return self.fxa.extra_data.get('displayName')
 
+    @property
+    def custom_domain(self):
+        return f'{self.subdomain}@{settings.MOZMAIL_DOMAIN}'
```
groovecoder, line 183:
```suggestion
        return f'@{self.subdomain}.{settings.MOZMAIL_DOMAIN}'
```

PR #1244, opened 2021-10-20
UI refinements for settings

Vinnl:
- Icons are now purple
- More margin below the warning with the yellow border.

Still need to make it fit the full screen, but as [mentioned before](https://github.com/mozilla/fx-private-relay/pull/1180#issuecomment-946560337), I'll need some help with that.


static/images/help-purple.svg
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<svg
```
maxxcrawford, line 6: When following how other SVG terms are added, it's _before_ the `<xml>` tag. You may swap these two lines' order. 

Firefox SVG reference: https://searchfox.org/mozilla-central/source/toolkit/themes/shared/icons/arrow-down-12.svg

static/images/performance-purple.svg
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
```
maxxcrawford, line 4:
```suggestion
<!-- This Source Code Form is subject to the terms of the Mozilla Public
   - License, v. 2.0. If a copy of the MPL was not distributed with this
   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```

PR #1245, opened 2021-10-20
Only show summary stats when there are aliases

Vinnl:
This makes the onboarding less overwhelming.


maxxcrawford:
Can you get this logic/UX approved with Tony and Eduardo? I like this idea.

PR #1307, opened 2021-10-28
UI Bugs - Landing Page 

codemist:
This PR:

- [x] Replaces third image in the How It Works section with the correct image
- [x] Adds 'limited time pricing' to the top of the premium pricing option (now localized: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/41/files)
- [x] Adds subtext to third feature of the premium pricing option
- [x] Change hero section CTA to 'Sign Up'
- [x] Standardize widths of pricing options across desktop sizes
- [x] Standardize text widths across desktop sizes

Demo (1440px / 810px / 375px) :
https://drive.google.com/file/d/1coh5EKB6FKrMYzX3c_kcLvrNWelgSiv5/view?usp=sharing


static/scss/partials/landing/a.scss
```diff
@@ -226,6 +247,13 @@
             border-radius: $border-radius-md;
             text-align: center;
             padding: $spacing-xl;
+            width: 100%;
+
+            @media #{$mq-sm} {
+                justify-items: center;
+                margin-top: $spacing-xl;
+                width: calc((#{$content-xs} + #{$spacing-lg}));
```
codemist, line 255: This keeps the pricing option width standardized across desktop sizes.

privaterelay/templates/newlanding/a/plans.html
```diff
@@ -27,6 +29,7 @@
                 </div>
                 <!--Premium Option-->
                 <div class="c-option c-premium">
+                    <span class="c-premium-detail">Recommended for better protection</span>
```
codemist, line 32: Note: still waiting on localized string

static/scss/partials/landing/a.scss
```diff
@@ -167,7 +174,7 @@
 .c-gray-bg {
     background: linear-gradient(180deg, $color-light-gray-20 0%, $color-light-gray-10 3%);
     height: auto;
-    padding: $layout-md 0;
+    padding-top: $layout-md 0 $layout-xs 0;
```
Vinnl, line 177: There are four values here, but also a change to `padding-top`. If this works now, presumably it should be just:

```suggestion
    padding-top: $layout-md;
```

?

static/scss/partials/landing/a.scss
```diff
@@ -201,20 +205,44 @@
...
+            grid-row-start: 1;
+            margin-top: $layout-sm;
 
             @media #{$mq-md} {
                 grid-row-start: auto;
+                margin-top: 0;
+                justify-self: left;
+            }
+
+            li:nth-child(3) {
```
Vinnl, line 221: Nit, but instead of targeting the third element specifically, I'd add a `has-subtext` class, and rename `c-feature-3-subtext` to `c-feature-subtext`, just because the design doesn't specifically add a subtext to the third item, and if another item later gets a subtext it'll be more obvious for whoever implements that that they can reuse these classes.

static/scss/partials/landing/a.scss
```diff
@@ -201,20 +205,44 @@
...
+                margin: 0;
+                color: $color-grey-40;
+            }
+
+            .c-premium-detail {
+                @include text-body-sm;
+                color: $color-blue-50;
+                font-weight: 700;
+                position: absolute;
+                margin-top: -$layout-lg;
```
Vinnl, line 236: More of a general tip to keep in mind for the future: if you can, try to use `padding` instead of `margin` to avoid affecting the layout of elements next to the component. That will make it easier to reuse the class in other contexts. (_Especially_ when considering the headache that is [margin collapsing](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing).)

I haven't fully internalised it yet myself, but still a good thing to keep in mind. This article is a good read on it: https://non-traditional.dev/encapsulated-css-the-key-to-composable-layouts-94f11c177cc1

Vinnl (approved):
Some small suggestions for you to consider inline, but also wanted to give an overall shoutout that the things you work on always look really tight with more attention to detail than I can usually muster myself, so good job :clap:

maxxcrawford (changes requested):
Blocking for untranslated strings in this PR. See comments below.

codemist:
@maxxcrawford  Yes, I placed that string in this PR: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/41 Not sure if we'd like to add anything more to that (I suggested throwing in the other sections for the landing page, though I'm not sure if we're keeping this a fast follow)

static/scss/partials/landing/a.scss
codemist: Yup, I too am a bigger fan of padding > margin, however for this case specifically I set it to `position: absolute` and intentionally shifted it out of the parent module, hence the negative `margin-top` value.

static/scss/partials/landing/a.scss
codemist: Nice call out! Revised :)

static/scss/partials/landing/a.scss
codemist: Nice call out - this was an error on my part, it's supposed to just be `padding-top` with one value.

codemist:
@Vinnl Thank you Vincent :D Means a lot coming from you! I revised this PR based on your recommendations, appreciate it.

static/scss/partials/landing/a.scss
Vinnl: Ah gotcha, carry on then :)

Vinnl (approved):
@maxxcrawford Think you'll have to unblock now that the strings are merged :)

codemist:
ftlmsg is in!

static/scss/partials/landing/a.scss
```diff
@@ -201,20 +205,44 @@
...
             }
         }
 
         .c-free {
             border: 2px solid $color-light-gray-20;
+
+            @media #{$mq-md} {
+                justify-self: right;
```
maxxcrawford, line 244: Would `end` work here instead of `right`?

maxxcrawford (approved):
Good to go! Thanks!

PR #1314, opened 2021-11-02
Don't submit search form when pressing Enter

Vinnl:
The aliases are filtered as-you-type, so we don't need to submit
the form when the user submits it.


maxxcrawford (approved):
Nice fix!

PR #1315, opened 2021-11-02
Add end onboarding banner logic

codemist:
Test cases:

- **Non-premium user:**
	- No aliases created:
        - On non-fx browser: Show download Firefox banner and upgrade banner at the bottom
        - On fx browser: Show download Relay Add on and upgrade banner at the bottom
    - At least 1 alias created:
        - Same as top two cases but banners show at the top
    - Premium disabled: No upgrade banner

- **Premium user:**
     - No aliases created:
        - On non-fx browser: Show download Firefox banner and upgrade banner at the bottom
        - On fx browser: Show download Relay Add on and upgrade banner at the bottom
    - At least 1 alias created:
        - Same as top two cases but banners show at the top
    - No subdomain: Show subdomain registration banner


privaterelay/templates/profile.html
```diff
@@ -87,8 +87,21 @@
         <span class="dashboard-greeting-message">{% ftlmsg 'profile-label-welcome-html' email=request.user.email %}
       </div>
       {% endif %}
-      {% include "includes/banners.html" %}
 
+      <!-- Data Notification Banner -->
+      {% include "includes/banners/announcement-banner.html" %}
```
codemist, line 92: All 'permanently at the top' announcement banners can be edited here. I placed the data notification module in here.

privaterelay/templates/profile.html
```diff
@@ -87,8 +87,21 @@
...
+      
+      <!-- Register Subdomain Banner -->
+      {% with request.user.profile_set.first as user_profile %}
+      {% if user_profile.has_premium and not user_profile.subdomain and settings.PREMIUM_ENABLED %}
+        {% include "includes/banners/subdomain-banner.html" %}
+      {% endif %}
+      {% endwith %}
```
codemist, line 99: The register subdomain banner is always at the top, so I placed it into its own partial.

privaterelay/templates/profile.html
```diff
@@ -87,8 +87,21 @@
...
+  
+      <!-- Upgrade, Firefox and Add On Banners -->
```
codemist, line 101: These default banners (upgrade to premium/download Firefox/download the add on) are stored in this partial since these banners are consistent for premium/non-premium states (with the exception of upgrade being only for non-premium).

privaterelay/templates/profile.html
```diff
@@ -400,6 +413,10 @@
       </div> <!-- end informtaion footer -->
       
     </div><!-- end dashboard-container -->
+
+    {% if relay_addresses|length == 0 %}
+      {% include "includes/banners/default-banners.html" %}
```
codemist, line 418: The reason I stored those default banners into its own partial is so I can call it here, where it switches to a bottom placement when a user has at least one alias.

codemist:
I broke down `banners.html` into 3 partials stored in a dedicated `/banners` folder within `/includes`. These partials are 

1) an announcement banner for custom banners such as the data notification which would always be at the top, 
2) the default banners with upgrade prompts for premium/Firefox/add on
3) subdomain banner for premium users without one set

maxxcrawford (approved):

maxxcrawford:
@codemist Great work! This covers so many items. 💯

PR #1316, opened 2021-11-02
Fix #1301 - Create common "alias-card" template for relay and domain addresses

### LINKED ISSUE #1301: DRY: Profile "Alias" Markup duplicates code
The profile has duplicated markup to generated each type of alias: 

- Generated aliases: [/templates/profile.html#L101-L226](https://github.com/mozilla/fx-private-relay/blob/main/privaterelay/templates/profile.html#L101-L226)
- Custom aliases [/templates/profile.html#L228-L356](https://github.com/mozilla/fx-private-relay/blob/main/privaterelay/templates/profile.html#L228-L356)

Noted from this comment: https://github.com/mozilla/fx-private-relay/pull/1298#issue-766499302

lennartschoch:
This should solve #1301 by creating a common template for both relay addresses and domain addresses.

Also, one quick question: What configuration are you using for formatting HTML templates? Is there any linting tool I can use? I found an ESLint config for JS and Stylelint for CSS but wasn't sure about HTML.


Vinnl (approved):
Yes! Thanks for this @lennartschoch - it's going to save us quite a few bugs! (In fact, it's already solving one - in reviewing this I already some one change that I made to the Relay addresses earlier but forgot to apply to the domain aliases.)

Unfortunately we don't have any formatting tools for the HTML/Django templates at the moment, so indeed it's just looking at how things are done elsewhere and manually replicating.

PR #1317, opened 2021-11-02
Fix #1238 - Unify banner logic

### LINKED ISSUE #1238: Unify Banner Cookie Logic
The following functions are all clones of each other. We should migrate this to a single function with some parameters.

Params: 
- Cookie ID Name
- Banner Target ID/Class
- Dismiss Target Class 

https://github.com/mozilla/fx-private-relay/blob/b443409dcd9b3423643ee507c2d8d40d01dcfc59/static/js/app.js#L480
https://github.com/mozilla/fx-private-relay/blob/b443409dcd9b3423643ee507c2d8d40d01dcfc59/static/js/app.js#L431
https://github.com/mozilla/fx-private-relay/blob/b1e954a9d0d9a8afded6d64d428047c2c5d27a66/static/js/app.js#L529

lennartschoch:
This solves #1238 by creating a single function called `bannerLogic` that unifies the logic of the three banners that set a cookie when dismissed.
The VPN banner is doing some extra things (e.g. calling Google Analytics), so (additionally to what @maxxcrawford mentioned in the issue) I added callbacks for the banner functions which should allow for a bit more flexibility. Also renamed the VPN banner's `closed` class to `is-hidden` to be consistent with the other banners.

### DIFF
--- a/privaterelay/templates/includes/banners.html
+++ b/privaterelay/templates/includes/banners.html
@@ -26,7 +26,7 @@
       {% if not user_profile.server_storage %}
         <div id="sync-labels" class="c-notification-banner c-data-notification-banner t-warning is-hidden">
           <div class="dismiss-wrapper">
-            <button class="notification-dismiss js-data-collection-dismiss"><img src="{% static 'images/x-close.svg' %}"></button>
+            <button id= "data-collection-dismiss" class="notification-dismiss"><img src="{% static 'images/x-close.svg' %}"></button>
           </div>
           <div class="notification-banner-bg">
             <div class="notification-banner-content">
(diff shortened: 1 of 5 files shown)


maxxcrawford:
@lennartschoch Thank you for submitting this. We've had some pretty substantial changes to our code base since you first submitted this. Could you catch it up to `main` and I can take another look? Thanks!

maxxcrawford (changes requested):
Firstly — thank you so much this contribution. We've had some pretty substantial changes to our codebase since you first submitted this. Could you catch this up to `main` and I can take another look. 🙌

groovecoder:
Closing as out-of-date.

PR #1319, opened 2021-11-03
better and tighter _reply_allowed() logic

groovecoder:
This refactors code out of `_handle_reply` and into a new `_reply_allowed()` function.

The new `_reply_allowed()` function now checks both conditions:

* If the reply is coming *from* a Relay user to external recipient, make sure the Relay user is premium.
* If the reply is coming from an external sender *to* a Relay recipient, make sure the Relay user is premium.

This closes a loop-hole where a free user could send replies with a `From:` address different from their FXA address. Relay would send the reply because Relay thought the reply was coming from an external sender.

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -298,7 +298,7 @@ def _sns_message(message_json):
         user_profile = address.user.profile_set.first()
     except (ObjectDoesNotExist):
         if to_local_portion == 'replies':
-            return _handle_reply(from_address, message_json)
+            return _handle_reply(from_address, message_json, to_address)
 
         return HttpResponse("Address does not exist", status=404)
 
@@ -404,16 +404,10 @@ def _strip_localpart_tag(address):
     return f'{subaddress_parts[0]}@{domain}'
 
 
-def _handle_reply(from_address, message_json):
+def _reply_allowed(from_address, to_address, reply_record):
     stripped_from_address = _strip_localpart_tag(from_address)
-
-    mail = message_json['mail']
-    (lookup_key, encryption_key) = _get_keys_from_headers(mail['headers'])
-    reply_record = _get_reply_record_from_lookup_key(lookup_key)
-    address = reply_record.address
-    reply_record_email = address.user.email
+    reply_record_email = reply_record.address.user.email
     stripped_reply_record_address = _strip_localpart_tag(reply_record_email)
-
     if (
         (from_address == reply_record_email) or
         (stripped_from_address == stripped_reply_record_address)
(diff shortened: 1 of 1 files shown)


emails/views.py
```diff
@@ -424,9 +418,34 @@ def _handle_reply(from_address, message_json):
...
+    return True
+
+
+def _handle_reply(from_address, message_json, to_address):
+    mail = message_json['mail']
+    (lookup_key, encryption_key) = _get_keys_from_headers(mail['headers'])
+    reply_record = _get_reply_record_from_lookup_key(lookup_key)
+    address = reply_record.address
+
+    if not _reply_allowed(from_address, to_address, reply_record):
```
say-yawn, line 445: We should verify the in-reply-to field instead of the `to_address` since the `to_address` will ONLY be`replies` on the local portion [here](https://github.com/mozilla/fx-private-relay/blob/main/emails/views.py#L300-L301).

emails/views.py
```diff
@@ -424,9 +418,34 @@ def _handle_reply(from_address, message_json):
...
+        # premium Relay user
+        try:
+            [to_local_portion, to_domain_portion] = to_address.split('@')
+            address = _get_address(
+                to_address, to_local_portion, to_domain_portion
             )
+            user_profile = address.user.profile_set.first()
+            if not user_profile.has_premium:
+                incr_if_enabled('free_user_reply_attempt', 1)
+                return False
```
say-yawn, line 433: Should be check the in reply to address being a premium relay user that the external is sending to and NOT the `to_address` which will have `replies` portion according to the check [here](https://github.com/mozilla/fx-private-relay/blob/main/emails/views.py#L300-L301).

If we want to disable external messages that are responses to the premium reply feature, we should be checking the in reply to while checking the local portion at inbound email checks [here](https://github.com/mozilla/fx-private-relay/blob/main/emails/views.py#L300-L301)

say-yawn (changes requested):
Changes requested on checking that the in-reply-to is a premium user happens at a different place and is checking the `in-reply-to` header instead of the `to` header.

groovecoder:
`to_address` is `replies@` when our user replies to an external sender. But, when an external sender replies to our user, to `to_address` is their relay address. (Should it be? Not sure. I tried to change the code the make the `Reply-To` address *ALWAYS* be our `replies@` address, but after 15m it felt like too much refactoring for this PR.)

So, I added another `_reply_allowed()` check for all emails sent to our aliases with an `In-Reply-To` field.

I also inverted the `_reply_allowed()` so only those 2 specific conditions will `return True` and the default is to `return False`

say-yawn (approved):
LGTM!

PR #1320, opened 2021-11-03
Consider domain aliases where relevant

### LINKED ISSUE #1264: "Create your first alias..." text and section could be improved when no aliases were created before but a custom alias is created instead
**Build:**
Stage

**Affected platforms:**
Windows 10 64 bit

**Browsers tested on:**
Firefox

**Preconditions:**
The user has an active subscription and email domain registered.
No email aliases are added.

**Steps to reproduce:**

1. Navigate to Stage server and log in with an existing account.
2. Go to your inbox and send an email to hey@bi-anca.mozmail.fxprivaterelay.nonprod.cloudops.mozgcp.net.
3. Go to Relay page and observe the created custom alias.
4. Observe the section with "Create your first alias...".

**Actual result:**
The first alias is already created by email so the "Create your first alias..." text is not adequate anymore.

**Expected result:**
The text is changed to a more appropriate one in this case. A fourth way to create an alias could be added (custom alias created by email).

premium user - create first alias section

Vinnl:
Fixes #1264.

Two elements were dependent on aliases being created or not, but
they only checked for Relay addresses, not domain addresses.

### DIFF
--- a/privaterelay/templates/profile.html
+++ b/privaterelay/templates/profile.html
@@ -90,11 +90,11 @@
       {% include "includes/banners.html" %}
 
 
-      {% if relay_addresses %}      
+      {% if relay_addresses or domain_addresses %}
         {% include "includes/dashboard-filter.html" %}
       {% endif %}
 
-      {% if relay_addresses|length == 0 %}
+      {% if relay_addresses|length == 0 and domain_addresses|length == 0 %}
         {% include "includes/dashboard_onboarding.html" %}
       {% endif %}
 


privaterelay/templates/profile.html
```diff
@@ -90,12 +90,14 @@
...
 
 
-      {% if relay_addresses %}      
+      {% if relay_addresses or domain_addresses %}
         {% include "includes/dashboard-filter.html" %}
       {% endif %}
 
       {% if relay_addresses|length == 0 %}
-        {% include "includes/dashboard_onboarding.html" %}
+        {% if domain_addresses|length == 0 %}
```
Vinnl, line 98: If there's a way to do conjunctions in combination with `| length`, let me know. (Yes, I miss JSX.)

maxxcrawford: Oh! I think we can? https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#boolean-operators

image

maxxcrawford (approved):
Approved! See my comment below. We may be able to add a `AND`/`OR` here, safely.

privaterelay/templates/profile.html
Vinnl: Heh, for some reason I was under the impression that that would have to be done _before_ passing it to ` | length`, and was so convinced that I didn't even try it :P

PR #1322, opened 2021-11-04
get_delayed_emails_from_sqs command

groovecoder:
to get failed SNS messages from our SQS dead-letter queue, and
re-process them thru the same _sns_inbound_logic function.


emails/management/commands/get_delayed_emails_from_sqs.py
```diff
@@ -0,0 +1,45 @@
...
+        messages = dl_queue.receive_messages(
+            MaxNumberOfMessages=10, WaitTimeSeconds=1
+        )
+        while len(messages) > 0:
+            for message in messages:
+                json_body = json.loads(message.body)
+                verified_json_body = verify_from_sns(json_body)
+                topic_arn = verified_json_body['TopicArn']
+                message_type = verified_json_body['Type']
+                validate_sns_header(topic_arn, message_type)
```
moz-astults, line 38: @groovecoder Do we need this since it's not coming from SNS? I would assume SQS payloads fail this check, but there's no HTTP response to give?

moz-astults: Oh actually looking at an example payload it keeps all of the SNS data so this is probably good validation. Disregard :slightly_smiling_face:

emails/apps.py
```diff
@@ -20,6 +20,8 @@ def __init__(self, app_name, app_module):
                 'ses', region_name=settings.AWS_REGION
             )
             self.s3_client = boto3.client('s3', region_name=settings.AWS_REGION)
+            self.sqs_client = boto3.resource('sqs', region_name=settings.AWS_REGION)
+            self.dl_queue = self.sqs_client.Queue(settings.AWS_SQS_QUEUE_URL)
```
moz-astults, line 24: @groovecoder do you know if an SQS auth attempt is done here? If so maybe we should have a separate try/except section? Otherwise I think it would always log as `exception during SES connect`

emails/management/commands/get_delayed_emails_from_sqs.py
```diff
@@ -0,0 +1,45 @@
...
+                _sns_inbound_logic(topic_arn, message_type, verified_json_body)
```
say-yawn, line 39: We should catch the response from `_sns_inbound_logic` and check the 200 before we delete.

privaterelay/settings.py
```diff
@@ -105,6 +105,7 @@
 AWS_SNS_TOPIC = config('AWS_SNS_TOPIC', None)
 AWS_SNS_KEY_CACHE = config('AWS_SNS_KEY_CACHE', 'default')
 AWS_SES_CONFIGSET = config('AWS_SES_CONFIGSET', None)
+AWS_SQS_QUEUE_URL = config('AWS_SQS_QUEUE_URL', None)
```
say-yawn, line 108: Same for this, since this is only being used in the command should we consider getting this inside the command script?

emails/management/commands/get_delayed_emails_from_sqs.py
```diff
@@ -0,0 +1,45 @@
...
+                logger.info(f'processed sqs message ID: {message.message_id}')
+                message.delete()
+            messages = dl_queue.receive_messages(
+                MaxNumberOfMessages=10, WaitTimeSeconds=1
+            )
```
say-yawn, line 44: why only 10 at a time with 1 sec of wait?

say-yawn (changes requested):
Some questions/recs and a change requested. Would also recommend we add metrics on how many emails we processed with SQS? Or if this information is already available or not important to collect, ignore.

emails/management/commands/get_delayed_emails_from_sqs.py
moz-astults, line 39: I'm curious about what happens if something fails here. It seems like there should be a `try/except` statement around this, but going through the functions it looks like we would get a log and have an `HttpResponse` returned. If that happened, wouldn't this just log that the message was processed and delete it?

The main scenario I'm thinking of is around rate limiting/throttling responses from SES. If we have a backlog of messages in this queue and are sending them out as fast as possible in this `while` loop, I'm not sure how we would handle a throttling response, or if in this current state would just result in us logging the error and deleting the message from the queue.

emails/management/commands/get_delayed_emails_from_sqs.py
groovecoder: Yeah, the SNS-receiving code checks these via HTTP request headers that SNS sends. I didn't want to remove these checks completely, and the same values are also sent in the JSON body, so I just grabbed them out of the JSON body and run the same validation code on them.

emails/management/commands/get_delayed_emails_from_sqs.py
groovecoder: This queue should have 0 messages in it after we recover from down-time and run this script. 

If we only delete on 200's, we'll leave messages in SQS where we would otherwise have responded 4xx to SNS and then dropped the email data completely. 

Scenarios where we respond with 4xx:
* SNS message type we don't support
* messages without `commonHeaders`
* address does not exist
* replies require premium account
* cannot find message contents from s3 (future)

No matter what happens in `_sns_inbound_logic`, we should delete the message from SQS so we're still operating with the same "we don't keep any of the email data" architecture we have with SNS.

emails/apps.py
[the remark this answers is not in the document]
groovecoder: Good call! No need to have an SQS client connection open unless we need it.

emails/management/commands/get_delayed_emails_from_sqs.py
moz-astults: fwiw it looks like 10 is the max number allowed here, and 1 second poll seems reasonable because in our use-case there will either be something in the queue waiting or there won't be

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Client.receive_message

groovecoder: 10 is an [SQS-enforced max](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Client.receive_message). 😢

emails/management/commands/get_delayed_emails_from_sqs.py
groovecoder: Note: I did update the code to NOT delete the message from SQS, but only in the case of an `AccountSendingPausedException` as [@moz-astults pointed out a specific likely rate-limiting scenario](https://github.com/mozilla/fx-private-relay/pull/1322#discussion_r742958158) that would cause an error different from the regular errors that would otherwise drop the email data.

emails/apps.py
groovecoder: Moved this into the command itself, as @say-yawn suggested. It will now log as `sqs_client_error`.

emails/management/commands/get_delayed_emails_from_sqs.py
```diff
@@ -0,0 +1,57 @@
...
+    verified_json_body = verify_from_sns(json_body)
+    topic_arn = verified_json_body['TopicArn']
+    message_type = verified_json_body['Type']
+    validate_sns_header(topic_arn, message_type)
+    try:
+        _sns_inbound_logic(topic_arn, message_type, verified_json_body)
+        logger.info(f'processed sqs message ID: {message.message_id}')
+        message.delete()
+    except ClientError as e:
+        if e.response['Error']['Code'] == 'AccountSendingPausedException':
```
moz-astults, line 30: @groovecoder I don't think this is quite the right error code. I think we would be looking for an error code with the word `Throttling` in it: https://aws.amazon.com/blogs/messaging-and-targeting/how-to-handle-a-throttling-maximum-sending-rate-exceeded-error/

If we catch this particular error, we can probably just sleep for 1 second and try again.

privaterelay/settings.py
groovecoder: Hmm ... I like having ALL the app configs in a single module - e.g., `settings`. There are a number of other app configs that are only used in a single module but we still instantiate them in `settings`.

emails/management/commands/get_delayed_emails_from_sqs.py
groovecoder: Ah ... I updated the code to check for either `throttling` or `pause` in the error code to try to catch either case where the error might be temporary. Hmm ... sleeping and trying again would be cool, but I'm feeling a recursive function call coming on. I'll give it a shot though.

groovecoder: Okay, I gave it a `time.sleep(1)` and a 2nd try before it logs the error.

groovecoder:
Spot-checked this on dev and it worked!! I also added more logging and metrics. Ready for final r?

say-yawn (approved):
LGTM

PR #1326, opened 2021-11-04
Fix #956 - Shift main wrapper down when mobile menu is toggled

### LINKED ISSUE #956: Not all info is visible when "Menu" is expanded on mobile/ Scroll issue
**Build:**
Stage

**Affected platforms:**
Android 11
iOS 14.6

**Browsers tested on:**
Firefox

**Steps to reproduce:**

2. Expand the "Menu".
3. Observe the info on the page.
4. Scroll down the page.

**Actual result:**
Not all information is visible when "Menu" is expanded. When you scroll down the page the info is visible but it is not fixed.

**Expected result:**
All information is visible when "Menu" is expanded or it is visible and fixed when scrolling down the page.

https://user-images.githubusercontent.com/79697848/126962269-1bb1742a-11ae-49b0-a44b-fd52ef56d972.mp4

codemist:
This PR fixes #956, and pushes the main wrapper of the site down when the mobile menu is toggled.

Demo:

https://user-images.githubusercontent.com/13066134/148368494-48f9dbea-d530-41c6-b64e-35396816981a.mov

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -235,17 +235,14 @@ function addEventListeners() {
     });
   }
 
-  function readExpandedHeight() {
-    const mobileMenuWrapperLinks = document.querySelector(".mobile-menu-links");
-    const mobileMenuWrapperLinksHeight = mobileMenuWrapperLinks.offsetHeight;
-    return mobileMenuWrapperLinksHeight;
-}
 
-function moveDownMainContainer(setMenuLinksHeightVal, mainWrapper) {
+function moveDownMainContainer(setMenuLinksHeightVal) {
+
+  const mainWrapper = document.querySelector("main");
   
-  const moveDownBy = parseInt(setMenuLinksHeightVal) + 70;
-  console.log(moveDownBy);
+  const moveDownBy = parseInt(setMenuLinksHeightVal);
 
+  console.log(moveDownBy);
 
   if (mainWrapper.style.marginTop == 0 || mainWrapper.style.marginTop == "0px" ) {
     mainWrapper.style.marginTop = moveDownBy + "px";
@@ -253,20 +250,25 @@ function moveDownMainContainer(setMenuLinksHeightVal, mainWrapper) {
   else {
     mainWrapper.style.marginTop = 0;
   }
+
 }
-
   const mobileMenuWrapper = document.querySelector(".mobile-menu");
-  if (mobileMenuWrapper) {
-    const mobileMenuButton = document.querySelector(".mobile-menu-toggle");
 
-    const mainWrapper = document.querySelector("main");
+  if (mobileMenuWrapper) {
+
+    const mobileMenuWrapperLinks = document.querySelector(".mobile-menu-links");
+
+    const mobileMenuButton = document.querySelector(".mobile-menu-toggle");
 
     mobileMenuButton.addEventListener("click", () => {
       mobileMenuWrapper.classList.toggle("menu-open");
-      const setMenuLinksHeightVal = setTimeout(readExpandedHeight, 1000);
-      setTimeout(moveDownMainContainer, 1200, setMenuLinksHeightVal, mainWrapper);
-     });
 
+      setTimeout(() => {
+        const mobileMenuWrapperLinksHeight = mobileMenuWrapperLinks.offsetHeight;
+        moveDownMainContainer(mobileMenuWrapperLinksHeight);
+      }, 100);
+
+     });
 
   }
 
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -238,9 +238,21 @@ function addEventListeners() {
   const mobileMenuWrapper = document.querySelector(".mobile-menu");
   if (mobileMenuWrapper) {
     const mobileMenuButton = document.querySelector(".mobile-menu-toggle");
+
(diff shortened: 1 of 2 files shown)


maxxcrawford (commented):
Before even really reviewing – is there a way to do this without settings hardcoded numbers via JS? Can this be class toggled?

codemist:
@maxxcrawford I initially wanted to add a class to the main container like `.add("shift-down")` but the issue was the menu heights varied (its longer at pre-sign in with the added 'Sign Up' button), so I opted for `.offsetHeight` instead

static/scss/partials/main.scss
```diff
@@ -323,9 +323,9 @@ button:active {
     max-width: $content-sm;
 }
 
-main,
-table {
+main, table {
     width: 100%;
+    transition: all .4s ease;
```
Vinnl, line 328: Not too much of an issue, but [MDN recommends](https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property):

> The set of properties that can be animated is subject to change. As such, you should avoid including any properties in the list that don't currently animate, as someday they might, causing unexpected results.

Vinnl (approved):
I think I'd just have done away with the overlapping behaviour altogether, but this works :P 

The margin below the menu is quite a bit smaller when expanded than when collapsed though, so it might be worth considering enlarging that a bit.

Also, the "signed in" message is still overlapped, though that's a bit of an edge case so probably fine to ignore?

https://user-images.githubusercontent.com/4251/140924098-5fa7a45f-ba61-4fbb-af0c-589f26d8b04c.mp4

and more in Chrome:

https://user-images.githubusercontent.com/4251/140924119-4d7e1ef5-d978-4b63-9b84-f6e60efb49f5.mp4

I'm also wondering to what extent reflowing the entire page, in an animation, affects performance, but unfortunately I didn't succeed in opening my local copy on my phone, so I'm just hoping it's good enough :)

static/js/app.js
```diff
@@ -238,9 +238,21 @@ function addEventListeners() {
...
     const mobileMenuButton = document.querySelector(".mobile-menu-toggle");
+
+    const mainWrapper = document.querySelector("main");
+    const mobileMenuWrapperHeight = mobileMenuWrapper.offsetHeight + 24;
+    
     mobileMenuButton.addEventListener("click", () => {
       mobileMenuWrapper.classList.toggle("menu-open");
-    });
+
+      if (mainWrapper.style.marginTop == 0 || mainWrapper.style.marginTop == "0px" ) {
```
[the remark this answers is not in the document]
codemist: Good catch

static/js/app.js
```diff
@@ -238,9 +238,21 @@ function addEventListeners() {
   const mobileMenuWrapper = document.querySelector(".mobile-menu");
   if (mobileMenuWrapper) {
     const mobileMenuButton = document.querySelector(".mobile-menu-toggle");
+
+    const mainWrapper = document.querySelector("main");
+    const mobileMenuWrapperHeight = mobileMenuWrapper.offsetHeight + 24;
```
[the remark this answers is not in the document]
codemist: Magic number is gone now (a while later, haha)

static/js/app.js
```diff
@@ -235,12 +235,35 @@ function addEventListeners() {
     });
   }
 
+
+function moveDownMainContainer(setMenuLinksHeightVal) {
+  const mainWrapper = document.querySelector("main");
+  const moveDownBy = parseInt(setMenuLinksHeightVal);
```
Vinnl, line 241: I don't think it matters here, but it's generally good practice to always specify the radix (usually 10) when using `parseInt`, because otherwise it will use a radix of 16 when the given string happens to start with `0x`, so that's a good habit to get into: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

```suggestion
  const moveDownBy = parseInt(setMenuLinksHeightVal, 10);
```

Vinnl (approved):

static/scss/partials/main.scss
Vinnl: This still applies :)

static/js/app.js
codemist: Done!

codemist: Never heard of that before, good to know.

codemist:
@Vinnl  Is it okay to submit it even with the submodule updates?

static/scss/partials/main.scss
```diff
@@ -294,9 +294,39 @@ button:active {
     margin: auto auto 0 0;
 }
 
-main,
-table {
+.dashboard-mpp-promo {
```
codemist, line 297: @maxxcrawford @Vinnl I tried to rebase it main and found some merge conflicts and accepted incoming changes - do these `.dashboard-mpp-promo` classes need to be here, or can they be dropped?

Vinnl: I don't see them in [current `main`](https://github.com/mozilla/fx-private-relay/blob/main/static/scss/partials/main.scss), so I think they can be dropped?

As for the submodule updates: I think so? I think it means you've committed to `/privaterelay/locales` yourself, which I think probably shouldn't happen, but will also be automatically overwritten by the next update. If so, it shouldn't be a problem.

PR #1327, opened 2021-11-04
set in_reply_to None by default

groovecoder:


groovecoder:
Merging this to un-break stage, but would still like a review on it.

PR #1334, opened 2021-11-08
Use updated FAQs

Vinnl:
l10n dependency: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/44

### DIFF
--- a/privaterelay/templates/faq.html
+++ b/privaterelay/templates/faq.html
@@ -54,7 +54,7 @@
           {% else %}
             <section class="faq" id="faq-what-is">
               <h2 class="faq-headline">{% ftlmsg 'faq-question-what-is-question' %}</h2>
-              <p class="faq-answer js-set-href">
+              <p class="faq-answer">
                 {% ftlmsg 'faq-question-what-is-answer' %}
               </p>
             </section>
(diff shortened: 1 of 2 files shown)


privaterelay/templates/faq.html
```diff
@@ -136,6 +137,10 @@ <h2 class="faq-headline">{% ftlmsg 'faq-question-unsubscribe-domain-question' %}
               <h2 class="faq-headline">{% ftlmsg 'faq-question-8-question' %}</h2>
               <p class="faq-answer js-set-href">{% ftlmsg 'faq-question-8-answer-html' url='https://www.mozilla.org/privacy/firefox-relay/' attrs='class="text-link" data-url="https://www.mozilla.org/privacy/firefox-relay/"' %}</p>
             </section>
+            <section class="faq" id="faq-data-collection">
```
maxxcrawford, line 140: Duplicated ID from preceding question 

```suggestion
            <section class="faq" id="faq-privacy-notice">
```

privaterelay/templates/faq.html
```diff
@@ -136,6 +137,10 @@ <h2 class="faq-headline">{% ftlmsg 'faq-question-unsubscribe-domain-question' %}
...
+              <h2 class="faq-headline">{% ftlmsg 'faq-question-email-storage-question' %}</h2>
```
maxxcrawford, line 141: This question is only visible if premium is enabled. It needs to be in the above block too.

maxxcrawford (changes requested):

Vinnl (commented):
Geez rookie mistakes. I can't approve my own PR, but assuming the new question should indeed also be shown before Premium is enabled, your added commits look good to me :+1:

codemist (commented):
Looks good, though I have to ask if 'When should I use Relay aliases' qn should disappear when `PREMIUM_ENABLED` is set to `False`, but remains visible on the landing page?

codemist (approved):

maxxcrawford (approved):
🤝

PR #1335, opened 2021-11-08
Replace icons by Eduardo's own exported SVGs

Vinnl:
Somehow the ones I used were much more inconsistent in terms of
boldness. I think I might have used Nebula icons instead? Anyway, it looks much more consistent now:

image

versus what it used to look like:

image


maxxcrawford:
Discussed with design. This is good to go!

PR #1339, opened 2021-11-08
Fix #1330 - Change contact tool tip to match menu

### LINKED ISSUE #1330: Small difference for "Contact us" tooltip
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit
macOS 10.14

**Browsers tested on:**
Firefox

**Preconditions:**

**Steps to reproduce:**

1. Navigate to Stage server and log in with an existing account.
2. Click the "Profile" icon and hover over the "Contact us" menu item.
3. Observe the tooltip.
4. Go to "Settings" and hover over the "Contact us" link.
5. Observe the tooltip.

**Actual result:**
There is a small difference between the 2 tooltips.

**Expected result:**
The tooltip is the same.

tooltip diff

codemist:
@Vinnl Just checking here that the 'Contact Us' on the settings page is only available for Premium users?

### DIFF
--- a/privaterelay/templates/settings.html
+++ b/privaterelay/templates/settings.html
@@ -22,7 +22,7 @@
           href="{{ settings.FXA_SUPPORT_URL }}?utm_source={{ settings.SITE_ORIGIN }}"
           target="_blank"
           rel="noopener noreferrer"
-          title="{% ftlmsg 'settings-meta-contact-tooltip' %}"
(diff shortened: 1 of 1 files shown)


Vinnl (approved):
That is correct! The link is to a page in your Firefox account section only available to people with a subscription.

Maybe also remove the original string from the l10n project after this is merged, so translators that haven't translated this yet won't waste their time doing so?

Issue #1337, opened 2021-11-08
The onboarding section is confusing regarding Subdomains containing uppercase characters


AOiegas:
#### Build:
- Stage;

#### Affected Platforms:
- Across platforms;

#### Browsers tested on:
- Firefox, latest version;

#### Prerequisites:
- Be signed in into a Firefox Relay account with an active subscription;
- Have not passed through the new onboarding;
- Have not set a Subdomain yet;

#### Steps to reproduce:
1. Go to the Relay Stage site;
2. Click the "Get Started" button on the new onboarding;
3. At step 2, type a Subdomain name that contains at least one uppercase letter;
4. Go through the Subdomain name confirmation flow;
5. Observe how the Subdomain is displayed;
6. Confirm and register the Subdomain;
7. Go to the Relay dashboard and observe your Subdomain;

#### Expected result:
- The onboarding section should inform the user that the Subdomain name will default to lowercase letters OR at least to display the Subdomain name chosen properly;

#### Actual result:
- The user is not informed of the case and the Subdomain name is displayed with Uppercase letters throughout the confirmation flow of the onboarding;

#### Notes:
- This happens when trying to register a Subdomain after the onboarding as well;
- The Subdomain defaults to lowercase letters only as intended when getting to the Dashboard;
![uppercase1 1](https://user-images.githubusercontent.com/85748693/140776351-ebb15171-f8b3-48ef-95f8-292a0591735f.png)
![uppercase1 2](https://user-images.githubusercontent.com/85748693/140776376-e9bb3571-a8f7-41c3-b290-4401d7216902.png)
![uppercase1 3](https://user-images.githubusercontent.com/85748693/140776410-4d21a60b-eb74-4c2b-878a-afb914750261.png)
![uppercase1 4](https://user-images.githubusercontent.com/85748693/140776435-043b5a34-bbd8-4aee-96d9-5554cc06e326.png)

![image](https://user-images.githubusercontent.com/85748693/140776554-0fc6b199-22a7-4ae6-ae99-95b8274fd2aa.png)

AOiegas:
Verified this as fixed on Stage, Windows 10. The Subdomain registration flow from the second step in the onboarding makes it visibly clear that the Subdomain name defaults any uppercase letter to lowercase.
![subdomain verify](https://user-images.githubusercontent.com/85748693/141141463-1c15d589-e9f4-4842-bb33-5e8c8d07ba90.gif)

Issue #1332, opened 2021-11-08
Wrong message displayed when adding a domain that is not allowed


Bianca-O:
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit
macOS 10.14

**Browsers tested on:**
Firefox

**Preconditions:**
The user has an active subscription but no email domain added.

**Steps to reproduce:**

1. Navigate to Stage server and log in with an existing account.
2. Click the "Get Started" button on the new onboarding.
3. At step 2 add white space as a domain or a word that is on the badwords list.
4. Observe the warning message displayed.

**Actual result:**
The last message displayed shows up as a warning message. (If "...successfully signed in" message was last displayed then this is displayed as the warning message too)

**Expected result:**
The correct warning message is displayed.

![wrong message displayed1](https://user-images.githubusercontent.com/79697848/140710003-34d3e85f-fa9f-4fd6-a3af-f4d6a7fcdc5a.gif)

AOiegas:
This also happens when trying to use not allowed characters, like most special characters.

codemist:
@Bianca-O I can't seem to reproduce this error: 
<img width="1617" alt="image" src="https://user-images.githubusercontent.com/13066134/140816978-1445ee39-313c-4b4d-95be-4eef76022f0f.png">

maxxcrawford:
@codemist It only occurs when you login and have a green "logged in" success message visible. 

I have a fix for this!

PR #1345, opened 2021-11-10
Multi-part Premium Onboarding UI Bug Fixes

maxxcrawford:
This PR solves two issues flagged by QA yesterday: 

- [#1332](https://github.com/mozilla/fx-private-relay/issues/1332) / [MPP-1335](https://mozilla-hub.atlassian.net/browse/MPP-1335)
- [#1337](https://github.com/mozilla/fx-private-relay/issues/1337) / [MPP-1339](https://mozilla-hub.atlassian.net/browse/MPP-1339)

Testing: 

## Test 1: Double-notifications Fix

_Pre-requisites for both tests below:_
- Premium account
- Onboarding is visible
- NO Subdomain set

- Log into Relay
- **Expected:** Green "Logged in" message should be visible
- Go to Step 2 of Onboarding (Domain Registration) 
- Try to register a previously registered domain (test, etc) OR a bad word (butt) 
- **Expected:** You should only get ONE message, with the correct error message ("The domain @### is not available…"). 
- Extra Steps: You should be able to dismiss the message, search for another error-causing domain, and see the new erro message. 

## Test 2: Double-notifications Fix
- Log into Relay
- Go to Step 2 of Onboarding (Domain Registration) 
- Try to register an available domain with mixed/all capitals (MoZiLlAaa,  MyDomain, MAILMAILMAIL, etc.)
- **Expected:** The search input value is converted to lowercase on submit, the preview modals are shown in lowercase.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,7 +26,7 @@ module.exports = {
     },
   ],
   parserOptions: {
-    ecmaVersion: 2018,
+    ecmaVersion: 2020,
   },
   root: true,
   rules: {
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -2,7 +2,7 @@
 
 function dismissNotification() {
 	const notification = document.querySelector(".js-notification");
-	notification.classList.toggle("is-hidden");
+	notification.classList.add("is-hidden");
 }
 
 if (typeof(sendGaPing) === "undefined") {
--- a/privaterelay/templates/includes/messages-frontend.html
+++ b/privaterelay/templates/includes/messages-frontend.html
@@ -1,6 +1,6 @@
 {% load ftl %}
 {% load relay_tags %}
-<div class="js-notification messages is-hidden">
+<div class="js-notification-html messages is-hidden">
 	<fluent 
 		data-error-subdomain-not-available="{% ftlmsg 'error-subdomain-not-available' unavailable_subdomain='REPLACE'  %}"
 		data-success-subdomain-registered="{% ftlmsg 'success-subdomain-registered' unavailable_subdomain='REPLACE'  %}"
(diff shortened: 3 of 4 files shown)


.eslintrc.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
     },
   ],
   parserOptions: {
-    ecmaVersion: 2018,
+    ecmaVersion: 2020,
```
maxxcrawford, line 29: https://eslint.org/blog/2020/07/eslint-v7.5.0-released#optional-chaining-support

static/js/app.js
```diff
@@ -2,7 +2,7 @@
 
 function dismissNotification() {
 	const notification = document.querySelector(".js-notification");
-	notification.classList.toggle("is-hidden");
+	notification.classList.add("is-hidden");
```
maxxcrawford, line 5: The usage of `toggle` here created an edge case where if the code runs when its hidden, it'll become visible Replaced to have explicit HIDE ME properties.

maxxcrawford (commented):
Self-review!

.eslintrc.js
Vinnl: Note that we have to be somewhat more careful about this on the website than in the add-on, but [it looks like](https://caniuse.com/mdn-javascript_operators_optional_chaining) only old versions of Safari and Chrome accounting for <3% of global usage don't support optional chaining, so I guess that's fine? Might not hold for everything in ES2020 though.

static/js/app.js
Vinnl: I'd say avoiding toggling code is a good thing in general, since it can often lead to unexpected behaviour when e.g. a user quickly hits the same button multiple times. In other words: nice!

Vinnl (approved):
Is the `-html` suffix intended to indicate that the element was added client-side rather than by Django? (Because I also seem to recall it being used for FTL strings that needed HTML inserted into them? Might be something to add to the class name conventions I suppose. Though if we move to React and it's all client-side, it should go away by itself :) )

PR #1349, opened 2021-11-10
Add a pull request template

Vinnl:
I keep forgetting to check things before I submit a PR. This adds a checklist based on [the template of a project I worked on previously](https://github.com/inrupt/solid-client-js/blob/main/.github/pull_request_template.md), intended only as a reminder for yourself (i.e. as far as I'm concerned, people can just remove whatever's not relevant to their PR, which could be the entire template). And of course, we can always update this template in the future.

Let me know what you think.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,17 @@
+<!-- When fixing a bug: -->
+
+This PR fixes #<issue ID>.
+
+- [ ] l10n dependencies have been merged, if any.
+- [ ] I've added a unit test to test for potential regressions of this bug (or this is a front-end change, where we don't yet have unit test infrastructure).
+- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
+
+<!-- When adding a new feature: -->
+
+# New feature description
+
+# Checklist
+
+- [ ] l10n dependencies have been merged, if any.
+- [ ] All acceptance criteria are met.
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,17 @@
+<!-- When fixing a bug: -->
+
+This PR fixes #<issue ID>.
+
+- [ ] l10n dependencies have been merged, if any.
+- [ ] I've added a unit test to test for potential regressions of this bug (or this is a front-end change, where we don't yet have unit test infrastructure).
+- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
+
```
maxxcrawford, line 8:
```suggestion
- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md)
```

maxxcrawford (approved):
Love it! I added one recommendation highlight the coding standards. Should we be more direct any mention class naming or SASS token usage? That's a constant 'gotcha' in the reviews that I'd like to call out. 

Additionally  – Do we need to mention no lint errors, etc as the CI catches those?

.github/pull_request_template.md
Vinnl: Good suggestion, added in 1cd6a666. I'm not sure what you mean by "class naming" though - you mean using the right prefixes?

And yeah, in my view a checklist is useful for things you need to manually check. Anything that can be delegated to computers should be delegated to computers :)

groovecoder (approved):

PR #1350, opened 2021-11-10
Move the subdomain extension to the next line

Vinnl:
This ensures maximum space for the user's chosen subdomain. Note
that we're currently abusing FTL, setting the `subdomain` variable
in `modal-domain-register-success` to an empty string and then just
stuffing the actual chosen subdomain in front of it. This will
break in languages where the chosen subdomain should not be the
first word, but is the result of not loading the language strings
client-side and thus not being able to substitute in variables
generated on the client-side. This was already the case, and this
commit keeps it that way.


maxxcrawford (approved):
I like this call. Does it break everytime, or when necessary? It would be nice if a small name keeps it on one line (if there's room)

image

--- 


Before:
image

After:
image

Vinnl:
It breaks every time, because we've been going back and forth on trying to make the subdomain fit too many times already, so we just wanted to make absolutely sure it overflows as little as possible. We can look at tightening it up more later.

PR #1357, opened 2021-11-12
_get_relay_recipient_from_message_json

groovecoder:
Go thru all To, Cc, and Bcc fields and return the one that has a Relay domain

This seems to clean up all the scenarios in https://mozilla-hub.atlassian.net/browse/MPP-1226.

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -270,6 +270,31 @@ def _sns_notification(json_body):
     return _sns_message(message_json)
 
 
+def _get_relay_recipient_from_message_json(message_json):
+    # Go thru all To, Cc, and Bcc fields and
+    # return the one that has a Relay domain
+
+    # First check commmon headers for to or cc match
+    headers_to_check = 'to', 'cc'
+    domains_to_check = get_domains_from_settings().values()
+    common_headers = message_json['mail']['commonHeaders']
+    for header in headers_to_check:
+        if header in common_headers:
+            for recipient in common_headers[header]:
+                for domain in domains_to_check:
+                    if domain in recipient:
+                        return parseaddr(recipient)[1]
+
+    # SES-SNS sends bcc in a different part of the message
+    recipients = message_json['receipt']['recipients']
+    for recipient in recipients:
+        for domain in domains_to_check:
+            if domain in recipient:
+                return recipient
+
+    return None
+
+
 def _sns_message(message_json):
     incr_if_enabled('sns_inbound_Notification_Received', 1)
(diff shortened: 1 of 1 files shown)


emails/views.py
```diff
@@ -270,6 +270,31 @@ def _sns_notification(json_body):
...
+                for domain in domains_to_check:
+                    if domain in recipient:
+                        return parseaddr(recipient)[1]
+
+    # SES-SNS sends bcc in a different part of the message
+    recipients = message_json['receipt']['recipients']
+    for recipient in recipients:
+        for domain in domains_to_check:
+            if domain in recipient:
+                return recipient
```
say-yawn, line 293: We should separate this out to its own function since it's being duplicated in two places. A good name could be: `_get_recipient_with_relay_domains`

emails/views.py
```diff
@@ -282,15 +307,14 @@ def _sns_message(message_json):
...
-        logger.error('SNS message without commonHeaders "to".')
-        return HttpResponse(
-            'Received SNS notification without commonHeaders "to".',
-            status=400
-        )
+
+    to_address = _get_relay_recipient_from_message_json(message_json)
+    if to_address is None:
+        incr_if_enabled('no_relay_domain_in_recipient_fields', 1)
+        logger.error('No Relay domain in To, Cc, or Bcc fields.')
```
say-yawn, line 314:
```suggestion
```
Is the logger here necessary or is the metric enough?

say-yawn (changes requested):
LGTM, would like test on this since the code is fairly straight forward.

emails/views.py
groovecoder: It's not exactly duplicated. 😢 In the case of `commonHeaders` we `return parseaddr(recipient)[1]` but in the case of `recipients` we `return recipient`

emails/views.py
groovecoder: Removed the logger line.

emails/views.py
say-yawn: Oh, I didn't see that 🤦 but that means we can do something like:
```
def _get_recipient_with_relay_domains
    for recipient in recipients:
        for domain in domains_to_check:
            if domain in recipient:
                return recipient
    return None
```
and in `SES-SNS sends bcc in a different part of the message` block `return recipient` and in `commmon headers for to or cc match` we can `return parseaddr(recipient)[1] if recipient else recipient`

say-yawn (approved):
LGTM

PR #1362, opened 2021-11-16
Remove test flags

say-yawn:
# About this PR
We used to use the flag `TEST_MOZMAIL` to make sure that mozmail.com domain was only being used by premium users. Since we want this as a default behavior and is now being set using the `default_domain_numerical` we should remove all references to the `TEST_MOZMAIL`.

# Acceptance Criteria
- [x] All users are use mozmail.com domain on Relay and Domain addresses
- [x] mozmail.com domain is ONLY dependent on the `PREMIUM_RELEASE_DATE` (if now is later than `PREMIUM_RELEASE_DATE` mozmail.com should be used otherwise `relay.firefox.com` should be used

#
We should remove any env set for `TEST_MOZMAIL` on Stage and Prod
Need to do further cleaning later to remove flags for gating premium features with `PREMIUM_RELEASE_DATE`

### DIFF
--- a/.env-dist
+++ b/.env-dist
@@ -7,7 +7,6 @@ SECRET_KEY=unsafe-secret-key-for-dev-envs
 ADMIN_ENABLED=
 DEBUG=True
 DJANGO_INTERNAL_IPS=127.0.0.1, localhost
-TEST_MOZMAIL=False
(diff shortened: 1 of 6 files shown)


groovecoder (approved):
Code and spot-checks LGTM. I'll remove the config from dev server after merge.

PR #1368, opened 2021-11-16
Update UI footer Premium CTA

maxxcrawford:
Revised UI for Footer Upgrade CTA. 

## Screenshots

Desktop: 
image

Mobile: 
image


static/scss/components/banners.scss
```diff
@@ -333,3 +333,49 @@
     }
 
 }
+
+
+.c-footer-banner {
+    background-color: $color-white;
+
+    .mzp-l-content {
```
Vinnl, line 341: Isn't `.mzp-*` indicative of Protocol classes? If so, seems unwise to be modifying those - it might clash with future changes there, or with developer expectations for other users of those classes. (If not, ignore this :) )

Vinnl (approved):
:+1:

static/scss/components/banners.scss
maxxcrawford: Oof. I was feeling skirmy about this when I dropped it in. I was wanting to use Protocol, but had to override basically all the padding. I dropped the div/classes entirely and just used tokens to get the max-width to match.

PR #1372, opened 2021-11-18
use x_client_region header to detect country

groovecoder:
Removes the `/premium_subscribe_url` endpoint, `premium_subscribe_url_view` view, and the `setPremiumSubscribeUrlsByCountry` JS function that used them.

Instead, uses the new `x_client_region` HTTP header to pass `country_code` to the `premium_subscribe_url` template tag.

### DIFF
--- a/privaterelay/context_processors.py
+++ b/privaterelay/context_processors.py
@@ -12,11 +12,13 @@ def common(request):
     fxa = _get_fxa(request)
     avatar = fxa.extra_data['avatar'] if fxa else None
     accept_language = request.headers.get('Accept-Language', 'en-US')
+    country_code = request.headers.get('X-Client-Region', 'us').lower()
     return {
         'avatar': avatar,
         'ftl_mode': 'server',
         'accept_language': accept_language,
-        'monthly_price': premium_plan_price(accept_language)
+        'country_code': country_code,
+        'monthly_price': premium_plan_price(accept_language, country_code)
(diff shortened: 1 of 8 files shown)


privaterelay/context_processors.py
```diff
@@ -12,11 +12,13 @@ def common(request):
     fxa = _get_fxa(request)
     avatar = fxa.extra_data['avatar'] if fxa else None
     accept_language = request.headers.get('Accept-Language', 'en-US')
+    country_code = request.headers.get('x_client_region', 'us')
```
moz-astults, line 15: @groovecoder unless there's something changing `_` to `-`, the header name will come in as `x-client-region`

groovecoder: Thanks. I saw `x_client_region` in the GCP log explorer, so I figured that's what it would be. Can we tell if it will be all-lowercase like this too?

moz-astults: @groovecoder yea I think that's more of a GCP log thing, but as long as they are [following RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) these headers should be case-insensitive.

moz-astults (approved):

say-yawn (approved):
LGTM. Is the header isn't as reliable we could consider a hybrid effort on getting the location data. :)

PR #1374, opened 2021-11-18
add premium_available_in_country context var

groovecoder:
* Adds a new `/api/v1/premium_countries` endpoint with response like:
```
{
"country_code":"us",
"premium_countries":["at","be","ch","de","es","fr","ie","it","nl","us","gb","ca","nz","my","sg"],
"premium_available_in_country":true
}
```
* Also adds the same 3 items as context vars (`country_code`, `premium_countries`, and `premium_available_in_country`) for all templates
* Moves the common code for those 3 items into a new `privaterelay.utils.get_premium_countries_info_from_request` function.

### DIFF
--- a/privaterelay/context_processors.py
+++ b/privaterelay/context_processors.py
@@ -4,6 +4,7 @@ from functools import lru_cache
 from django.conf import settings
 
 from .templatetags.relay_tags import premium_plan_price
+from .utils import get_premium_countries_info_from_request
 
 
 def django_settings(request):
@@ -14,8 +15,8 @@ def common(request):
     avatar = fxa.extra_data['avatar'] if fxa else None
     accept_language = request.headers.get('Accept-Language', 'en-US')
     country_code = request.headers.get('X-Client-Region', 'us').lower()
-    premium_available_in_country = (
-        country_code in settings.PREMIUM_PLAN_COUNTRY_LANG_MAPPING.keys()
+    premium_countries_vars = (
+        get_premium_countries_info_from_request(request)
     )
 
     first_visit = request.COOKIES.get("first_visit")
@@ -25,15 +26,17 @@ def common(request):
         (datetime.now(timezone.utc) > datetime.fromisoformat(first_visit) + timedelta(days = 3))
     )
 
-    return {
+    common_vars = {
         'avatar': avatar,
         'ftl_mode': 'server',
         'accept_language': accept_language,
         'country_code': country_code,
-        'monthly_price': premium_plan_price(accept_language, country_code),
         'show_nps': show_nps,
-        'premium_available_in_country': premium_available_in_country
+        'monthly_price': premium_plan_price(
+            accept_language, premium_countries_vars['country_code']
+        ),
     }
+    return {**common_vars, **premium_countries_vars}
 
 @lru_cache(maxsize=None)
 def _get_fxa(request):
--- a/privaterelay/context_processors.py
+++ b/privaterelay/context_processors.py
@@ -14,6 +14,9 @@ def common(request):
     avatar = fxa.extra_data['avatar'] if fxa else None
     accept_language = request.headers.get('Accept-Language', 'en-US')
     country_code = request.headers.get('X-Client-Region', 'us').lower()
+    premium_available_in_country = (
+        country_code in settings.PREMIUM_PLAN_COUNTRY_LANG_MAPPING.keys()
+    )
 
     first_visit = request.COOKIES.get("first_visit")
     show_nps = (
@@ -29,6 +32,7 @@ def common(request):
         'country_code': country_code,
         'monthly_price': premium_plan_price(accept_language, country_code),
         'show_nps': show_nps,
+        'premium_available_in_country': premium_available_in_country
     }
 
 @lru_cache(maxsize=None)
--- a/privaterelay/utils.py
+++ b/privaterelay/utils.py
@@ -34,4 +34,4 @@ def get_premium_country_lang(accept_lang, cc=None):
         if lang in languages.keys():
             return cc, lang
         return cc, list(languages.keys())[0]
-    return 'us', 'en'
+    return cc, 'en'
--- a/privaterelay/utils.py
+++ b/privaterelay/utils.py
@@ -1,0 +1,37 @@
+from django.conf import settings
+
+def get_premium_countries_info_from_request(request):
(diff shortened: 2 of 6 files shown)


privaterelay/context_processors.py
```diff
@@ -12,13 +12,17 @@ def common(request):
     fxa = _get_fxa(request)
     avatar = fxa.extra_data['avatar'] if fxa else None
     accept_language = request.headers.get('Accept-Language', 'en-US')
-    country_code = request.headers.get('X-Client-Region', 'us').lower()
+    country_code = request.headers.get('X-Client-Region', '').lower()
```
Vinnl, line 15: I'd do one of two things here:

1. Fall back to `accept_language` if this header is not available (risking that we're showing the wrong price/currency to people, and/or showing Premium promotion to people who can't get it), or
2. Set `country_code` to `us` if the header is not set and `DEBUG==True`, otherwise we'll never see Premium banners when running locally, possibly breaking it without noticing.

groovecoder: Moved a lot of this into `utils.py` now too.

privaterelay/utils.py
```diff
@@ -0,0 +1,37 @@
...
+        'premium_countries': premium_countries,
+        'premium_available_in_country': premium_available_in_country
+    })
+
+
+def _get_cc_from_request(request):
+    if 'X-Client-Region' in request.headers:
+        return request.headers['X-Client-Region'].lower()
+    if 'Accept-Language' in request.headers:
+        return get_premium_country_lang(request.headers['Accept-Language'])
```
Vinnl, line 18: As I understand it `get_premium_country_lang` returns a tuple `(cc, lang)`. If we don't return just the country code, `get_premium_countries_info_from_request` will try to find the tuple as a whole in `premium_countries`, which it can't.

```suggestion
        return get_premium_country_lang(request.headers['Accept-Language'])[0]
```

privaterelay/context_processors.py
Vinnl: Looks good. I guess the second point isn't too relevant if you fall back to the Accept language too, since the browser I think always sends an accept language and thus we'll never hit that code path in practice, but it doesn't do any harm either :)

privaterelay/utils.py
maxxcrawford: @groovecoder This change is blocking #1376. Could you review?

groovecoder: Committed suggestion.

privaterelay/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    if cc is None:
+        cc = lang_parts[1] if len(lang_parts) == 2 else lang_parts[0]
+        cc = cc.lower()
+
+    if cc in settings.PREMIUM_PLAN_COUNTRY_LANG_MAPPING.keys():
+        languages = settings.PREMIUM_PLAN_COUNTRY_LANG_MAPPING[cc]
+        if lang in languages.keys():
+            return cc, lang
+        return cc, list(languages.keys())[0]
+    return 'us', 'en'
```
Vinnl, line 37: This causes a problem for detecting whether a country has Premium based on the accept language: if the country does not have Premium, this will auto-default to `'us', 'en'`, and then return that they _do_ have Premium.

privaterelay/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    })
+
+
+def _get_cc_from_request(request):
+    if 'X-Client-Region' in request.headers:
+        return request.headers['X-Client-Region'].lower()
+    if 'Accept-Language' in request.headers:
+        return get_premium_country_lang(request.headers['Accept-Language'])[0]
+    if settings.DEBUG:
+        return 'us'
```
say-yawn, line 20:
```suggestion
```
Is this test code or actually necessary?

say-yawn (approved):
LGTM. Minor change recommended.

privaterelay/utils.py
say-yawn: @Vinnl said good catch! @groovecoder, we should return the `cc` as is and catch and handle that in `premium_plan_id` and `premium_plan_price`.

groovecoder: Fixed and added test!

privaterelay/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    if cc is None:
+        cc = lang_parts[1] if len(lang_parts) == 2 else lang_parts[0]
+        cc = cc.lower()
+
+    if cc in settings.PREMIUM_PLAN_COUNTRY_LANG_MAPPING.keys():
+        languages = settings.PREMIUM_PLAN_COUNTRY_LANG_MAPPING[cc]
+        if lang in languages.keys():
+            return cc, lang
+        return cc, list(languages.keys())[0]
+    return cc, 'en'
```
Vinnl, line 37: @groovecoder Hmm, looks like we have other code that assumes that this function only returns country codes that are present in `PREMIUM_PLAN_COUNTRY_MAPPING`, e.g. here: https://github.com/mozilla/fx-private-relay/blob/68c209d2c447c5fbe69a2c4216c8685bccc51b8d/privaterelay/templatetags/relay_tags.py#L57

If I run this and set my browser's Accept-Language to `ro-ro`, I get a KeyError because `ro` is not present there :/

groovecoder: argh!!! okay let me add a test case and fix.

groovecoder: Okay, I updated the `relay_tags` to return sensible values when the country code isn't in the `MAPPING`. (I think I checked that all places that need the ID will work with a `''` value, and everywhere that needs a price can at least *render* a USD price instead.

Vinnl: Testing it locally, I'm no longer encountering any problems! :tada:

Vinnl:
Oh, could I make one more request for this API? It would be nice if it could also return the plan ID for the Premium plan relevant to this country.

groovecoder:
I rebased this branch on `main` which caused conflicts and broke comparison in GitHub, so I didn't merge #1396. But, I added `plan_country_lang_mapping` to `/api/v1/premium_countries` in this branch myself.

privaterelay/utils.py
groovecoder: We want this so that our local dev instances are "premium available" by default.

PR #1378, opened 2021-11-20
capture bounce details into sentry

groovecoder:
To help us understand why some receiving MTA's are bouncing emails back at Relay, capture bounced recipient details in sentry messages. (Note: this strips the localpart of the bounced recipient so we don't leak that PII into sentry.)

This also updates the bounce-handling code to match the bounce notification I got from my local SES. It should keep the existing logic working too, in case my bounce notifications are different from what's in dev/stage/prod.

Example of the message:
```
bounced recipient domain: simulator.amazonses.com, {'action': 'failed', 'status': '5.1.1', 'diagnosticCode': 'smtp; 550 5.1.1 user unknown'}
```

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -609,14 +609,16 @@ def _handle_bounce(message_json):
     bounce = message_json.get('bounce')
     bounced_recipients = bounce.get('bouncedRecipients')
     for recipient in bounced_recipients:
-        recipient_address = recipient.pop('emailAddress', None)
+        recipient_address = recipient.get('emailAddress', None)
+        if recipient_address is None:
+            continue
         recipient_address = parseaddr(recipient_address)[1]
         recipient_domain = recipient_address.split('@')[1]
         capture_message(
             f'bounced recipient domain: {recipient_domain}, {recipient}'
         )
         try:
-            user = User.objects.get(email=recipient.get('emailAddress'))
+            user = User.objects.get(email=recipient_address)
             profile = user.profile_set.first()
         except User.DoesNotExist:
             incr_if_enabled('email_bounce_relay_user_gone', 1)
--- a/emails/views.py
+++ b/emails/views.py
@@ -257,7 +257,10 @@ def _sns_notification(json_body):
     message_json = json.loads(json_body['Message'])
     event_type = message_json.get('eventType')
     notification_type = message_json.get('notificationType')
-    if notification_type != 'Received' and event_type != 'Bounce':
+    if (
+        notification_type not in['Received', 'Bounce'] and
+        event_type != 'Bounce'
+    ):
         logger.error(
             'SNS notification for unsupported type',
             extra={'notification_type': notification_type},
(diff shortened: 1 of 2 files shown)


emails/views.py
```diff
@@ -604,6 +609,12 @@ def _handle_bounce(message_json):
...
     bounced_recipients = bounce.get('bouncedRecipients')
     for recipient in bounced_recipients:
+        recipient_address = recipient.pop('emailAddress', None)
+        recipient_address = parseaddr(recipient_address)[1]
+        recipient_domain = recipient_address.split('@')[1]
+        capture_message(
+            f'bounced recipient domain: {recipient_domain}, {recipient}'
+        )
         try:
             user = User.objects.get(email=recipient.get('emailAddress'))
```
say-yawn, line 619:
```suggestion
            user = User.objects.get(email=recipient.get(recipient_full_address))
```

emails/views.py
```diff
@@ -604,6 +609,12 @@ def _handle_bounce(message_json):
     bounce = message_json.get('bounce')
     bounced_recipients = bounce.get('bouncedRecipients')
     for recipient in bounced_recipients:
+        recipient_address = recipient.pop('emailAddress', None)
+        recipient_address = parseaddr(recipient_address)[1]
```
say-yawn, line 613:
```suggestion
        recipient_full_address = recipient.pop('emailAddress', None)
        recipient_address = parseaddr(recipient_full_address)[1]
```
`emailAddress` value is looked up later in the code so we should assign this as a separate variable to use it in the try/catch block.

emails/views.py
say-yawn: Should we early return if the `email_full_address` was `None` or is this handled in the exception for `User.DoesNotExist` because we'll never have a `User` without an `email` value?

say-yawn (changes requested):
Changes requested to make sure the full email address is used for user look-up in the try/catch block.

groovecoder:
Addressed the PR comments (but not with the suggested changes.)

emails/views.py
```diff
@@ -609,14 +609,16 @@ def _handle_bounce(message_json):
...
+        if recipient_address is None:
+            continue
         recipient_address = parseaddr(recipient_address)[1]
         recipient_domain = recipient_address.split('@')[1]
         capture_message(
             f'bounced recipient domain: {recipient_domain}, {recipient}'
         )
         try:
-            user = User.objects.get(email=recipient.get('emailAddress'))
+            user = User.objects.get(email=recipient_address)
```
say-yawn, line 621: We have a same problem with `recipient_address` being only the domain portion since in line 615:
```
recipient_address = parseaddr(recipient_address)[1]
```

say-yawn (changes requested):
Change requested.

emails/views.py
groovecoder: [The return value of `parseaddr` is a tuple](https://docs.python.org/3.8/library/email.utils.html#email.utils.parseaddr) where the `0`th item is the "real name" and the `1`st item is the email address.

I confirmed by triggering a bounce to my local app. When SES/SNS sent Relay the bounce message, the `emailAddress` field comes in like:

```
ipdb> recipient.get('emailAddress', None)
'bounce@simulator.amazonses.com'
ipdb> recipient_address
'bounce@simulator.amazonses.com'
ipdb> parseaddr(recipient_address)
('', 'bounce@simulator.amazonses.com')
ipdb> parseaddr(recipient_address)[1]
'bounce@simulator.amazonses.com'
```

say-yawn: oh my bad. I should have looked into that more. Thanks for clarifying that.

say-yawn (approved):
LGTM

PR #1367, opened 2021-11-16
Add separate metric for replies

say-yawn:
# About this PR
Since we offer premium feature to reply to relayed emails, we want to separately increment the number of replies made instead of grouping it with forwarded.

# Acceptance Criteria
- [ ] Client side code can get number of replied per Relay address
- [ ] Client side code can get number of replied per Domain address
- [ ] Client side can get total number of replied per profile
- [ ] Forward metric is NO longer incremented when a user responds to email

#
[MPP-1146](https://mozilla-hub.atlassian.net/browse/MPP-1146)

PR #1391, opened 2021-11-30
Keep track of when someone subscribes to Premium

Vinnl:
# New feature description

This adds a column `date_subscribed` to the profile to keep track of when we first saw a Premium subscription for the user.

Don't hold back when reviewing this - I have practically zero back-end, Python and Django experience :) Things I was unsure about:

- Is the view the right place to write data to the DB? (And if not, what is?)
- (How) do I write a test for this?
- How can I manually test the removal of the date when unsubscribing? If I unsubscribe from the FxA interface, the subscription won't be cancelled for the next month. And related:
- What will cause the FxA data be refreshed? I understand that FxA redirects back to us with a query parameter that triggers a refresh when the user subscribes, but I don't see any other ways this gets triggered.

# Checklist

- [ ] All acceptance criteria are met. - This is step one in being able to show an NPS survey three days after subscribing.


api/serializers.py
```diff
@@ -40,5 +40,5 @@ class Meta:
 class ProfileSerializer(serializers.ModelSerializer):
     class Meta:
         model = Profile
-        fields = ['id', 'server_storage', 'subdomain', 'onboarding_state']
-        read_only_fields = ['id', 'subdomain']
+        fields = ['id', 'server_storage', 'subdomain', 'onboarding_state', 'subscription_date']
+        read_only_fields = ['id', 'subdomain', 'subscription_date']
```
groovecoder, line 44: Did this change in dev? Looks like the model field is called `date_subscribed`?

```suggestion
        fields = ['id', 'server_storage', 'subdomain', 'onboarding_state', 'date_subscribed']
        read_only_fields = ['id', 'subdomain', 'date_subscribed']
```

privaterelay/views.py
```diff
@@ -86,6 +86,12 @@ def profile(request):
         _handle_fxa_profile_change(fxa)
         now_has_premium = profile.has_premium
         newly_premium = not had_premium and now_has_premium
+        if newly_premium:
+            profile.date_subscribed = datetime.now(timezone.utc)
+            profile.save()
+        elif had_premium and not now_has_premium:
+            profile.date_unsubscribed = None
```
groovecoder, line 93:
```suggestion
            profile.date_subscribed = None
```

groovecoder: I wonder if we should set `date_subscribed = None` at all. If we leave it as the last datetime that the user subscribed, we could do a query to find "churned" customers - i.e. (pseudo-code):
```
churned_users = SocialAccount.objects
    .exclude(extra_data__contains='premium-relay')
    .include(user_id__in=User.objects.exclude(
        profile__date_subscribed=None
    ))
```

groovecoder (changes requested):

api/serializers.py
Vinnl: Whoops no, that's me realising at the last moment that I forgot the API field, mixing up the names, and forgetting to test it :/ Fixed now.

privaterelay/views.py
Vinnl: Hmm good point. I was thinking of people unsubscribing and later resubscribing, but of course just overwriting the field too. Maybe I should rename it to `date_last_subscribed?

Only risk is that we then really should be using `has_premium` to determine whether someone has Premium, i.e. that future contributors should not look at this field and think they can use it for determining whether someone has Premium.

Anyway, I removed it now.

groovecoder (changes requested):
Approved, but marking as "request changes" because we need to coordinate which to merge first: this PR, or https://github.com/mozilla/fx-private-relay/pull/1367 which also contains a migration.

Vinnl:
@groovecoder Now that #1367 has been closed, can we merge this?

groovecoder (approved):
Code and spot-checks look good.

PR #1400, opened 2021-12-04
_record_receipt_verdicts to record ses receipt verdicts

groovecoder:
SES makes checks for spf, dkim, dmarc, spam, and viruses. It sends
the check results in fields like spamVerdict, dmarcVerdict, etc.

We may want to block emails based on these verdicts, but we want to
know how many emails are getting each verdict. So, record into statsd.

## Example
The receipt verdict counters will be recorded like so:
```
spamVerdict_PASS
virusVerdict_PASS
spfVerdict_FAIL
dkimVerdict_GRAY
dmarcVerdict_FAIL
```

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -317,11 +317,13 @@ def _sns_message(message_json):
         )
     common_headers = mail['commonHeaders']
 
+    _record_receipt_verdicts(message_json, 'all')
     to_address = _get_relay_recipient_from_message_json(message_json)
     if to_address is None:
         incr_if_enabled('no_relay_domain_in_recipient_fields', 1)
         return HttpResponse("Address does not exist", status=404)
 
+    _record_receipt_verdicts(message_json, 'relay_recipient')
     from_address = parseaddr(common_headers['from'][0])[1]
     [to_local_portion, to_domain_portion] = to_address.split('@')
     if to_local_portion == 'noreply':
@@ -339,9 +341,11 @@ def _sns_message(message_json):
 
         return HttpResponse("Address does not exist", status=404)
 
+    _record_receipt_verdicts(message_json, 'valid_user')
     # first see if this user is over bounce limits
     bounce_paused, bounce_type = user_profile.check_bounce_pause()
     if bounce_paused:
+        _record_receipt_verdicts(message_json, 'user_bounce_paused')
         incr_if_enabled('email_suppressed_for_%s_bounce' % bounce_type, 1)
         return HttpResponse("Address is temporarily disabled.")
 
@@ -365,8 +369,10 @@ def _sns_message(message_json):
         incr_if_enabled('email_for_disabled_address', 1)
         address.num_blocked += 1
         address.save(update_fields=['num_blocked'])
+        _record_receipt_verdicts(message_json, 'disabled_alias')
         return HttpResponse("Address is temporarily disabled.")
 
+    _record_receipt_verdicts(message_json, 'active_alias')
     incr_if_enabled('email_for_active_address', 1)
 
     subject = common_headers.get('subject', '')
@@ -433,6 +439,18 @@ def _sns_message(message_json):
         remove_message_from_s3(bucket, object_key)
 
     return response
+
+
+def _record_receipt_verdicts(message_json, state):
+    verdict_tags = []
+    for key, value in message_json['receipt'].items():
+        if key.endswith('Verdict'):
+            value = value['status']
+            verdict_tags.append(f'{key}:{value}')
+            incr_if_enabled(
+                f'relay.emails.verdicts.{key}', 1, [f'state:{state}']
+            )
+    incr_if_enabled(f'relay.emails.state.{state}', 1, verdict_tags)
 
 
 def _get_keys_from_headers(headers):
--- a/emails/utils.py
+++ b/emails/utils.py
@@ -40,9 +40,9 @@ def time_if_enabled(name):
     return timing_decorator
 
 
-def incr_if_enabled(name, value=1):
+def incr_if_enabled(name, value=1, tags=None):
     if settings.STATSD_ENABLED:
-        metrics.incr(name, value)
+        metrics.incr(name, value, tags)
 
 
 def histogram_if_enabled(name, value, tags=None):


emails/views.py
```diff
@@ -301,6 +301,7 @@ def _get_relay_recipient_from_message_json(message_json):
 
 def _sns_message(message_json):
     incr_if_enabled('sns_inbound_Notification_Received', 1)
+    _record_receipt_verdicts(message_json)
```
moz-astults, line 304: @groovecoder should this happen after the `commonHeaders` check to ensure it's a well formatted message before trying to parse the JSON payload?

say-yawn (approved):
LGTM given that @moz-astults change recommendation be applied if we are checking the verdict values only for emails we are relaying and not for **all** the emails being received by our inbound notification. Alternatively, we could gather data on both before and after commonHeaders check to find any correlations that can better handle emails in the receipt rules if we want to optimize for that.

say-yawn:
Sample SNS message [here](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html) and different types of verdicts and status found [here](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/notification-contents.html)

groovecoder:
Oh, neat idea. I updated this to call the `_record_receipt_verdicts` function with an indicator prefix a number of times thru-out the process:

* After checking `commonHeaders` (indicator `all`)
* After determining a relay recipient is in the message headers (indicator `relay_recipient`)
* After validating the relay recipient belongs to a relay user (indicator `valid_user`)
* If the relay user is bounce-paused (indicator `user_bounce_paused`)
* If the alias is disabled (indicator `disabled_alias`)
* If the alias is active (indicator `active_alias`)

By looking at the verdicts at each stage of processing, we can learn, e.g., if there are a high number of certain failures going to users who are bounce-paused. Which could tell us if, e.g., emails that fail SPF, DKIM, and/or DMARC are disproportionately related to bounce pauses. If that's the case, blocking those emails could result in fewer bounces and bounce-pauses!

moz-astults (approved):

emails/views.py
```diff
@@ -435,6 +441,13 @@ def _sns_message(message_json):
     return response
 
 
+def _record_receipt_verdicts(message_json, stage):
+    for key, value in message_json['receipt'].items():
+        if key.endswith('Verdict'):
+            value = value['status']
+            incr_if_enabled(f'{stage}_{key}_{value}')
```
say-yawn, line 448:
```suggestion
            histogram_if_enabled(key, value, stage)
```
I think this is a good application for Markus's [histogram](https://markus.readthedocs.io/en/latest/usage.html#markus.main.MetricsInterface.histogram) ([sample](https://markus.readthedocs.io/en/latest/metricsoverview.html?highlight=histogram#getting-statistical-distributions-timing-histogram) scenarios to use histogram). 
Doing this will help answer question on “Does the different verdicts and their result correlate to checks passing/failing?” but we will need to pass multiple `stage`s so we don't duplicate the data point. Since that may be hard to do I think doing:
```
            histogram_if_enabled(stage, key, value)
```
Which means we can ask the same question like:
"In a stage of a process (`stage`), what are the distribution of different verdict types (`key`) broken down by types of verdict results (`value`s)." and can be easier to see if the combinations of verdicts being available or not will also impact the likelihood the email pass/fail-ing the different stage. I reached out to William (willkg) to see if this would be a good way to format the data.

say-yawn: Yeah, I did get to talk to him and we think it's best to stick with `incr` but with tags on the data.
```python
metrics.incr(
    "relay.emails.status.checkbounce",
    value=1,
    tags=["dkimverdict:pass", "dmarcverdict:fail", ...]
)

metrics.incr(
    "relay.emails.verdicts.dkim",
    value=1,
    tags=["status:checkbounce"]
)

# Then in grafan, you can graph that with something like:
from relay.emails.status.checkbounce
select field(value) sum()
group by time($__interval) tag:dkimverdict fill(null)
```
Something like this will give us data on:
- each stage/process and how many verdicts were present and pass/failed.
- for verdict what was the process that passed (not necessary, but it's another way to see the previous bullet point but easier to see patterns for a given verdict instead of stage)

groovecoder:
I've updated the code to make increments with tags. Haven't spot-checked yet though.

groovecoder:
Spot-check looks good:
```
relay.emails.verdicts.spamVerdict +1
state:all
relay.emails.verdicts.virusVerdict +1
state:all
relay.emails.verdicts.spfVerdict +1
state:all
relay.emails.verdicts.dkimVerdict +1
state:all
relay.emails.verdicts.dmarcVerdict +1
state:all
relay.emails.verdicts.all +1
['spamVerdict:PASS', 'virusVerdict:PASS', 'spfVerdict:PASS', 'dkimVerdict:PASS', 'dmarcVerdict:PASS']
relay.emails.verdicts.spamVerdict +1
state:relay_recipient
relay.emails.verdicts.virusVerdict +1
state:relay_recipient
relay.emails.verdicts.spfVerdict +1
state:relay_recipient
relay.emails.verdicts.dkimVerdict +1
state:relay_recipient
relay.emails.verdicts.dmarcVerdict +1
state:relay_recipient
relay.emails.verdicts.relay_recipient +1
['spamVerdict:PASS', 'virusVerdict:PASS', 'spfVerdict:PASS', 'dkimVerdict:PASS', 'dmarcVerdict:PASS']
relay.emails.verdicts.spamVerdict +1
state:valid_user
relay.emails.verdicts.virusVerdict +1
state:valid_user
relay.emails.verdicts.spfVerdict +1
state:valid_user
relay.emails.verdicts.dkimVerdict +1
state:valid_user
relay.emails.verdicts.dmarcVerdict +1
state:valid_user
relay.emails.verdicts.valid_user +1
['spamVerdict:PASS', 'virusVerdict:PASS', 'spfVerdict:PASS', 'dkimVerdict:PASS', 'dmarcVerdict:PASS']
relay.emails.verdicts.spamVerdict +1
state:active_alias
relay.emails.verdicts.virusVerdict +1
state:active_alias
relay.emails.verdicts.spfVerdict +1
state:active_alias
relay.emails.verdicts.dkimVerdict +1
state:active_alias
relay.emails.verdicts.dmarcVerdict +1
state:active_alias
relay.emails.verdicts.active_alias +1
['spamVerdict:PASS', 'virusVerdict:PASS', 'spfVerdict:PASS', 'dkimVerdict:PASS', 'dmarcVerdict:PASS']
```

PR #1403, opened 2021-12-06
Inline email images

Vinnl:
This way, we cannot (even theoretically) track when someone opens
the emails we forward. Additionally, the image should no longer be
hidden by email clients that disable image embedding by default for
the above reason.

How to test:

Ideally, have Relay forward an email and verify that the Relay image is still visible in the footer. Also, try setting your email client to block embedded images, and verify that the logo is still visible.

Unfortunately I wasn't able to do that myself, since I don't know how to deploy to `dev` before merging in a PR. I just opened the email template in my browser and verified that the image still displays OK and is powered by a data URL in the source.


maxxcrawford (commented):
Q: Does embedded images count/increase file size over 150kb? (or is it outside of that limitation?)

Vinnl:
Oh crap, that's a good point. I would think so, but pinging @groovecoder to make sure - another 5.5 kB could push an email over the edge. If so, I guess it's best to wait until we can raise the limit.

Vinnl:
People on StackOverflow [are also reporting](https://stackoverflow.com/a/8580387) that Gmail removes data URLs from images... Closing this.

PR #1416, opened 2021-12-10
pop emailAddress so it isn't logged into sentry

groovecoder:
This PR fixes https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/15904470/?referrer=alert_email.

How to test:

Regression testing: make sure email forwarded still works in dev/stage before prod

Need to check the sentry error in  production


say-yawn (approved):
LGTM `_handle_bounce` only gets used if the notification or event had the type `bounce` and should not affect the email being relayed.

PR #1422, opened 2021-12-13
Separate logs stream to stderr and stdout mpp 1275

say-yawn:
# About this PR
Logs are all going to stderr which makes it hard for our SREs to find err messages. This PR sets all none error logs to go to stdout instead of stderr by creating a separate handler `console_out` stream logs from `request.summary`, `eventsinfo`, and `abusemetrics` to go to stdout.

# How to test
Because there is no well documented ways to mimic how Google's Log Explorer highlights the stdout from stderr, this should be tested on Stage and Prod. Instead here are some documentations to prove that this fix should solve the problem.

Before the fix we were using [logging.StreamHandler](https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler) in our settings.py which defaults to logging to `sys.stderr`. Setting the `stream` value in the settings.py to `sys.stdout` should update the StreamHandler to stream the logs to `sys.stdout` [[1](https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler.setStream)]. While GCP, may handle any log with ERROR messages stream to stderr, I separated the handlers and the loggers so that each logger is printing only to one or the other. Here are the list of places where each logger and the type of messages are used:
 . | request.summary | request.summary | request.summary | events | events | events | abusemetrics | abusemetrics | abusemetrics |
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- |
  | INFO | WARNING | ERROR | INFO | WARNING | ERROR | INFO | WARNING | ERROR
emails/views.py |   |   |   | 1 |   | 10 |   |   |  
models.py |   |   |   |   |   |   | 1 |   |  
utils.py |   |   |   |   |   | 5 |   |   |  
sns.py |   |   |   |   |   | 1 |   |   |  
commands/ |   |   |   | 2 |   | 4 |   |   |  
privaterelay/views.py |   |   |   | 3 |   | 1 |   |   |  
models.py |   |   |   |   |   |   |   |   |  
utils.py |   |   |   |   |   |   |   |   |  

#
This PR fixes [MPP-1275](https://mozilla-hub.atlassian.net/browse/MPP-1275).

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -11,7 +11,7 @@ https://docs.djangoproject.com/en/2.2/ref/settings/
 """
 
 import ipaddress
-import os
+import os, sys
 from datetime import datetime
 
 from decouple import config
@@ -421,6 +421,8 @@ LANGUAGE_CODE = 'en'
 LANGUAGES = settings.LANGUAGES + [
     ('zh-tw', 'Chinese'),
     ('zh-cn', 'Chinese'),
+    ('fy', 'Frisian'),
+    # ('fy-nl', 'Frisian'),
 ]
 
 TIME_ZONE = 'UTC'
@@ -484,7 +486,13 @@ LOGGING = {
         }
     },
     'handlers': {
-        'console': {
+        'console_out': {
+            'level': 'DEBUG',
+            'class': 'logging.StreamHandler',
+            'stream': sys.stdout,
+            'formatter': 'json'
+        },
+        'console_err': {
             'level': 'DEBUG',
             'class': 'logging.StreamHandler',
             'formatter': 'json'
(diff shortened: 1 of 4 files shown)


privaterelay/settings.py
```diff
@@ -421,6 +421,8 @@ def _get_initial_middleware():
 LANGUAGES = settings.LANGUAGES + [
     ('zh-tw', 'Chinese'),
     ('zh-cn', 'Chinese'),
+    ('fy', 'Frisian'),
+    # ('fy-nl', 'Frisian'),
```
jwhitlock, line 425: Adding Frisian should be in a different PR

jwhitlock (approved):
Thanks @say-yawn, this looks good to me. It seems there should be other ways for the SREs to filter for just the "interesting" logs, but this accomplishes the goals as you stated them.

privaterelay/settings.py
say-yawn: Good catch. this was a code snippet I was testing for a different bug. I should remove this.

say-yawn: Done on 4321891

groovecoder (approved):
Code and spot-checks look good. Looking forward to seeing how it helps GCP logging in stage!

PR #1424, opened 2021-12-14
don't allow DELETE requests to profiles

groovecoder:
This PR fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1742217.

How to test:
1. Visit http://127.0.0.1:8000/api/v1/docs/
2. Make sure the "DELETE" endpoint is gone from the profiles section
3. View source of http://127.0.0.1:8000/accounts/profile/
4. Get the value from `data-profile-id`
5. In network dev tool, right click the /accounts/profile/ request and choose "Edit and Resend"
6. Change the method to "DELETE" and the url to "http://127.0.0.1:8000/api/v1/profiles/${data-profile-id}/"
7. Send the request
   * [ ] The server should respond with `403 Forbidden`

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -49,6 +49,7 @@ class DomainAddressViewSet(SaveToRequestUser, viewsets.ModelViewSet):
 class ProfileViewSet(viewsets.ModelViewSet):
     serializer_class = ProfileSerializer
     permission_classes = [permissions.IsAuthenticated, IsOwner]
+    http_method_names = ['get', 'post', 'head', 'put', 'patch']
 
     def get_queryset(self):
         return Profile.objects.filter(user=self.request.user)


api/views.py
```diff
@@ -49,6 +49,7 @@ def get_queryset(self):
 class ProfileViewSet(viewsets.ModelViewSet):
     serializer_class = ProfileSerializer
     permission_classes = [permissions.IsAuthenticated, IsOwner]
+    http_method_names = ['get', 'post', 'head', 'put', 'patch']
```
say-yawn, line 52:
```suggestion
    http_method_names = ['get', 'patch']
```
Can we consider restricting this more? Not sure what methods the add-on/client uses to update the `server_storage` and the `onboarding_state` but we could consider removing `post`, `patch`, and `put` if it only uses `patch` for example. Would prefer the `patch` over `put` since `put` historically was used to create object if it didn't exist (no longer the case after 3.0, see doc [here](https://www.django-rest-framework.org/api-guide/generic-views/#put-as-create))

say-yawn (approved):
Test instruction returns 403 and remove DELETE from Profile as expected. A recommendation made to more strictly scope the methods allowed for `ProfileViewSet` but not required.

api/views.py
groovecoder: I'd like to leave the API endpoint more open-ended to accomodate more client use-cases. (I'm also thinking we should go back thru and expose MORE of the model in the endpoints, e.g., I had to [add `full_address` to the address API endpionts.](https://github.com/mozilla/fx-private-relay/pull/1425))

PR #1428, opened 2021-12-15
Remove the `premiumEnabled` feature flag

Vinnl:
Premium has now been enabled for everyone. This is a companion PR to https://github.com/mozilla/fx-private-relay-add-on/pull/246, which should be merged first.


Vinnl:
I was thinking that too, but AFAIK it will just return `undefined` if it looks up a non-existent data att- Oh wait, that will cause it not to display Premium features. Yeap, let's release that first :P

maxxcrawford:
Actually – the thing is keys up on is [still here](https://github.com/mozilla/fx-private-relay/blob/remove-premium-feature-flag/privaterelay/templates/profile.html#L29). We could remove it, or just set the value to `True` for now and remove it later. That would unblock the add-on PR dependency.

maxxcrawford (changes requested):
Blocking approval temporarily for the instance that is still in the code: https://github.com/mozilla/fx-private-relay/blob/remove-premium-feature-flag/privaterelay/templates/profile.html#L29

Question: Are any of the FAQs no longer in use now?

Vinnl:
Huh, I was sure I did a search across the codebase. But that's a good idea, set it to True in f9d68b31!

And yes, there are a couple of FAQs that are no longer being shown. I was thinking we could remove them after this is merged, but actually, since they already aren't shown on the website, we could remove them right away I suppose. Only reason I'm a bit hesitant is that there'd be a lot of eyeballing trying to figure out which are and aren't shown anymore, so I'm not sure if the cleanup is worth the risk of accidentally removing an entry that _is_ still shown?

privaterelay/templates/profile.html
```diff
@@ -26,7 +26,7 @@
   {% with request.user.profile_set.first as user_profile %}
     data-premium-subdomain-set="{{ user_profile.subdomain }}"
   {% endwith %}
-  data-premium-enabled="{{ settings.PREMIUM_ENABLED }}"
+  data-premium-enabled="True"
```
maxxcrawford, line 29: Not sure how ya feel about TODOs in the code, but I would recommend adding a one-liner about why this is here, and that it should be removed at a later date. 
```suggestion
{# TODO: Remove the `premium-enabled` attribute once https://github.com/mozilla/fx-private-relay-add-on/pull/246 has been merged/deployed   #}
  data-premium-enabled="True"
```

maxxcrawford (approved):
Good to go. I have one note about a commented TODO.

privaterelay/templates/profile.html
Vinnl: Excellent callout, don't know why I didn't do that here.

PR #1429, opened 2021-12-15
use SuspiciousOperation exception to cause 400 errors

groovecoder:
This PR fixes https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/15055581/?query=is%3Aunresolved.

## How to test:
1. Sign in with a local Relay account
2. (In another container/PBM window) Go to http://127.0.0.1:8000/admin/emails/profile/
3. Click the profile of the Relay account
4. Set its "Last account flagged" values to Today/Now
5. Go to http://127.0.0.1:8000/api/v1/docs/
6. At the POST /relayaddresses/ section, click "Try it out", and then "Execute"

### Expected Result:
Should get `Error: Bad Request` with status code of 400

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -10,6 +10,7 @@ import uuid
 from django.apps import apps
 from django.conf import settings
 from django.contrib.auth.models import User
+from django.core.exceptions import SuspiciousOperation
 from django.db import models
 from django.dispatch import receiver
 from django.utils.translation.trans_real import (
@@ -394,7 +395,9 @@ class RegisteredSubdomain(models.Model):
         return self.subdomain_hash
 
 
-class CannotMakeSubdomainException(Exception):
+# extend from SuspiciousOperation to trigger 400 status code error
+# TODO: in Django 3.2+ change this to BadRequest
+class CannotMakeSubdomainException(SuspiciousOperation):
     """Exception raised by Profile due to error on subdomain creation.
 
     Attributes:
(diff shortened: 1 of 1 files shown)


say-yawn (changes requested):
If these are only for exceptions thrown because the account was on pause, I suggest using the `SuspiciousOperations` where the exception is being thrown on emails/models.py. Or if we want to just better handle the exceptions thrown because of the API requests, this PR and its changes [here](https://github.com/mozilla/fx-private-relay/pull/1437/files#diff-7fb412808754eb12ee1b172c34c44eb9ce58b3bba43d62ffcb986ad7917588a8R55) should wrap the error into permission denied.

groovecoder:
#1437 definitely improves error-handling to a better degree than this PR.

But we want to return a `4xx` response _anytime_ our code encounters a "known" exception (e.g., account is on pause, API request is mal-formed, DB conflict, etc.) so that these "known" exceptions don't trigger 5xx alerts that ping SRE's. 

Anytime we throw one of our own exceptions, we KNOW it's an erroroneous code path caused by a bad request. So when we're seing a spike of bad requests that we already know cause errors, we want Django to return a `4xx` response instead of a `5xx` response that causes SRE alerts.

say-yawn (approved):
@groovecoder that makes sense. I understand better what you are saying. The known exceptions which inherits from the `Exception` allows the errors to respond as `5xx` when in fact it should be `4xx` because they are known. This will remove the noise from `5xx`, since the known errors will no longer cause the `5xx` to spike. Thanks for elaborating on this further.

PR #1432, opened 2021-12-16
fix MPP-1489 add ie and at country codes to PREMIUM_PLAN_COUNTRY_LANG…

groovecoder:
…_MAPPING

This PR fixes https://mozilla-hub.atlassian.net/browse/MPP-1489.

How to test:
1. Change your Firefox language to `en-ie`
2. Go to http://127.0.0.1:8000/premium
   * [ ] You should see the price in €
3. Change your Fireefox language to `de-at`
4. Go to http://127.0.0.1:8000/premium
   * [ ] You should see the price in €

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -257,6 +257,10 @@ PREMIUM_PLAN_ID_MATRIX = {
         },
     },
     "euro": {
+        "at": {
+            "id": "price_1JmRTDJNcmPzuWtRnJavIXXX",
+            "price": "0,99 €",
+        },
         "de": {
             "id": "price_1JmRTDJNcmPzuWtRnJavIXXX",
             "price": "0,99 €",
@@ -279,6 +283,10 @@ PREMIUM_PLAN_ID_MATRIX = {
         },
         "nl": {
             "id": "price_1JmROfJNcmPzuWtR6od8OfDW",
+            "price": "0,99 €",
+        },
+        "ie": {
+            "id": "price_1JmRCQJNcmPzuWtRprMnmtax",
             "price": "0,99 €",
         },
     },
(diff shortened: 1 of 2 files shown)


Vinnl (commented):
Hmm, I don't understand how this would solve the issue, and your test instructions already succeed on `main` (i.e. if I set my language to `en-IE` I _already_ see Euro's without this change). I think this is because PREMIUM_PLAN_COUNTRY_LANG_MAPPING already maps `ie,en` to `euro,en` and `at,de` to `euro,de`. (And if this _does_ fix it, I presume those mappings would need to be updated as well?)

I think that something may be going wrong in the location detection using the `X-Client-Region` header - possibly that contains different values than expected in Ireland and Austria?

groovecoder:
I'm afraid that might be the case. If that's the case, we can't do anything about it. GCP sends us the value in `X-Client-Region` so if we're receiving inaccurate country code there, we can't do anything about it. We may have to close this out as "CANTFIX" ?

Vinnl:
Yeah maybe, although the things that make me hesitant:

- Supposedly VPN does show the correct currencies, which I think uses the same locale detection?
- The plan IDs I think _are_ correct, given that the correct currencies supposedly _are_ shown when people go into the subscription flow?
- Although I'm also wondering if they're not seeing those from an invalid plan ID that happens to use the correct currency, in which case I'm wondering if people in Ireland and Austria can even purchase Premium in the first place? Can we check anywhere whether we have customers there?

I guess I should be getting myself some VPN so I can verify this myself too. In any case, I guess we might want to discuss impact with Tony as well.

groovecoder:
Since the country-code value will trump everything else, I'm going to merge this to see if it fixes stage. We can revert if it doesn't or if it causes other problems. But can't really test this "IRL" until it's on stage.

PR #1433, opened 2021-12-16
new delete_old_reply_records command

groovecoder:
Deletes all Reply objects older than <days_old>

usage: python manage.py delete_old_reply_record <days_old>
e.g., python manage.py dlete_old_reply_record 30

This PR fixes https://mozilla-hub.atlassian.net/browse/MPP-1079.

How to test:

1. Run `python manage.py delete_old_reply_records 30`
   * [ ] Verify it deletes all Reply records with `created_at` dates older than 30 days


emails/management/commands/delete_old_reply_records.py
```diff
@@ -0,0 +1,23 @@
...
+    def handle(self, *args, **options):
+        delete_date = (
+            datetime.now(timezone.utc) - timedelta(options['days_old'][0])
+        )
+        replies_to_delete = (
+            Reply.objects.filter(created_at__lt=delete_date)
+            .only('id')
+        )
+        print(f'Deleting {len(replies_to_delete)} reply records')
+        replies_to_delete.delete()
```
say-yawn, line 23: Consider using `atomic` for possible race condition and/or adding logging here to surface errors to delete replies. If it's not deleted immediately it will be deleted the next time the script will run, likely the day after.

say-yawn (changes requested):
Change requested. I also wonder if we should add a snippet on the header or the footer of the email Relayed on when is the last day they can reply to a relayed email.

emails/management/commands/delete_old_reply_records.py
groovecoder: `delete()` should run a bulk (i.e., single) `DELETE` statement, so I don't think `atomic` is required, but I'll include it for extra defensive coding.

Any exception should surface up into Sentry as usual, IIUC. @moz-astults - exceptions & errors during cron jobs still pop up in Sentry, right?

moz-astults: @groovecoder I'm not sure if you need to wrap the exception with the sentry_sdk [like this](https://github.com/mozilla/fx-private-relay/blob/8d5194841a971a98e376dc836dee049b51827aa7/privaterelay/views.py#L242) or not, but if there's some way it handles exceptions globally then I think it should work. The `SENTRY_DNS` env var will be in place on the cron job.

groovecoder: Yup, we call `sentry_sdk.init` from `privaterelay.settings` module which is run for management commands like this one. So Sentry should receive any errors during this command.

say-yawn (approved):
LGTM

PR #1439, opened 2021-12-21
add X-Detected-Client-Region to response headers

groovecoder:
# New feature description
Add a new `X-Detected-Client-Region` response header to help debug which country is detected in the `X-Client-Region` header passed from load balancer to app.

# How to test
1. Open dev tool network inspector
2. Go to a Relay page
3. Right-click the request in the network inspector, and choose 'Edit and Resend"
4. In the request headers, add a line like `X-Client-Region: ie`
5. Send the request
6. Click the new request in the network inspector
   * [ ] Check that the "Response Headers" section shows the same value in the `X-Detected-Client-Region` header

# Checklist

### DIFF
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -50,6 +50,19 @@ class FxAToRequest:
         return self.get_response(request)
 
 
+class AddDetectedCountryToResponseHeaders:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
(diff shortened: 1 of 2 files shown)


Vinnl (approved):
LGTM and works here. Just wondering whether we should maybe abstain from lowercasing it, given that this is mostly to aid debugging? Just in case the casing might affect the code's behaviour.

groovecoder:
Good idea. I removed the `lower()` call.

PR #1442, opened 2021-12-27
Don't inline lightbulb image

Vinnl:
Apparently Gmail strips data URIs, so the lightbulb image isn't visible there without this change: https://stackoverflow.com/a/8580387

How to test: I think this is a straightforward enough change that we can just deploy this to stage, send an email to an alias for a Premium account there, and verify that the image shows up in Gmail. Otherwise, render the email locally and verify that you can see the image and that its `src` points to your local server.

### DIFF
--- a/emails/templates/emails/wrapped_email.html
+++ b/emails/templates/emails/wrapped_email.html
@@ -59,7 +59,7 @@
                   {% endif %}
                   {% if recipient_profile.has_premium %}
                     <br />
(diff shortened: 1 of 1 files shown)


emails/templates/emails/wrapped_email.html
```diff
@@ -59,7 +59,7 @@
                   {% endif %}
                   {% if recipient_profile.has_premium %}
                     <br />
-                    <img alt="" width="24" style="display: inline-block; margin-bottom: 0px; width: 24px;"  src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAYAAADgdz34AAAACXBIWXMAAA7DAAAOwwHHb6hkAAAA GXRFWHRTb2Z0d2FyZQB3d3cuaW5rc2NhcGUub3Jnm+48GgAAAX1JREFUSIntlLEvQ2EUxX/3a+uN JcZuXRm6EIn4AyTWSqh0a5sOIozCPyFEWhual7B2EQtRQ1kk5tpsTS0MbT3X4BHE1+cZMPRMN985 95x77/BBH38NCRJsTKvjDJNDWQBGARSuRagMtSnPHkrnxwHbWU0YjyqQskiuniLMFHfl1uZhek3+ Zi40RJmLeiSiHgkV5oEbIGU8qgdpHbD5RG2Ef5YUQgOPibwrzXe0W5rTYwx1IHXnkAc2Q23g3xyF 9cJHcwAKrjQVVgFUydhs7AEwAhB75NQmcGIcAYivDRvwANCNMGgTdNsk/PL+JwEnAEbI2gRqXjjx taECjLAFoMrK9ryOf+b9t2VAPV60oQJy+3KmsAPEjOCW0hp/5UppjRvBBWJAuViR89ABAJ0Wi8Al kFSHtdd3v04KXLRbLPXyCPwqyhkdU7iwjDdR2JN6r/6eGwDkK3KJUPtitFqQ+a8g8EQA5YzWFCY/ ddYK+zIV1Bt4IgAV9Du6Pv4nngE8HXRqmKzo6wAAAABJRU5ErkJggg==" />
+                    <img alt="" width="24" style="display: inline-block; margin-bottom: 0px; width: 24px;"  src="{{ SITE_ORIGIN }}/static/images/email-images/tip-purple.png" />
```
maxxcrawford, line 62: Should we add content to the empty `alt` tag?

```suggestion
                    <img alt="Lightbulb" width="24" style="display: inline-block; margin-bottom: 0px; width: 24px;"  src="{{ SITE_ORIGIN }}/static/images/email-images/tip-purple.png" />
```

maxxcrawford (approved):
I've not seen the [email-specific CSS guide](https://www.campaignmonitor.com/css/) before. Great find! 

I dig this fix. It's straight forward. ~Are there any concerns linking directly to an image here? Is there a better/more optimized location for email specific assets?~ Disregard that statement. I see that `relay-logo-emails.png` and `smile-purple.png` are already linked this way. Proceed!

emails/templates/emails/wrapped_email.html
Vinnl: I would say no:

>  Setting this attribute to an empty string (alt="") indicates that this image is not a key part of the content (it’s decoration or a tracking pixel), and that non-visual browsers may omit it from rendering. Visual browsers will also hide the broken image icon if the alt is empty and the image failed to display.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#attr-alt

Imagine someone using a screenreader reading this email, and the screenreader saying "Lightbulb": that doesn't actually convey any extra information. I could go with something like "Tip:", but I don't think that adds that much value: I think the lightbulb is primarily decorative.

Vinnl:
Haha, I didn't even click through to that guide - but yes, that is a pretty great resource, thanks for highlighting it!

PR #1449, opened 2022-01-05
Revert "Remove the `premiumEnabled` feature flag"

Vinnl:
Reverts mozilla/fx-private-relay#1428 - sorry, should have merged https://github.com/mozilla/fx-private-relay-add-on/pull/246 first. Merging this, then re-creating the original PR.


Vinnl:
Oh haha no, I'm losing track after the holidays :stuck_out_tongue: We changed the original PR so it would no longer break, which is why there were no "Do not merge" flags I'd have missed.

PR #1455, opened 2022-01-07
Handle errors for dlq cronjob mpp 1533

say-yawn:
This PR fixes [MPP-1533](https://mozilla-hub.atlassian.net/browse/MPP-1533). It properly returns a `400 Bad Request` when the `to`, `bcc`, or `cc` field has malformed email address (does not have have `@`) AND handles any exception raised during DLQ cron job by logging the exception and deleting the message so that it does not linger in the queue.

How to test:
Test on stage/prod since we need malformed emails to be queued in DLQ and there are currently no good ways to trigger malformed emails to be sent.

#
Resolves Sentry Error [13518576](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/13518576/activity/?query=is%3Aunresolved%20valueerror)
Resolves Sentry Error [14874921](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/14874921/activity/?query=is%3Aunresolved%20valueerror)
Resolves [MPP-1533](https://mozilla-hub.atlassian.net/browse/MPP-1533)

### DIFF
--- a/emails/management/commands/process_delayed_emails_from_sqs.py
+++ b/emails/management/commands/process_delayed_emails_from_sqs.py
@@ -29,7 +29,6 @@ def _verify_and_run_sns_inbound_on_message(message):
     try:
         _sns_inbound_logic(topic_arn, message_type, verified_json_body)
         info_logger.info(f'processed sqs message ID: {message.message_id}')
-        message.delete()
     except ClientError as e:
         incr_if_enabled('rerun_message_from_sqs_error', 1)
         logger.error('sqs_client_error: ', extra=e.response['Error'])
@@ -45,7 +44,6 @@ def _verify_and_run_sns_inbound_on_message(message):
             try:
                 _sns_inbound_logic(topic_arn, message_type, verified_json_body)
                 info_logger.info(f'processed sqs message ID: {message.message_id}')
-                message.delete()
             except ClientError as e:
                 incr_if_enabled('rerun_message_from_sqs_error', 1)
                 logger.error('sqs_client_error: ', extra=e.response['Error'])
@@ -74,6 +72,7 @@ class Command(BaseCommand):
                 except:
                     exc_type, _, _ = sys.exc_info()
                     logger.exception(f'dlq_processing_error_{exc_type}')
+                finally:
                     message.delete()
             messages = dl_queue.receive_messages(
                 MaxNumberOfMessages=10, WaitTimeSeconds=1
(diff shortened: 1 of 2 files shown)


emails/management/commands/process_delayed_emails_from_sqs.py
```diff
@@ -69,8 +69,12 @@ def handle(self, *args, **options):
         )
         while len(messages) > 0:
             for message in messages:
-                _verify_and_run_sns_inbound_on_message(message)
-
+                try:
+                    _verify_and_run_sns_inbound_on_message(message)
+                except:
+                    exc_type, _, _ = sys.exc_info()
+                    logger.exception(f'dlq_processing_error_{exc_type}')
```
say-yawn, line 76: Will log the exception type, value, and traceback, see this [Sentry doc](https://docs.sentry.io/platforms/python/guides/logging/#usage) for more info.

groovecoder: When we see these in Sentry, let's make sure these exceptions don't accidentally store any PII that we don't want to be storing.

emails/management/commands/process_delayed_emails_from_sqs.py
```diff
@@ -69,8 +69,12 @@ def handle(self, *args, **options):
...
+                    message.delete()
```
groovecoder, line 77: I wonder if it would be best to use `try...except...finally` here? So remove the 2 calls to `message.delete` that are inside `_verify_and_run_sns_inbound_on_message`, move this call into a `finally:` block?

groovecoder (changes requested):

emails/management/commands/process_delayed_emails_from_sqs.py
say-yawn: I thought about this too. Hopefully we can test it on Stage but if not I'll watch out for it the first time we run on Prod. But I believe this won't happen because we disabled the breadcrumbs which is where the PII shows up.

emails/management/commands/process_delayed_emails_from_sqs.py
say-yawn: Sounds good. That will make it more clear that deleting will happen irregardless.

say-yawn: Done on 54ca46b

groovecoder (approved):

PR #1457, opened 2022-01-10
fix MPP-1512: send ga event for any premium user with "clicked-purchase"

groovecoder:
FXA sends the subscription state change event directly to our back-end
as soon as the user completes a purchase, and we process those events as
soon as we get them.

So, our check for "newly_premium" never evaluated to true, because the
event had already updated the profile to premium by the time the user
got back to the profile page.

So instead, always send the "user_purchased_premium" event to GA when a
user with a "clicked-purchase" cookie lands back on the profile page.

Additionally, move the code that updates date_subscribed to the
event-handling code.

This PR fixes [MPP-1512](https://mozilla-hub.atlassian.net/browse/MPP-1512).

## How to test:
0. Make sure you're not running something that blocks Google Analytics
1. Go to local relay dashboard as a free user
2. Open the developer tools "Console" tab
3. Click one of the "upgrade to relay premium"
   * [x] In the console, you should see an event sent to GA: Outbound Link, Click, Upgrade to Relay Premium
4. In the new tab that opened, open the developer tools "Console" tab
5. Go all the way thru the purchase flow in the new tab
6. Click the "Get Started" button to get back to Relay
   * [x] In the console, you should see an event sent to GA: server event, fired, user pruchased premium
7. Clear the console output to make the next step easier
8. Refresh the Relay dashboard page
   * [x] In the console, **YOU SHOULD NOT SEE AN EVENT** sent to GA: server event, fired, user pruchased premium. This makes sure we're not double-counting conversions.

### DIFF
--- a/privaterelay/tests/views_tests.py
+++ b/privaterelay/tests/views_tests.py
@@ -1,3 +1,5 @@
+import json
+
 from django.contrib.auth.models import User
 from django.test import TestCase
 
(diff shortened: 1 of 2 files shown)


maxxcrawford (approved):
Both events fired as expected through the flow and did not fire again after a refresh! 🚀

PR #1437, opened 2021-12-20
Domain address POST API logic mpp 1462

say-yawn:
# New feature description
POST to `/api/v1/domainaddresses/` should create a new domain address if the user is a valid premium subscriber to premium Relay.

# How to test
## POST to `/api/v1/domainaddresses/` does NOT create a new domain address for the user WITHOUT a valid subscription to premium Relay.
1. Create a test account without a subscription to premium Relay
2. Go to `/api/v1/docs/` and make a similar request like the following:
```
curl -X POST "http://127.0.0.1:8000/api/v1/domainaddresses/" -H  "accept: application/json" -H  "Content-Type: application/json" -H  "X-CSRFToken: csrftoken" -d "{  \"enabled\": true,  \"description\": \"test\",  \"address\": \"foo\"}"
```
3. Check that you get a `403` response with the body that looks like:
```
{
  "detail": "You must be a premium subscriber to create subdomain aliases."
}
```

## POST to `/api/v1/domainaddresses/` creates a new domain address for the user with a valid premium subscriber to premium Relay.
1. For the test account, purchase the premium Relay subscription
2. Go to `/api/v1/docs/` and make a similar request like the following:
```
curl -X POST "http://127.0.0.1:8000/api/v1/domainaddresses/" -H  "accept: application/json" -H  "Content-Type: application/json" -H  "X-CSRFToken: csrftoken" -d "{  \"enabled\": true,  \"description\": \"test\",  \"address\": \"foo\"}"
```
3. Check that you get a `201` response with the body that looks like:
```
{
  "enabled": true,
  "description": "test",
  "id": 11,
  "address": "string",
  "domain": 2,
  "full_address": "foo@subdomain.mozmail.com",
  "created_at": "2021-12-20T16:45:02.149999Z",
  "last_modified_at": "2021-12-20T16:45:02.150027Z",
  "last_used_at": null,
  "num_forwarded": 0,
  "num_blocked": 0,
  "num_spam": 0
}
```

## POST to `/api/v1/domainaddresses/` does NOT create a new domain address for the user with a valid premium subscriber to premium Relay when the domain address already exists.
1. Use the same test account that created a subdomain with the local portion `foo`
2. Go to `/api/v1/docs/` and make a similar request like the following:
```
[...]

PR #1464, opened 2022-01-12
fix #1153: use spamVerdict to automatically block spam

### LINKED ISSUE #1153: outlook.com classified a forwarded from relay.firefox.com email as spam
1. go https://store.irobot.com/default/register
2. use email alias generated by Relay and redirecting to name@outlook.com
3. Relay reports 2 emails forwarded and that there was a `soft "bounce" from your email provider`
4. outlook.com shows only 1 email in spam folder, classified as spam most likely because of the domain

groovecoder:
If spam is sent to and thru a Relay alias, MTAs sometimes send bounces
back to Relay with a message like: "Message considered as spam ..."

So, when we receive a bounce for a user and the bounce details includes
the term 'spam', instead of setting a bounce pause on the account,
set a new `auto_block_spam` field.

Then, when we receive any email for a user with `auto_block_spam`, and the
SES `spamVerdict` status is `FAIL`, don’t forward the email.

This should automatically stop us from forwarding any more spam to a
user after their MTA has bounced a first spam message back.

This PR fixes #1153.

## How to test:

Regression testing:  Check that emails are still delivering as expected. May need to do this on dev server.


say-yawn (changes requested):
One change requested and needs migration file for `Profile` changes that adds the `auto_block_spam` field

emails/views.py
```diff
@@ -656,13 +666,19 @@ def _handle_bounce(message_json):
...
         incr_if_enabled(
             'email_bounce_%s_%s' % (
                 bounce.get('bounceType'), bounce.get('bounceSubType')
             ),
             1
         )
+        if 'spam' in recipient.lower():
+            profile.auto_block_spam = True
+            profile.save()
+            return HttpResponse("OK", status=200)
```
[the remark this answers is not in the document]
groovecoder: Good catch. Fixed in https://github.com/mozilla/fx-private-relay/pull/1464/commits/6fd0559e5d2bdad132e138933fb5c022ad1cb2e5.

groovecoder:
I updated the testing code around this so I could test the new `auto_block_spam` logic during both the bounce-handling and the email-handling. I also ended up refactoring some to make everything more test-able, and I added a few more fixtures we can use to expand on the email-processing testing in the future too.

emails/migrations/0038_profile_auto_block_spam.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.24 on 2022-01-11 21:15
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('emails', '0037_reply_add_index_on_created_at'),
```
groovecoder, line 9: I'll update this and the migration number + name after we merge https://github.com/mozilla/fx-private-relay/pull/1437.

emails/models.py
```diff
@@ -32,13 +32,18 @@
...
+            'RELAY_FIREFOX_DOMAIN': 'default.com',
+            'MOZMAIL_DOMAIN': 'test.com'
+        }
     return {
         'RELAY_FIREFOX_DOMAIN': settings.RELAY_FIREFOX_DOMAIN,
         'MOZMAIL_DOMAIN': settings.MOZMAIL_DOMAIN
     }
 
 
-DOMAINS = get_domains_from_settings()
```
groovecoder, line 41: Testing this module-level variable was really nasty, so I removed it and replaced all uses with `get_domains_from_settings` instead. I also made `get_domains_from_settings` return mock/test values while in test mode.

emails/tests/views_tests.py
```diff
@@ -1,83 +1,126 @@
...
 from emails.models import (
     address_hash,
     DeletedAddress,
+    Profile,
     RelayAddress,
 )
-from emails.views import _get_address
+from emails.views import _get_address, _sns_notification
 
-TEST_DOMAINS = {'RELAY_FIREFOX_DOMAIN': 'default.com', 'MOZMAIL_DOMAIN': 'test.com'}
```
groovecoder, line 24: Removed this in favor of the new `get_domains_from_settings` function that returns these test values while in test mode.

say-yawn (changes requested):
LGTM but waiting on merging of #1437 so you can update your migration. Really great to see new tests and glad we are making sure we respect the user's email config and increasing deliverability of emails.

emails/models.py
say-yawn: I was unsure about using conditionals for tests in feature code but I read some docs talked with @jwhitlock and this is a good approach. When he joins our team I hope we can talk about how we want to structure our test framework or our code so that code is more clear/concise and test conditionals can live in the test code.

emails/views.py
```diff
@@ -663,6 +673,12 @@ def _handle_bounce(message_json):
             ),
             1
         )
+        # if an email bounced as spam, set to auto block spam for this user
+        # and DON'T set them into bounce pause state
+        if (any('spam' in val.lower() for val in recipient.values())):
+            profile.auto_block_spam = True
+            profile.save()
+            continue
```
say-yawn, line 681: We record bounce on the user attribute `last_soft_bounce` if it’s user’s email server responding that it is not receiving emails right now (bounced). But when we get a bounce from the email server with the reason being that we are relaying spam, we instead record `auto_block_spam`. This will help us continue to relay emails reliably without being marked as a spammer or putting users relay account on pause.

groovecoder:
Merged https://github.com/mozilla/fx-private-relay/pull/1437, rebased, and updated the migration number + name here to depend on that new one.

say-yawn (approved):
LGTM

PR #1468, opened 2022-01-14
Create codeql-analysis.yml

groovecoder:


groovecoder:
Okay, I like this enough to leave it on our PRs until something comes up that makes it too noisy or annoying with false positives. It helped me find a couple places where we could add some additional defensive coding:

* add `noopener noreferrer` to our email wrapper links
* html- and shell-escape input that comes from outside the app

In all cases, the alerts here were not real vulnerabilities, but I like tools-that-teach-secure coding, so I like this. 😀

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
```
Vinnl, line 16: Any reason not to run this for all branches? Would be nice to catch issues _before_ they get merged into `main`.

```suggestion
    branches: ['*']
```

Vinnl (approved):
LGTM.

PR #1478, opened 2022-01-17
Back-end changes to allow website feature-parity through the API

Vinnl:
# New feature description

To allow the entire website to be generated based on data from the API, a couple of changes were needed. This PR includes those changes. Specifically, it:

- exposes the user avatar as read-only (I added a new property to the Profile model, since this comes from FxA).
- exposes whether the user has Premium as read-only (`has_premium`).
- exposes data used in the bounce warning as read-only (`next_email_try` and `bounce_status`).
- exposes the user's API token as read-only - this can be removed again once the add-on has its own OAuth flow.
- makes the subdomain writable (*note:* there's a bit of validation in `add_subdomain` that I think does not get called now - can/should that be added as a validator on the field directly, or should it be called from django_rest_framework somehow?).
- makes `accounts/profile/subdomain` callable as an API (so a local devserver hosting the front-end can call it).
- adds an API endpoint for `Users` and exposes the user's email through the API.
- adds an endpoint `/accounts/profile/refresh` that refetches data from FXA (i.e. called when `?fxa_refresh=1`).
- adds CORS headers when running with `DEBUG` that allows http://localhost:3000 to make API calls.

# How to test

I've pushed a rebased version of the branch `react-experiment` that builds on top of this one. It has an updated README that demonstrates how to run the front-end (essentially: `cd frontend; npm install; npm run watch` in addition to running the backend).

# Checklist

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.db import IntegrityError
+from django.contrib.auth.models import User
 
 from drf_yasg.views import get_schema_view
 from drf_yasg import openapi
@@ -15,7 +16,7 @@ from privaterelay.utils import get_premium_countries_info_from_request
 
 from .permissions import IsOwner
 from .serializers import (
-    DomainAddressSerializer, ProfileSerializer, RelayAddressSerializer
+    DomainAddressSerializer, ProfileSerializer, RelayAddressSerializer, UserSerializer
 )
 from .exceptions import ConflictError
 
@@ -71,6 +72,13 @@ class ProfileViewSet(viewsets.ModelViewSet):
 
     def get_queryset(self):
         return Profile.objects.filter(user=self.request.user)
+
+class UserViewSet(viewsets.ModelViewSet):
+    serializer_class = UserSerializer
+    permission_classes = [permissions.IsAuthenticated, IsOwner]
+
+    def get_queryset(self):
+        return User.objects.filter(id=self.request.user.id)
 
 @decorators.api_view()
 @decorators.permission_classes([permissions.AllowAny])
--- a/api/views.py
+++ b/api/views.py
@@ -76,6 +76,7 @@ class ProfileViewSet(viewsets.ModelViewSet):
 class UserViewSet(viewsets.ModelViewSet):
     serializer_class = UserSerializer
     permission_classes = [permissions.IsAuthenticated, IsOwner]
+    http_method_names = ['get', 'head']
 
     def get_queryset(self):
         return User.objects.filter(id=self.request.user.id)
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -150,6 +150,7 @@ INSTALLED_APPS = [
 
     'rest_framework',
     'rest_framework.authtoken',
+    'corsheaders',
 
     'privaterelay.apps.PrivateRelayConfig',
 ]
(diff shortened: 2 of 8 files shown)


api/views.py
```diff
@@ -54,6 +55,13 @@ class ProfileViewSet(viewsets.ModelViewSet):
     def get_queryset(self):
         return Profile.objects.filter(user=self.request.user)
 
+class UserViewSet(viewsets.ModelViewSet):
+    serializer_class = UserSerializer
+    permission_classes = [permissions.IsAuthenticated, IsOwner]
```
groovecoder, line 60: We should probably include `http_method_names` like we do in the `ProfileViewSet`, so clients can't delete user objects without going thru site UI.
```
http_method_names = ['get', 'post', 'head', 'put', 'patch']
```

privaterelay/settings.py
```diff
@@ -547,6 +549,22 @@ def _get_initial_middleware():
...
+    # origin and thus has access to the same cookies.
+    CORS_ALLOW_CREDENTIALS = True
+    SESSION_COOKIE_SAMESITE = None
+    CORS_ALLOWED_ORIGINS += [
+        'http://localhost:3000',
+    ]
+    CSRF_TRUSTED_ORIGINS += [
+        'http://localhost:3000',
+    ]
+
```
groovecoder, line 567: I think this is fine, but I might want a bit of time to think about how bad the failure mode is here. E.g., if `DEBUG` is somehow set to true in prod, what additional vulnerabilities does this open up?

privaterelay/views.py
```diff
@@ -133,6 +134,23 @@ def _get_fxa(request):
     return request.user.socialaccount_set.filter(provider='fxa').first()
 
 
+@api_view()
+@require_http_methods(['GET'])
+def profile_refresh(request):
```
groovecoder, line 139: Hmm ... what do we need this view for? What user flow?

groovecoder (changes requested):
Overall looks good and feels exciting. Few things I want to be sure about as we open more API surface though.

api/views.py
Vinnl: Oh hmm, I wonder why I didn't include that in the first place. Fixed in df47f8a2.

privaterelay/settings.py
Vinnl: Yeah that sounds good. I think the main failure mode is DEBUG being set to true and a malicious app on the user's machine runs a server on localhost:3000. If they obtain the user's API token, they will then be able to act on behalf of that user. (Alternative attack method would be to insert itself into the browser, of course.)

Is there anyone with more domain knowledge we could ask here for a second opinion?

privaterelay/views.py
Vinnl: The equivalent of this is currently run when server the `/accounts/profile` view, so it's the equivalent of

https://github.com/mozilla/fx-private-relay/blob/df47f8a217ddc1cf6b057bb8ac7139664976dd2f/privaterelay/views.py#L79-L81

and

https://github.com/mozilla/fx-private-relay/blob/df47f8a217ddc1cf6b057bb8ac7139664976dd2f/privaterelay/views.py#L105-L109

privaterelay/settings.py
groovecoder: At that level of ask it would probably involve a consulting fee. 😄

groovecoder: This sounds like the default port for all React apps during development, and those apps are likely running without security protections? So, I'm worried about the React developer who is also a Relay user, and an attacker crafts a malicious request at their http://localhost:3000 site AND we're running in `DEBUG=True` mode.

Seems like a low-risk scenario.

Vinnl:
Conflict resolved!

groovecoder (approved):

PR #1479, opened 2022-01-17
Add Accordion Anchor Links

codemist:
This PR addresses MPP-1509.

This PR generates an anchor link for each use case in the accordion. When the window hash changes, the corresponding use case element will get triggered. Note that the hash change should also anchor to the use case section.

Preview (updated Jan 24)

https://user-images.githubusercontent.com/13066134/150854901-69329aaa-e52d-4887-964c-b03690ec6584.mov

How to test:

# New feature description
This PR generates an anchor link for each use case in the accordion. When the window hash changes, the corresponding use case element will get triggered.

# How to test

1) Go to the use case section, select any of them and check for a hash change. 
2) Go to `http://127.0.0.1:8000/#use-cases/shopping`, and the page should scroll to the use case section. While on the 'shopping' tab, go to `http://127.0.0.1:8000/#use-cases/gaming` and check that the active use case changes to 'gaming'.

# Checklist

### DIFF
--- a/static/js/app.js
+++ b/static/js/app.js
@@ -686,55 +686,56 @@ if ( document.getElementById("survey-dismiss") ) {
 }
 
 
-//Landing page use case accordion 
+//Landing page use case accordion
+const useCaseSection = document.getElementById("use-cases");
 const useCaseTitle = document.querySelectorAll(".use-case-title");
 
 function toggleClass(elem) {
    useCaseTitle.forEach( item => {
       item.classList.remove("is-active");
   });
+  useCaseSection.scrollIntoView();
   elem.target.classList.add("is-active");
   window.location.hash = "#use-cases" + "/" + elem.target.dataset.useCase;
 }
 
-function jumpToUseCase(){
-  const url = location.hash;
-  if (url.includes("#use-cases")) {
-    window.location.hash = "#use-cases";
-  }
-}
-
 function hashChangeAccordion(){
-  if (location.hash === "#use-cases/shopping"){
-    useCaseTitle.forEach( item => {
-      item.classList.remove("is-active");
-    });
-    document.querySelector(".use-case-shopping").classList.add("is-active");
+  const urlHash = location.hash;
+  if (urlHash.includes("#use-cases")){
+    useCaseSection.scrollIntoView();
+    
+    if (urlHash === "#use-cases/shopping"){
+      useCaseTitle.forEach( item => {
+        item.classList.remove("is-active");
+      });
+      document.querySelector(".use-case-shopping").classList.add("is-active");
+    }
+  
+    if (urlHash === "#use-cases/social-networks"){
+      useCaseTitle.forEach( item => {
+        item.classList.remove("is-active");
+      });
+      document.querySelector(".use-case-social-networks").classList.add("is-active");
+    }
+  
+    if (urlHash === "#use-cases/offline"){
+      useCaseTitle.forEach( item => {
+        item.classList.remove("is-active");
+      });
+      document.querySelector(".use-case-offline").classList.add("is-active");
+    }
+  
+    if (urlHash === "#use-cases/gaming"){
+      useCaseTitle.forEach( item => {
+        item.classList.remove("is-active");
+      });
+      document.querySelector(".use-case-gaming").classList.add("is-active");
+    }
   }
-
-  if (location.hash === "#use-cases/social-networks"){
-    useCaseTitle.forEach( item => {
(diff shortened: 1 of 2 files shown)


static/js/app.js
```diff
@@ -694,8 +694,54 @@ function toggleClass(elem) {
...
+  const accordionHash = window.location.hash;
+
+  if (accordionHash === ""){
+    window.location.href = urlRelay.concat("#" + accordionClassArray[1]);
+  }
+
+  else {
+    window.location.href = urlRelay.replace(accordionHash, "#" + accordionClassArray[1]);
+  }
 }
```
maxxcrawford, line 711: You can drop all the manipulation of `window.location.href` and manually adding the `#`. By setting a string as the `location.hash`, you get that functionality for free! 

```suggestion
  window.location.hash = accordionClassArray[1];
}
```

Additionally, how you're querying the class/hash ID (the second class name in the array) is really fragile. I would propose moving (or cloning) the class names in a `data-use-case` element in the [`use-cases.html`](https://github.com/mozilla/fx-private-relay/blob/anchor-links-accordion/privaterelay/templates/newlanding/a/use-cases.html) file. That way you're being more exact.

static/js/app.js
```diff
@@ -694,8 +694,54 @@ function toggleClass(elem) {
...
+    });
+    document.querySelector(".use-case-offline").classList.add("is-active");
+  }
+
+  if (location.hash === "#use-case-gaming"){
+    useCaseTitle.forEach( item => {
+      item.classList.remove("is-active");
+    });
+    document.querySelector(".use-case-gaming").classList.add("is-active");
+  }
```
maxxcrawford, line 740: This can be migrated to a switch case instead a stack of `if` statements. 

```
 switch (location.hash) {
    case "#use-case-shopping":
      useCaseTitle.forEach( item => {
        item.classList.remove("is-active");
      });
      document.querySelector(".use-case-shopping").classList.add("is-active");
      break;
    …
  }
```

maxxcrawford (changes requested):
Thanks for diving into this issue. It's one of those tasks where there's many different ways to a solution. I called out a few issues but there's one missing feature — if a hash is present in the URL, I expect the page to scroll down to the section, open to that section. That is currently not working for me. Should it be?

static/js/app.js
Vinnl: As a counterpoint to that (to make things more confusing :sweat_smile: ), I think I'm generally partial to a stack of `if` s tatements rather than `switch`, as there's no risk of forgetting a `break`, and the consistent structure makes it otherwise just as easy to scan :)

codemist: @maxxcrawford @Vinnl Thank you both for your two cents...


I guess now I'm at a loss about which method to go with 🥴

static/js/app.js
codemist: Thanks for the suggestion, I also felt that it was a fragile approach. I looked into using a data attribute and used that method instead!

static/js/app.js
```diff
@@ -694,8 +694,43 @@ function toggleClass(elem) {
       item.classList.remove("is-active");
   });
   elem.target.classList.add("is-active");
+  window.location.hash = elem.target.dataset.useCase;
```
codemist, line 697: Opted for a `dataset` property over reading the element's `classList`.

static/js/app.js
codemist: I'm keeping the `if` statements in for the time being

static/js/app.js
```diff
@@ -686,16 +686,59 @@ if ( document.getElementById("survey-dismiss") ) {
...
+function hashChangeAccordion(){
+  const urlHash = location.hash;
+  if (urlHash.includes("#use-cases")){
+    useCaseSection.scrollIntoView();
+    
+    if (urlHash === "#use-cases/shopping"){
+      useCaseTitle.forEach( item => {
+        item.classList.remove("is-active");
+      });
+      document.querySelector(".use-case-shopping").classList.add("is-active");
```
maxxcrawford, line 711: nit: Since we're repeating this so much, I recommend moving it to its own function — that way we can change it once and update each if. 

You could even pass in the class name below to clean that up too — 

```suggestion
      useCaseTitle.resetAccordianAndSetActiveSection(".use-case-shopping")
```

maxxcrawford (approved):
Good to go! I added one nit/optimization nit since you didn't opt for that [`switch` case](https://media.giphy.com/media/YFVCgwTRr0W1a/giphy.gif).

PR #1496, opened 2022-01-25
Add `.` in between subdomain and domain

Vinnl:
This fix doesn't _feel_ like the best approach, but the "best" approach involves adding a new string with a `.` added, and having all translators do the same. This approach bothers far fewer people and works too, so...

How to test: Set "onboarding state" to 1 for a Premium user's profile at http://127.0.0.1:8000/admin/, then view the second onboarding page on the dashboard. 

image

Before:

image

After:

image


maxxcrawford (approved):
Thoughtful approach, @Vinnl. I think this is the quietest way to solve it. Good looking out!

PR #1513, opened 2022-01-31
Fix #1461 - Update UI of subdomain onboarding module

### LINKED ISSUE #1461: Mobile premium onboarding: On the second step, the user’s chosen and confirmed custom domain is displayed differently from the reference
#### Build:
- Stage;

#### Affected Platforms:
- Android 9, Android 11;

#### Browsers tested on:
- Firefox, latest version;

#### Prerequisites:
- Do not have a Relay account yet;

#### Steps to reproduce:
1. Go to the Stage Relay page;
2. Create a new Premium Relay account (Mozilla account) by signing up;
3. Verify the account and proceed to the onboarding;
4. Continue to the second step;
5. Choose a custom domain name and confirm it;
6. Observe the updated step;

#### Expected result:
- The updated step is displayed accordingly;

#### Actual result:
- The user’s chosen and confirmed custom domain is displayed differently from the reference, missing an “Email domain:” text and the “.mozmail.com” text on a separate row below;

#### Notes:
- See attached media:
Stage:
image

Reference:
image

codemist:
This PR fixes #1461.

# New feature description

This updates the design of the subdomain module on the premium onboarding flow for users who already have a pre-existing subdomain.

# Screenshot (if applicable)

Mobile view: 
image

Desktop view:
image

# How to test

Have an with a premium account with a subdomain set, on the inspect tool turn off the `display: none` property on `.c-multipart-premium-onboarding` and `.c-premium-onboarding-step`.

# Checklist

### DIFF
--- a/privaterelay/templates/includes/premium-onboarding.html
+++ b/privaterelay/templates/includes/premium-onboarding.html
@@ -59,8 +59,11 @@
                         {% endif %}
                         
                         <div class="c-premium-onboarding-action-completed {% if user_profile.subdomain %} is-visible {% endif %}">
-                            <img class="margin-right" src="/static/images/icon-check.svg" alt="">
-                            <samp class="js-premium-onboarding-domain-registration-preview">{{ user_profile.custom_domain }}</samp>
+                            <div class="c-premium-onboarding-subdomain-set-module">
+                                <p class="c-premium-onboarding-label">{% ftlmsg 'profile-label-domain' %}</p>
+                                <p class="c-premium-onboarding-subdomain">@{{ user_profile.subdomain }}</p>
+                                <span class="c-premium-onboarding-mozmail">.mozmail.com</span>
+                            </div>
                         </div>
                     </div>
                 </div>
--- a/privaterelay/templates/includes/premium-onboarding.html
+++ b/privaterelay/templates/includes/premium-onboarding.html
@@ -61,7 +61,7 @@
                         <div class="c-premium-onboarding-action-completed {% if user_profile.subdomain %} is-visible {% endif %}">
                             <div class="c-premium-onboarding-subdomain-set-module">
                                 <p class="c-premium-onboarding-label">{% ftlmsg 'profile-label-domain' %}</p>
-                                <p class="c-premium-onboarding-subdomain">@{{ user_profile.subdomain }}</p>
+                                <span>@</span><p class="c-premium-onboarding-subdomain js-premium-onboarding-domain-registration-preview">{{ user_profile.subdomain }}</p>
                                 <span class="c-premium-onboarding-mozmail">.mozmail.com</span>
                             </div>
                         </div>
--- a/privaterelay/templates/includes/premium-onboarding.html
+++ b/privaterelay/templates/includes/premium-onboarding.html
@@ -61,7 +61,7 @@
                         <div class="c-premium-onboarding-action-completed {% if user_profile.subdomain %} is-visible {% endif %}">
                             <div class="c-premium-onboarding-subdomain-set-module">
                                 <p class="c-premium-onboarding-label">{% ftlmsg 'profile-label-domain' %}</p>
-                                <span>@</span><p class="c-premium-onboarding-subdomain js-premium-onboarding-domain-registration-preview">{{ user_profile.subdomain }}</p>
+                                <p class="c-premium-onboarding-subdomain js-premium-onboarding-domain-registration-preview">{{ user_profile.subdomain }}</p>
                                 <span class="c-premium-onboarding-mozmail">.mozmail.com</span>
                             </div>
                         </div>
--- a/privaterelay/templates/includes/premium-onboarding.html
+++ b/privaterelay/templates/includes/premium-onboarding.html
@@ -40,8 +40,8 @@
                         {% if not user_profile.subdomain %}
                             <div class="js-premium-onboarding-domain-registration-form">
                                 <p>{% ftlmsg 'multi-part-onboarding-premium-domain-cta' %}</p>
-                                <samp>***@<span>{% ftlmsg 'banner-register-subdomain-example-address' %}</span>.mozmail.com</samp>
-                                <form id="onboardingDomainRegistration" class="c-premium-onboarding-domain-registration-form" method="post" action="{% url 'profile_subdomain' %}">
+                                <samp class="c-premium-onboarding-subdomain-registration">***@<span>{% ftlmsg 'banner-register-subdomain-example-address' %}</span>.mozmail.com</samp>
+                                <form id="onboardingDomainRegistration" method="post" action="{% url 'profile_subdomain' %}">
                                     {% csrf_token %}
                                     <input
                                     class="js-subdomain-value subdomain-banner-input"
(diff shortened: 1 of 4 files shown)


privaterelay/templates/includes/premium-onboarding.html
```diff
@@ -59,8 +59,11 @@ <h1>{% ftlmsg 'multi-part-onboarding-premium-get-domain' %}</h1>
                         {% endif %}
                         
                         <div class="c-premium-onboarding-action-completed {% if user_profile.subdomain %} is-visible {% endif %}">
-                            <img class="margin-right" src="/static/images/icon-check.svg" alt="">
-                            <samp class="js-premium-onboarding-domain-registration-preview">{{ user_profile.custom_domain }}</samp>
```
Vinnl, line 63: I couldn't test it because the removal of `js-premium-onboarding-domain-registration-preview` caused a JS error:

https://user-images.githubusercontent.com/4251/151981857-63bd4653-dcf3-420d-b356-fc981cc8182a.mp4

Vinnl (changes requested):
This appears to introduce a bug; see inline comment.

privaterelay/templates/includes/premium-onboarding.html
codemist: I removed instances of `domainPreview` and any unused parameters.

codemist:
Vincent's comment via Slack: 
So it might just be that that's no longer necessary? I haven't dived into that code too closely, but I think the previous preview worked by inserting the chosen domain through JavaScript (i.e. that line, which now breaks), whereas you're inserting the chosen domain directly in the back-end. If the latter replaces the former, then it might just be a matter of removing the former.

privaterelay/templates/includes/premium-onboarding.html
```diff
@@ -59,8 +59,11 @@ <h1>{% ftlmsg 'multi-part-onboarding-premium-get-domain' %}</h1>
...
+                            <div class="c-premium-onboarding-subdomain-set-module">
+                                <p class="c-premium-onboarding-label">{% ftlmsg 'profile-label-domain' %}</p>
+                                <p class="c-premium-onboarding-subdomain">@{{ user_profile.subdomain }}</p>
+                                <span class="c-premium-onboarding-mozmail">.mozmail.com</span>
+                            </div>
```
Vinnl, line 66: Ah, I'm running into another error, and I see why you're not running into it (and not in the other one either): they only occur when you claim a subdomain during the onboarding, not if you already have a subdomain before starting the onboarding. That's presumably also why there was JavaScript to set the subdomain in there, rather than just in the Django template.

This is what it looks like for me after registering a subdomain:

image

The reason for that is that at the time the back-end serves up this page, `user_profile.subdomain` is not set yet. And when the subdomain is registered, the page is not reloaded, so the contents of this element is still `@None.mozmail.com` when it is made visible.

Vinnl (commented):
Unfortunately I encountered another bug. I've described the cause inline, but you can reproduce it yourself if you register a subdomain _during_ the onboarding flow, rather than beforehand. Could you give that a try?

privaterelay/templates/includes/premium-onboarding.html
codemist: @Vinnl  Huh weird, wouldn't it show this:
image

if a user does not have a subdomain set yet?

codemist: Ohh! I understand now, thank you so much. I re-added the Javascript that I deleted as well as the `.js-premium-onboarding-domain-registration-preview' class into the HTML. Would this work now?

static/js/domain-registration.js
```diff
@@ -189,7 +189,7 @@
                 modalRegistrationSuccessState.classList.remove("is-hidden");
 
                 const domainPreview = document.querySelector(".js-premium-onboarding-domain-registration-preview");
-                domainPreview.textContent = domain + ".mozmail.com";
+                domainPreview.textContent = domain;
```
codemist, line 192: Removed ".mozmail.com" here since I'm adding that string directly into the HTML.

Vinnl (commented):
Works now! Only thing is that the `@` is placed on a separate line, and it look a bit weird on desktop view where its width seems to be restricted for no particular reason (though maybe that looks less weird if the `@` is not on a separate line):

image

privaterelay/templates/includes/premium-onboarding.html
```diff
@@ -59,8 +59,11 @@ <h1>{% ftlmsg 'multi-part-onboarding-premium-get-domain' %}</h1>
                         {% endif %}
                         
                         <div class="c-premium-onboarding-action-completed {% if user_profile.subdomain %} is-visible {% endif %}">
-                            <img class="margin-right" src="/static/images/icon-check.svg" alt="">
-                            <samp class="js-premium-onboarding-domain-registration-preview">{{ user_profile.custom_domain }}</samp>
+                            <div class="c-premium-onboarding-subdomain-set-module">
+                                <p class="c-premium-onboarding-label">{% ftlmsg 'profile-label-domain' %}</p>
+                                <span>@</span><p class="c-premium-onboarding-subdomain js-premium-onboarding-domain-registration-preview">{{ user_profile.subdomain }}</p>
```
maxxcrawford, line 64: Suggestion: If we're previewing the name, I recommend we stick with the [`samp` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/samp) like we're using else where.

static/scss/partials/onboarding.scss
```diff
@@ -393,29 +393,57 @@ firefox-private-relay-addon[data-addon-installed="true"] + firefox-private-relay
...
-    padding: $spacing-md;
+    padding: $spacing-lg;
     padding-inline-end: $spacing-lg;
     background-color: $color-white;
     display: none;
     align-items: center;
     border-radius: $border-radius-md;
-    width: auto;
-
-    samp {
```
maxxcrawford, line 404: Issue: If you switch the tag back to a `samp` tag, you may need to move some of this back.

static/scss/partials/onboarding.scss
```diff
@@ -393,29 +393,57 @@ firefox-private-relay-addon[data-addon-installed="true"] + firefox-private-relay
...
+        content: "";
+        background-image: url("/static/images/icon-check.svg");
+        background-repeat: no-repeat;
+        background-position: center center;
+        padding-right: $spacing-sm;
+    }
+
+    .c-premium-onboarding-subdomain {
+        @include text-body-lg;
+        font-weight: bold;
```
maxxcrawford, line 440: nitpick: Font-weights should be numeric. 
```suggestion
        font-weight: 700;
```

maxxcrawford: Also, egg on my face - I checked the code base for any other outliers. Could you resolve these while your making your updates?

- https://github.com/mozilla/fx-private-relay/blob/main/static/scss/partials/onboarding.scss#L51
- https://github.com/mozilla/fx-private-relay/blob/main/static/scss/partials/onboarding.scss#L98

maxxcrawford (changes requested):
Reviewed!

static/scss/partials/onboarding.scss
Vinnl: Oh, I didn't know that we did that! Should we see if we can update [Protocol's CSS coding guide](https://protocol.mozilla.org/docs/css-guide.html#units)? Or is that just because that's how we load our fonts?

privaterelay/templates/includes/premium-onboarding.html
codemist: Done!

static/scss/partials/onboarding.scss
maxxcrawford: > Oh, I didn't know that we did that! Should we see if we can update [Protocol's CSS coding guide](https://protocol.mozilla.org/docs/css-guide.html#units)? Or is that just because that's how we load our fonts?

We don't but we should. I'll update the README.

Vinnl (approved):

static/scss/partials/onboarding.scss
codemist: Switched back to a samp tag!

PR #1514, opened 2022-02-01
add FxaTokenAuthentication to REST API auth classes

groovecoder:
This PR addresses #MPP-1531.

# New feature description
This change adds a new `FxaTokenAuthentication` backend for the API.

# Screenshot
image

# How to test
* Check out this branch of the Relay server and `runserver`.
* Check out [this branch](https://github.com/mozilla/fx-private-relay-add-on/pull/268) of the add-on.
* `web-ext run -s src`
* Open a tab to `about:debugging`, and inspect the Relay add-on so you can see its Console
* Back on the add-on first run page, Sign in
   * [ ] You should be able to complete the OAuth flow with Stage FXA server
   * [ ] On the debugging Console, you should see the sequence shown in the screenshot above, showing that the Add-on was able to authorize its requests to the server with an FXA access token and receive the array of aliases.


api/tests/authentication_tests.py
```diff
@@ -0,0 +1,59 @@
+from allauth.socialaccount.models import SocialAccount
+from model_bakery import baker
+import responses
```
groovecoder, line 3: @say-yawn @jwhitlock how do you like [this `responses` library](https://github.com/getsentry/responses#basics) for mocking `requests`?

api/tests/authentication_tests.py
```diff
@@ -0,0 +1,89 @@
...
+            json={'active': True, 'sub': self.uid, 'exp': exp_time}
+        )
+
+        headers = {'HTTP_AUTHORIZATION': f'Bearer {user_token}'}
+        get_addresses_req = self.factory.get(self.path, **headers)
+        assert self.auth.authenticate(
+            self.auth, get_addresses_req
+        ) == (self.sa.user, None)
+        print(f'asserting {get_cache_key(user_token)} in {cache}')
+        assert cache.get(get_cache_key(user_token)) is not None
```
groovecoder, line 89: This is causing [the test failure](https://app.circleci.com/pipelines/github/mozilla/fx-private-relay/3929/workflows/a7247be0-9e12-4567-ab14-cb5d71800bc5/jobs/7623) that I don't understand. 😢 
```
>       assert cache.get(get_cache_key(user_token)) is not None
E       AssertionError: assert None is not None
E        +  where None = <bound method LocMemCache.get of <django.core.cache.backends.locmem.LocMemCache object at 0x7f0a05e3be50>>('fxa_token_user-123')
E        +    where <bound method LocMemCache.get of <django.core.cache.backends.locmem.LocMemCache object at 0x7f0a05e3be50>> = <django.core.cache.DefaultCacheProxy object at 0x7f0a088b26d0>.get
E        +    and   'fxa_token_user-123' = get_cache_key('user-123')
```

groovecoder:
Green checks all around again!

api/tests/authentication_tests.py
say-yawn: Oh interesting! You can validate the request data. Looks good to me. Should make testing on APIs be easier too. Was there a lot of places we mocked request? Wondering if we should file a ticket to replace all the old requests or if it's small enough to replace it in this pr.

docs/api_auth.md
```diff
@@ -0,0 +1,35 @@
...
+endpoint][fxa-oauth-token-verify].
+
+This auth scheme can be used by other clients too. E.g., Firefox browser has a
+[`getOAuthToken`][searchfox-getoauthtoken] function which can be used to
+perform the same API authentication as the add-on.
+
+![FXA OAuth Token Authentication Sequence Diagram](api_auth.png)
+
+### TODOs
+* Cache the token verification check so we don't hit FXA on every client request.
```
say-yawn, line 26: Did we not already do this [here](https://github.com/mozilla/fx-private-relay/pull/1514/files#diff-26e3f029d42ad868ed2e12764aa9e1263473a75b7d2125d3930de69e18665febR58)?

say-yawn (changes requested):
First pass

api/authentication.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class FxaTokenAuthentication(BaseAuthentication):
+    def authenticate(self, request):
+        if 'Authorization' not in request.headers:
+            return None
+
+        authorization = request.headers.get('Authorization')
+        if not authorization.startswith('Bearer '):
+            return None
```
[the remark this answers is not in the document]
groovecoder: I'm doing `return None` on purpose, so that DRF will try to authenticate the request with one of the other authentication backends. (per [docs for custom auth backends](https://www.django-rest-framework.org/api-guide/authentication/#custom-authentication).)

api/authentication.py
```diff
@@ -0,0 +1,67 @@
...
+                '%s/introspect' %
+                settings.SOCIALACCOUNT_PROVIDERS['fxa']['OAUTH_ENDPOINT']
+            )
+            fxa_resp = requests.post(introspect_token_url, json={'token': token})
+            if not fxa_resp.status_code == 200:
+                # cache anything besides 200 for 60s - it might be an error
+                # we need to re-try, but we don't want to send un-throttled
+                # retries at FXA
+                cache.set(cache_key, fxa_resp, 60)
+                return None
```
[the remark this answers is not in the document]
groovecoder: Same comment as above: I'm doing `return None` on purpose, so that DRF will try to authenticate the request with one of the other authentication backends. (per [docs for custom auth backends](https://www.django-rest-framework.org/api-guide/authentication/#custom-authentication).) The cache is defensive, so that an out-of-control or infinite-looping client doesn't trigger a full FXA request every time - only once ever 60s.

api/authentication.py
```diff
@@ -0,0 +1,67 @@
...
+            fxa_uid = fxa_resp_json.get('sub')
+            sa = SocialAccount.objects.get(uid=fxa_uid)
+
+            # cache fxa_resp_json for as long as access_token is valid
+            # Note: FXA iat and exp are timestamps in *milliseconds*
+            fxa_token_exp_time = int(fxa_resp_json.get('exp')/1000)
+            now_time = int(datetime.now().timestamp())
+            timeout = fxa_token_exp_time - now_time
+            cache.set(cache_key, fxa_resp, timeout)
+            return (sa.user, None)
```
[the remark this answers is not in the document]
groovecoder: Good call.

api/authentication.py
```diff
@@ -0,0 +1,66 @@
...
+            introspect_token_url = (
+                '%s/introspect' %
+                settings.SOCIALACCOUNT_PROVIDERS['fxa']['OAUTH_ENDPOINT']
+            )
+            fxa_resp = requests.post(introspect_token_url, json={'token': token})
+            if not fxa_resp.status_code == 200:
+                # cache anything besides 200 for 60s - it might be an error
+                # we need to re-try, but we don't want to send un-throttled
+                # retries at FXA
+                cache.set(cache_key, fxa_resp, 60)
```
jwhitlock, line 37: It looks to me like we're using this data from the FxA response:

* The status code - If not 200, then wait 60 seconds to try again
* The JSON payload - which is probably not present in some cases, like a 5xx error

I think we should be caching just that data, rather than the whole `Response` object (which includes the request, the response headers, timing info, etc.)

api/authentication.py
```diff
@@ -0,0 +1,66 @@
...
+                return None
+
+        fxa_resp_json = json.loads(fxa_resp.content)
```
jwhitlock, line 40: This can fail in lots of ways, but mainly if the response is HTML, such as a 5xx response from Apache / gunicorn. You should use [fxa_resp.json()](https://docs.python-requests.org/en/latest/user/quickstart/#json-response-content) and a try/except block.

api/tests/views_tests.py
```diff
@@ -1,5 +1,3 @@
-from django.test import TestCase
```
jwhitlock, line 1: Thanks for cleaning this up!

docs/api_auth.md
```diff
@@ -0,0 +1,32 @@
...
+## FXA OAuth Token Authentication
+The Relay add-on uses the [`identity.launchWebAuthFlow` API][mdn-webauthflow]
+to perform an OAuth2 flow with [the FXA OAuth service][fxa-oauth], including
+[PKCE][fxa-pkce].
+
+After the OAuth flow is complete, the add-on has an FXA access token and a
+long-living FXA refresh token, and authenticates all requests to the Relay
+server by including an `Authorization: Bearer {fxa-access-token}` header in all
+API requests. The Relay server checks the token against [the FXA OAuth `/verify`
+endpoint][fxa-oauth-token-verify].
```
jwhitlock, line 17: This embeds the newline in the link text - add an early line break to avoid it?
```suggestion
API requests. The Relay server checks the token against
[the FXA OAuth `/verify` endpoint][fxa-oauth-token-verify].
```

api/authentication.py
```diff
@@ -0,0 +1,66 @@
...
+class FxaTokenAuthentication(BaseAuthentication):
+    def authenticate(self, request):
+        authorization = get_authorization_header(request).decode()
+        if not authorization or not authorization.startswith('Bearer '):
+            return None
+
+        token = request.headers.get('Authorization').split(' ')[1]
+        cache_key = get_cache_key(token)
+        fxa_resp = cache.get(cache_key)
+        if not fxa_resp:
```
jwhitlock, line 27: I think the cache hit cases are unchecked. I'd expect tests for:

* A 200 valid response with an active FxA user is cached
* A non-200 response is cached - I suspect this would break, since the status code is only checked on a request.

jwhitlock (changes requested):
Thanks @groovecoder! I think there are some changes needed to what we're caching and what we do the second time through when FxA returns an error.

api/authentication.py
```diff
@@ -0,0 +1,82 @@
...
+
+from allauth.socialaccount.models import SocialAccount
+from rest_framework.authentication import (
+    BaseAuthentication, get_authorization_header
+)
+
+
+logger = logging.getLogger('events')
+
+def get_cache_key(token, field):
```
jwhitlock, line 17: What does the `field` parameter do?

groovecoder: whoops good catch.

api/authentication.py
```diff
@@ -0,0 +1,82 @@
...
+            # we need to re-try, but we don't want to send un-throttled
+            # retries at FXA
+            cache.set(status_cache_key, fxa_resp_status, 60)
+            return None
+
+        json_cache_key = get_cache_key(token, 'json')
+        fxa_resp_json = cache.get(json_cache_key)
+        if not fxa_resp_json:
+            try:
+                fxa_resp_json = fxa_resp.json()
```
jwhitlock, line 51: I'm concerned that there are cases where `fxa_resp` will not be set when you get to here, but maybe it is OK. I'm having to read the whole function to figure that out, so maybe I'm missing the logic.

What if we stored this in the cache?

```json
{
   "status": 200,
   "json": {"whatever": "the json is"}
}
```

that way there is no way for the status and the JSON to get out of sync.

api/authentication.py
groovecoder: Fixed.

api/tests/authentication_tests.py
groovecoder: There was even a really cool `assert_call_count` method I used to make sure the `cache` logic was preventing the additional `requests` calls!

api/authentication.py
```diff
@@ -0,0 +1,83 @@
...
+        try:
+            fxa_uid = fxa_resp_json.get('sub')
+            sa = SocialAccount.objects.get(uid=fxa_uid)
+
+            # cache fxa_resp_json for as long as access_token is valid
+            # Note: FXA iat and exp are timestamps in *milliseconds*
+            fxa_token_exp_time = int(fxa_resp_json.get('exp')/1000)
+            now_time = int(datetime.now().timestamp())
+            timeout = fxa_token_exp_time - now_time
+            cache.set(status_cache_key, fxa_resp_status, timeout)
```
jwhitlock, line 73: ah here it is - this is where you are storing the `200`

jwhitlock (approved):
Thank @groovecoder , 🚢 SHIP IT

say-yawn (approved):
LGTM. Reviewed this with @jwhitlock and agree with his concerns but I feel that it can be addressed afterwards.

PR #1522, opened 2022-02-03
refactor csat context processor logic

groovecoder:


privaterelay/context_processors.py
```diff
@@ -77,3 +50,43 @@ def _get_fxa(request):
         return fxa
     except AttributeError:
         return None
+
+
+@lru_cache(maxsize=None)
```
groovecoder, line 55: If nothing else, I'd like to move the logic into a cached function like this so we can avoid the performance hit of all this logic on every request.

privaterelay/context_processors.py
```diff
@@ -77,3 +50,44 @@ def _get_fxa(request):
...
+        days_since = _get_days_since(days_since_args[reason])
+        if reason == 'premium' and days_since == None:
+            return '', None
+        for num_days in day_periods:
+            if (days_since >= num_days):
+                csat_dismissal_cookie = (
+                    f'csat-survey-{reason}-{num_days}days_dismissed'
+                )
+                if (request.COOKIES.get(csat_dismissal_cookie)):
+                    return csat_dismissal_cookie, None
```
groovecoder, line 79: Dang, this `return` will always execute as soon as someone dismisses the `free7days` or the `premium7days` prompt. We'll never go into the next iteration over the `day_periods` loop. Needs some re-thinking again.

privaterelay/context_processors.py
```diff
@@ -77,3 +50,44 @@ def _get_fxa(request):
...
+                return csat_dismissal_cookie, f'{reason}{num_days}days'
+
+    return '', None
+
+
+def _get_days_since(days_since_arg):
+    now = datetime.now(timezone.utc)
+    if type(days_since_arg == Profile):
```
Vinnl, line 87: I'm assuming that there's a wrongly placed bracket here?

```suggestion
    if type(days_since_arg) == Profile:
```

privaterelay/context_processors.py
Vinnl: OK, I did that in 50dbfab5.

groovecoder:
Closing this. Refactor can come later.

PR #1524, opened 2022-02-09
Open the newest alias created on the dashboard

codemist:
# New feature description

When visiting the relay dashboard, the **newest** created alias should have the full menu expanded so users can see all the info for the relay. All the aliases below the first should remain closed unless a user interacts and opens them.

# Screenshot (if applicable)

https://user-images.githubusercontent.com/13066134/153282416-87ab092c-a992-4ebe-9121-f6fccc0fdfd2.mov

# How to test

Generate a new alias and the newest alias card should be open, showing more details.

# Checklist


Vinnl (commented):
It mostly works, but the expand/collapse arrow is the wrong way around (i.e. it points down when it's expanded and up when it's collapsed, whereas the other aliases have it the other way around).

static/js/app.js
```diff
@@ -239,10 +246,22 @@ function addEventListeners() {
       }
 
       sendGaPing("Generate New Alias", "Engage", "btn-enabled");
+    
       e.target.submit();
+      localStorage.setItem("generateBtn", "True");
```
Vinnl, line 251: Hmm, I don't understand why we need this? (Does the backend set something like this? Since it's capitalised like Python's `True`?)

codemist: I used `localStorage` to check that the "Generate an Alias" button was selected even when the site refreshes when an alias is generated.

codemist: (so upon using React... we probably don't need any more than a setting an attribute)

Vinnl: Hmm, but isn't the top alias _always_ supposed to be expanded? (Which will be the just-generated alias if you refresh the page.)

Vinnl: Ah gotcha, that's good to know so I know what to test :) (And also because the React implementation currently works differently, so I'll also send a PR for that :) )

Vinnl (approved):

PR #1530, opened 2022-02-10
Start critical emails feature mpp 1347

groovecoder:
This PR fixes https://mozilla-hub.atlassian.net/browse/MPP-1642, https://mozilla-hub.atlassian.net/browse/MPP-1641, https://mozilla-hub.atlassian.net/browse/MPP-1600.

## How to test:
The only way to test this end-to-end on your local Relay setup is to go all-out and get an email domain you control to go thru AWS SNS and something like ngrok to reach your local machine. (I can help walk thru this setup if anyone is interested)

So, we'll likely need to test this end-to-end on the dev server. But I’m also finding it hard to find a test list server that is 1. easy to sign up for testing 2. uses the List-Unsubscribe header on its first email, so I can verify an alias blocking list emails does NOT get it.


emails/tests/fixtures/single_recipient_list_email_sns_body.json
```diff
@@ -0,0 +1 @@
...F0dXJlX2Jsb2NrLWVtcHR5\\r\\nIj48L2Rpdj48ZGl2IGNsYXNzPSJwcm90b25tYWlsX3NpZ25hdHVyZV9ibG9jay1wcm90b24iPlNl\\r\\nbnQgd2l0aCA8YSByZWw9Im5vb3BlbmVyIG5vcmVmZXJyZXIiIGhyZWY9Imh0dHBzOi8vcHJvdG9u\\r\\nbWFpbC5jb20vIiB0YXJnZXQ9Il9ibGFuayI+UHJvdG9uTWFpbDwvYT4gU2VjdXJlIEVtYWlsLjwv\\r\\nZGl2PjwvZGl2PjwvZGl2PjxkaXYgc3R5bGU9ImZvbnQtZmFtaWx5OiBhcmlhbDsgZm9udC1zaXpl\\r\\nOiAxNHB4OyI+PGJyPjwvZGl2Pg==\\r\\n\\r\\n\\r\\n--b1_1tMoOzirX5rNUK0QIqMdSQB8hwcBeFGHtIfEtmvEG0k--\\r\\n\\r\\n\"}", "Timestamp": "2022-01-13T21:02:46.911Z", "SignatureVersion": "1", "Signature": "BvlKJoLKcxcGsVloKnY7PKuqgY1EsSi1D5SVlHYtxRcsvuXVUDrdEvTlcWYNBiCpbaM/3kGYeXyb8uuWsdcNnT20zKLlpIP1dZhEV4o70aiM/35MxNfgAvA3NXkddRCQ2l+iplqtuRYdJ2XFkYJypECsoif9sRxB2jBW5bIz4vwjoeZdQCwMwBKQFsSGOr5Zc1pPE/5js9gZpBZlpsgRi2N4wj2sjBF0PyyO26UhGgWGwK1F7y1hpo90+PHRnD4W/XYc1flLu29hIaamgtXIbxol0xthMg0TtSfV9Q0MDJkpnsfOt/AsqvfZquIzGWyP2xpppuDLPf4MkftIbS4tpQ==", "SigningCertURL": "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-7ff5318490ec183fbaddaa2a969abfda.pem", "UnsubscribeURL": "https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:168781634622:ses-inbound-grelay:f1e4420c-3b67-4cb2-b6db-4d22fd45916e"}
```
groovecoder, line 1: The main additions here are:
```
\"headers\":[{\"name\":\"List-Unsubscribe\",\"value\":\"https://example.com/unsubscribe\"} ...
```
and
```
\"content\":\"List-Unsubscribe: https://example.com/unsubscribe\\r\\nReturn-Path: <fxastage@protonmail.com>\\r\\n ...
```

emails/tests/views_tests.py
```diff
@@ -28,10 +28,14 @@
...
+    'single_recipient_list'
+]
 EMAIL_SNS_BODIES = {}
-for email_type in [
-    'spamVerdict_FAIL', 'single_recipient', 'domain_recipient'
-]:
+for email_type in EMAIL_FIXTURES:
     email_file = os.path.join(
         real_abs_cwd, 'fixtures', '%s_email_sns_body.json' % email_type
     )
```
jwhitlock, line 41: Next time we add a fixture, we should load all by filename:

```python
suffix = '_email_sns_body.json'
for email_file in glob.glob(os.path.join(real_abs_cwd, 'fixtures', '*' + suffix)):
    filename = os.path.basename(email_file)
    email_type = basename[:-len(suffix)]
    with open(email_file, 'r') as f:
        email_sns_body = json.load(f)
        EMAIL_SNS_BODIES[email_type] = email_sns_body
```

emails/views.py
```diff
@@ -396,6 +398,14 @@ def _sns_message(message_json):
     _record_receipt_verdicts(receipt, 'active_alias')
     incr_if_enabled('email_for_active_address', 1)
 
+    # if address is blocking list emails, and email is from list, early return
+    email_is_from_list = _check_email_from_list(mail['headers'])
+    if address and address.block_list_emails and email_is_from_list:
```
jwhitlock, line 403: Nit: This may need the S3 deletion logic from PR #1543. But then every early return may need it...

https://github.com/mozilla/fx-private-relay/blob/9e154ab4ffb938b948bee295972dd81140ef857b/emails/views.py#L470-L473

emails/views.py
```diff
@@ -466,6 +476,13 @@ def _get_verdict(receipt, verdict_type):
     return receipt['%sVerdict' % verdict_type]['status']
 
 
+def _check_email_from_list(headers):
+    for header in headers:
+        if header['name'].lower().startswith('list-'):
```
jwhitlock, line 481: This includes almost all my work email:

* GitHub notifications
* Bugzilla CC
* Sentry emails
* Newsletter subscriptions

I may be doing email wrong 😁

emails/tests/views_tests.py
```diff
@@ -74,6 +78,31 @@ def test_single_recipient_sns_notification(self, mock_ses_relay_email):
         assert self.ra.num_forwarded == 1
         assert self.ra.last_used_at.date() == datetime.today().date()
 
+    @patch('emails.views.ses_relay_email')
+    def test_list_email_sns_notification(self, mock_ses_relay_email):
```
jwhitlock, line 82: Nice tests!

jwhitlock (approved):
Thanks @groovecoder, and sorry for the long wait for the first review on this PR! This is a code-only review, because I wanted to get you some feedback, and the developer setup is a bit involved, and this feature is off by default, so it should be safe to merge and ship without the UI to enable it.

emails/views.py
groovecoder: The work email is wrong - too many list emails involved in our work.

emails/migrations/0040_add_block_list_emails_fields.py
```diff
@@ -0,0 +1,23 @@
...
+        migrations.AddField(
+            model_name='domainaddress',
+            name='block_list_emails',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AddField(
+            model_name='relayaddress',
+            name='block_list_emails',
+            field=models.BooleanField(default=False),
+        ),
```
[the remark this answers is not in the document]
groovecoder: We deploy a single node into the production cluster and then it runs migrations. So there's a short interval of time (seconds?) where the load balancer might route some requests to the single node with new code before it's done running the migration. We usually see a few seconds where a dozen or so requests throw DB-related errors. :/

emails/tests/views_tests.py
groovecoder: did it this time!

emails/migrations/0040_add_block_list_emails_fields.py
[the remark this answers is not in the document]
jwhitlock: OK, that seems like we're a few infrastructure tweaks away from "migrations before code". I can hold my nose and take a few DB errors for now, unless you want to start splitting migration PRs across deploys.

emails/views.py
say-yawn: Sounds good. If we do a follow-up PR to clean that can you add the work for early return while checking the email being relayed is a reply on line 393?

emails/views.py
say-yawn: Huh... that could be just that the email industry is inundated with subscriptions... but also, why is their subscriptions related info on Sentry emails (presuming this is for Sentry notification)?

api/serializers.py
```diff
@@ -4,11 +4,21 @@
 from django.contrib.auth.models import User
 
 
-class RelayAddressSerializer(serializers.ModelSerializer):
+class PremiumValidatorsMixin:
```
say-yawn, line 7: Will you consider adding test for when the user tries to set the block_list_emails when they are not a premium user?

say-yawn (approved):
The logic looks good. Some recommendations on the `views.py` file but they can be moved to a fast follow PR.

emails/views.py
jwhitlock: Yes, it is a sentry notification. The one header that starts with ``list-`` is:

```
List-Id: fx-private-relay-prod.operations.localhost
```
I think this is the pattern ``<project>.<org>.<server hostname>``, and may change if we switch Sentry servers.

In turn, this header is mentioned in the values for ``ARC-Message-Signature`` and ``DKIM-Signature``. It may be that this helps keep some separation between different Sentry clients, so that if one use marks a message as spam, it doesn't necessarily block the other projects on the same server. But, it is my first week working email, so 🤷‍♂️.

emails/migrations/0040_add_block_list_emails_fields.py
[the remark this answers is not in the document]
groovecoder: Nah, separating deploys from migrations to code deploys feels like it could be even more cumbersome or error-prone. Some day we can work with SRE to change up the deploy so it runs prod migrations before putting the new node into the load balancer rotation.

emails/views.py
groovecoder: Yeah I think `list-*` headers are used for more than just "marketing lists". Because email clients sometimes parse and process them (e.g., when Gmail detects a `list-unsubscribe` header, it can show an "Unsubscribe" button in the UI), lots of tools that send automated emails are also using `list-*` headers.

api/serializers.py
groovecoder: Good call! Adding the tests uncovered a couple of fixes to make!

privaterelay/context_processors.py
```diff
@@ -59,16 +59,17 @@ def _get_csat_cookie_and_reason(request):
...
+        if not profile.date_subscribed:
+            # There are two reasons why someone might not have a subscription date set:
+            # - They subscribed before we started tracking that.
+            # - They have Premium because they have a Mozilla email address.
+            # In the latter case, their first visit date is effectively their
+            # subscription date. In the former case, they will have had Premium for
+            # a while, so they can be shown the survey too. Their first visit will
+            # have been a while ago, so we'll just use that as a proxy for the
+            # subscription date:
+            days_since_subscription = (datetime.now(timezone.utc) - datetime.fromisoformat(first_visit)).days
```
jwhitlock, line 71: It's exiting code, but if the cookie isn't set, `first_visit` is `None`, so `datetime.fromisoformat(None)` fails.

privaterelay/context_processors.py
```diff
@@ -54,12 +54,14 @@ def _get_csat_cookie_and_reason(request):
     if not request.user.is_authenticated:
         return None, None
     profile = request.user.profile_set.first()
-    first_visit = request.COOKIES.get("first_visit")
+    first_visit = request.COOKIES.get(
+        "first_visit", datetime.now(timezone.utc).isoformat()
+    )
```
groovecoder, line 59: @jwhitlock look good?

jwhitlock: Thanks!

PR #1541, opened 2022-02-14
Add an FAQ entry on acceptable use of Relay

Vinnl:
# New feature description

Like https://github.com/mozilla/fx-private-relay/pull/1540, but applied to the React codebase from #1518 (this PR merges into that PR's branch). To get ourselves used to reviewing React PRs :slightly_smiling_face:  

# Screenshot (if applicable)

image

# How to test

Go to the `/faq` page, verify that the new question is shown there.

# Checklist

- [ ] l10n dependencies have been merged, if any. Not yet: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/60

### DIFF
--- a/frontend/src/pages/faq.page.tsx
+++ b/frontend/src/pages/faq.page.tsx
@@ -73,7 +73,13 @@ const Faq: NextPage = () => {
                     attrs: "",
                   }}
                   elems={{
-                    a: <a href="https://support.mozilla.org/products/relay/" />,
+                    a: (
+                      <a
+                        href="https://support.mozilla.org/products/relay/"
+                        target="_blank"
+                        rel="noopener noreferrer"
+                      />
+                    ),
                   }}
                 >
                   <p />
@@ -99,7 +105,11 @@ const Faq: NextPage = () => {
                   }}
                   elems={{
                     a: (
-                      <a href="https://addons.mozilla.org/firefox/addon/private-relay/" />
+                      <a
+                        href="https://addons.mozilla.org/firefox/addon/private-relay/"
+                        target="_blank"
+                        rel="noopener noreferrer"
+                      />
                     ),
                   }}
                 >
@@ -119,7 +129,11 @@ const Faq: NextPage = () => {
                   }}
                   elems={{
                     a: (
-                      <a href="https://addons.mozilla.org/firefox/addon/private-relay/" />
+                      <a
+                        href="https://addons.mozilla.org/firefox/addon/private-relay/"
+                        target="_blank"
+                        rel="noopener noreferrer"
+                      />
                     ),
                   }}
                 >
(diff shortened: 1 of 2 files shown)


frontend/src/pages/faq.page.tsx
```diff
@@ -229,6 +229,58 @@ const Faq: NextPage = () => {
...
+                    attrs: "",
+                  }}
+                  elems={{
+                    a: (
+                      <a href="https://www.mozilla.org/about/legal/acceptable-use/" />
+                    ),
+                  }}
+                >
+                  <p />
+                </Localized>
```
Vinnl, line 251: This is a small peculiarity of [`@fluent/react`](https://github.com/projectfluent/fluent.js/wiki/React-Bindings): it can only replace `<tags>` in translation strings via the `<Localized>` wrapper component, not via the `useLocalization` hook (that provides the `l10n` object).

The way that component works is: it takes its child element (`<p />`, in this case), and replaces its contents with the contents of the Fluent string (`faq-question-acceptable-use-answer-a-html`).

Also note that `vars` shouldn't actually be necessary: if the string just says `<my-tag>acceptable use policy</my-tag>`, then we could add `my-tag` to `elems` and set whatever attribute we need on the element there, just like we set the `href` attribute on the `<a>` element we replace the `<a>` tag with. However, because this string is also used on the current site, I had to use the `<a href="{ $url }" { $attrs }>` trick too.

maxxcrawford:
Quick note for @codemist and @Vinnl: I created a new label for any React focused issues or issues while we are managing both front ends. If we need a label for the Django front-end, lmk and we can make another tag too.

frontend/src/pages/faq.page.tsx
```diff
@@ -229,6 +229,58 @@ const Faq: NextPage = () => {
...
+                </ul>
+                <Localized
+                  id="faq-question-acceptable-use-answer-b-html"
+                  vars={{
+                    url: "https://www.mozilla.org/about/legal/terms/firefox-relay/",
+                    attrs: "",
+                  }}
+                  elems={{
+                    a: (
+                      <a href="https://www.mozilla.org/about/legal/terms/firefox-relay/" />
```
maxxcrawford, line 277: suggestion: Can open all FAQ links in a new tab? 

From the Djagon frontend, we add the following attrs to each link in the FAQ: https://github.com/mozilla/fx-private-relay/blob/main/static/js/app.js#L362-L373

Note that the `noopener` may [no longer be necessary](https://github.com/whatwg/html/issues/4078).

maxxcrawford (approved):
I have one non-blocking change!

frontend/src/pages/faq.page.tsx
Vinnl: 7b1a05b5

Note that the linter (configured to use the recommended default rules by ESLint and Next.js) added both parameters automatically on commit. Better safe than sorry, I suppose :)

PR #1543, opened 2022-02-14
Remove email from s3 when email address does not exist mpp 1257

say-yawn:
This PR fixes [MPP-1257](https://mozilla-hub.atlassian.net/browse/MPP-1257) and [MPP-1256](https://mozilla-hub.atlassian.net/browse/MPP-1256)

# How to test
## Inbound email to an address that does not exist is deleted from S3 after attempt to process the email
1. Go to the S3 bucket that stores the email and take note of the files stored here
2. Go to the Relay dashboard and generate a new Relay alias
3. Send an email to this alias to check that the email relaying works as expected
4. Go back to the dashboard and delete the generated alias on step (2)
5. Send an email to the deleted Relay address and check that the S3 bucket stayed the same
6. Check the S3 metrics [here](https://s3.console.aws.amazon.com/s3/bucket/fxprivaterelay-dev/metrics/bucket_metrics?region=us-east-1&tab=request&period=1d) that the email was stored and then deleted

# Checklist
- [ ] I've added a unit test.


emails/tests/views_tests.py
```diff
@@ -135,6 +145,67 @@ def test_sns_message_with_spam_bounce_sets_auto_block_spam(self):
...
+class SnsMessageTest(TestCase):
+    def setUp(self) -> None:
+        self.bucket = 'test-bucket'
+        self.key = '/emails/objectkey123'
+
+        patcher = patch('emails.views._get_address', side_effect=ObjectDoesNotExist())
+        patcher.start()
+        self.addCleanup(patcher.stop)
+
+    @patch('emails.views.remove_message_from_s3')
```
groovecoder, line 158: All these test make sure we call `utils.remove_message_from_s3` but we're not testing that function. We could add tests around that function too, or make these integration tests using `moto` to make sure the expected aws api functions are called too.

groovecoder (changes requested):

emails/tests/views_tests.py
groovecoder: We can do this in a follow-up.

groovecoder (approved):

PR #1544, opened 2022-02-14
end-to-end-local-dev docs

groovecoder:


jwhitlock (approved):
Thanks @groovecoder, the embedded mermaid diagrams are an awesome new feature!

I've read through the docs, but I haven't run them yet, and it may be a while before I can do all the steps. However, this is a big step forward in documenting your working setup, so let's get it into main!

PR #1552, opened 2022-02-17
In README, add migration step to enabling FxA

jwhitlock:
After setting `ADMIN_ENABLED=True`, a migration is needed to create the admin tables, and a restart to add the admin URL routes.

Other cleanup for the section:
* Use `<h3>` level, same as "Working With Translations", rather than `<h4>` to be under that topic
* Remove some whitespace

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ of the translations submodule:
 
 * `git push`
 
-#### Recommended: Enable Firefox Accounts authentication
+### Recommended: Enable Firefox Accounts authentication
 To enable Firefox Accounts authentication on your local server, you can use the
 "Firefox Private Relay local dev" OAuth app on accounts.stage.mozaws.net.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -122,33 +122,36 @@ of the translations submodule:
...
 
 To do so:
 
 1. Set `ADMIN_ENABLED=True` in your `.env` file
 
-2. Go to [the django admin page to change the default
+2. Shutdown the server if running, and run `python manage.py migrate` to add
+   the admin tables. Start the server with `python manage.py runserver`.
+
+3. Go to [the django admin page to change the default
```
groovecoder, line 136:
```suggestion
2. Shutdown the server if running, and run `python manage.py migrate` to add
   the admin tables. 

3. Re-start the server with `python manage.py runserver`.

4. Go to [the django admin page to change the default
```

groovecoder (changes requested):

README.md
groovecoder: Can we split the 2 different commands into 2 different steps?

groovecoder (approved):

PR #1553, opened 2022-02-17
Add tests for `_get_attachment`, and do not use attachment filename for temporary storage

### LINKED ISSUE #1551: Attachment name used in temporary local file leads to invalid filenames
When an attachment exceeds 150KB, it is written to disk rather than held in memory:

https://github.com/mozilla/fx-private-relay/blob/04fbbdd5888aeb8a6e8a9485b47d4523ac4978a7/emails/views.py#L748-L752

The `prefix` and `suffix` come from the attachment name, which is user-provided and may not result in a valid filename, and an error when the contents are written to disk.

We should not base the filename on text from the incoming email, and should double-check that these files are closed at the end of the request to free disk space.

jwhitlock:
This PR fixes #1551 by using the prefix `"relay_attachment_"`, and not basing the filename prefix or suffix on user data from the attachment filename. It also adds some tests for `_get_attachment()`.

`_get_attachment()` uses [SpooledTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile), which holds an attachment in memory until it reaches 150 KiB, when it is stored in a temporary file. Previously, we used the attachment filename to make the prefix and suffix for the temporary file. However, this fails when the attachment filename has invalid characters, such as a colon (`:`). Now that we support larger files, we're starting to see some large attachments with a URL for the filename (`https://example.com/image.jpg`), and failures when writing to a tempfile with that name, such as:

```
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/_y/72__kkbs2tqd5ch4gwrgnwtr0000gn/T/https://example.com/image7piavr1v.jpg`
```

With this change, the temporary file will now have a randomly generated name, not related to the attachment name.

There is another potential bug for attachments with no filename, but I have not seen the signature in production yet, where calling `prefix=os.path.splitext(fn)[0]` results in the error:

```
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

### DIFF
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -409,48 +409,39 @@ class GetAttachmentTests(TestCase):
         message.add_attachment(data, maintype=maintype, subtype=subtype, filename=filename)
         return message
 
+    def get_name_and_stream(self, message):
+        """Get the first attachment's filename and data stream from a message."""
+        for part in message.walk():
+            if part.is_attachment():
+                return _get_attachment(part)
+        return None, None
+
     def test_short_attachment(self):
         """A short attachment is stored in memory"""
         message = self.create_message(b"A short attachment", "text/plain", "short.txt")
-        att_name, att = None, None
-        for part in message.walk():
-            if part.is_attachment():
-                att_name, att = _get_attachment(part)
-                break
-        assert att_name == 'short.txt'
-        assert isinstance(att._file, io.BytesIO)
+        name, stream = self.get_name_and_stream(message)
+        assert name == 'short.txt'
+        assert isinstance(stream._file, io.BytesIO)
 
     def test_long_attachment(self):
         """A long attachment is stored on disk"""
         message = self.create_message(self.long_data, "application/octet-stream", "long.txt")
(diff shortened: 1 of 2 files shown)


groovecoder (dismissed):

jwhitlock:
I found Sentry issues for omitted filenames as well

https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/16863848

The line is:

https://github.com/mozilla/fx-private-relay/blob/04cfa86602af0a9ac966fa414b6288cffadea519/emails/views.py#L768

and the exception is:

```TypeError: expected str, bytes or os.PathLike object, not NoneType````

This change should also fix that, but we should have a test case for omitted filenames as well. Any, we may need to adjust downstream processing of attachments without filenames, but let's deal with that later.

jwhitlock:
I've rebased on main, refactored the common code to `GetAttachmentTests.get_name_and_stream()`, and added a test for an omitted filename.

say-yawn (approved):
LGTM.

PR #1554, opened 2022-02-18
Only expand newly-created aliases

Vinnl:
We used to always expand the top-most alias, but this changes that
to only expand aliases that have just been created, to align with
https://github.com/mozilla/fx-private-relay/pull/1524.

# Code walkthrough

This is in the `<AliasList>` component, which essentially is only in charge of displaying the list of aliases - actual fetching and saving of alias data is handled by the Profile page that includes this component through callbacks like `onCreate` and `onUpdate`. This means that when the `onCreate` callback is called, the Profile page will send a request to the API, get the updated list of aliases, and will then re-render `<AliasList>`, simply passing it a new list of aliases.

In other words: `<AliasList>` just gets a list of aliases, and can't see which of those are "new". To solve this, I modified `<AliasList>` to keep track of the aliases it has already seen (in the `existingAliases` state variable). It updates this variable whenever someone clicks the `<GenerateAliasButton>` (by calling `setExistingAliases`), but _before_ it passes that signal on to the Profile page through the `onCreate` handler.

Then when it gets re-rendered, `existingAliases` contains the list of aliases that were known when the user clicked the `<GenerateAliasButton>`, and `props.aliases` contains the updated list of aliases passed to it by the Profile. It can then choose to expand alias cards that are in the latter but not in the former (`defaultOpen={!existingAliases.includes(alias)}`).

Explaining code can get quite length, but I hope it's still useful :sweat_smile: 

# Screenshot (if applicable)

https://user-images.githubusercontent.com/4251/154667344-7b1648a4-6ddb-41dc-abe3-465a1bc68659.mp4

# How to test

Open the dashboard, verify that no aliases are expanded by default. Then create a new alias, and verify that it is expanded.

# Checklist

### DIFF
--- a/frontend/src/components/dashboard/aliases/AliasList.tsx
+++ b/frontend/src/components/dashboard/aliases/AliasList.tsx
@@ -32,6 +32,9 @@ export const AliasList = (props: Props) => {
   const [stringFilterInput, setStringFilterInput] = useState("");
   const [categoryFilters, setCategoryFilters] = useState<SelectedFilters>({});
   const [localLabels, storeLocalLabel] = useLocalLabels();
+  // Whenever a new alias is created, this value tracks the aliases that existed
+  // before that. That allows us to expand newly-created aliases by default.
+  const [existingAliases, setExistingAliases] = useState(props.aliases);
 
   if (props.aliases.length === 0) {
     return null;
@@ -44,7 +47,7 @@ export const AliasList = (props: Props) => {
     })
   );
 
-  const aliasCards = aliases.map((alias, index) => {
+  const aliasCards = aliases.map((alias) => {
     const aliasWithLocalLabel = { ...alias };
     if (
       alias.description.length === 0 &&
@@ -81,7 +84,7 @@ export const AliasList = (props: Props) => {
           profile={props.profile}
           onUpdate={onUpdate}
           onDelete={() => props.onDelete(alias)}
-          defaultOpen={index === 0}
+          defaultOpen={!existingAliases.includes(alias)}
           showLabelEditor={props.profile.server_storage || localLabels !== null}
         />
       </li>
@@ -139,6 +142,12 @@ export const AliasList = (props: Props) => {
       </Localized>
     ) : null;
 
+  const onCreate: typeof props.onCreate = (options) => {
+    setExistingAliases(props.aliases);
+
+    return props.onCreate(options);
+  };
+
   return (
     <section>
       <div className={styles.controls}>
@@ -148,7 +157,7 @@ export const AliasList = (props: Props) => {
             aliases={props.aliases}
             profile={props.profile}
             runtimeData={props.runtimeData}
-            onCreate={props.onCreate}
+            onCreate={onCreate}
           />
         </div>
       </div>


Vinnl:
You can now test this here! https://deploy-preview-1554--amazing-goodall-634c5f.netlify.app/?mockId=some

groovecoder (commented):
Thanks for the explanation. I'm not sure if I understand React enough yet to articulate a code "smell", but ...

When I read "It updates this variable whenever someone clicks the `<GenerateAliasButton>` (by calling `setExistingAliases`), but _before_ it passes that signal on to the `Profile` page through the `onCreate` handler." ... it sounds like less-predictable state conditions that I always thought React was meant to eliminate?

So, I tried to figure out what was happening with `AliasList.onCreate` BEFORE this change, and it took me quite a while (and some over-the-shoulder help) to trace from `AliasList.onCreate` to `Profile.createAlias`, which got me thinking -why is there so much imperative code (`if`, `else`, `response =`, `throw`, `toast()`) in a declarative component like `Profile`? 

It seems like everything in `Profile.createAlias` should be pushed "up" into ... maybe the `randomAliasCreater` logic? Which, instead of returning the API fetch response, should add a flag like `newlyCreated: true` to the `CommonAliasData` of the newly-created alias in the `AliasList`, which can then do `defaultOpen={alias.newlyCreated}`.

I'm not really savvy enough to intelligently criticize React architectures, but beyond just this `AliasList` component, it looks like there's a LOT of imperative code mixed in with our declarative components. The only other React app I built also used redux as a centralized state container, and it felt much "tighter" than this code does. I'm not sure if Redux and SWR are redundant? Or can we use them together? Or can we refactor our use of SWR to work more like a centralized state container too?

Vinnl:
@groovecoder Your intuition is definitely right that the code you're referring to is "smelly". There's a common thread to all of them that lies at the root of that: the impedance mismatch between the largely declarative world of React, and the touch points where we have to perform side effects to interact with the outside world.

In other words, behaviour that cleanly maps to a mental model of the view being a function of application state (e.g. "the first element of the list of aliases should be expanded by default"), which is most of the code, slots nicely into the declarative world. Behaviour that's reliant on side effects ("send an API request", "show an error message when a response comes in", "only show this until some time passes", ...) doesn't slot in as cleanly.

(Interesting to note: the upcoming v18 of React is [set to add](https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html#whats-coming-in-react-18) some more primitives to make rendering conditional on data fetching, and in response the ecosystem [appears to be coming up](https://remix.run/blog/remixing-react-router) with cleaner ways to map data fetching onto React. That will still take quite a bit of maturing though, but something to keep an eye on.)

It's definitely possible to abstract some of that logic out some more. Unfortunately that still won't get you into a fully declarative world, because the mismatch with side effects is still there, so if e.g. we were to use Redux, we'd have to layer further abstractions like [thunks](https://redux.js.org/tutorials/essentials/part-5-async-logic#thunks-and-async-logic) on top of it. So far I've been trying to avoid that, to avoid making our code even harder to learn with yet another layer of indirection.

We could move more of the `createAlias` logic into the alias data fetching hooks. In fact, I think we'll probably want to do that. However, I didn't want to do that prematurely. The main pain point of those hooks are that the code paths for aliases are pretty much the same everywhere, yet we have separate API's for random and custom aliases. IIRC there was already talk about unifying those in preparation for Fx integration, and so that's probably a better place to resolve that tension rather than pulling up an abstraction in front of that on the client-side first?

That said, I'm not sure about integrating error handling into the data fetching code, since the way we want to communicate errors to the user (in this case with a toast message) might differ depending on the context.

Anyway, those were the trade-offs I considered and the choices I made when considering them. Given that there's pros and cons to the different approaches, I'd be OK with reworking some of that, assuming we'd do it consistently across the codebase. Especially when it comes to learnability, though, I'd like input from Maxx and Kaitlyn on which approach sounds least confusing :slightly_smiling_face:

groovecoder:
Thanks for the thoughtful response; TIL "thunks" and I think I really like them! (? Maybe ... haven't used them so don't know what kind of mess they might be like "in production".)

> We could move more of the `createAlias` logic into the alias data fetching hooks. In fact, I think we'll probably want to do that. However, I didn't want to do that prematurely. The main pain point of those hooks are that the code paths for aliases are pretty much the same everywhere, yet we have separate API's for random and custom aliases.

Could we make a `markAsNewAlias` function and pass it as a callback for the `.mutate()` calls already in the `useAliases` hooks?

> IIRC there was already talk about unifying those in preparation for Fx integration, and so that's probably a better place to resolve that tension rather than pulling up an abstraction in front of that on the client-side first?

We'll start Fx integration with a single unified `GET` endpoint for all aliases - to reduce the requests coming from so many Fx clients. But we don't plan to consolidate into a single API endpoint like `GET|POST|PATCH|PUT|DELETE /api/v1/alias` anytime soon.

Vinnl:
> Could we make a `markAsNewAlias` function and pass it as a callback for the `.mutate()` calls already in the `useAliases` hooks?

I spent some time looking at this today, but this looks to be a bit gnarly. Indeed, it [is possible](https://swr.vercel.app/docs/mutation#returned-data-from-mutate) to pass in a callback to `mutate` to update the local cache of API data, in which we could set another property. However, the intended use case for that is to optimistically update the local UI while awaiting changes being persisted to the back-end. Now, SWR also regularly revalidates data (e.g. when losing a network connection and then reconnecting, which is useful for long-running tabs), including after an update. That means the flow would look something like this:

1. We send a `POST` to update the data.
2. When the response comes in, we add the newly-created alias to the local cache of aliases, but setting the property `just_created: true`.
3. SWR then does another `GET` request, to ensure it has the latest alias data (e.g. so that aliases created on other devices or tabs are also visible).
4. As soon as that response comes in, it overwrites the local cache of aliases, including our `just_created` property.

We could try to keep some global client-side state keeping track of which aliases were created in that session, and then amending the response data every time it comes in, but that feels like it would end up _less_ declarative than what we've got now. Let me know what you think.

> But we don't plan to consolidate into a single API endpoint like GET|POST|PATCH|PUT|DELETE /api/v1/alias anytime soon.

OK, I'll take a look at consolidating that client-side then, but I'll do that in a separate PR, since that exceeds the scope of this one quite a bit :)

PR #1561, opened 2022-02-21
Add announcement area for new features

Vinnl:
# New feature description

This adds a menu that highlights new features. Note that this is slightly limited compared to the acceptance criteria: since only a few entries are shown at this time, the menu shows the entire history. I've prepared a separate PR to implement the full feature set in React, which will be used by the time there are more entries: https://github.com/mozilla/fx-private-relay/pull/1562.

# Screenshot (if applicable)

https://user-images.githubusercontent.com/4251/154987854-dce94cf3-54d6-4b11-888c-3ba99f22269d.mp4

# How to test

- Sign in with a pro account and with the add-on enabled.
- See that there's a "News" entry in the menu, with a counter set to 3.
- Verify that the counter is decremented when opening one of the entries for the first time.
- With different accounts, verify that the counter is reset again. Additionally:
- With the extension not installed, verify that there is no entry for signing back in.
- Without Premium, verify that there is no entry about critical emails.

# Checklist

- [ ] l10n dependencies have been merged, if any. - Not yet: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/63
- [ ] All acceptance criteria are met. - No, see above.


static/scss/components/whatsnew.scss
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        li {
+            display: none;
+
+            &.is-visible {
+                display: initial;
+            }
+
+            button {
```
maxxcrawford, line 75: Suggestion: this should be broken out into its own class/section: `.c-whatsnew-entry-content`

static/scss/components/whatsnew.scss
```diff
@@ -0,0 +1,157 @@
...
+                }
+
+                p {
+                    @include text-body-sm;
+                }
+
+                &:hover {
+                    background-color: $color-blue-50;
+                    color: $color-white;
+                }
```
maxxcrawford, line 99: issue (blocking): The hover state for each news item only the color of the description, but not the title (`h3`). They both should be updated to the same color.

image

maxxcrawford (changes requested):
This PR isn't reviewable as there's no companion PR in the [L10N repo](https://github.com/mozilla-l10n/fx-private-relay-l10n/pulls) to check against, so all the strings are `???` (However, it looks like the [React PR](https://github.com/mozilla/fx-private-relay/pull/1561) has all the strings in the [`pendingTranslations.ftl`](https://github.com/mozilla/fx-private-relay/blob/new-stuff_react/frontend/pendingTranslations.ftl) file. 

Added two CSS notes!

static/scss/components/whatsnew.scss
Vinnl: Nice catch: bb9fa6c9

But then Eduardo pointed out that the hover colour should be different, and the text colour is black again, so I removed that again :P e0a201d3

static/scss/components/whatsnew.scss
Vinnl: b1e3a8e0

Vinnl:
l10n PR wasn't submitted yet (the strings in the React PR were tentative) because the content wasn't final yet, but it has now been submitted! https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/63

maxxcrawford (commented):
I did a round of just running/testing and did some quick design comparisons with the Figma layout. There are a a few major things missing, LMK if any of these were omitted intentionally (since we have the new front-end coming soon).  

image

UX: 
- History/News toggle state
- Close/cancel button in top right corner
- Clear all button in bottom right corner 

UI Differences: 
- Border in between two news items (and in the history/news toggle header) 
- Padding for elements inside the feed panel 
  - Note: It's 13px/6px for non-hovered/hovered states, so I would put a `$spacing-sm` on both the parent and child elements in the feed. 
- The hover color is wrong (but you're using the correct color (`$color-violet-05`). Can you confirm that the variables are actually overwriting the Protocol tokens? 
- Missing triangle bubble on top of news feed
- The number (encased in yellow circle) is in italics
- Additional horizontal padding on "News" item when on hover/active (at least left side, maybe both?) 

Will do a deeper dive / code review tomorrow!

Vinnl:
@maxxcrawford Yes, see my initial PR description :)

> Note that this is slightly limited compared to the acceptance criteria: since only a few entries are shown at this time, the menu shows the entire history. I've prepared a separate PR to implement the full feature set in React, which will be used by the time there are more entries

Tony and Eduardo were OK with this approach. I've omitted the close button as well since there's no header now, and it can also be closed by clicking outside of the menu.

As for the remaining issues:

> - Border in between two news items
> - Padding for elements inside the feed panel 

74c02b2a

> The hover color is wrong (but you're using the correct color ($color-violet-05). Can you confirm that the variables are actually overwriting the Protocol tokens?

So I investigated this, and it looks like

```scss
@import "libs/protocol/css/components/forms/choice.scss";
```

in `filter-bar.scss` again leads to an import of the Protocol tokens, which then overwrites the Nebula colours again. I could fix that for this specific colour by moving the filter-bar import below the whatsnew import (1947f472), but that does mean that much of our other styles don't actually seem to use the Nebula colours.

Removing that import from `filter-bar` broke a bunch of other things, so I propose just waiting for this to be fixed with the React frontend, where styles don't easily leak into other stylesheets.

> Missing triangle bubble on top of news feed

ea2859dd

> The number (encased in yellow circle) is in italics

c500a522

> Additional horizontal padding on "News" item when on hover/active (at least left side, maybe both?)

I tried fixing this, but the horizontal padding on the other menu items is a lot smaller, so if I were to make space for this, the "News" entry would look very out-of-place. I'm also a bit afraid to touch this, since the menu items are already a bit inconsistent between the Premium and Free themes, and so I'm not sure what I'd break if I were to start messing with those spacings. So I propose accepting the smaller padding for now.

codemist (commented):
A couple things, I can't seem to close the tab when clicking the news button again.
https://user-images.githubusercontent.com/13066134/156044262-c05fdbda-f307-4420-ae94-cd3ba2f1d379.mov

Another thing is that when I shrink the device width, the panel gets cut off at a certain breakpoint (around 762px). Are we keeping the whats new feature on mobile as well? 
https://user-images.githubusercontent.com/13066134/156044637-352b3131-0126-4fa5-82e1-38f8a1748bb3.mov

static/scss/components/whatsnew.scss
```diff
@@ -0,0 +1,177 @@
...
+        padding: 0 $spacing-md;
+
+        &.is-hidden {
+            display: none;
+        }
+
+        li {
+            display: none;
+            padding: $spacing-sm 0;
+            border-bottom: 1px solid $color-light-gray-30;
```
maxxcrawford, line 87: You'll want to add an extra rule to remove this for the last item. 

I made the border color red to show its visibility: 

image

maxxcrawford (approved):
Nice fixups. I had a ton of nits, and I appreciate you knocking them out.

I found one non-blocking CSS issue.  I'm approving, but please be sure to check out @codemist's comments above. Thanks!

maxxcrawford:
Quick note: We're at "code freeze" today, so I'm going to merge this (after I fix the analytics conflict) and let you do a follow up PR for the other issues (including what I called out").

PR #1567, opened 2022-02-23
Handle s3 response not available

say-yawn:
Related to Sentry error spike [here](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/18255995/?query=is%3Aunresolved).

- [ ] I've added a unit test to test for potential regressions of this bug

### DIFF
--- a/emails/apps.py
+++ b/emails/apps.py
@@ -2,6 +2,7 @@ import logging
 import os
 
 import boto3
+from botocore.config import Config
 
 from django.apps import AppConfig
 from django.conf import settings
--- a/emails/apps.py
+++ b/emails/apps.py
@@ -19,7 +19,12 @@ class EmailsConfig(AppConfig):
             self.ses_client = boto3.client(
                 'ses', region_name=settings.AWS_REGION
             )
-            self.s3_client = boto3.client('s3', region_name=settings.AWS_REGION)
+            self.s3_client = boto3.client(
+                's3',
+                region_name=settings.AWS_REGION,
+                max_attempts=1,  # this includes the initial attempt to get the email
+                retry_mode='standard',
+            )
         except Exception:
             logger.exception("exception during SES connect")
 
--- a/emails/apps.py
+++ b/emails/apps.py
@@ -19,12 +19,13 @@ class EmailsConfig(AppConfig):
             self.ses_client = boto3.client(
                 'ses', region_name=settings.AWS_REGION
             )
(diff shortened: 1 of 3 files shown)


emails/apps.py
```diff
@@ -19,7 +19,12 @@ def __init__(self, app_name, app_module):
             self.ses_client = boto3.client(
                 'ses', region_name=settings.AWS_REGION
             )
-            self.s3_client = boto3.client('s3', region_name=settings.AWS_REGION)
+            self.s3_client = boto3.client(
+                's3',
+                region_name=settings.AWS_REGION,
+                max_attempts=1,  # this includes the initial attempt to get the email
+                retry_mode='standard',
+            )
```
jwhitlock, line 27: I think we should model after the example code in [Defining a retry configuration in a Config object for your Boto3 client](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#defining-a-retry-configuration-in-a-config-object-for-your-boto3-client):

```python
import boto3
from botocore.config import Config

config = Config(
   retries = {
      'max_attempts': 10,
      'mode': 'standard'
   }
)

ec2 = boto3.client('ec2', config=config)
```

That looks like this (without the ``from botocore.config import Config``):


```suggestion
            s3_config = Config(
                region_name=settings.AWS_REGION,
                retries = {
                    "max_attempts": 1,  # this includes the initial attempt to get the email
                    "mode": "standard",
                })
            self.s3_client = boto3.client('s3', config=s3_config)
```

Also see the [botocore.config.Config](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html) docs.

emails/views.py
```diff
@@ -584,9 +584,14 @@ def _handle_reply(from_address, message_json, to_address):
...
     )
 
-    text_content, html_content, attachments = _get_text_html_attachments(
-        message_json
-    )
+    try:
+        text_content, html_content, attachments = _get_text_html_attachments(
+            message_json
+        )
+    except ClientError as e:
```
jwhitlock, line 591: Nice catch! Easy to forget that `_get_text_html_attachments()` now calls S3...

jwhitlock (changes requested):
Thanks @say-yawn!

I _think_ the S3 configuration will work (maybe it acts like it was defined in the AWS configuration file?), but I'd be more comfortable using the ``Config`` object like the docs.

jwhitlock (approved):

PR #1571, opened 2022-02-25
Add release to Sentry

jwhitlock:
When building the Docker image, inject the tag, branch, and commit SHA as environment variables. Use this to create a release number, either the tag or "branch:commit".

This enables us to track releases in Sentry, and use features like "resolved in next release".


jwhitlock:
Trying again with branch on the main repository rather than a fork, so that CI will run.

PR #1572, opened 2022-02-25
Tag latest image on merge to main

jwhitlock:
The CircleCI configuration was not updated for the change from "master" to "main", so it hasn't been pushed since January 2021.


jwhitlock:
Opening again as branch on main repository, so that CI will run on the changes.

PR #1574, opened 2022-02-25
Tag latest image on merge to main

jwhitlock:
The CircleCI configuration was not updated for the change from "master" to "main", so it hasn't been pushed since January 2021.


groovecoder (approved):
Weird. I guess we don't use the `latest` tag on DockerHub for anything, so we hadn't noticed this until now. Nice catch.

PR #1576, opened 2022-02-25
Generate and store Circle-CI XML test reports

jwhitlock:
Adjust the test commands to store the test results as XML, which are then stored as test results. 

This may give more detailed information when a test fails. This could be extended to use a site like codecov to track coverage over time:

https://about.codecov.io/tool/circleci/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,32 @@ jobs:
 
       - run:
           name: Test Code
-          command: docker run --entrypoint "/app/.local/bin/coverage" fx-private-relay run --source=. -m pytest --cov=./ --cov-fail-under=60
+          command: |
+            # Create a volume owned by the app user
+            docker run \
+              --volume /tmp/test-results \
+              --name test-results \
+              alpine \
+              /bin/sh -c \
+                "chmod 0777 /tmp/test-results && \
+                 chown 10001:10001 /tmp/test-results"
+            # Run coverage tests, outputting the results in XML format
+            docker run \
+              --entrypoint "/bin/bash" \
+              --volumes-from test-results \
+              fx-private-relay \
+              -c \
+                'mkdir -p /tmp/test-results/pytest && \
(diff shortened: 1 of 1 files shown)


jwhitlock:
Like a lot of CI work, that took several rounds to get a working set of commands. 

The hard part is that CircleCI runs docker commands on a remote instance, so you have to do some tricks to use a docker volume rather than mounting local folders (which is how I got this working locally). I used the [Mounting folders](https://circleci.com/docs/2.0/building-docker-images/#mounting-folders) doc to get me started.

The next hard part was writing YAML, but I'm pretty happy with the results.

As a bonus, the CircleCI display now knows how many tests ran:

CircleCI Tests Tab

.circleci/config.yml
```diff
@@ -60,7 +60,32 @@ jobs:
...
+            # Run coverage tests, outputting the results in XML format
+            docker run \
+              --entrypoint "/bin/bash" \
+              --volumes-from test-results \
+              fx-private-relay \
+              -c \
+                'mkdir -p /tmp/test-results/pytest && \
+                 mkdir -p /tmp/test-results/coverage && \
+                 /app/.local/bin/coverage run --source=. --branch -m pytest \
+                   --cov=./ --cov-fail-under=60 \
```
groovecoder, line 81: question: Now that this is done with multiple commands in a single `docker run` - will the `--cov-fail-under=60` still fail the step if the coverage drops under 60%?

groovecoder (changes requested):

.circleci/config.yml
jwhitlock: the command runs with `#!/bin/sh -eo pipefail`. `-e` means exit immediately when a command fails, and `-o pipefail` includes failures in the middle of a pipe operation. So, I think yes, it will fail if coverage falls below 60. But, there's only one way tko be 100% certain - remove some tests and watch it fail.

groovecoder (approved):

Issue #1558, opened 2022-02-18
Validate user can make more alias on validation


say-yawn:
# About this Issue
After fixing the 5xx to 4xx error for `CannotMakeAddress` exception in #1429 we stopped seeing errors [here](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/15055581/?query=is%3Aunresolved) but in the API endpoint, this exception is being raised on save() instead of validate() causing Sentry events [here](https://docs.djangoproject.com/en/4.0/ref/exceptions/#django.core.exceptions.SuspiciousOperation). And because the exception is raised in save(), the [Django Rest Framework takes what should have been a 4xx and returns a 5xx](https://gist.github.com/twidi/9d55486c36b6a51bdcb05ce3a763e79f).

# Acceptance Criteria
Check that the user can generate new alias on the validate and return exception that resolves to 4xx error

EMMLynch:
This issue is being automatically closed as it is older than 2 years and has not been addressed. If this issue is still relevant, please feel free to reopen it with updated information.

Issue #1556, opened 2022-02-18
Do not send events from specific loggers to Sentry


say-yawn:
# About this Issue
Some of the loggers are showing duplicate events on Sentry along with the actual exception with stack trace being recorded on Sentry (see example [here](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/18254312/?query=is%3Aunresolved) and [here](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/14916780/?query=is%3Aunresolved)). Since certain loggers (like `request.summary`) are being recorded on Sentry with no stack trace while the actual exception is also being captured, we should stop sending events with no stack trace that are a duplicate of the actual exception from being logged on Sentry.

# Acceptance Criteria
`request.summary` should stop sending [events](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/?query=logger%3Adjango.security.DisallowedHost) to Sentry
`django.security.DisallowedHost` should stop sending [events](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/?query=logger%3Adjango.security.DisallowedHost) to Sentry (except for 3 hosts that need investigation #1557)

jwhitlock:
In PR #1578, we're raising ``SuspiciousOperation`` to protect from inconsistent data (setting a premium feature flag for a non-premium feature), for the side-effect of returning a 400 error to the user. We'll need to ensure that these types of bug checks 1) result in a traceback in Sentry, and 2) return a 400 rather than a 50x to our users.

PR #1578, opened 2022-02-28
validate premium to set block_list_emails=True

groovecoder:

### DIFF
--- a/emails/tests/models_tests.py
+++ b/emails/tests/models_tests.py
@@ -7,6 +7,7 @@ from unittest.mock import patch
 
 from django.conf import settings
 from django.contrib.auth.models import User
+from django.core.exceptions import ValidationError
 from django.test import (
     override_settings,
     TestCase,
@@ -236,6 +237,23 @@ class RelayAddressTest(TestCase):
         assert not valid_address(
             relay_address.address, relay_address.domain_value
         )
+
+    def test_free_user_cant_set_block_list_emails(self):
+        relay_address = RelayAddress.objects.create(user=self.user)
+        try:
+            relay_address.block_list_emails = True
+            relay_address.save()
+        except ValidationError:
+            relay_address.refresh_from_db()
+            assert relay_address.block_list_emails == False
+
+    def test_premium_user_can_set_block_list_emails(self):
+        relay_address = RelayAddress.objects.create(user=self.premium_user)
+        assert relay_address.block_list_emails == False
+        relay_address.block_list_emails = True
+        relay_address.save()
+        relay_address.refresh_from_db()
+        assert relay_address.block_list_emails == True
 
 
 class ProfileTest(TestCase):
@@ -957,3 +975,11 @@ class DomainAddressTest(TestCase):
         )
         assert deleted_address_qs.count() == 1
         assert deleted_address_qs.first().address_hash == domain_address_hash
+
+    def test_premium_user_can_set_block_list_emails(self):
+        domain_address = DomainAddress.objects.create(user=self.user)
+        assert domain_address.block_list_emails == False
+        domain_address.block_list_emails = True
+        domain_address.save()
+        domain_address.refresh_from_db()
+        assert domain_address.block_list_emails == True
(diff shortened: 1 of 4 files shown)


emails/tests/models_tests.py
```diff
@@ -237,6 +238,23 @@ def test_valid_address_dupe_of_deleted_invalid(self):
...
         )
 
+    def test_free_user_cant_set_block_list_emails(self):
+        relay_address = RelayAddress.objects.create(user=self.user)
+        try:
+            relay_address.block_list_emails = True
+            relay_address.save()
+        except ValidationError:
+            relay_address.refresh_from_db()
+            assert relay_address.block_list_emails == False
```
jwhitlock, line 248: This test will pass if the exception is not raised. Please use `assertRaises` to ensure `ValiationError` is raised:

```suggestion
        relay_address.block_list_emails = True
        with self.assertRaises(ValidationError):
            relay_address.save()
        relay_address.refresh_from_db()
        assert relay_address.block_list_emails == False
```

jwhitlock (changes requested):
Looks good, thanks @groovecoder! One change needed to an existing test, and the suggested change to the new test.

I'm seeing this test fail locally and in CI:

```
emails/tests/views_tests.py::SNSNotificationTest::test_block_list_email_sns_notification
```

`self.ra` is associated with a free user, some more work is needed to upgrade the user or associate it with ``self.premium_user ``:

https://github.com/mozilla/fx-private-relay/blob/c6fbdd123683c9fe614e581bfa94c80a43fee47f/emails/tests/views_tests.py#L99-L103

There is an existing API validator ``PremiumValidatorsMixin``, so this won't have the same problem as issue #1558 when using the API.

groovecoder:
Updated to use `SuspiciousOperation` to cause a 4xx error instead of 5xx, and fixed the test.

jwhitlock:
I'm not a fan of using [SuspicousOperation](https://docs.djangoproject.com/en/4.0/ref/exceptions/#suspiciousoperation). It does return a 400, but it also logs at Error level, because it usually means someone is doing something that means they are trying to get around security. `PermissionDenied` or `ValidationError` seems right, but some additional work would be needed to return the 400 instead of a 50x.

Do we have time to move the check higher in the application logic, and make this more like a "assert this never happens" check? Or does this need to go for the sprint, and we can refactor later?

groovecoder:
I don't think this change is critical to this release - the only code-path to change this field goes thru the API which already has a validator to make sure the user is premium.

Sorry I didn’t include it in code comments or even [the other PR](https://github.com/mozilla/fx-private-relay/pull/1429) where I introduced `SuspiciousOperation` for 4xx status code exceptions. But I looked up some different ways to change exceptions to return 4xx status codes. It started to get messy and hack-y to the point that I thought it could cause more problems than it would solve for us. And then I saw that when we update to Django 3.2+ we’ll get to change these to [`BadRequest` exceptions](https://docs.djangoproject.com/en/4.0/ref/exceptions/#badrequest), which gives us a better-matching 4xx exception.

So basically:

1. I don't think this needs to land for the release - it can follow the initial release
2. We can do additional work to return 4xx status codes some other way, or we can defer that to our Django 3.2+ work

jwhitlock (approved):
The distinction I'd like to make is "did the user do something wrong, or do we have a bug?" If the user did something wrong, a Bad Request response with a 400 status code is the right thing. If we have a bug, I could go with a 40x or 50x, but I'd want to see it in Sentry.

This PR meets my criteria - if the code gets all the way to the save method with this invalid combination, then ``SuspiciousOperation`` will log at Error level, which currently shows up in Sentry. The user gets a 400, which is also consistent with the same error from the API.

We may adjust logging error responses with issue #1556. Other stuff like invalid host headers show up as ``SuspiciousOperation``, and appear as events in Sentry that we can't do anything about. And, we can adjust with Django 3.2+. So, there are opportunities to readjust our strategy.

I'll leave it up to you if you want to get this in for the release. It seems safer to have it, to preserve data consistency, then to leave it out.

groovecoder:
I'll merge it so it can ride the QA fix trains and go thru QA cycle while we're at it.

Issue #1588, opened 2022-03-01
Handle SNS notification for unsupported type


say-yawn:
# About this Issue
We are currently [logging error](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/10505337/) where the SNS notifications from AWS SNS has the notification type `None`. 

# Acceptance Criteria
- We should log the `eventType` to see if there is specific type of events that is passing `notificationType` to be sent with `None` and handle them appropriately.

Related to https://github.com/mozilla/fx-private-relay/issues/1587

jwhitlock:
Let's see if we can determine the shape of these unknown notifications with the additional information.

PR #1589, opened 2022-03-01
Log unsupported sns types 1588

### LINKED ISSUE #1587: Shlex errors logged
Since errors logged from `validate_sns_header` is happening before we verify that the notification is coming from a valid AWS SNS, we want to escape any shell executable that can be passed that is used on logging the error like [here](https://github.com/mozilla/fx-private-relay/blob/644f3f4ddf58302a2f4a92fc15a83cf4e762fe3a/emails/views.py#L214)

groovecoder:
This PR fixes #1587 #1588.


jwhitlock (approved):
Thanks @groovecoder!

PR #1592, opened 2022-03-02
Cache has_premium property to fix slow requests

maxxcrawford:
Issue: I was having slow page load times (~12 seconds) when loading a premium account with a high number of aliases. 

As introduced in 56755320375d5dcc799d3f8c92b3d89e43883069, the alias-card loop includes additional "checks" when displaying certain UI. 

To test: 
- Create account that is premium with 40+ aliases
- Log in
- **Expected:** Page load time should be under ~2 seconds

Try this same setup from `main`: 
- **Expected:** The page load should be extremely long.

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -1,5 +1,6 @@
 from collections import namedtuple
 from datetime import datetime, timedelta, timezone
+from functools import lru_cache
 from hashlib import sha256
 import logging
 import random
@@ -252,6 +253,7 @@ class Profile(models.Model):
         return f'@{self.subdomain}.{settings.MOZMAIL_DOMAIN}'
 
     @property
+    @lru_cache(maxsize=None)
     def has_premium(self):
         # FIXME: as we don't have all the tiers defined we are over-defining
         # this to mark the user as a premium user as well


emails/models.py
```diff
@@ -252,6 +253,7 @@ def custom_domain(self):
         return f'@{self.subdomain}.{settings.MOZMAIL_DOMAIN}'
 
     @property
+    @lru_cache(maxsize=None)
```
maxxcrawford, line 256: Not 100% on the `maxsize=None` bit. Please lmk if there is more needed.

maxxcrawford (commented):
Self-review

emails/models.py
jwhitlock: I don't know if `@lru_cache` works well on class methods or properties. If we were on Python 3.8 you could use `@cached_property`.

I think it would make sense to calculate once and pass as template context. I'm taking a look at both caching and the context option.

groovecoder: Hmm ... I think @jwhitlock may know this better than I do. The boolean return value is small, but the `self` object passed in as an argument is large (the entire `Profile` object). There might be a better way to do this.

E.g., I ran a profile (with `django-silk`) of the dashboard page and saw waaaay too many queries to `socialaccount_socialaccount`. This is caused any time we reference the `Profile.fxa` property. It may be a better overall performance gain to optimize the `Profile.fxa` instead.

@jwhitlock - what's your favorite django model object property caching/performance tweak?

jwhitlock: Web search to the rescue! Let's use [Django's @cached_property](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.functional.cached_property):

```python
from django.utils.functional import cached_property

...

@cached_property
def has_premium(self):
    ...
```

This adds a problem of cache invalidation, but the slowness seems like a bigger issue right now. I think we can do a light version by adding a `del` to the `save()` method:

```python
        ret = super().save(*args, **kwargs)
        try:
            del self.has_premium  # clear the cached value
        except AttributeError:
            pass # the cached value hasn't been accessed
```

Also - great work tracing down the source of the slowness! Full-stack @maxxcrawford!

I think it would be challenging to add it to the context instead, so I'm not going down that path.

jwhitlock: @groovecoder Looking at that traceback, we may need to ensure the related profiles are fetched before rendering the template. Which may speed up the `has_premium` call, and avoid getting into caching.

PR #1608, opened 2022-03-04
Optimize profile object

groovecoder:
This PR fixes some performance issues that are especially pronounced with some of the new changes to the profile page.

* Changes the `profile` view to call `prefetch_related` for the profile's FXA `SocialAccount` set, and changes the `Profile.fxa` property to use this pre-fetched `socialaccount_set` queryset. This prevents a number of duplicate DB queries everywhere else that templates were accessing the `fxa` or `has_premium` properties of the profile.
* Adds `Profile.relay_addresses` and `Profile.domain_addresses` as `cached_property`, and updates other `Profile` methods to use these properties. This prevents a number of duplicate DB queries.
* Uplifts `user_profile` and `user_has_premium` template vars into `base.html`, and changes templates to use these vars rather than access the `Profile` object directly. This prevents a number of duplicate DB queries in templates that were accessing `Profile` in loops.

How to test:

* [x] All tests should continue to pass
* [ ] A user should still be able to complete all user flows (particularly on their profile) as expected.


jwhitlock (approved):
Thanks @groovecoder! This is a big step forward!

If you have time, please clean out the `silk` stuff rather than comment it out.

PR #1611, opened 2022-03-05
Store attachments as a list

### LINKED ISSUE #1559: If multiple attachments have the same filename, only one is forwarded
Attachments are indexed in a dictionary by filename:

https://github.com/mozilla/fx-private-relay/blob/04cfa86602af0a9ac966fa414b6288cffadea519/emails/views.py#L782-L785

If two attachments have the same filename, then the last one "wins".

The filename is optional, so if multiple attachments have no filename, the last one "wins" as well.

We can fix this by changing `attachments` to a different data structure, such as a list of `(filename, data)` tuples (or even `(filename, mimetype, data)`) tuples. We could also start by counting the number of parts in the incoming message versus the outgoing message, to see how significant the issue is.

jwhitlock:
Store and load attachments as a list of tuples `(name, data_stream)`, rather than a dict `{name: data_stream}`. This allows multiple attachments to have the same name (or `None` if filenames are omitted).

This PR fixes #1559

How to test:

- Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

### DIFF
--- a/emails/utils.py
+++ b/emails/utils.py
@@ -134,7 +134,7 @@ def _add_body_to_message(msg, message_body):
 
 def _add_attachments_to_message(msg, attachments):
     # attach attachments
-    for actual_att_name, attachment in attachments.items():
+    for actual_att_name, attachment in attachments:
         # Define the attachment part and encode it using MIMEApplication.
         attachment.seek(0)
         att = MIMEApplication(attachment.read())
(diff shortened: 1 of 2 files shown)


groovecoder:
I rebased this on the latest `main` to fix the tests.

say-yawn (approved):
LGTM. Can we wait to merge this? I am waiting for some changes for tomorrow's release and I would like this to bake in Stage before pushing to Prod.

say-yawn:
Merged because the release date is Thursday, not tomorrow.

PR #1631, opened 2022-03-10
cache the DomainAddress.user_profile property

groovecoder:
The new `alias-card.html` template references `alias.full_address` 6 times. `Domain.full_address` calls `self.user_profile` which calls `Profile.objects.get()`. So, each domain alias in the dashboard makes 6 separate DB queries to `emails_profile` to fetch the profile.
image

This PR changes the `DomainAddress` model to cache its `user_profile` property, so the calls to `full_address` no longer makes so many DB queries.
image

How to test:

* [ ] Make sure dashboard page still works as expected.
* [ ] Make sure replies still work as expected.

### DIFF
--- a/privaterelay/templatetags/relay_tags.py
+++ b/privaterelay/templatetags/relay_tags.py
@@ -1,6 +1,7 @@
 from django import template
 from django.conf import settings
 
+from emails.models import RelayAddress
 from emails.utils import get_email_domain_from_settings
 
 from ..utils import get_premium_country_lang
@@ -51,3 +52,10 @@ def premium_plan_price(accept_lang, cc=None):
 def premium_subscribe_url(accept_lang=None, cc=None):
     plan_id = premium_plan_id(accept_lang, cc)
     return f'{settings.FXA_SUBSCRIPTIONS_URL}/products/{settings.PREMIUM_PROD_ID}?plan={plan_id}'
+
+
+@register.filter
+def full_address(alias, user_profile):
+    if type(alias) == RelayAddress:
+        return '%s@%s' % (alias.address, alias.domain_value)
+    return '%s@%s.%s' % (alias.address, user_profile.subdomain, alias.domain_value)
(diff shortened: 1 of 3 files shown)


groovecoder:
Updated `alias-card.html` to also use a new `full_address` filter that receives the single (cached) `user_profile` object as an argument. This eliminates ALL the duplicate queries that were caused per domain address in alias card.

maxxcrawford (approved):
Under 1000ms for a 45+ filter account

privaterelay/templatetags/relay_tags.py
```diff
@@ -51,3 +52,10 @@ def premium_plan_price(accept_lang, cc=None):
 def premium_subscribe_url(accept_lang=None, cc=None):
     plan_id = premium_plan_id(accept_lang, cc)
     return f'{settings.FXA_SUBSCRIPTIONS_URL}/products/{settings.PREMIUM_PROD_ID}?plan={plan_id}'
+
+
+@register.filter
+def full_address(alias, user_profile):
+    if type(alias) == RelayAddress:
+        return '%s@%s' % (alias.address, alias.domain_value)
+    return '%s@%s.%s' % (alias.address, user_profile.subdomain, alias.domain_value)
```
say-yawn, line 61: I was hoping we don’t use full_address  template tag  because I wanted to prevent the formatting of the strings to be done in multiple different function. Wondering if we can use the `cached_property` for the `full_address` property too?

say-yawn (changes requested):
Changes mostly look good. One recommendation to use cached_property instead of template tag to deal with full_address of the address.

privaterelay/templatetags/relay_tags.py
groovecoder: Putting `cached_property` on `Domain.full_address` didn't work when I tried it. 😢 Near as I can tell (though I didn't debug thru the Django code itself), calling a `cached_property` from different points in a template passes a different `self` argument value to the decorator, so it doesn't return from the cache - it makes the full DB query each time.

groovecoder:
Closing this in favor of working on the React UI which will use the API endpoints for showing the dashboard.

PR #1632, opened 2022-03-10
use profile_set.prefetch_related during sns-inbound

groovecoder:
The `wrapped_email.html` template references `Profile.has_premium` 3 times. In https://github.com/mozilla/fx-private-relay/pull/1608, we updated the dashboard view to add `prefetch_related` so that the `socialaccount_set` (i.e., the `fxa` property which is used in `has_premium`) only triggers a single DB query.

In this case, it saves us 3 DB queries ...

## OLD:
image

## New Hotness:
image

How to test:
(Probably on dev server)
- [ ] Check that emails flow thru the server as expected.

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -377,7 +377,7 @@ def _sns_message(message_json):
         # RelayAddress or DomainAddress types makes the Rustacean in me throw
         # up a bit.
         address = _get_address(to_address, to_local_portion, to_domain_portion)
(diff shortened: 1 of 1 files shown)


jwhitlock (approved):
Sorry for the delay! This is a minimal change, and looks safe to me. The new SQS polling method will continue calling this code, so this optimization is still useful.

PR #1633, opened 2022-03-10
Return 404 if the S3 object no longer exists

### LINKED ISSUE #1590: Handle S3 client error NoSuchKey
# About this Issue
We have [errors](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/18255993/) that shows that email object is no longer in S3 ([NoSuchKey](https://aws.amazon.com/premiumsupport/knowledge-center/404-error-nosuchkey-s3/)).

# Acceptance Criteria
- Log the key and the actions we did to the key to have a history of why we are trying to relay an email that is no longer in S3 (was it never stored in S3)
- Return 4xx error instead of 5xx since we cannot retry to send an email that is not in S3 [here](https://github.com/mozilla/fx-private-relay/blob/main/emails/views.py#L591-L594) and [here](https://github.com/mozilla/fx-private-relay/blob/main/emails/views.py#L423-L426).

say-yawn:
# About this PR
We are getting [S3 client error](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/18255993/?query=is%3Aunresolved) with `NoSuchKey` code in the error response, likely because we have already processed the email but we are not responding to the SNS in time resulting in SNS retry. If the email is not in S3, instead of returning a 503, we should return a 404 so the SNS does not send us the retry ping.

Fixes #1590

### DIFF
--- a/emails/utils.py
+++ b/emails/utils.py
@@ -330,6 +330,9 @@ def remove_message_from_s3(bucket, object_key):
         )
         return response.get('DeleteMarker')
     except ClientError as e:
-        logger.error('s3_client_error_delete_email', extra=e.response['Error'])
+        if e.response['Error'].get('Code', '') == 'NoSuchKey':
+            logger.error('s3_object_does_not_exist', extra=e.response['Error'])
+        else:
+            logger.error('s3_client_error_delete_email', extra=e.response['Error'])
         incr_if_enabled('message_not_removed_from_s3', 1)
     return False
--- a/emails/utils.py
+++ b/emails/utils.py
@@ -331,7 +331,7 @@ def remove_message_from_s3(bucket, object_key):
         return response.get('DeleteMarker')
     except ClientError as e:
         if e.response['Error'].get('Code', '') == 'NoSuchKey':
-            logger.error('s3_object_does_not_exist', extra=e.response['Error'])
+            logger.error('s3_delete_object_does_not_exist', extra=e.response['Error'])
         else:
             logger.error('s3_client_error_delete_email', extra=e.response['Error'])
         incr_if_enabled('message_not_removed_from_s3', 1)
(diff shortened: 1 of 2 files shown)


emails/utils.py
```diff
@@ -330,6 +330,9 @@ def remove_message_from_s3(bucket, object_key):
         )
         return response.get('DeleteMarker')
     except ClientError as e:
-        logger.error('s3_client_error_delete_email', extra=e.response['Error'])
+        if e.response['Error'].get('Code', '') == 'NoSuchKey':
+            logger.error('s3_object_does_not_exist', extra=e.response['Error'])
```
jwhitlock, line 334: I think we'll see this error twice  - once when we try to read the email from S3 in `_sns_response`, and once when try to delete the item. That's probably OK.

emails/views.py
```diff
@@ -443,6 +443,9 @@ def _sns_message(message_json):
             message_json
         )
     except ClientError as e:
+        if e.response['Error'].get('Code', '') == 'NoSuchKey':
+            logger.error('s3_object_does_not_exist', extra=e.response['Error'])
```
jwhitlock, line 447: maybe we should use `logger.exception` so that a traceback is included, and we know which line raised the error. But I'm not sure if that will be logged cleanly, or as an ugly multi-line entry. So, let's stay on `logger.error`.

jwhitlock (approved):
Thanks @say-yawn - 404 seems right to me, and to stop trying to process this email.

emails/utils.py
say-yawn: I'll name this different errors so we can distinguish between the two.

emails/views.py
say-yawn: We can do this as a follow-up.

jwhitlock (approved):

PR #1635, opened 2022-03-11
Work around not-loaded Swedish and Frysian locales

Vinnl:
This PR fixes/works around Django's LocaleMiddleware not loading `fy-NL` localisations when the user's Accept-Language is set to `fy` (it _would_ do so if it was the other way around, i.e. it would load `fy` if the Accept-Language was set to `fy-NL` though).

It's more a workaround than a "proper" fix, for these reasons:

- It's probably not feasible (especially not for me) to change this behaviour upstream - it might not even be considered desirable by them/other locales, and might be confusing for their other downstreams if it were to change.
- Renaming/moving the locales in Pontoon supposedly is problematic too.
- The React front-end _does_ handle this correctly (by virtue of being able to use Fluent's `@fluent/langneg` library), so for most of the UI, it's only needed temporarily. On the other hand, we'll likely keep needing this for the email templates.

How to test:

Unfortunately I don't know how to run the Relay Docker image locally and being able to visit it from your browser. However, you can run the command locally manually, visit the site with your `Accept-Language` set to `fy`, and see the Frysian translation.

You can also verify that the command executes successfully in Docker by building the image (`docker build --tag my/relay .`), then opening a shell into that image (`sudo docker run --interactive --tty --rm --entrypoint=bash my/relay`), and verifying that the `fy-NL` directory has indeed been copied to `fy`: `ls /app/privaterelay/locales/ | grep fy`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,12 @@ COPY --from=builder --chown=app /app/static ./static
 
 COPY --chown=app ./requirements.txt /app/requirements.txt
 RUN pip install -r requirements.txt
+# When the user's Accept-Language is set to `fy`, Django's LocaleMiddleware
+# doesn't load `fy-NL`. This is a workaround to force the Frysian and Swedish
+# localisations to load anyway when appropriate.
 COPY --chown=app . /app
+RUN cp -r /app/privaterelay/locales/fy-NL/ privaterelay/locales/fy/
+RUN cp -r /app/privaterelay/locales/sv-SE/ privaterelay/locales/sv/
 COPY --chown=app .env-dist /app/.env
 
 RUN mkdir -p /app/staticfiles


Dockerfile
```diff
@@ -34,7 +34,12 @@ COPY --from=builder --chown=app /app/static ./static
 
 COPY --chown=app ./requirements.txt /app/requirements.txt
 RUN pip install -r requirements.txt
+# When the user's Accept-Language is set to `fy`, Django's LocaleMiddleware
+# doesn't load `fy-NL`. This is a workaround to force the Frysian and Swedish
+# localisations to load anyway when appropriate.
 COPY --chown=app . /app
+RUN cp -r /app/privaterelay/locales/fy-NL/ privaterelay/locales/fy/
```
flodolo, line 41: Would a link be enough?

```suggestion
RUN ln -s /app/privaterelay/locales/fy-NL/ privaterelay/locales/fy/
```

Vinnl: Looks point, looks like it would be: d18d8224

Dockerfile
```diff
@@ -34,7 +34,12 @@ COPY --from=builder --chown=app /app/static ./static
 
 COPY --chown=app ./requirements.txt /app/requirements.txt
 RUN pip install -r requirements.txt
+# When the user's Accept-Language is set to `fy`, Django's LocaleMiddleware
+# doesn't load `fy-NL`. This is a workaround to force the Frysian and Swedish
+# localisations to load anyway when appropriate.
 COPY --chown=app . /app
+RUN ln --symbolic /app/privaterelay/locales/fy-NL/ privaterelay/locales/fy
+RUN ln --symbolic /app/privaterelay/locales/sv-SE/ privaterelay/locales/sv
```
groovecoder, line 42: In my PR to add premium product plan IDs for Sweden and Finland, I also found that Swedish had this problem. I tried to fix it by adding those language codes to Django's `LANGUAGES` map (like we do for Chinese), but I think this approach seems more robust and reliable. I also discovered that Spanish has this same issue? So if a user's `Accept-Language` is set to `es`, we end up showing English.

question: Can we expand this solution to cover those languages too?
```suggestion
RUN ln --symbolic /app/privaterelay/locales/fy-NL/ privaterelay/locales/fy
RUN ln --symbolic /app/privaterelay/locales/sv-SE/ privaterelay/locales/sv
RUN ln --symbolic /app/privaterelay/locales/es-ES/ privaterelay/locales/es
RUN ln --symbolic /app/privaterelay/locales/zh-CN/ privaterelay/locales/zh
```

groovecoder (changes requested):

Dockerfile
Vinnl: Yeah we can certainly add more languages; Sweden and Frisian were just the ones I was aware of.

I'm not sure about the issues with Chinese, but reading [the comment](https://github.com/mozilla/fx-private-relay/blob/3a335a266b7058c5c3921d773850f09926203559/privaterelay/settings.py#L442-L443), it sounds like it might be a different issue (i.e. `zh-tw` and `zh-cn` not being known to Django)? Could you confirm that it's the same issue? Then I'll add them (both, I assume?) tomorrow.

groovecoder: Oops, you're right. Chinese was a separate issue.

groovecoder: Although ... this is ALSO affects Chinese. If you set your browser language to `zh` it will render English rather than "fall back" to, e.g., `zh-CN`. But if the `zh -> zh-CN` symbolic link exists, then `zh` WILL fall "back" to `zh-CN`. So it's probably worth keeping this AND keeping the fix in `settings.py`.

Vinnl: OK, I was going to apply this for all locales with a country code, but now I'm getting nervous about the ones that have multiple locales for the same language (i.e. `zh-` and `es-`). How would someone in Taiwan feel about getting served `zh-CN` instead of `zh-TW` when their language is set to `zh`? Or someone in Mexico about getting served `es-ES`?

There's also `pt-BR`, which is our only locale for Portuguese, but I do vaguely recall someone saying that many `pt-PT` people would not want `pt-BR` and might even be a bit sensitive about it (since `pt-PT` can be considered the "original", I guess).

All of which is to say: maybe we should just merge these, since it's pretty safe for Swedish and Frisian, and consult with someone more attuned to locale-specific sensitivities before doing the same for other locales?

flodolo: This is probably a good reference (typically done by size of population, in case of pt or zh) https://github.com/mozilla/bedrock/blob/9ad6bab20acdca88505ff3614384b34f24b2eec2/bedrock/settings/base.py#L314

Vinnl: Thanks @flodolo! Based on that, I set `pt-BR` as the default for `pt` and `es-ES` for `es`: 1e8cd376

Given that that doesn't list a default for `zh`, I didn't do that here either.

Vinnl:
Rebased to resolve conflicts.

groovecoder (approved):

PR #1636, opened 2022-03-11
Add coveralls upload

jwhitlock:
This reformats the `pytest` command to just use the `pytest-cov` version instead of running under `coverage run`.

It then caches the `.coverage` file, and adds a step that uploads it to coveralls.io. The tool expects to run in a git checkout, so it doesn't work well to run it inside the docker container that runs the tests.

How to test:
- See if it uploaded a coverage report to https://coveralls.io/github/mozilla/fx-private-relay

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,10 +82,14 @@ jobs:
               -c \
                 'mkdir -p /tmp/test-results/pytest && \
                  mkdir -p /tmp/test-results/coverage && \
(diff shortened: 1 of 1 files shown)


jwhitlock:
The new test steps pass, and coverage is at https://coveralls.io/github/mozilla/fx-private-relay

I renamed `test` to `test_backend`, to match the language in PR #1518. @Vinnl, we'll still need some gymnastics to combine the two coverage runs into a single coverage. I'm happy to help with that.

PR #1655, opened 2022-03-17
Set up preview deployments with mocked APIs

Vinnl:
# New feature description

This will create a preview deployment of the (React) UI whenever we create a pull request, with the API mocked out (because the back-end is harder to willy-nilly provision). This allows us (both engineers, as well as product and design) to easily preview what those changes will look like and whether they satisfy (non-data-related) requirements.

One thing that might be good to do is for the other engineers to create a Netlify account as well, and then to set up a team in Netlify so that other people can manage that as well. (Not that there's much to manage, but still.)

# Screenshot (if applicable)

# How to test

# Checklist

- [ ] I've added or updated relevant docs in the docs/ directory. - Not sure if any are needed, but let me know if there's something you'd like docs about.


Vinnl:
Ah, these notifications were enabled:

image

I've removed the PR comments when starting the build, and when it is successful. Comments on fail seem useful.

image

frontend/src/apiMocks/handlers.ts
```diff
@@ -0,0 +1,237 @@
...
+  };
+
+  addGetHandler("/accounts/logout", (_req, res, ctx) => {
+    return res();
+  });
+  addGetHandler("/api/v1/runtime_data", (_req, res, ctx) => {
+    const runtimeData: RuntimeData = {
+      FXA_ORIGIN: "https://example.com",
+      GOOGLE_ANALYTICS_ID: "UA-123456789-0",
+      PREMIUM_PRODUCT_ID: "prod_123456789",
```
maxxcrawford, line 57: Question: Can we move these IDs to env variables? Not that I expect stuff to change often, but its' a risk putting the static values here.

frontend/src/apiMocks/handlers.ts
```diff
@@ -0,0 +1,237 @@
...
+        plan_country_lang_mapping: {
+          nl: {
+            nl: {
+              id: "price_1JmROfJNcmPzuWtR6od8OfDW",
+              price: "€0,99",
+            },
+          },
+        },
+        premium_countries: ["nl"],
+        premium_available_in_country: true,
```
maxxcrawford, line 69: Do we need a US version of this too?

maxxcrawford (approved):
Approving with a few questions. 

Overall: This is an AWESOME feature and we needed it since the project started. Great job. 🎉

frontend/src/apiMocks/handlers.ts
```diff
@@ -0,0 +1,237 @@
+import { rest, RestHandler, RestRequest } from "msw";
+import { CustomAliasData, RandomAliasData } from "../hooks/api/aliases";
+import { ProfileData } from "../hooks/api/profile";
+import { RuntimeData } from "../hooks/api/runtimeData";
+import { mockIds, domainaddresses, profiles, relayaddresses } from "./mockData";
+
+export function getHandlers(
+  defaultMockId: null | typeof mockIds[number] = null
+): RestHandler[] {
+  const handlers: RestHandler[] = [];
```
Vinnl, line 10: 3. Here we define the handlers. They're loaded by the Service Worker, and respond to the listed HTTP requests.

frontend/src/apiMocks/handlers.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  const getMockId = (req: RestRequest): typeof mockIds[number] | null => {
+    const authHeader = req.headers.get("Authorization");
+    if (typeof authHeader !== "string") {
+      return defaultMockId;
+    }
+    const token = authHeader.split(" ")[1];
```
Vinnl, line 17: 4. We then look at the `Authorization` header, which normally (when running in `dev` mode) includes a user's API token to authenticate them, for a `mockId` which specifies which mock data we want to load.

frontend/src/apiMocks/mockData.ts
```diff
@@ -0,0 +1,148 @@
+import { CustomAliasData, RandomAliasData } from "../hooks/api/aliases";
+import { ProfileData } from "../hooks/api/profile";
+
+export const mockIds = ["empty", "onboarding", "some", "full"] as const;
```
Vinnl, line 4: 5. The available mockIds are listed here. By enumerating them here, we can ensure that every handler has data to return for every listed ID. (Also, you can append `?mockId=<mockId>` to a request to automatically fake sign in as the given user.)

frontend/src/apiMocks/mockData.ts
```diff
@@ -0,0 +1,148 @@
...
+    bounce_status: [true, "soft"],
+    date_subscribed: "2020-04-09T00:00:00.000Z",
+    has_premium: true,
+    id: 0,
+    next_email_try: "2020-04-09T00:00:00.000Z",
+    onboarding_state: 3,
+    server_storage: true,
+    subdomain: "mydomain",
+  },
+};
```
Vinnl, line 55: 6. This is an example of the default mock data for the `/api/v1/profiles/` endpoint. Every key is a different mock ID, and it's typed, so if the API response changes, TypeScript will warn us about updating these mocks. Below this is the data for the other endpoints. As you can see, the user identified by the mockId `empty` has no random aliases, neither does the one identified by `onboarding`, whereas the user identified by `some` has two.

frontend/src/pages/mock/login.page.tsx
```diff
@@ -18,6 +19,16 @@ const MockLogin: NextPage = () => {
...
+    if (process.env.NEXT_PUBLIC_MOCK_API === "true") {
+      // When the API is mocked out, the API tokens are fake as well:
+      const mockIdsAsTokens: UsedToken[] = mockIds.map((id) => ({
+        lastUsed: 0,
+        token: id,
+        user: `${id}@example.com`,
+      }));
+      setUsedTokens(mockIdsAsTokens);
+      return;
+    }
```
Vinnl, line 31: In `dev` mode, the fake login page lists API tokens previously used. When running against a mocked backend, those tokens are not valid, and we just list all the available mock IDs.

Vinnl (commented):
Self-review to give a small tour of how this works (numbers correspond to inline comments):

1. In a `postinstall` script, [Mock Service Worker](https://mswjs.io/) generates a... Service Worker. It's generated in `/frontend/public`, which means that it'll be included as a static file with the front-end deployment.
2. This Service Worker is then initialised on page load. It intercepts all HTTP requests, passing them to...
3. out handlers, which specify, for every API endpoint, what response to return. The specific response is determined by...
4. the auth token, which normally is used in the dev server to pass a user-specific token known to the back-end, but in this case is just a hardcoded string identifying one of the...
5. `mockIds`. Which are then used as keys for
6. the mock data. Also note that this data is annotated with the API types, so if the API adds a new field, TypeScript will warn us to update these mocks.

frontend/src/apiMocks/handlers.ts
Vinnl: We can, but if you look at the actual IDs, these are fake values, and their actual value doesn't matter. So should be too risky, I think.

PR #1660, opened 2022-03-18
Set up visual regression testing using Playwright

Vinnl:
# New feature description

This adds visual regression testing. I've [run it several times in a row](https://github.com/mozilla/fx-private-relay/actions/runs/2004708839), and it does not appear to be flaky - a common problem with visual regression tests. That said, updating the screenshots (in case of intended changes) is a bit cumbersome, because you'll need to trigger a run in CI to make sure the environment (resolution, zoom level, available fonts, etc.) are consistent. So consider this PR a proof-of-concept: it shows that we could do this, but it's a totally valid option to discard it and choose not to go with this. Let me know what you think.

# Screenshot (if applicable)

Well... They're in the diff :grin: 

# How to test

You could try running it locally (don't forget to `npm install`; also `npx playwright install` to download dev versions of the relevant browsers): `npm run ui-test`. The first time it will probably fail, because it's generating the screenshots on your machine. (Assuming you're not on Linux; otherwise it will compare the screenshots to the reference ones, which I'd expect to slightly differ.) Subsequent runs _should_ succeed, however: it will generate the screenshots again, and compare them to the ones it generated before. Now make a visual change to one of the elements in one of the screenshots, run the tests again, and watch them fail, and a test report being generated. (And if this turns out flaky, let me know! We'd probably want to close this if so.)

# Checklist


frontend/src/pages/faq.page.tsx
```diff
@@ -10,308 +10,294 @@ const Faq: NextPage = () => {
 
   return (
     <Layout theme="free">
-      <main>
-        <div className={styles["faq-page"]}>
```
Vinnl, line 14: This looks like a very large diff, but literally the only thing I did was remove this `<div>` and move the class `.faq-page` to `<main>` - but of course, that changed the indentation.

The reason I did this was because the `<div>` was superfluous, and while it had a `height: 100%` set on it (to make sure the footer sticks to the bottom), the containing `<main>` did not have that. Since a percentage height is relative to the container, this (I think) caused the height of the `<div>` to change a little bit during page load on Firefox, causing things to jump around and thereby the screenshot to be of a different height depending on how soon it was taken. (Resulting in a flaky test.) Since it didn't do anything, I just removed it.

Vinnl:
Closing this for now. We can always resurrect this if we keep hitting many visual regressions :)

PR #1669, opened 2022-03-22
Correctly communicate authn status to add-on

Vinnl:
I'd previously assumed that all attributes on
<firefox-private-relay-addon> were set by the add-on and read by
the website (as opposed to <firefox-private-relay-addon-data>,
whose attributes are set by the website and read by the add-on).
However, apparently that is not true for data-user-logged-in: that
attribute is set by the website and read by the add-on. What's
more, it's set to the strings "True" or "False" (like booleans in
Python), rather than "true" or "false".

That means that when the user logged out, the add-on did not know
about it, when running the React front-end. This fixes that.

How to test: Run the website with `SERVE_REACT` set to `True` and `npm run watch`. Run the add-on locally. Sign in, verify that the add-on shows you as signed in, then sign out and verify that the add-on shows you as having signed out.

- [ ] I've added a unit test to test for potential regressions of this bug. N/A - would be an integration test, for which we don't have the infra.


maxxcrawford (approved):
Logged in/out as expected! 🎉

PR #1678, opened 2022-03-23
Generate and upload frontend coverage

jwhitlock:
Generate a coverage report with jest, and output coverage and test results. Merge the test results with the backend results and upload to coveralls.io.

PR #1679, opened 2022-03-24
Add frontend coverage to CircleCI, coveralls.io

jwhitlock:
This PR runs frontend code coverage with some minimum levels, and merges it with the backend code coverage when sending to coveralls.io. This results in a drop in total code coverage, which coveralls.io will complain about.

I switched from the [Caching](https://circleci.com/docs/2.0/caching/) feature to the [Workspace](https://circleci.com/docs/2.0/workspaces/) feature for test results. The workspace is for files needed by later jobs in the same test run, while the cache is better suited for files needed in later test runs. I'm using `/tmp/workspace` as the base folder for workspace files, and placing test files in `/tmp/workspace/test-results`. This will allow us to put other stuff in the workspace in the future. The `docker.tar` file looks promising, and maybe the built frontend files.

I expanded the frontend test command:
* Use [--ci](https://jestjs.io/docs/cli#--ci) mode, failing on previous snapshots
* Use [--coverage](https://jestjs.io/docs/cli#--coverageboolean), to generate coverage reports. This was probably in an earlier version of the Big One, but was dropped at some point.
* Use ``--coverageDirectory`` to write coverage files directly to ``/tmp/workspace/test-results/frontend-coverage``. This is undocumented as a CLI option, but [is documented](https://jestjs.io/docs/configuration#coveragedirectory-string) as a config option.
* Use the ``--reporters=default``, so test results are shown in CI output

The [coveralls-python documentation](https://coveralls-python.readthedocs.io/en/latest/usage/multilang.html) (thanks @Vinnl!) suggests using the Ruby tool [coveralls-lcov](https://github.com/okkez/coveralls-lcov) to convert the JS tests to the coveralls JSON format. I added another job ``convert_frontend_coverage`` that runs the Ruby image and this tool. The ``lcov`` file paths are relative to the ``frontend`` folder, and I used ``sed`` to add that to the file paths so that all paths will be relative to the repository. The output is added to the workspace, so that coveralls-python can gather the Python test results and merge in the frontend results before uploading to coveralls.io.

Finally, I added some lower bounds for frontend coverage, based on the current coverage levels:
* Statements: 40% (currently 47.92%)
* Branches: 40% (currently 49.16%)
* Functions: 30% (currently 33.33%)
* Lines: 40% (currently 47.98%)

This is set in ``frontend/jest.config.js`` with [coverageThreshold](https://jestjs.io/docs/configuration#coveragethreshold-object), which has a ton of options for per-file coverage threshholds. I also put [collectCoverageFrom](https://jestjs.io/docs/configuration#collectcoveragefrom-array) in this file, as suggested by [Configuring code coverage in Jest, the right way](https://www.valentinog.com/blog/jest-coverage/), which was really helpful.

As with any build changes, none of this worked the first 10 times. The messier version is in https://github.com/mozilla/fx-private-relay/pull/1678.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,13 +111,51 @@ jobs:
           working_directory: ./frontend/
       - run:
           name: Test Code
-          command: npm test -- --reporters=jest-junit
+          command: |
+            mkdir --parents /tmp/workspace/test-results/frontend-coverage
+            npm test -- \
+              --ci \
+              --coverage \
+              --coverageDirectory=/tmp/workspace/test-results/frontend-coverage \
+              --reporters=jest-junit \
+              --reporters=default
           working_directory: ./frontend/
       - store_test_results:
           path: frontend/junit.xml
       - store_artifacts:
-          path: frontend/coverage/
+          path: /tmp/workspace/test-results/frontend-coverage
           destination: frontend_test_coverage
+      - persist_to_workspace:
+          root: /tmp/workspace
+          paths:
+            - test-results/frontend-coverage
+
+  convert_frontend_coverage:
+    docker:
+      - image: cimg/ruby:2.7.2
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - checkout
+      - run: gem install coveralls-lcov
+      - run:
+          name: Add frontend prefix to lcov.info
+          command: |
+            sed 's|^SF:|SF:frontend/|' \
+              /tmp/workspace/test-results/frontend-coverage/lcov.info \
+              > /tmp/workspace/test-results/frontend-coverage/lcov-prefixed.info
+      - run:
+          name: Generate coveralls.json report
+          command: |
+            coveralls-lcov \
+              --verbose --dry-run \
+              /tmp/workspace/test-results/frontend-coverage/lcov-prefixed.info \
+              > /tmp/workspace/test-results/frontend-coverage/coveralls.json
+      - persist_to_workspace:
+          root: /tmp/workspace
+          paths:
+            - test-results/frontend-coverage/lcov-prefixed.info
+            - test-results/frontend-coverage/coveralls.json
 
   test_backend:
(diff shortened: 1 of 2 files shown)


jwhitlock:
The coverage files are available as [artifacts](https://8967-190652903-gh.circle-artifacts.com/0/frontend_test_coverage/lcov-report/index.html), and the combined coverage in [coveralls.io](https://coveralls.io/builds/47638114).

.circleci/config.yml
```diff
@@ -111,13 +111,51 @@ jobs:
...
           name: Test Code
-          command: npm test -- --reporters=jest-junit
+          command: |
+            mkdir --parents /tmp/workspace/test-results/frontend-coverage
+            npm test -- \
+              --ci \
+              --coverage \
+              --coverageDirectory=/tmp/workspace/test-results/frontend-coverage \
+              --reporters=jest-junit \
+              --reporters=default
```
Vinnl, line 121: I'd set the `--reports` and `--coverage` options directly in `jest.config.js`, since it's fine having those enabled locally as well (at least at this point in time, when the overhead is still pretty low), and having CI be less of a snowflake will make the effects of future modifications to the test script more predictable.

Vinnl (approved):
Fantastic work @jwhitlock, happy to see the frontend coverage in the report as well. If you want to set the threshold a bit higher, we can also mark the `apiMocks` directory as ignored for coverage, since there aren't really sensible unit tests to write for those.

jwhitlock:
Omitting `apiMocks` improved the coverage and raised the thresholds:

* Statements: 40% → 50%  (47.92% → 53.06%)
* Branches: 40% → 50% (49.16% → 53.28%)
* Functions: still 30% (33.33% → 36.3%)
* Lines: 40% → 50% (47.98% → 53.11%)

The combined coverage is now 59.91%, a decrease of -5.5%, versus the old decrease of 8%.

PR #1680, opened 2022-03-24
Add missing events to metrics docs

Vinnl:
We didn't have a new data review and only added new events that
fall into the same category of data as data we were already
collecting (Category 2, "interaction data"), but it's both good
hygiene to attempt to keep this up-to-date, as well as possibly
interesting to people; hence the update.

I wasn't quite sure who to select as reviewers, so you're the lucky ones :)

### DIFF
--- a/METRICS.md
+++ b/METRICS.md
@@ -1,4 +1,4 @@
-_Last updated: May 20 2020_
+_Last updated: March 24 2022_
 &nbsp;
 &nbsp;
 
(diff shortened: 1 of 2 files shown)


METRICS.md
```diff
@@ -152,13 +154,87 @@ We collect data for the following events:
...
+  * `eventAction` : View
+  * `eventLabel` : Recruitment text
+
+- When the recruitment link is clicked
+  * `hitType` : event
+  * `eventCategory` : Recruitment
+  * `eventAction` : Engage
+  * `eventLabel` : Recruitment text
+
+### NPS/CSAT surveys
```
groovecoder, line 173: Let's write these out since here, since this is the first time we introduce them in the doc: "Net Promoter Score (NPS)" and "Customer Satisfaction (CSAT)"

groovecoder (approved):
Approved after 1 minor change request. Looks good!

groovecoder (approved):

PR #1703, opened 2022-03-29
Also support the dev server running at 0.0.0.0

Vinnl:
It gets started at both `localhost:3000` and `0.0.0.0:3000`. I always
ran it at localhost, and thus didn't notice that requests from
`0.0.0.0:3000` got rejected.


maxxcrawford (approved):
Nice fix!

maxxcrawford:
cc/ @codemist

PR #1708, opened 2022-03-30
Mpp 1664 chrome extension banner react

groovecoder:

### DIFF
--- a/frontend/src/components/dashboard/PremiumOnboarding.tsx
+++ b/frontend/src/components/dashboard/PremiumOnboarding.tsx
@@ -1,4 +1,4 @@
-import { useState, createContext, useContext } from "react";
+import { useState } from "react";
 import { useLocalization } from "@fluent/react";
 import { event as gaEvent } from "react-ga";
 import { useOverlayTriggerState } from "react-stately";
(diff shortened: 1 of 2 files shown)


frontend/src/components/dashboard/PremiumOnboarding.tsx
```diff
@@ -26,6 +26,54 @@ export type Props = {
...
+  }
+  return {
+    extSupported: false,
+    headerMessageId: "",
+    paragraphMessageId: "",
+    linkHref: "",
+    linkMessageId: "",
+  };
+};
+
```
groovecoder, line 76: I thought about trying to consolidate all of this into EITHER a context OR a helper function. 🤷

Vinnl (commented):
Neat, great to see you diving into React land - more knowledge sharing is better!

The general approach looks good. Note that prefixing functions with a `_` isn't really necessary: everything inside a module (i.e. a file) is already scoped to that module, unless it's explicitly `export`ed.

Other than that I only have one major remark, that should make this a bit more idiomatic React (or at least, I think so - I'm aware that that term is fuzzy enough to also allow it to be abused to push through personal opinions), and at least allows for the removal of quite a bit of code while doing the same.

I'll start by quoting [the docs on Context](https://reactjs.org/docs/context.html):

> In a typical React application, data is passed top-down (parent to child) via props, but such usage can be cumbersome for certain types of props (e.g. locale preference, UI theme) that are required by many components within an application. Context provides a way to share values like these between components without having to explicitly pass a prop through every level of the tree.

Since in this case, you're really only passing values one level down, I'd just pass them as props. That way, the dependency is explicit, there's less "noise", and we avoid Context's [potential gotchas](https://reactjs.org/docs/context.html#caveats).

But in fact, since calculating `supportsFirefoxExtension()` and `supportsChromeExtension()` is practically free, I'd just call those functions inline where you need them, rather than calculating them once and passing them down through the different functions/components. (Though of course, a `supportsExtension()` that calls out to those two could certainly be added, if that's useful.)

Additionally, since the properties of `AddonDescriptionProps` are only generated once and used once, that allows you to just generate them where they are used. For example, in `<AddonDescriptionHeader>`, you could do something like this:

```ts
  return (
    <h3>{
      l10n.getString(supportsFirefoxExtension()
        ? "multi-part-onboarding-premium-extension-get-title"
        : "multi-part-onboarding-premium-chrome-extension-get-title")
    }</h3>
  );
```

That means you could get rid of `AddonDescriptionProps` and `BrowserContext` while still switching over to a more declarative component-based approach and getting rid of the `let`s and `if/else if`s.

Anyway, the overall approach of splitting these out into separate components is sound! In fact, I think you could even take it a step further: rather than creating a function called `_getStepThreeTitle`, I think you could just turn that into a separate component as well, to replace the `<h2>{title}</h2>` in `<StepThree>`. (In fact, that might even be necessary? I'm not 100% that you shouldn't be able to call hooks (`useLocalization` in this case) outside of React components, but it does trigger my spidey sense.)

groovecoder:
Good call. I refactored to drop the context and instead use a new `supportsAnExtension` function in a524d57ba5eb25ea101bd77915ab3baaf5ccfd83.

Vinnl (approved):
Looking good!

groovecoder:
Can you merge this into your branch and then (I'll) merge that into `main`?

Vinnl:
Sure!

PR #1677, opened 2022-03-23
fix #1671: change From: to replies@ for premium customers

groovecoder:
This PR fixes #1671.

How to test:
End-to-end manual test is easiest to do on the dev server:

1. Sign in as a free user
2. Generate an alias
3. Send a test email to the free alias
   * Verify the "From:" of the email is the "noreplies" address for the dev server
4. Sign in as a premium customer
5. Generate an alias
6. Send a test email to the premium alias
   * Verify the "From:" of the email is the "repies" address for the dev server

- [x] I've added a unit test to test for potential regressions of this bug.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #1714, opened 2022-03-31
Add docs on env vars in the UI

Vinnl:
https://github.com/mozilla/fx-private-relay/pull/1677/commits/2ece3c0ccd452025b87e692e27a14a1464376dfa exposed that there was no documentation yet on how to access environment variables in the front-end. This PR adds those docs.

### DIFF
--- a/docs/frontend-architecture.md
+++ b/docs/frontend-architecture.md
@@ -46,6 +46,26 @@ Import tokens from [Protocol](https://protocol.mozilla.org).
 Uses [`react-ga`](https://www.npmjs.com/package/react-ga).
 See `useGaViewPing` and `trackPurchaseStart`.
 
+## Add/modify environment-specific data
+
+We only compile our front-end once, and then deploy the built code to
+our different environments (stage, production). This is different from an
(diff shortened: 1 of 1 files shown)


docs/frontend-architecture.md
```diff
@@ -46,6 +46,27 @@ Import tokens from [Protocol](https://protocol.mozilla.org).
 Uses [`react-ga`](https://www.npmjs.com/package/react-ga).
 See `useGaViewPing` and `trackPurchaseStart`.
 
+## Add/modify environment-specific data
+
+To make life easy for our Site Reliability Engineers (SREs), we only compile our
+front-end once, and they deploy the built code to our different environments
+(stage, production). This is different from an approach that you might be
+familiar with from other front-end projects, in which the code is built for
+each environment separately.
```
groovecoder, line 55:
```suggestion
We only compile our front-end once, and then deploy the built code to
our different environments (stage, production). This is different from an
approach that you might be familiar with from other front-end projects,
in which the code is built for each environment separately.
```

groovecoder (changes requested):

PR #1722, opened 2022-04-04
Set asserted value in test

Vinnl:
Rather than having a test depend on a value implicitly set in the
function, the test now explicitly passes the value it's matching on
into the mock creator. This can avoid the test accidentally passing
if the mocks are changed in the future. (For example, for this test
specifically, it's checking whether the string `some` is in a
mocked email's subdomain, but it would also succeed if that string
was in the address.)

Obviously in this one instance, it's not super important, but I
thought it could be useful to clarify what I meant by my comment in
https://github.com/mozilla/fx-private-relay/pull/1677#discussion_r840517332.

### DIFF
--- a/frontend/src/functions/filterAliases.test.ts
+++ b/frontend/src/functions/filterAliases.test.ts
@@ -8,7 +8,7 @@ import { filterAliases, Filters } from "./filterAliases";
 
 const getMockedAliasMatching = (
   filters: Partial<Filters>,
-  customSubdomain = "some"
+  customSubdomain = "arbitrary_subdomain"
 ): AliasData => {
   const alias = {
     address: filters.string ?? "arbitrary_string",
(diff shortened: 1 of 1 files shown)


frontend/src/functions/filterAliases.test.ts
```diff
@@ -100,7 +100,7 @@ it("filters out aliases that do not match the given string", () => {
 it("can match on an alias's subdomain when using a string filter", () => {
   const aliases = [
     getMockedAliasMatching({ domainType: "random" }),
-    getMockedAliasMatching({ domainType: "custom" }),
+    getMockedAliasMatching({ domainType: "custom" }, "some_subdomain"),
   ];
 
   expect(filterAliases(aliases, { string: "some" })).toStrictEqual([
```
Vinnl, line 106: This `"some"` now refers to the `some` on line 103, rather than the one that happened to be set on line 11, outside of this test.

groovecoder: Weird the inline diffe says line 103, but I think you meant line 125? In any case, looks good and tests still pass! Ship it!

groovecoder (approved):

PR #1727, opened 2022-04-04
Change codeql to only scan pushes to main

groovecoder:


jwhitlock (approved):
Thanks @groovecoder - wish I'd approved this one a few hours ago!

Vinnl:
@groovecoder @jwhitlock Hmm, one disadvantage of this is that we'll only catch newly-introduced security errors once they get merged to `main`... How would you feel about keeping this as it was before, but instead pin the CodeQL Action version, and manage its updates via dependabot? That means that findings on PRs would only relate to code in that PR, and new security checks that apply to us would be highlighted in the Dependabot PR.

PR #1728, opened 2022-04-04
Collect traditional and React static files

jwhitlock:
~All `SERVE_REACT` to be set as a Docker build argument, since `./manage.py collectstatic` uses different ``STATICFILES_DIRS`` settings with this setting.~

Instead, collect files from `static` and from `frontend/out` for all settings of `SERVE_REACT`, so that the Docker image works for both settings.


jwhitlock:
The changes to `.circleci/config.yml` and `Dockerfile` are probably unnecessary, let me see if I can remove those...

groovecoder (approved):

PR #1730, opened 2022-04-05
Enable CSP for Whitenoise-served files

Vinnl:
WhiteNoise's docs mention that its middleware should be added
directly after Django's SecurityMiddleware, and before all other
middleware [1], presumably so that its protections also apply to
WhiteNoise-served paths. However, the same presumably holds for the
CSP headers that Django-CSP needs to add.

This supersedes https://github.com/mozilla/fx-private-relay/pull/1729.

[1] https://whitenoise.evans.io/en/stable/django.html#enable-whitenoise

How to test (shamelessly copied from John):

* Build the React toolchain, or put a basic file at `frontend/out/index.html`
* Run `SERVE_REACT=1 ./manage.py runserver 8000`.
* Run `curl -I http://127.0.0.1:8000/`, and look at the `Content-Security-Policy` header:
   - Before this PR - No header
   - After this PR - `connect-src 'self' https://www.google-analytics.com/ https://accounts.firefox.com https://location.services.mozilla.com; style-src 'self'; default-src 'self'; script-src 'self' https://www.google-analytics.com/; img-src 'self' mozillausercontent.com https://profile.stage.mozaws.net`


groovecoder (approved):
Code and spot-check looks good.

PR #1732, opened 2022-04-05
Allow Next.js to add styles inline

Vinnl:
When switching pages, Next.js loads the relevant styles on-demand to avoid bloating the initial bundle. These styles are injected as inline styles, so our Content Security Policy should allow for that. This setting should be a lot safer for styles than if it were used for scripts.

How to test: Before this PR, run the Django server with React enabled, then switch pages - you'll see the styles disappearing (this can also be tested out on stage). With this PR, the styles should be applied just fine.

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -86,7 +86,7 @@ CSP_SCRIPT_SRC = (
 # if settings.DEBUG:
 #     CSP_SCRIPT_SRC += ("'unsafe-inline'",)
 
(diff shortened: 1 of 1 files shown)


groovecoder:
We should find a solution that that doesn't require `unsafe-inline`. Even if we have to run with `style-src: 'unsafe-inline'` temporarily, we should have a long-term solution that doesn't require `unsafe-inline`.

PR #1710, opened 2022-03-30
Content refresh

Vinnl:
# New feature description

~~Do not merge this until after April 19th, so that localisers have time to deal with the unfortunate massive workload we're generating here :(~~

This applies three primary changes across the entire website:
- An "alias" is now called a "mask"
- A "real email" is now called a "true email"
- We're using "subdomain" consistently now, rather than "domain"

For that, it uses the strings from https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/66.

Fixes #1814.

# How to test

Check out https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/66, then verify that there's no mention of "alias", "real email" and "domain" anywhere anymore.

# Checklist

- [x] l10n dependencies have been merged, if any. Not yet: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/66
- [x] All acceptance criteria are met.
- [x] I've added or updated relevant docs in the docs/ directory.
- [x] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/static/scss/libs/protocol/css/includes/tokens/dist/index.scss`).
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #1746, opened 2022-04-06
🛑 DNM: Add Aliases are now masks on news tab

codemist:
# New feature description
This adds the 'Aliases to masks' news feature on the News tab.

l10n PR: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/72/files

# Screenshot (if applicable)

image

image

# How to test

On a free or premium account, check that the news tab indicate a new item, and that the 'Aliases are now masks' is listed as a news item.

# Checklist


frontend/src/components/layout/whatsnew/WhatsNewMenu.tsx
```diff
@@ -160,6 +162,31 @@ export const WhatsNewMenu = (props: Props) => {
...
+    ),
+    hero: aliasToMaskHero.src,
+    icon: aliasToMaskIcon.src,
+    dismissal: useLocalDismissal(
+      `whatsnew-feature_sign-back-in_${props.profile.id}`
+    ),
+    announcementDate: {
+      year: 2022,
+      month: 4,
+      day: 6,
```
codemist, line 185: Is there a protocol for what the `announcementDate` attribute should be set to when adding announcements?

frontend/src/components/layout/whatsnew/WhatsNewMenu.tsx
```diff
@@ -160,6 +162,31 @@ export const WhatsNewMenu = (props: Props) => {
...
+    },
+  };
+  entries.push(aliasToMask);
```
codemist, line 188: I didn't wrap this news component around an `if` statement since the news is visible to all users (free and premium), is this good practice?

frontend/src/components/layout/whatsnew/WhatsNewMenu.tsx
```diff
@@ -160,6 +162,31 @@ export const WhatsNewMenu = (props: Props) => {
...
+        )}
+        heading={l10n.getString("whatsnew-feature-alias-to-mask-heading")}
+        image={aliasToMaskHero.src}
+      />
+    ),
+    hero: aliasToMaskHero.src,
+    icon: aliasToMaskIcon.src,
+    dismissal: useLocalDismissal(
+      `whatsnew-feature_sign-back-in_${props.profile.id}`
+    ),
```
codemist, line 181: I'm unsure as to what this `dismissal` object is for, and why all of the news items have it set to `whatsnew-feature_sign-back-in_${.....`

frontend/pendingTranslations.ftl
```diff
@@ -208,3 +208,7 @@ whatsnew-feature-forward-some-heading = Promotional email blocking
 # length as the English version.
 whatsnew-feature-forward-some-snippet = { -brand-name-relay-premium } allows you to block only promotional emails...
 whatsnew-feature-forward-some-description = { -brand-name-relay-premium } allows you to block only promotional emails sent to an alias. You’ll receive emails like receipts but not marketing emails.
+
+whatsnew-feature-alias-to-mask-heading = Aliases are now masks
+whatsnew-feature-alias-to-mask-snippet = Notice a change? We’re now calling aliases “masks” to make { -brand-name-firefox-relay }...
```
codemist, line 213: Was thinking instead of adding a snippet string, we could just truncate the longer description with the `text-overflow` [property](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow), or use the `substring()` [method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring).

codemist:
Probably wanna sync this PR up with https://github.com/mozilla/fx-private-relay/pull/1710

frontend/src/components/layout/whatsnew/WhatsNewMenu.tsx
Vinnl: Yep, perfect!

frontend/src/components/layout/whatsnew/WhatsNewMenu.tsx
Vinnl: [It is](https://github.com/mozilla/fx-private-relay/blob/ab13f0ccec87ec56bbf6e6e561a376afbf30d728/frontend/src/hooks/localDismissal.ts#L22) the ID that will be used for the cookie that tracks whether the current user has already seen this news entry. So you'll want to use an ID here that is unique for this entry, e.g.

```suggestion
    dismissal: useLocalDismissal(
      `whatsnew-feature_alias-to-mask_${props.profile.id}`
    ),
```

frontend/pendingTranslations.ftl
Vinnl: Definitely something that would be interesting to investigate! I think `substring` would be problematic in different locales, where "characters" don't quite correspond to "letters" (e.g. Japanese characters might get split in half and suddenly look entirely different, or a Japanese snippet might be truncated way earlier than an English one), but I'm not sure about that - so that would be something to verify if you're open to looking into this.

So yeah, if you're up for looking into that and verifying it in different locales, that could definitely be a nice improvement.

Vinnl: Oh, and one thing I got back last time when I submitted these to the l10n repo: there's a single character that represents an ellipsis instead of three separate dots :

```suggestion
whatsnew-feature-alias-to-mask-snippet = Notice a change? We’re now calling aliases “masks” to make { -brand-name-firefox-relay }…
```

Vinnl (approved):
Nailed it! **One required change**: make sure the dismissal ID is unique, otherwise someone who's already seen the "Sign back in with your alias" news entry won't be notified that this one was added.

Vinnl:
Wanted to surface https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/72#issuecomment-1102800481 by @flodolo here: we might want create a list of all locales that actually changed alias to mask, and only show this in those locales?

flodolo:
> Wanted to surface [mozilla-l10n/fx-private-relay-l10n#72 (comment)](https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/72#issuecomment-1102800481) by @flodolo here: we might want create a list of all locales that actually changed alias to mask, and only show this in those locales?

That would definitely reduce the confusion for users. But we'd still have the problem that those strings need to be translated, e.g. "Aliases are now aliases" 😕

Vinnl:
@codemist I added d1920890 to make sure the comments I added in https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/74/commits/7bf2817f9f793f09dac6d0249c7ccfa7ce4b79cc are actually true. There are a couple more locales that I'm fairly certain could be added, but better safe than sorry for now. (@flodolo And hopefully those comments will make it clear to the localisers that they don't need to localise those strings.)

PR #1749, opened 2022-04-07
Use an allowlist for inline styles

Vinnl:
Previously we allowed _all_ inline styles, allowing for malicious
interventions making our pages look different, potentially
misleading the user. With this change, we explicitly list the
inline styles that we inject, disallowing all others.

How to test:

1. Replace line 95 in `settings.py` by `if (False):` to make sure the production-behaviour is replicated.
2. Ensure `SERVE_REACT` is True, and that you've built (`npm run build`) the frontend.
3. Start Django.
4. Visit 127.0.0.1:8000, then click the link to the FAQ page. Observe that styling is still applied.

- [ ] I've added a unit test to test for potential regressions of this bug. - Note sure how, but would appreciate pointers.


groovecoder (approved):
It's so pretty!!!!

image

PR #1751, opened 2022-04-07
React UI - Typography fix

codemist:
# New feature description

Fixing up typography inconsistences on the React site.

# Checklist


frontend/src/pages/accounts/profile.module.scss
```diff
@@ -131,6 +132,7 @@
     gap: $spacing-md;
 
     h3 {
+      @include font-mozilla;
```
codemist, line 135: Hey @Vinnl trying to get your take on this, for some reason this line doesn't load the appropriate [Zilla Slab font](https://protocol.mozilla.org/fundamentals/typography.html#typeface-zilla-slab). I tried to import the font face in the code base, following the pattern of Inter UI and Metropolis. However, this is still not working. On the contrary, `@include font-firefox` works just fine.

frontend/src/pages/accounts/profile.module.scss
```diff
@@ -1,5 +1,6 @@
 @import "../../styles/tokens.scss";
 @import "~@mozilla-protocol/core/protocol/css/includes/lib";
+@import "~@mozilla-protocol/core/protocol/css/includes/mixins/typography";
```
codemist, line 3: @Vinnl I tried using this to import the Zilla Slab font (this line exists in `Button.module.scss`), but was wondering if its redundant.

Vinnl: I think it's technically redundant if `Button.module.scss` gets loaded as well, but since that means removing it from there would also break it here, and also because that means people reading this file can inspect the imports to see where the mixin came from, I'd indeed include this.

frontend/src/pages/accounts/profile.module.scss
Vinnl: Heh, you had me stumped there for a bit, since you did all the right things. But there are two things going wrong, both in the `zilla-slab.scss` file:

1. The URLs to the fonts are wrong: they refer to `/fonts/ZillaSlab/`, but the folder you created in `frontend/public/fonts` is called `Zilla_Slab` instead of `ZillaSlab`.
2. The name of the `font-family` does not align with the names used by Protocol. If you inspect the styles of this `<h3>`, you'll see that it refers to "Zilla Slab", whereas in `zilla-slab.scss`, it matches `font-family: "ZillaSlab"` (i.e. that should be `font-family: "Zilla Slab"`).

image

When I fix those, I do get the correct font:

image

image

(Also: we're using _three_ different fonts? Hoowee!)

Vinnl (approved):
Changes look good! I added comments on how to fix the Zilla Slab loading inline.

frontend/src/pages/accounts/profile.module.scss
codemist: Whoop, thanks for catching that!

static/scss/app.scss
```diff
@@ -11,6 +11,7 @@ $font-path: "/static/fonts";
 @import "includes/utils";
 @import "includes/fonts/inter-ui";
 @import "includes/fonts/metropolis";
+@import "includes/fonts/zilla-slab";
```
maxxcrawford, line 15: This links to a file that does not exist. It is causing the following error: 

```bash
SERVE_REACT=True python manage.py runserver

...

Error in plugin "sass"
Message:
    static/scss/app.scss
Error: Can't find stylesheet to import.
   ╷
14 │ @import "includes/fonts/zilla-slab";
   │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
  static/scss/app.scss 14:9  root stylesheet
```

PR #1753, opened 2022-04-07
React UI - Buttons fix

codemist:
# New feature description

This updates the button UI and states according to the Nebula design system and Protocol.

# Checklist

### DIFF
--- a/frontend/src/components/Button.module.scss
+++ b/frontend/src/components/Button.module.scss
@@ -8,7 +8,7 @@
   justify-content: center;
   cursor: pointer;
   background-color: $color-blue-50;
-  padding: $spacing-sm $spacing-2xl;
+  padding: $spacing-sm $spacing-md;
   border-radius: $border-radius-sm;
   border: 2px solid transparent;
   color: $color-white;
--- a/frontend/src/components/Button.module.scss
+++ b/frontend/src/components/Button.module.scss
@@ -19,6 +19,7 @@
 
   &:hover:not(:disabled) {
     background-color: $color-blue-60;
+    opacity: 1;
   }
 
   &:focus {
--- a/frontend/src/components/Button.module.scss
+++ b/frontend/src/components/Button.module.scss
@@ -29,7 +29,6 @@
   }
 
   &:disabled {
-    opacity: 0.6;
     cursor: not-allowed;
     background-color: $color-light-gray-80;
   }
@@ -39,10 +38,15 @@
   }
 
   &.is-destructive {
-    background-color: $color-red-50;
+    background-color: $color-red-60;
+
+    &:disabled {
+      background-color: $color-red-10;
+    }
 
     &:hover:not(:disabled) {
-      background-color: $color-red-60;
+      background-color: $color-red-70;
+      opacity: 1;
     }
   }
 }
--- a/frontend/src/components/dashboard/SubdomainPicker.module.scss
+++ b/frontend/src/components/dashboard/SubdomainPicker.module.scss
@@ -27,7 +27,7 @@
 
   a {
     @include text-body-sm;
-    color: $color-purple-60;
+    color: $color-blue-50;
 
     &:hover,
     &:focus {
--- a/frontend/src/components/dashboard/SubdomainPicker.module.scss
+++ b/frontend/src/components/dashboard/SubdomainPicker.module.scss
@@ -31,7 +31,7 @@
 
     &:hover,
     &:focus {
-      color: $color-purple-80;
+      color: $color-blue-60;
     }
   }
 
--- a/frontend/src/components/dashboard/aliases/AddressPickerModal.module.scss
+++ b/frontend/src/components/dashboard/aliases/AddressPickerModal.module.scss
@@ -94,15 +94,14 @@
         cursor: pointer;
         border-radius: $border-radius-sm;
         color: $color-blue-50;
-        padding: $spacing-sm $spacing-2xl;
+        padding: $spacing-sm $spacing-md;
 
         &:hover {
-          background-color: $color-blue-50;
-          color: $color-white;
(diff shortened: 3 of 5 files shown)


frontend/src/components/Button.module.scss
```diff
@@ -39,10 +38,15 @@
...
+    background-color: $color-red-60;
+
+    &:disabled {
+      background-color: $color-red-10;
+    }
 
     &:hover:not(:disabled) {
-      background-color: $color-red-60;
+      background-color: $color-red-70;
+      opacity: 1;
```
Vinnl, line 49: Hmm, this line seems to be overwriting a style that is not defined in this module, which I think is a red flag? But I don't see any differences when I remove this line, so I'm not sure where to verify.

frontend/src/components/dashboard/SubdomainPicker.module.scss
```diff
@@ -27,7 +27,7 @@
 
   a {
     @include text-body-sm;
-    color: $color-purple-60;
+    color: $color-blue-50;
 
     &:hover,
     &:focus {
```
Vinnl, line 33: If the regular colour is blue, should the hover state be dark blue instead of dark purple as well? (IIRC that's what we do elsewhere.)

frontend/src/components/dashboard/aliases/AddressPickerModal.module.scss
```diff
@@ -94,15 +94,14 @@
         cursor: pointer;
         border-radius: $border-radius-sm;
         color: $color-blue-50;
-        padding: $spacing-sm $spacing-2xl;
+        padding: $spacing-sm $spacing-md;
 
         &:hover {
-          background-color: $color-blue-50;
-          color: $color-white;
+          color: $color-link-hover;
```
Vinnl, line 100: Oh nice, I didn't know about this var. That said, I don't think this hover state is accessible though, since the colour differs so little from the original colour (at first I even thought there was no hover state at all), in addition to colour being the only modality. Which I guess we're doing elsewhere too, but maybe something to bring up with Eduardo.

(Likewise for the "cancel" buttons in `AliasDeletionButton.module.scss` and `ConfirmationForm.module.scss` I suppose.)

frontend/src/components/dashboard/aliases/AliasDeletionButton.module.scss
```diff
@@ -100,15 +100,14 @@
           cursor: pointer;
           border-radius: $border-radius-sm;
           color: $color-blue-50;
-          padding: $spacing-sm $spacing-2xl;
+          padding: $spacing-sm $spacing-md;
 
           &:hover {
-            background-color: $color-blue-50;
-            color: $color-white;
+            color: $color-link-hover;
```
Vinnl, line 106: Same comments as in the AddressPickerModal :)

frontend/src/components/dashboard/subdomain/ConfirmationForm.module.scss
```diff
@@ -40,21 +40,34 @@
     justify-content: space-between;
     padding-top: $spacing-xl;
 
+    button[type="submit"] {
```
Vinnl, line 43: Praise: nice job matching on a required attribute rather than adding a new class, so that we'll easily notice if we inadvertently remove this attribute.

frontend/src/components/dashboard/subdomain/ConfirmationForm.module.scss
```diff
@@ -40,21 +40,34 @@
...
+      background: $color-blue-50;
+      opacity: 0.4;
```
Vinnl, line 45: Might be more future-proof if we only set this opacity on disabled buttons, rather than first setting it and then removing it if it's _not_ disabled?

frontend/src/components/dashboard/subdomain/ConfirmationForm.module.scss
```diff
@@ -40,21 +40,34 @@
...
+
+      &:not(:disabled) {
+        opacity: 1;
+      }
+
+      &:hover:not(:disabled) {
+        background-color: $color-blue-60;
+        opacity: 1;
```
Vinnl, line 53: Also this line seems redundant with the one from line 48?

frontend/src/components/dashboard/subdomain/ConfirmationForm.module.scss
Vinnl: Btw, might we not want this disabled styling for all buttons, rather than just the one in this dialog? i.e. shouldn't these changes be in `Button.module.scss`?

Vinnl (approved):
Overall the changes look good (especially the less-wide buttons!), just some small, non-blocking suggestions inline.

frontend/src/components/dashboard/SubdomainPicker.module.scss
codemist: Ah yes, good callout!

frontend/src/components/dashboard/aliases/AddressPickerModal.module.scss
codemist: Yeah - i'll bring this hover state up with Eduardo. Just following the Nebula docs for now.

frontend/src/components/dashboard/subdomain/ConfirmationForm.module.scss
codemist: This applies a `:hover` style on non-disabled buttons.

frontend/src/components/dashboard/subdomain/ConfirmationForm.module.scss
codemist: Good point! I'll add this into `Button.module.scss` instead.

PR #1754, opened 2022-04-08
fix #1748 MPP 1837: Change wrapped_email.html to use static

### LINKED ISSUE #1748: Logos not visible in header and footer of received emails
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox, latest release

**Prerequisites:**
Have a second email address available.

**Steps to reproduce:**

1. Navigate to Relay page and sign in.
2. Go to aliases list and copy an alias.
3. Using a second email address send an email to the copied alias.
4. Check the inbox of the email that is connected to the alias for the email received.
5. Observe the header and the footer of the email.

**Actual result:**
The smiley face and Relay logo are not visible anymore.

**Expected result:**
The smiley face and Relay logo are visible.

**Notes:**
The issue occurs for both free and premium users.

**React:**
logo header_footer not visible

**Before:**
before

groovecoder:
This PR fixes #1748 ([MPP-1837](https://mozilla-hub.atlassian.net/browse/MPP-1837)).

How to test:
1. Go to http://127.0.0.1:8000/emails/wrapped_email_test
   * [ ] The images in the email wrapper should show up

### DIFF
--- a/emails/templates/emails/wrapped_email.html
+++ b/emails/templates/emails/wrapped_email.html
@@ -1,5 +1,6 @@
 {% load ftl %}
 {% load email_extras %}
+{% load static %}
 {% withftl mode='server' bundle='privaterelay.ftl_bundles.main' language=recipient_profile.language %}
 
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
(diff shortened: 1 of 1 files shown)


Vinnl (changes requested):
I don't think this fixes the issue; `static` generates relative paths (i.e. they start with `/`), so while the images show up when the email HTML is served up on the same domain as the assets, they won't load when they show up in an email client. (This is the reason they didn't use `static` before.)

I described a fix that should work in https://github.com/mozilla/fx-private-relay/issues/1748#issuecomment-1091613903.

PR #1766, opened 2022-04-11
Header fixes

Vinnl:
Bundling a couple of the requested UI changes for the header in a single PR. This:

- Increases the padding at the bottom of the "What's new" dropdown.
- Makes icons in the user menu smaller.
- Lightens the header font for new Premium users.
- Makes the header sticky.
- Increases horizontal padding for menu iems.
- Better aligns the chevron of the "What's new menu". (Properly attaching that is still a pain though, I don't think we have a better strategy other than spitballing the offset a bit, do we?)
- Remove underlines from hovered "What's new" entries.

~~I haven't figured out yet how to remove the underline in the "What's new" menu without removing it on focus (which Eduardo is fine with). I'll update this PR if I do.~~

### DIFF
--- a/frontend/src/components/layout/Layout.module.scss
+++ b/frontend/src/components/layout/Layout.module.scss
@@ -22,6 +22,7 @@
 
 .header {
   position: sticky;
+  top: 0;
   // The sticky position causes a new stacking context,
   // which is stacked below elements lower in the document
   // with their own stacking contexts (specifically,
@@ -95,6 +96,7 @@
     }
   }
   &.is-light {
+    background-color: $color-white;
     color: $color-grey-40;
   }
 
--- a/frontend/src/components/layout/Layout.module.scss
+++ b/frontend/src/components/layout/Layout.module.scss
@@ -94,6 +94,9 @@
       color: $color-light-gray-30;
     }
   }
+  &.is-light {
+    color: $color-grey-40;
+  }
 
   @media screen and #{$mq-md} {
     .logo {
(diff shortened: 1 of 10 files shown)


Vinnl:
I also just realised that I hadn't enabled the CSAT survey yet (it wasn't enabled on the current website when implemented). I just did, which disables the NPS survey: aa7a8efe.

frontend/src/components/layout/Layout.module.scss
```diff
@@ -22,6 +22,7 @@
 
 .header {
   position: sticky;
+  top: 0;
```
maxxcrawford, line 25: Praise: This is one of those pesky CSS fixes. Nice one-liner to get the sticky sticking. 😵‍💫

frontend/src/components/layout/whatsnew/WhatsNewList.module.scss
```diff
@@ -4,7 +4,8 @@
 .empty-message {
   display: flex;
   flex-direction: column;
-  padding: $spacing-lg;
+  padding-top: $spacing-lg;
+  padding-bottom: $spacing-2xl;
```
maxxcrawford, line 8: thought: I really like this approach to padding, rather than `padding: top x-axis bottom` shorthand. What are your thoughts on making this a [style rule](https://stylelint.io/user-guide/rules/list/declaration-block-no-shorthand-property-overrides/)?

maxxcrawford (approved):
All straight forward fixes. 🎯

codemist (approved):
LGTM!

PR #1777, opened 2022-04-13
Fix logging out when using the dev server locally

Vinnl:
On the production website, a logout redirects the user to Firefox
Accounts, thereby inducing a full page refresh. However, when
mocking that logout flow locally, we remain in the context of the
Single-Page App, and therefore user data remained cached in memory.
This change forces a revalidation of user data after clearing the
session key, which then results in that data no longer being
available and thus the user's logged-out status being properly
recognised.

It also more thoroughly detects whether the user is logged in on
the homepage, by not only checking if user data is available, but
also that refetching it did not result in an error (which it does
if the user is logged out).

How to test: run the dev server (`npm run dev`), log in as a user, then log out again. Observe that the "Sign up" and "Sign in" buttons in the top menu are present without refreshing, as opposed to them disappearing like they do on `main`.

- [ ] I've added a unit test to test for potential regressions of this bug: No, mocking both the router and the user API for what is otherwise a static page felt like overkill, but let me know if you disagree.

### DIFF
--- a/frontend/src/pages/index.page.tsx
+++ b/frontend/src/pages/index.page.tsx
@@ -35,7 +35,7 @@ const Home: NextPage = () => {
     label: "home-hero-cta",
   });
 
-  if (typeof userData.data?.[0] === "object") {
+  if (typeof userData.data?.[0] === "object" && !userData.error) {
     router.push("/accounts/profile/");
   }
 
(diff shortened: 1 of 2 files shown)


groovecoder (approved):
Code looks good, and this is certainly better than before, so I'm happy to approve.

suggestion (non-blocking): 

The background is light color (as if signed in) and the `UserMenu` component still shows until a page refresh:

image

After page refresh:
image

groovecoder:
Can fix the remaining issues in a follow-up. Merging.

PR #1783, opened 2022-04-13
React UI - Landing Page

codemist:
# New feature description

This PR fixes landing page specific React UI bugs.

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist

### DIFF
--- a/frontend/src/components/landing/Carousel.module.scss
+++ b/frontend/src/components/landing/Carousel.module.scss
@@ -79,11 +79,9 @@
       display: flex;
       align-items: center;
       gap: $spacing-sm;
-      padding: $spacing-xs $spacing-sm;
+      padding: $spacing-xs $spacing-sm 0 $spacing-sm;
       box-shadow: $box-shadow-sm;
       border-radius: $border-radius-md;
-      border-bottom: 8px solid transparent;
-      border-right: 8px solid transparent;
       font-weight: 700;
 
       &.yellow {
@@ -139,7 +137,7 @@
     }
 
     &:focus .title {
-      border-color: hsla(0, 0%, 80%, 0.5);
+      box-shadow: none;
     }
 
     &:not(.is-selected):hover .title {
@@ -154,9 +152,10 @@
 
   .content {
     grid-area: content;
-    background-color: $color-light-gray-20;
+    background-color: $color-light-gray-10;
     border-radius: $border-radius-md;
     padding: $spacing-lg;
+    margin-top: $spacing-md;
 
     .content-heading {
       display: none;
@@ -179,8 +178,15 @@
         text-align: center;
 
         img {
-          width: 150px;
-          height: 150px;
+          width: calc(#{$layout-2xl} - #{$layout-lg});
+          height: calc(#{$layout-2xl} - #{$layout-lg});
+
+          // calc(#{$screen-sm} - #{$spacing-md} - #{$spacing-md})
+
+          @media screen and #{$mq-xl} {
+            width: $layout-2xl;
+            height: $layout-2xl;
+          }
         }
       }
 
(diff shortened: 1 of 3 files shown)


frontend/src/pages/index.module.scss
```diff
@@ -79,21 +79,40 @@
   .steps {
     display: flex;
     flex-direction: column;
-    gap: $spacing-2xl;
+    gap: $layout-xl;
```
Vinnl, line 82: I think the idea behind using Protocol tokens is to keep things consistent, even if they were to change in the future. And though that ship has probably mostly sailed already, I think it's still good to try and not make a habit of using tokens in situations other than their intended use, if they happen to have the value (at this point in time) that you're looking for.

In other words, I'd avoid using a `$layout-` token for spacing. That means we can either say we're bound to the smaller spacing by Protocol, or (probably more constructively :slightly_smiling_face:) go for this compromise:

```suggestion
    gap: 2 * $spacing-2xl;
```

frontend/src/pages/index.module.scss
```diff
@@ -79,21 +79,40 @@
...
     list-style-type: none;
-    margin: 0;
     padding: $spacing-2xl 0;
-    text-align: center;
+    text-align: left;
```
Vinnl, line 85: I don't know if we support right-to-left scripts, but still a good habit to form:

```suggestion
    text-align: start;
```

frontend/src/pages/index.module.scss
```diff
@@ -79,21 +79,40 @@
...
+    p {
+      padding-top: $spacing-sm;
     }
 
     .step {
-      flex: 1 0 0;
+      flex: 1;
+
+      @media screen and #{$mq-md} {
+        margin-top: $spacing-2xl;
```
Vinnl, line 110: To make styles more composable (i.e. try as much as we can to have an element only affect layout _inside_ that element), I'd add this to `.steps`' `padding-top` instead.

Vinnl (commented):
Nice, this is quite a bit tighter. I added mostly non-blocking comments inline, but the one thing I don't really like is removing focus styles without providing a replacement, letting design concerns trump accessibility. Unless the rest of the team disagrees, I'd say we can only remove those if we have an alternative design. (And hopefully also make a habit out of designs including focus and hover state designs at some point - one can dream :grin: )

frontend/src/pages/index.module.scss
codemist: Oh! Yes! Good callout

frontend/src/pages/index.module.scss
codemist: Hmm.. I think I'm still entirely unclear about the use cases for spacing v. layout. I remember a good rule of thumb was spacing for individual modules (margins/padding) and layout for spacing between elements/modules.

Vinnl: Oh! Actually, it might be that _I_ misinterpreted this token - I think I was confusing it with `$content-`. Then carry on :)

@maxxcrawford Could you confirm whether the interpretation of those tokens above is correct? Then I can use them correctly in the future as well :)

frontend/src/components/landing/Carousel.module.scss
```diff
@@ -139,7 +137,7 @@
     }
 
     &:focus .title {
-      border-color: hsla(0, 0%, 80%, 0.5);
+      box-shadow: none;
```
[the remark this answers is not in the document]
Vinnl: There's still no focus state here, because you only restored the `border-color`, but not the actual border on `.title`, so the colour applies to nothing :)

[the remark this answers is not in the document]
codemist: I added the `outline` property instead as the focus state.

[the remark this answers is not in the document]
Vinnl: Woah, TIL about `outline`, neat!

Vinnl (approved):

PR #1787, opened 2022-04-14
Fix ever-expanding messages cookie

### LINKED ISSUE #1781: Re-logging in successfully for 7 times displays a 502 Bad Gateway Error message to the user
#### Build:
- Stage;

#### Affected Platforms:
- Windows 10;

#### Browsers tested on:
- Firefox, Release version;

#### Prerequisites:
- Have a Relay account;
- Do NOT be logged in the Relay account;

#### Steps to reproduce:
1. Go to the Relay homepage;
2. Log into the Relay account;
3. Log out from the Relay account;
4. Repeat until 7 logins have been done;
5. Observe the results;

#### Expected result:
- The user can log in without any issue;

#### Actual result:
- The user sees a 502 Bad Gateway Error;

#### Notes:
- This problem doesn’t come up on Production;
- It happens even if we use different, already existing accounts throughout the 7 logins;
- Switching to a different browser profile, clearing cookies or restarting the browser resolves this problem;
- I tried reproducing the error also by not doing all the 7 logins at a time. For example, I logged in a few times, then waited around 20 minutes and then got up to 7 and still got the error, or by waiting 1-2 minutes between each log in. Same result;
- See attached media:
image

Vinnl:
django-allauth adds login and logout messages to the `messages`
cookie, but with Django no longer rendering the view, it is never
reading out and clearing that cookie. Since the cookie is
HTTP-only, the client can't read it out or clear it either. Hence,
here we tell it not to set those messages by overriding its message
templates, as suggested in
https://django-allauth.readthedocs.io/en/latest/advanced.html#messages.

How to test: The best we can do locally, I think, is to inspect your cookie storage via e.g. your developer tools - before this change, there was a `messages` cookie that kept getting expanded. Now, if you log in and back out (make sure to be serving React and doing this on 127.0.0.1:8000), that cookie should no longer be modified, and if you remove it, it should not come back.

We'll have to verify when this is on stage that it actually fixes #1781, as it should.

- [ ] I've added a unit test to test for potential regressions of this bug. N/A?


groovecoder:
Did we only see this bug from QA? If so, they'll need to clear the cookies before testing this again. When I ran this code locally, it did not clear out any EXISTING messages cookies - it only prevented new messages from going into the cookie.

Vinnl:
Yes, only QA luckily, and they'd already found out that they could clear their cookies :)

PR #1788, opened 2022-04-14
Use a special endpoint to set the subdomain

### LINKED ISSUE #1770: Subdomain already taken is initially accepted for a different user
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox, latest release

**Prerequisites:**
Have a user with an active subscription but no domain added.
Have VPN set up in a country where Relay Premium is available.

**Steps to reproduce:**

2. Click on "Search your new domain" field and type a domain that was previouly added by another user.
3. Click "Search" button.
4. Check the checkbox "Yes, I want to register @your_domain_name" and click "Register Domain".

**Actual result:**
"Good news! @your_domain_name is available!" modal is displayed and then the red banner "The domain @your_domain_name is not available. Please try again with another domain." is displayed.

**Expected result:**
"Good News! @your_domain_name is available!" modal is not displayed at all. Only the red banner "The domain @your_domain_name is not available. Please try again with another domain." is displayed.

**Notes:**
N/A

domain already taken initially accepted

Vinnl:
When sending a PATCH to /profiles/, we directly modify the profile
table. However, that foregoes some additional actions performed in
the subdomain-specific endpoint that allow us to inform other
users when they're trying to register a taken subdomain.

This PR fixes #1770. I'm not sure if this is the best fix - ideally these checks would also be executed on the backend for PATCH requests. That said, it might be good to apply this first and then revert it if we have a lower-level fix later, to ensure that we don't have data loss (hashed domains) while we await that.

How to test: Register a new subdomain locally, then with a different user, try to register the same subdomain. You should now getting an error when looking it up, rather than when trying to register it.

- [ ] I've added a unit test to test for potential regressions of this bug. N/A - bug is in backend.

### DIFF
--- a/frontend/src/hooks/api/api.ts
+++ b/frontend/src/hooks/api/api.ts
@@ -23,8 +23,11 @@ export const authenticatedFetch = async (
     authToken = localStorage.getItem("authToken");
   }
   const headers = new Headers(init?.headers ?? undefined);
-  headers.set("Content-Type", "application/json");
-  headers.set("Accept", "application/json");
+  headers.set(
+    "Content-Type",
+    headers.get("Content-Type") ?? "application/json"
+  );
+  headers.set("Accept", headers.get("Accept") ?? "application/json");
   if (typeof authToken === "string") {
     headers.set("Authorization", `Token ${authToken}`);
   }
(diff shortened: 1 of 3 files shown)


jwhitlock (approved):
I was able to reproduce the issue locally, and confirm it goes away with this PR, and the requests / responses are as expected in developer tools. Thanks @Vinnl!

I've written https://github.com/mozilla/fx-private-relay/issues/1790 for the follow-on work of updating the `PATCH` method. However, I'm not sure about this, since `/accounts/profile/subdomain` is still used to query if a subdomain is available. A full solution should move both calls to the API.

Issue #1744, opened 2022-04-06
Green successful messages for Signed in/ Signed out not displayed


Bianca-O:
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox, latest release

**Prerequisites:**
N/A

**Steps to reproduce:**

1. Navigate to Relay page and sign in.
2. On dashboard observe if the green successful message "Successfully signed in..." is displayed.
3. Sign out.
4. On homepage observe if the green successful message "You have signed out" is displayed.

**Actual result:**
Green successful messages for signed in/ signed out are not displayed anymore.

**Expected result:**
Green successful messages for signed in/ signed out are displayed.

**Notes:**
N/A

**React:**
![no green messages](https://user-images.githubusercontent.com/79697848/161991928-e3eada5d-7839-4e99-bb79-a632c899de92.gif)

maxxcrawford:
Note that we want the toast to match VPN with this update: https://www.figma.com/file/cBaBmcPisqZ0QQX5LLzpjU/?node-id=300%3A0

Vinnl:
@maxxcrawford Oh! I'll create a separate ticket for that, so it doesn't get lost.

As for implementation, it'll likely involve something like this. First, the back-end work, courtesy of @groovecoder. If we want to use our own strings (submitting to l10n as well):

> - disable the [built-in django-allauth messages](https://django-allauth.readthedocs.io/en/latest/advanced.html?highlight=messages#messages)
> - add more logic to [our existing signal receivers](https://github.com/mozilla/fx-private-relay/blob/main/privaterelay/signals.py) for [the corresponding ](https://django-allauth.readthedocs.io/en/latest/signals.html)django-allauth[ signals](https://django-allauth.readthedocs.io/en/latest/signals.html) which can set either cookies or headers that the front-end can pick up to display

If we want to re-use django-allauth's strings:

> you can probably leave the messages enabled, but may need to add new machinery for “send all django messages to the client”

Then you'll probably want to show the messages in a component that's used on every page: `<Layout>`. You'll probably want to reuse the functions from `frontend/src/functions/cookies.ts` to determine whether and what message to show.

You'll also only want to show it only at the first render, in the browser, which you can do by wrapping it in a [`useEffect` hook](https://reactjs.org/docs/hooks-effect.html) with an empty [dependencies array](https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect).

[...]

PR #1792, opened 2022-04-14
 Toast notification for Signed in/out users

codemist:
L10N pr: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/76

# New feature description
This adds the toast notification for signed in/out users.

# Screenshot (if applicable)

image
image

# Checklist

### DIFF
--- a/frontend/src/components/layout/Toast.tsx
+++ b/frontend/src/components/layout/Toast.tsx
@@ -16,7 +16,7 @@ export function makeToast(l10n: ReactLocalization, usersData: UserData) {
   if (checkUserSignIn) {
     clearCookie("user-sign-in");
     return toast.success(
-      l10n.getString("success-signed-in-message") + "  " + usersData.email
+      l10n.getString("success-signed-in-message", { username: usersData.email })
     );
   }
 }
--- a/frontend/src/components/layout/Toast.tsx
+++ b/frontend/src/components/layout/Toast.tsx
@@ -3,7 +3,7 @@ import { toast } from "react-toastify";
 import { ReactLocalization } from "@fluent/react";
 import { UserData } from "../../hooks/api/user";
 
-export function makeToast(l10n: ReactLocalization, usersData: UserData) {
+export function makeToast(l10n: ReactLocalization, usersData?: UserData) {
   const checkUserSignOut = getCookie("user-sign-out");
 
   const checkUserSignIn = getCookie("user-sign-in");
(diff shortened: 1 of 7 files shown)


Vinnl:
Fixes #1744 (I think by adding this comment, that issue will get auto-closed when this is merged).

codemist:
@Vinnl Is there a hook for just grabbing the users' username? I've only seen the email instance being use (`user.email`) in `profile.page.tsx`

Vinnl:
@codemist That's not currently exposed through the API, so no. That said, if you do want it and are not afraid of wading a little bit into the back-end, you can expose it yourself! You can do so as follows.

In the file `/api/serializers.py`, you can find a class called `ProfileSerializer` - this class is used to determine which fields of `Profile` are visible and changeable via the API. We're interested in the field `display_name`, so you can add that to the array called `fields` - but make sure to also add it to the array called `read_only_fields`, as this name can only be updated in Firefox Accounts, and hence shouldn't be changeable through the API.

After you've done that (and possibly restarted your server), you should be able to see it in requests to `/api/v1/profiles/` in your devtools under the Network tab. That's it in terms of back-end changes :)

Now you can update the relevant type for the respective API hook in `/frontend/src/hooks/api/profile.ts` to add `display_name: string | null` to the `ProfileData` type (`| null`, because not everybody has a display name set in Firefox Accounts).

frontend/src/components/layout/Toast.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+  if (checkUserSignOut) {
+    clearCookie("user-sign-out");
+    return toast.success(l10n.getString("success-signed-out-message"));
+  }
+
+  if (checkUserSignIn) {
+    clearCookie("user-sign-in");
+    return toast.success(
+      l10n.getString("success-signed-in-message") + "  " + usersData.email
```
Vinnl, line 19: As mentioned above, the robust thing to do here is to use a variable to pass the email to the translation.

frontend/src/components/layout/Toast.tsx
```diff
@@ -0,0 +1,22 @@
+import { clearCookie, getCookie } from "../../functions/cookies";
```
Vinnl, line 1: Since this ended up not being a component but simply a function that triggers the display of the toast, this file should probably be `/frontend/src/functions/makeToast.ts` instead of `Toast.tsx` :)

Vinnl (commented):
The main blocker here is not using a variable in the l10n string, but otherwise this looks good! And if you've got the energy, I'm curious to see if it's clear enough now on how to show the user's username (if available, otherwise falling back to their email address).

Vinnl (approved):
Great, no more blockers, just the file rename would be nice :)

maxxcrawford:
Closing this PR as it was incorporated into https://github.com/mozilla/fx-private-relay/pull/1854

PR #1794, opened 2022-04-14
Measure trackers on emails

say-yawn:
# New feature description
For better understanding of breakage on emails related to email trackers, this PR counts the number of email trackers in emails we relay. We will use sample to:
- reduce the potential processing cost on counting trackers
- extrapolate from the sample what would be the average number of trackers in emails we relay
- correlate the number of trackers back to number of trackers found in breakage study to extrapolate potential breakage for all of the relayed email
We are using logging instead of `markus` because the influxdb/telegraf aggregates the data collected per minute resulting on loss of data: e.g. one minute we could process 100 emails with average 5 trackers while the another minute we could process 200 emails with average 1 tracker but aggregated metrics data does not show how big the size was resulting in improper representation of data when looking for average number of trackers in emails.

# Setup

- Run `pip install -r requirements.txt` to install django-waffle
- Run `./manage.py migrate waffle` to setup DB so service can use `waffle`
- ***NOTE: wait to run this*** `./manage.py waffle_sample tracker-sample X --create` to create waffle sample named `tracker-sample` that samples `X %` (int or floating value between 0.0 and 100.0) of relayed emails ([see doc](https://waffle.readthedocs.io/en/stable/usage/cli.html#samples)). Note you can use the sample like a switch by setting it to 100% or 0% respectively to turn it on or off

# How to test
## Sample does not exist, ignores sample data collection
1. After running migration to add `waffle` table AND BEFORE running the command line to create the sample
2. Email should be relayed successfully
3. No metrics or logging was added
4. Run Django shell with `./manage.py shell`
5. Run `from waffle import sample_is_active`
6. Run `sample_is_active('tracker-sample')` check that this returns `False`

## Sample exists, emails relayed gets data collected by the sample size
1. Run migration to collect email tracker data for 100% of the emails relayed `./manage.py waffle_sample tracker-sample 100 --create`
2. Check `Creating sample: tracker-sample` is printed on console
3.  Run Django shell with `./manage.py shell`
4. Run `from waffle import sample_is_active`
5. Run `sample_is_active('tracker-sample')` check that this returns `True`
6. Email is relayed successfully
7. Metrics and `studymetrics` logs are gathered and log looks something like:
```python
{
  'general': {'count': tracker_total, 'trackers': {domain: count, ...}},
  'strict': {'count': tracker_total, 'trackers': {domain: count, ...}}
}
}
```

# Checklist

#
Use dockerfile to create or enable/disable the sample? Since when the dockerfile is ran the DB is not setup this may not be possible.


emails/utils.py
```diff
@@ -49,7 +55,7 @@ def incr_if_enabled(name, value=1, tags=None):
 
 def histogram_if_enabled(name, value, tags=None):
     if settings.STATSD_ENABLED:
-        metrics.histogram(name, value=value, tags=None)
+        metrics.histogram(name, value=value, tags=tags)
```
jwhitlock, line 58: 👏 good fix!

emails/utils.py
```diff
@@ -341,3 +347,31 @@ def remove_message_from_s3(bucket, object_key):
             logger.error('s3_client_error_delete_email', extra=e.response['Error'])
         incr_if_enabled('message_not_removed_from_s3', 1)
     return False
+
+
+
+def count_tracker(html_content, tracker_domain, level='general'):
+    # html_content needs to be str for count()
+    tracker_count = html_content.count(tracker_domain)
+    study_logger.info(
```
jwhitlock, line 356: This logged a LOT, including when the tracker was not found in the email. I'm not sure this level of logging will be useful. What about including the non-zero counts in the `email_tracker_summary` message, and not logging for each tracker domain?

emails/views.py
```diff
@@ -501,6 +503,9 @@ def _sns_message(message_json):
     address.save(
         update_fields=['num_forwarded', 'last_used_at']
     )
+    # successful email relayed count number of trackers
+    if sample_is_active('tracker-sample') and html_content:
```
jwhitlock, line 507: 👍 waffle samples are a good choice for this

jwhitlock (changes requested):
Thanks @say-yawn! This worked locally for me. I forwarded a newsletter to my relay address, and it found 30 strict trackers and 0 general trackers.

I think we need some documentation on the source of the tracker list. I also think the `email_tracker` message is not useful - 351 log messages to mostly say a tracker was _not_ found. I think it will be more useful if the `email_tracker_summary` message has a summary of the found trackers, like:

```json
{
   "general": 0,
   "strict": 30,
   "trackers": {
       "strict": {
           "list-manage.com": 30
       }
    }
}
```

emails/tracker_lists/general-tracker.json
```diff
@@ -0,0 +1,216 @@
+[
+  "10web.io",
```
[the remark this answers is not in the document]
say-yawn: Oh, so these lists are coming from shavar-prod-lists' blocklist [here](https://github.com/mozilla-services/shavar-prod-lists/blob/master/disconnect-blacklist.json). So ideally we may want to do a github action to either (1) parse the list to create files on our repo (2) submodule the entire repo and parse when service starts. We may end up having to do (2) because of license reasons but while we figure out long term solution and not block testing I decided to add it to our repo.

Not sure about how often the list gets uploaded but for testing we should use a static list (test/study is expected to end on May 4th and this is when I hope to have a long term solution for maintaining this list.

emails/views.py
say-yawn: Thanks!

say-yawn:
@jwhitlock made updates based on your recommendation to improve logging only once after iterating and collecting non-zero tracker count. The `email_tracker_summary` logging will now look like:
```python
{
  'general': {'count': tracker_total, 'trackers': {domain: count, ...}},
  'strict': {'count': tracker_total, 'trackers': {domain: count, ...}}
}
```
where `tracker_total` is the total number of trackers found in that level, and the `'trackers'` key will have a dictionary of all the trackers and the number of the domain found in the email for the given level.

jwhitlock (approved):
Thanks @say-yawn, looks ready for the experiment!

PR #1795, opened 2022-04-14
Update cryptography 36.0.2, pyOpenSSL

jwhitlock:
PR #1647 is failing because it requires a [cryptography](https://github.com/pyca/cryptography) update, so this PR updates both.

Some highlights from [v3.3.2](https://cryptography.io/en/latest/changelog/#v3-3-2) to [v36.0.2](https://cryptography.io/en/latest/changelog/#v36-0-2):

* New numbering scheme with jump from 3.4.8 to 35.0.0, inspired by [Firefox versions](https://cryptography.io/en/latest/api-stability/#version-35-0-0), along with a new [deprecation schedule](https://cryptography.io/en/latest/api-stability/#deprecation)
* Use Rust for x.509 chain, require Rust 1.41.0 to build from source
* Drop support for Python 2
* Add typing

We use `cryptography` for the reply keys, and this part doesn't appear to have any changes:

https://github.com/mozilla/fx-private-relay/blob/a16c5df5cb586856230ae6448196cf677d3be068/emails/utils.py#L256-L263

This updates [pyopenssl](https://github.com/pyca/pyopenssl) from 19.1.0 to 22.0.0. Highlights from the [changelog](https://github.com/pyca/pyopenssl/blob/main/CHANGELOG.rst):
* Drop Python 2.7, 3.4, and 3.5
* Fix compatibility with OpenSSL 1.1.0, drop support for 1.0.1 and 1.0.2

We're using OpenSSL.crypto to verify the signature of the SNS messages:

https://github.com/mozilla/fx-private-relay/blob/a16c5df5cb586856230ae6448196cf677d3be068/emails/sns.py#L67-L80

However, the [OpenSSL.crypto docs](https://www.pyopenssl.org/en/stable/api/crypto.html) come with a big warning:

> [pyca/cryptography](https://cryptography.io/) is likely a better choice than using this module. It contains a complete set of cryptographic primitives as well as a significantly better and more powerful X509 API. If necessary you can convert to and from cryptography objects using the to_cryptography and from_cryptography methods on X509, X509Req, CRL, and PKey.

If this function remains after transitioning to SQS polling, we may want to convert and drop pyOpenSSL.


groovecoder:
Spot-check of SNS verification and replies still works. Looks like this needs a conflict resolution?

groovecoder (approved):

PR #1809, opened 2022-04-19
Add optional django-silk

jwhitlock:
[Silk](https://github.com/jazzband/django-silk) is a live profiling and inspection tool for Django, suitable for local development. It is disabled when ``DEBUG=False``. It is also disabled when running tests, since it adds database queries for writing profile data to the database, using the mechanism suggested at https://github.com/pytest-dev/pytest/issues/9502.

This is based on https://gist.github.com/groovecoder/5d2963d753cdfa690507a415b565dbaa

I'm not convinced we should add it by default, but if you want to use it, this makes it a lot easier.

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -22,9 +22,24 @@ from sentry_sdk.integrations.logging import ignore_logger
 from hashlib import sha512
 import base64
 
+
+# This needs to be before markus, which imports pytest
+IN_PYTEST = "pytest" in sys.modules
+if IN_PYTEST:
+    assert 1==2
+
+
 from django.conf import settings
 
 import dj_database_url
+
+try:
+    # Silk is a live profiling and inspection tool for the Django framework
+    # https://github.com/jazzband/django-silk
+    import silk
+    HAS_SILK = True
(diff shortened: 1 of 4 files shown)


privaterelay/settings.py
```diff
@@ -22,10 +22,25 @@
 from hashlib import sha512
 import base64
 
+
+# This needs to be before markus, which imports pytest
+IN_PYTEST = "pytest" in sys.modules
+if IN_PYTEST:
+    assert 1==2
```
groovecoder, line 29: question: I don't understand why we're doing this? This looks like it's just going to fail everything when we're "in pytest"?

groovecoder (commented):
praise: nice way to clean up code comments and make silk conditional! question on the `IN_PYTEST` always-fail block.

privaterelay/settings.py
jwhitlock: That is because I didn't remove my test code from the PR 😅 
Also, I should move this branch to the main repo so CI will run the tests.

jwhitlock:
Replaced by PR #1829.

PR #1813, opened 2022-04-19
Handle unexpected SNS message types in `_sns_notifications`

### LINKED ISSUE #1812: TypeError: expected string or bytes-like object in /emails/sns-inbound
This Sentry issue occurs about 100 times a day:

https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/18992993

The error is ``TypeError("expected string or bytes-like object")``. The code that triggers it is line 299:

https://github.com/mozilla/fx-private-relay/blob/f72f1e7997c1bdbf6bc4489bc4aa4c77fc4ce3e9/emails/views.py#L290-L306

It should instead be:

```python
        logger.error(
            'SNS notification for unsupported type',
            extra={
                'notification_type': shlex.quote(notification_type),
                'event_type': shlex.quote(event_type),
                'keys': [shlex.quote(key) for key in message_json.keys()],
            },
        )
```

Synced with MPP-1885

jwhitlock:
This PR fixes #1812, aka MPP-1885.

How to test:

I think in both these cases, we're prefer to log the entire `json_body`, so we have access to the confirmation URL or the complaint. However, in the past we've preferred to not log this in case it is large or has sensitive information.

### DIFF
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -376,7 +376,6 @@ class SNSNotificationInvalidMessageTest(TestCase):
         response = _sns_notification(json_body)
         assert response.status_code == 400
 
-    @pytest.mark.xfail(reason="raises JSONDecodeError")
     def test_subscription_confirmation(self):
         """A subscription confirmation returns a 400 error"""
         json_body = INVALID_SNS_BODIES['subscription_confirmation']
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -382,7 +382,6 @@ class SNSNotificationInvalidMessageTest(TestCase):
         response = _sns_notification(json_body)
         assert response.status_code == 400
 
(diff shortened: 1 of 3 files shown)


emails/tests/views_tests.py
```diff
@@ -357,6 +369,41 @@ def test_noreply_email_in_s3_deleted(
...
+    def test_subscription_confirmation(self):
+        """A subscription confirmation returns a 400 error"""
+        json_body = INVALID_SNS_BODIES['subscription_confirmation']
+        response = _sns_notification(json_body)
+        assert response.status_code == 400
+
+    def test_notification_type_complaint(self):
+        """A notificationType of complaint returns a 400 error"""
+        # Manual json_body because no instances captured, from
+        # https://docs.aws.amazon.com/ses/latest/dg/notification-contents.html#complaint-object
```
groovecoder, line 388: praise: yay more doc links in code comments!!

groovecoder (approved):

PR #1815, opened 2022-04-20
Remove cached data when logging out

Vinnl:
This is a follow-up to https://github.com/mozilla/fx-private-relay/pull/1777#pullrequestreview-942422686 by @groovecoder.

Although we now properly detect that a user is logged out after
mocking logout, the user's data would still be cached, such as the
fact that they're a Premium user, causing the header to still be
in the Premium theme.

With this change, user data is also cleared.

Note that this only affected when running using the dev server, as
authentication in production involves full page reloads and hence
the clearing of all in-memory data.

How to test: Log in as a Premium user when running using the dev server. Now logout. Observe that the "Sign up" and "Sign in" buttons are visible again, and that the header is now dark.

- [ ] I've added a unit test to test for potential regressions of this bug. No, mocking both the router and the user API for what is otherwise a static page felt like overkill, but let me know if you disagree.

### DIFF
--- a/frontend/src/pages/mock/logout.page.tsx
+++ b/frontend/src/pages/mock/logout.page.tsx
@@ -1,6 +1,7 @@
 import { NextPage } from "next";
 import { useRouter } from "next/router";
 import { MouseEventHandler } from "react";
+import { useSWRConfig } from "swr";
 import { getRuntimeConfig } from "../../config";
 import { authenticatedFetch } from "../../hooks/api/api";
 import { useProfiles } from "../../hooks/api/profile";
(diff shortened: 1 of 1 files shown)


frontend/src/pages/mock/logout.page.tsx
```diff
@@ -23,7 +25,11 @@ const MockLogout: NextPage = () => {
       method: "GET",
       redirect: "manual",
     });
-    // Revalidate account data to reflect logged out status:
+    // Revalidate account data to reflect logged out status.
+    // Note that the `clear` method exists, even though it's not exposed on the
+    // type yet:
+    // https://github.com/vercel/swr/issues/161#issuecomment-1079198998
```
groovecoder, line 31: praise: thanks for sending a PR upstream! I subscribed to it to hopefully help remember when we can remove this work-around.

PR #1822, opened 2022-04-20
check request.COOKIES for session cookie

groovecoder:
We want to keep Session and Authentication middleware AFTER Whitenoise,
so we're not running the session & auth logic on every request for
static files. But we also want to redirect logged-in users to
/accounts/profile/ BEFORE they see a flash of not-logged-in content on
the React index.html.

So, rather than check for request.user, which depends on the Session and
Authentication middleware running first, we can check if there's any
session cookie in the request at all - which indicates the user is
signed-in.

This fixes the flash-of-unauthenticated-content with a light-weight
middleware that is easier on the requests for static files.

### DIFF
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -4,6 +4,7 @@ import time
 import markus
 from oauthlib.oauth2.rfc6749.errors import CustomOAuth2Error
 
+from django.conf import settings
 from django.contrib.auth import logout
 from django.shortcuts import redirect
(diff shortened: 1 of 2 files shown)


Vinnl (approved):
Ooh nice job! Tested it locally and works for me. Only remark I'd have is that the auth middleware is in a different order in `main` - I don't know if it matters, but if you don't either, might be best to preserve that order.

PR #1824, opened 2022-04-20
React site - Profile UI

codemist:
# New feature description

This aligns the profile/dashboard UI of the React site with the Nebula Design System.

Key changes:
- Hide the upgrade banner during a non-premium onboarding phase, only showing the footer banner.
- Change delete button UI in alias card to follow Nebula Design System's version.
- Refine spacing/white space in alias cards.

# Checklist


frontend/src/components/dashboard/ProfileBanners.tsx
```diff
@@ -22,12 +22,14 @@ import { trackPurchaseStart } from "../../functions/trackPurchase";
...
 import { SubdomainPicker } from "./SubdomainPicker";
 import { useMinViewportWidth } from "../../hooks/mediaQuery";
+import { AliasData } from "../../hooks/api/aliases";
 
 export type Props = {
   profile: ProfileData;
   user: UserData;
   onCreateSubdomain: (chosenSubdomain: string) => Promise<void>;
   runtimeData?: RuntimeData;
+  aliases: AliasData[];
```
codemist, line 32: @Vinnl I tried adding `props.alias.length`  as a condition for showing the Premium banner under ProfileBanners.tsx, but I am getting the `can't access property "length", props.aliases is undefined` error - I've omitted it for now but was wondering why you think this error is being thrown, since the length property seems to work in other components (see `AliasGenerator.tsx` and `AliasList.tsx`)

Vinnl: If you look at [the failed check](https://app.circleci.com/pipelines/github/mozilla/fx-private-relay/4933/workflows/95c2fe6c-b954-41b3-a431-b9400f263571/jobs/10852) linked below, you can see the following error:

```
./src/pages/accounts/profile.page.tsx:303:8
Type error: Property 'aliases' is missing in type '{ profile: ProfileData; user: UserData; onCreateSubdomain: (customSubdomain: string) => Promise<void>; runtimeData: RuntimeData; }' but required in type 'Props'.

  301 |   const banners = (
  302 |     <section className={styles["banners-wrapper"]}>
> 303 |       <ProfileBanners
      |        ^
  304 |         profile={profile}
  305 |         user={user}
  306 |         onCreateSubdomain={setCustomSubdomain}
```

So while you indicated in the `Props` type of `<ProfileBanners>` that you want it to be passed a list of aliases, no such list actually gets passed in yet where the component is included (i.e. in `profile.page.tsx`). Luckily, that component does already have the list of the user's aliases (in the `allAliases` variable), so you have it give that to the `<ProfileBanners>` component.

Also, note that you named it `aliases`, so you'll need `props.aliases.length` rather than `props.alias.length` :)

codemist: Passing aliases to the `<ProfileBanners>` worked, thank you Vincent! Does this mean that it'll throw an error if I don't utilize all of the props when passing a component to pages where its being injected?

Vinnl: Basically you can define how your component should be called (i.e. when it will throw an error) via the `Props` type. In this case, you defined it as adding an `aliases` attribute ("prop") that should be passed an array of `AliasData` objects:

https://github.com/mozilla/fx-private-relay/blob/f94960c613f61ff74243bf84514f02f03e092fb5/frontend/src/components/dashboard/ProfileBanners.tsx#L32

You could also have defined it as accepting either that array, or `undefined`, effectively making it optional:

```ts
  aliases: AliasData[] | undefined;
```

That's actually such a common use case that there's a shorthand for it:

```ts
  aliases?: AliasData[];
```

(i.e. the `?` marks it as optional)

frontend/src/components/dashboard/aliases/BlockLevelSlider.module.scss
```diff
@@ -29,11 +29,11 @@ $stopLabelHeight: 2rem;
 
   .control {
     width: 100%;
-    max-width: $content-sm;
+    max-width: $content-xs;
     display: flex;
     flex-direction: column;
     align-items: center;
-    padding: 0 $spacing-lg;
+    padding-left: $spacing-md;
```
Vinnl, line 36: Nitpicky, but shouldn't we have this padding on both sides? It looks a little bit off on small screens for me now, I think:

image

Vinnl (approved):
Nice, no major remarks.

If you're up for a challenge, you could look at adding a test to `profile.test.tsx` to verify that the Banner is not shown if the user has 0 aliases. It would be similar to e.g. the test "it does not show a banner to upgrade to Premium if the user already has Premium", but using `setMockAliasesDataOnce` instead of `setMockRuntimeDataOnce`.

That's purely if you want to get some experience writing tests though; I don't think the risk of bugs here is such that it _needs_ a test. And we could pair on that as well, of course :)

maxxcrawford:
@Vinnl Looks like there was not an follow up from your review. I filed https://github.com/mozilla/fx-private-relay/issues/1860 to capture the test suggestion (good one!). I'll set a reminder for @codemist when she's back in the saddle about the padding issue. We can roll with it as is for now!

PR #1831, opened 2022-04-22
Add mypy for Python type hints

jwhitlock:
Add a working ``mypy`` configuration in ``pyproject.toml``, with minimal code changes. I've include type stubs for Django, Django REST Framework, and requests. The Django stubs are for 3.2, so it doesn't find `User.profile_set`, but it is necessary for the DRF stubs and ``pyproject.toml`` support, which we'll use for Black.

I've added ``mypy`` checking to CircleCI, using some of the changes I tried in PR #1830. It is overkill for this change, but will make it easier to add `black` and `flake8` or `pylint` to CI later.

I haven't added any pre-commit hooks for mypy yet. I've used [pre-commit](https://pre-commit.com) on other projects, it is fine. This [stackoverflow answer](https://stackoverflow.com/a/68946077/10612) suggests you can have the Husky commit hook call ``pre-commit``.

Testing:
* ``pip install -r requirements.txt``
* ``mypy .``
* Look at CircleCI workflow results


jwhitlock:
This build shows what a "soft failed" mypy looks like:

https://app.circleci.com/pipelines/github/mozilla/fx-private-relay/4977/workflows/7480b424-0bee-42ea-a8d9-3aa1353d8f26/jobs/11087

The last build shows a successful mypy build:

https://app.circleci.com/pipelines/github/mozilla/fx-private-relay/4979/workflows/55d6d653-38a5-42aa-893a-54f008cc7224/jobs/11099

.circleci/config.yml
```diff
@@ -269,6 +278,49 @@ jobs:
               docker push "${DOCKERHUB_REPO}:${CIRCLE_TAG}"
             fi
 
+  python_job:
```
groovecoder, line 281: praise: TIL creating custom jobs in/for Circle!

groovecoder (commented):
Whoa, is this really ALL the type-checking fixes we have to make across the whole code-base?

jwhitlock:
Re-based on main, and fixed some typing issues from the last few weeks.

I tried enabling `warn_unused_ignores`. This ignore is needed in my development environment, but is not needed in CircleCI:

https://github.com/mozilla/fx-private-relay/blob/dcd5225befb85133eb2b6390e2590976286039ba/emails/tests/views_tests.py#L393-L394

I found a related issue at https://github.com/typeddjango/django-stubs/issues/878. The fix is merged, and it looks like it will be released in the next version.

I'd like to merge this version, but I'd be OK including more conversion to types. Converting a recent management command, like [emails/management/commands/check_health.py](https://github.com/mozilla/fx-private-relay/blob/main/emails/management/commands/check_health.py), would be easiest. Converting a utility file, like [main/emails/utils.py](https://github.com/mozilla/fx-private-relay/blob/main/emails/utils.py), would be the highest impact.

jwhitlock:
A few changes from the last comment:

* I added a typing stub for `decouple`, and added more types to `privaterelay.settings`
* Updated to `mypy` 0.950, `django-stubs` 1.11.0, and `djangorestframework-stubs` 1.6.0. Weirdness between my dev and CircleCI has gone away.

I've learned a ton about mypy and typing, and it really helped when adding code to handle SES complaints and notifications, finding many issues as I worked. 

However, `mypy` is not like a lot of linters, where the changes are trivial and easy. There is a ton of new concepts, new tools, etc. I think the next step is a live introduction to `mypy`, showing how you'd get from our current state to no errors. I'll try to schedule that this month, for at least the backend devs.

jwhitlock:
I made some notes on [Converting fx-private-relay to use mypy](https://gist.github.com/jwhitlock/bb149f3d1525b21c563b25b695d7ad4d), which goes into a lot more details about the changes in this PR, and tips on how to use `mypy`. I used the notes for two-hour trainings with @say-yawn, @groovecoder, and @Vinnl.

Creating the notes and going through the training suggested a few changes, added to this PR:
* `mypy_stubs/decouple.pyi` is better with unused code deleted than commented out. `config` is easier to understand if typed as a function rather than an implementation of `Config`.
* It looks better to use `list[str]` than `tuple[str, ...]`, and other adjustments that make `mypy` happy _and_ make for better code.
* `django_stubs_ext.monkeypatch()` exists, is nice
* The `types-pyOpenSSL` package works nicely
* The `env` folder (created by `virtualenv` during project setup) should be ignored if it exists

I've rebased again, adjusted the configuration, and updated to the latest packages. I also annotated the functions in `privaterelay/settings.py`, so that everything in there is annotated. I can't mark it for `strict=True` yet, because it required annotations in models first.

I'm marking everyone that's gone through training as a possible reviewer - I'll just take one. If you think `mypy` is a bad addition, now is a good time to say that as well!

.circleci/config.yml
```diff
@@ -389,6 +392,24 @@ workflows:
             tags:
               only: /.*/
 
+      - python_job:
+          name: mypy type check
+          command: mypy --no-incremental --junit-xml job-results/mypy.xml .
+          has_results: true
+          allow_fail: true
```
Vinnl, line 399: Wouldn't we want this one to be blocking?

mypy_stubs/README.md
```diff
@@ -0,0 +1,44 @@
+# mypy type hints for 3rd party modules
```
Vinnl, line 1: Nice docs, and located where they're needed!

mypy_stubs/README.md
```diff
@@ -0,0 +1,44 @@
...
+The generated stubs are placeholders, and often require significant
+modifications to run without errors. They are also tied to the specific
+version of the package. For this reason, we don't require stubs for
+all third-party packages.
+
+## Documentation
+
+When generating stubs, add a docstring header to the file that specifies the
+`pip` package name and header. This gives us a chance to recognize that the
+stub is out-of-date when updating the package.
```
Vinnl, line 38: Is there a common convention for this? If so, I'm wondering if there's a lint rule that could catch it when this is missing. (Similar to [this one](https://github.com/microsoft/DefinitelyTyped-tools/blob/master/packages/dtslint/docs/dt-header.md) for TS type definitions.)

privaterelay/settings.py
```diff
@@ -29,6 +32,9 @@
 
 import dj_database_url
 
+if TYPE_CHECKING:
+    import wsgiref.headers
```
Vinnl, line 36: Might be my unfamiliarity, but it's not clear to me why this is needed - might warrant a comment.

Vinnl: Oh hmm, I guess it's just to get at a type definition without including it at runtime, similar to TypeScript's [type-only imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export).

Vinnl (approved):
Love to see this!

.circleci/config.yml
jwhitlock: Yes, eventually. I want to give developers a week or so (or maybe during an upcoming work week) to get the tools installed locally first.

mypy_stubs/README.md
jwhitlock: There is not a convention that I'm aware of, but that seems like a good idea. Practically speaking, I might be the only developer creating these for the next few months.

privaterelay/settings.py
jwhitlock: Yes, it is a good analogy. From [the docs](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING):

> A special constant that is assumed to be `True` by 3rd party static type checkers. It is `False` at runtime.

I thought `TYPE_CHECKING` would come up earlier, but this is the first time I needed it in our code.

PR #1852, opened 2022-04-26
Separate the query param properly

say-yawn:
How to test:
## Check emails with "Report breakage" in header has &control
1. Send an email to Relay account with Mozilla email address
2. Check the relayed email has "Report breakage"
3. Hover over the link and check that it ends with `&control=True` or `&control=False`

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -480,7 +480,7 @@ def _sns_message(message_json):
                 'https://www.surveygizmo.com/s3/6837234/Relay-General-Email-Tracker-Removal-2022?'
                 + f'general-found={general_count}&'
                 + f'general-removed={removed_count}&'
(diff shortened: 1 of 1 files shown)


jwhitlock (changes requested):
Sorry I didn't catch this the first time - this code should be using [urllib.parse.urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) to encode the dict as a query string, which will include putting the `&`  at the right spots and avoid other future problems.

jwhitlock (approved):
I'd rather get this fixed now then wait for it to get fixed the "right" way. I can refactor later if needed, but there's also a good chance this survey this will go away before that happens.

Issue #1586, opened 2022-03-01
Update to Django 3.2


jwhitlock:
We're on Django 2.2, which will reach the end of extended support in April 2022. Django 3.2 will be be the supported LTS release until April 2024.

References:
* [Download Django](https://www.djangoproject.com/download/) for LTS information
* [Upgrading Django to a newer version](https://docs.djangoproject.com/en/3.2/howto/upgrade-version/#upgrading-django-to-a-newer-version)
* [Deprecation Timeline](https://docs.djangoproject.com/en/3.2/internals/deprecation/)
* [Release notes](https://docs.djangoproject.com/en/3.2/releases/)
  * [3.0 release notes](https://docs.djangoproject.com/en/3.2/releases/3.0/)
  * [3.1 release notes](https://docs.djangoproject.com/en/3.2/releases/3.1)
  * [3.2 release notes](https://docs.djangoproject.com/en/3.2/releases/3.2)

jwhitlock:
This shipped in [3.5.1](https://github.com/mozilla/fx-private-relay/releases/tag/3.5.1). It works! Follow-on issues are tracked as new issues, like #1995.

PR #1853, opened 2022-04-26
MPP-1723: Update to Django 3.2

jwhitlock:
Update to Django 3.2, following the guide [How to upgrade Django to a newer version](https://docs.djangoproject.com/en/4.0/howto/upgrade-version/). This fixes issue #1586 / MPP-1723.

This required an update to two dependencies:
* ``django-allauth 0.31.0`` to ``0.50.0``. From the [release notes](https://django-allauth.readthedocs.io/en/latest/release-notes.html):
    - 0.32.0 - Views return JSON on AJAX ``GET`` as well as ``POST``
    - 0.35.0 - Support Django 2.0
    - 0.38.0 - Avoid XSS with ``{% user_display user %}`` (*not used in our code*)
    - 0.39.0 - Official Django 2.2 support
    - 0.41.0 - Allow storing OAuth creds in settings instead of DB
    - 0.43.0 - Switch from ``base.html`` to ``account/base.html``
    -  0.46.0 - Django 3.2 compatibility
    - 0.47.0 - Add ``SOCIALACCOUNT_LOGIN_ON_GET``, default False. **This would affect our code**, so added ``SOCIALACCOUNT_LOGIN_ON_GET=True``
    - 0.48.0 - Rate limiting on by default. **This may affect our code**, if an IP signs up more than the default of 20 times a minute. See [ACCOUNT_RATE_LIMITS](https://django-allauth.readthedocs.io/en/latest/configuration.html?highlight=ACCOUNT_RATE_LIMITS#configuration). A ``429.html`` template / view may be needed.
    - 0.48.0 - ``SOCIALACCOUNT_STORE_TOKENS`` defaults to False. **This would affect our code**, which stores FxA tokens for profile refreshes, so added ``SOCIALACCOUNT_STORE_TOKENS=True``
* ``django-csp 3.5`` to ``3.6``. From the [CHANGES](https://github.com/mozilla/django-csp/blob/main/CHANGES):
    - Add support/testing for Django 2.2 and 3.0
    - Add support/testing for Python 3.7 and 3.8
    - Disable CSP for Django ``NotFound`` debug view
    - Add new headers used in CSP level 3
    - Add support for the ``report-to`` directive
    
For the Django updates:

* Django 3.0 significant changes from [release notes](https://docs.djangoproject.com/en/4.0/releases/3.0/):
    - Adds ASGI support, PostgreSQL exclusion constrainers, ``BooleanField`` in filter expressions, more choice fields.
    - Removes PostgreSQL 9.4, Python 2 APIs and ``six``
    - Python 3.9 supported, starting in Django 3.0.11
    - ``SECURE_CONTENT_TYPE_NOSNIFF`` defaults to True
    - ``X_FRAME_OPTIONS`` defaults to 'DENY'
* Django 3.1 significant changes from [release notes](https://docs.djangoproject.com/en/4.0/releases/3.1/):
    - Adds support for async views and middleware, ``JSONField``, SHA-256 default hashing, bloom indexes, and more pathlib support.
    - Add migration of ``User.first_name`` from 30 to 150 characters.
    - Security change: ``SECURE_REFERRER_POLICY`` default is ``"same-origin"``, preventing ``Referer`` header sent to other origins.
    - Security change: Default signing algorithm is SHA-256. Signatures made with SHA-1 supported until Django 4.0.
    - Drop support for MariaDB 10.1, Internet Explorer
    - Deprecate PostgreSQL ``JSONField`` in favor or new ``JSONField``
    - Python 3.9 supported, starting 3.1.3
    - ``MEDIA_ROOT``, ``MEDIA_URL`` have to be different from ``STATIC_ROOT``, ``STATIC_URL``. Since we don't support uploads, this sets ``MEDIA_ROOT`` and ``MEDIA_URL`` to None.
* Django 3.2 significant changes from [release notes](https://docs.djangoproject.com/en/4.0/releases/3.2/):
    - Adds automatic AppConfig discovery, customize type of primary keys, functional indexes, new admin site decorators, HttpResponse.headers interface.
    - Drop support for PostgresSQL 9.5, MySQL 5.6
    - Use ``DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'`` to preserve current type, future default will be ``BigAutoField``


groovecoder (approved):
question (non-blocking): now that `SECURE_CONTENT_TYPE_NOSNIFF` defaults to `True`, should we remove our code that sets it to `True` based on an env var?

praise: great thorough work here! code and spot-checks look good! approved, but I'll let you handle the merge so you can coordinate stage + prod deploys.

jwhitlock:
> question (non-blocking): now that SECURE_CONTENT_TYPE_NOSNIFF defaults to True, should we remove our code that sets it to True based on an env var?

I'd like to. You added this override in https://github.com/mozilla/fx-private-relay/pull/27 - do you think we've tested that change enough?

@moz-astults do we override `SECURE_CONTENT_TYPE_NOSNIFF` in stage or prod? I checked in Heroku, we do not override it there.

requirements.txt
```diff
@@ -3,7 +3,7 @@ codetiming==1.3.0
 cryptography==36.0.2
 Django==2.2.27
 dj-database-url==0.5.0
-django-allauth==0.39.1
+django-allauth==0.50.0
```
say-yawn, line 6: `ACCOUNT_RATE_LIMITS` is introduced which uses per minute (guessing from this [code snippet](https://github.com/pennersr/django-allauth/blob/8a4201ade850a0819513cdbff50f251087e4651b/allauth/ratelimit.py#L12-L22)). It seems well have to provide `429.html` as mentioned [here](https://django-allauth.readthedocs.io/en/latest/release-notes.html?highlight=rate%20limiting#backwards-incompatible-changes-1).

say-yawn (approved):
The changes look good and thank you so much for doing this. I think I saw a breaking change while updating `django-allauth` that we may want to either set `ACCOUNT_RATE_LIMIT` to `{}` as mentioned in [configuration doc](https://django-allauth.readthedocs.io/en/latest/configuration.html?highlight=rate%20limiting#configuration) or either include `429.html`.

privaterelay/settings.py
```diff
@@ -708,3 +714,6 @@ def _get_initial_middleware():
 if USE_SILK:
     SILKY_PYTHON_PROFILER=True
     SILKY_PYTHON_PROFILER_BINARY=True
+
+# Django 3.2 switches default to BigAutoField
+DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
```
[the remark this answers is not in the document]
jwhitlock: I was curious if there was any reason to do this migration, unless you were hitting the limits of 32-bit indexes. This talks about some [theoretical improvements](https://serverfault.com/questions/160082/should-i-install-32-bit-database-or-64-bit-database), but [others find int8 slower than int4 slower when profiling](https://stackoverflow.com/questions/6633050/performance-difference-between-int4-and-int8-in-postgresql-on-64-bit-server). Most of those posts are about a decade old, so I'm assuming the perf issues have been worked out since then.

I'd start a new project on `BigAutoField`, but I'm not in a rush for Relay.

requirements.txt
jwhitlock: I'm not sure how rate-limited interacts with the React front-end. Before I re-enabled `SOCIALACCOUNT_LOGIN_ON_GET=True`, login involved a barely-styled auth form. 

I think a `429.html` is the right solution, so it is a usable error rather than a "template not found error". I'll spend an hour playing around and see if I can get rate-limited locally.

jwhitlock:
> @moz-astults do we override SECURE_CONTENT_TYPE_NOSNIFF in stage or prod? I checked in Heroku, we do not override it there.

@moz-astults confirmed in another channel that we're not setting this, so let's simplify to the new default without override.

requirements.txt
jwhitlock: TLDR: I'm adding a `429.html` which I don't think will ever render.

I went through the steps to get this working. In one shell:

```sh
brew install redis
/opt/homebrew/opt/redis/bin/redis-server /opt/homebrew/etc/redis.conf
```

In my `.env`, I added:
```sh
REDIS_URL="redis://127.0.0.1:6379"
```

I tested it:
```sh
pip install -r requirements.txt
./manage.py shell
>>> from django.core.cache import cache
>>> cache.set('foo', 'bar')
True
>>> cache.get('foo')
'bar'
```

And then added to settings:

```python
ACCOUNT_RATE_LIMITS = {
    "change_password": "1/d",
    "manage_email": "1/d",
    "reset_password": "1/d",
    "reset_password_email": "1/d",
    "reset_password_from_key": "1/d",
    "signup": "1/d",
}
ACCOUNT_LOGIN_ATTEMPTS_LIMIT=1
ACCOUNT_LOGIN_ATTEMPTS_TIMEOUT=300
```

I then tried to signup multiple times, and no rate limiting. 

I peeked at the keys:

```sh
> redis-cli keys \*
1) ":1:foo"
2) ":1:django.contrib.sessions.cachej8jyv03r2r7mcxvgcdnrxjabfrs7gkzt"
```

So, despite a cache being setup and a rate of once a day, allauth is not storing any usage data.

I peeked at the code:

https://github.com/pennersr/django-allauth/blob/d367458580f5ebe6e6a0903ba78f85e2e082e0f2/allauth/ratelimit.py#L58-L60

```python
if request.method == "GET" or not amount or not duration:
        pass
```

SO, rate limiting is only in effect if the request is a `POST`, and since we use `GET` for sign-in and sign-ups, django-allauth doesn't keep track of it.

I wrote a basic ``429.html`` that doesn't use Fluent or our website styling, so it will be very basic. I also think it will not ever be displayed.

PR #1854, opened 2022-04-27
MPP-1834 Update Toast Styling

### LINKED ISSUE #1745: Update toast styling
(Extracted from https://github.com/mozilla/fx-private-relay/issues/1744#issuecomment-1090370660.)

There's [a new design for toasts](https://www.figma.com/file/cBaBmcPisqZ0QQX5LLzpjU/?node-id=300%3A0). Toasts are rendered in `<Layout>` (see the `<ToastContainer>` element), so you can update its styling in `Layout.module.scss` (primarily via the `.toast` class, but if that's not enough, be sure to take a look at [the react-toastify docs](https://fkhadra.github.io/react-toastify/how-to-style)).

lloan:
This PR fixes #1745 . 

**How to test:** 
TBD

## Checklist 


lloan:
@maxxcrawford 

**Current status:** 
image

**Figma:** 
image

Would it be correct to assume that the height in the Figma specs wants us to set the overall height of the Toast notification to 40px? (line height and close button positioning are not modified here) - I modified the toastify-toast-min/max-height to 40px and this is what it looks like. 

For future reference, are these questions I should be asking our design team?

image

lloan:
@maxxcrawford @Vinnl one more thing - the modifications to the Toast component: I see we are using the CSS module approach. Where would we want to add modifications for component pieces like the close button? Without adding CSS to something like `globals.scss`, I had to make a custom close button for the `ToastContainer` but it seems overkill. I wanted to get your stance on where that kind of styling would go since you're early into decisions like this. For context, the styling would be something like: 

`.Toastify__close-button {
    padding: 11px;
}`

Vinnl:
Just responding to one semi React related thing since I'm off today:

> Where would we want to add modifications for component pieces like the close button? Without adding CSS to something like globals.scss, I had to make a custom close button for the ToastContainer but it seems overkill.

Components are cheap, so that seems fine by me. And [the Toastify docs say](https://fkhadra.github.io/react-toastify/how-to-style#override-existing-css-classes):

> /** Classes for the close button. Better use your own closeButton **/

(And in case it's useful, we do have [a file with icon components](https://github.com/mozilla/fx-private-relay/blob/b3d715119be4c5968a60aa71240871be5d476cf4/frontend/src/components/Icons.tsx#L33-L56) including a close button icon.)

lloan:
> Just responding to one semi React related thing since I'm off today:
> 
> > Where would we want to add modifications for component pieces like the close button? Without adding CSS to something like globals.scss, I had to make a custom close button for the ToastContainer but it seems overkill.
> 
> Components are cheap, so that seems fine by me. And [the Toastify docs say](https://fkhadra.github.io/react-toastify/how-to-style#override-existing-css-classes):
> 
> > /** Classes for the close button. Better use your own closeButton **/
> 
> (And in case it's useful, we do have [a file with icon components](https://github.com/mozilla/fx-private-relay/blob/b3d715119be4c5968a60aa71240871be5d476cf4/frontend/src/components/Icons.tsx#L33-L56) including a close button icon.)

Go enjoy your day off! :) 

And by the way, that's exactly what I used for the close button, its a close icon from the `icon.tsx` file :)

frontend/src/components/layout/Layout.module.scss
```diff
@@ -7,17 +7,23 @@
...
-  border-radius: $border-radius-md;
   text-align: center;
-  font-weight: 700;
+  font-weight: 600;
+  font-size: 13px;
+  line-height: 21px;
+  border-radius: $border-radius-sm;
   font-family: $font-stack-base;
+  padding-top: 9.5px;
+  padding-bottom: 9.5px;
```
maxxcrawford, line 26: Suggestion: Use Protocol tokens to set padding. We do not have to be pixel perfect, and we nudge designs to where they fit within the set variables. (Example: Random graytones get matched to the closest `$color-grey-##`, etc.) 

In this case: 
```suggestion
  padding-top: $spacing-md;
  padding-bottom: $spacing-md;
```

maxxcrawford: This also brings up a larger conversation with UX: The current toast design (from Nebula) is never wider than ~400px. I would check with UX and confirm they want full width toasts, or if there's any design changes necessary for a full-bodied 2000px wide toast. 🍞

frontend/src/components/layout/Layout.module.scss
```diff
@@ -7,17 +7,23 @@
...
-  --toastify-color-success: #{$color-green-60};
-  --toastify-text-color-success: #{$color-white};
-  --toastify-color-error: #{$color-red-50};
-  --toastify-text-color-error: #{$color-white};
+  --toastify-color-light: #{$color-white};
+  --toastify-color-dark: #{$color-grey-50};
+  --toastify-color-info: #{$color-blue-50};
+  --toastify-color-success: #{$color-green-50};
+  --toastify-color-warning: #{$color-yellow-50};
+  --toastify-color-error: #{$color-red-60};
```
maxxcrawford, line 15: Praise: Nice Protocol color token usage!

maxxcrawford (commented):
Not sure about how you're setting the `40px` height exactly. I would let the padding define it and go from there. 

I know this is a WIP review, but wanted to chime in. Be sure to isolate the close button from the main toast styling: 

image

frontend/src/components/layout/Layout.module.scss
lloan: Trying as best as possible to stick to Protocol :) - code updated.

lloan: Also spoke to Eduardo - This is what was adviced: 
image

Using `$content-xl` as the max-width for the toast container, you can see it goes over a bit, in comparison to the container below it (with content) - the container below it has a padding of 16px or so on each side. 

This is what it looks like if I remove `layout-sm` (32px) from the max-width value. Up to you what you want us to do. 
image

frontend/src/components/layout/Layout.module.scss
```diff
@@ -7,17 +7,72 @@
...
+  font-size: 13px;
+  line-height: 21px;
+  height: $layout-md;
+  border-radius: $border-radius-sm;
   font-family: $font-stack-base;
+  padding: 0;
+  max-width: $content-xl;
+  margin: 0 auto;
+
+  &:global {
```
lloan, line 36: @Vinnl is this alright here? 

It allows us to target based on toast state (error, etc.) and we can update `$toastify-toast-close-button` with the correct color since they're separate but one depends on the other. Unless there is a better way.

lloan: last question before you review @maxxcrawford - are we removing the progress bar too? I went ahead and removed it since it looks like we don't use it, based on the Figma.

Vinnl (commented):
Glad to see some attention being given to these, good work! And I'm looking forward to learning what `:global` does :grin: 

I added a couple of nitpicks, but the alt text and removal of `!important`s are the actual pressing issues.

The current behaviour is also a bit weird: when hovering a notification, everything _but_ the close button changes background colour, even though closing is what happens if you click it. I'd expect it the other way around: if you hover it, the close button gets highlighted.

https://user-images.githubusercontent.com/4251/165760798-88550ee3-9f75-473d-8a5e-07b3e9ccb78b.mp4

frontend/src/components/layout/Layout.module.scss
```diff
@@ -7,17 +7,146 @@
...
 
+// Allows us to dynamically set the background color based on Toast container state
+$toastify-toast-close-button: var(--toast-close-button);
+
 .toast {
-  border-radius: $border-radius-md;
   text-align: center;
-  font-weight: 700;
+  font-weight: 400;
+  font-size: 13px;
```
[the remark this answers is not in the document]
lloan: updated

frontend/src/components/layout/Layout.module.scss
```diff
@@ -7,17 +7,146 @@
...
+      &:disabled {
+        --toast-close-button: #{$color-yellow-70};
+      }
+    }
+  }
+}
+
+.close-toast-button-container {
+  padding: $spacing-md;
+  position: relative;
```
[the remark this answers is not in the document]
lloan: Correct - remnant of me testing centering component 🤦‍♂️ - thanks!

frontend/src/components/layout/Layout.module.scss
```diff
@@ -7,17 +7,146 @@
...
+  border-radius: 0 $border-radius-sm $border-radius-sm 0;
+  background-color: $toastify-toast-close-button;
+  display: flex;
+  flex-direction: column-reverse;
+}
+
+.close-toast-button-icon {
+  width: 20px !important;
+  height: 20px !important;
```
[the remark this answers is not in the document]
lloan: I set an ID for this custom close button to avoid fighting with Toastify styling and avoid using `!important` - hope that's an acceptable alternative. 

Also added the `declation-no-important` to stylelint.

lloan:
> Glad to see some attention being given to these, good work! And I'm looking forward to learning what `:global` does 😁
> 
> I added a couple of nitpicks (i.e. feel free to ignore at your prerogative), but the alt text and removal of `!important`s are the actual pressing issues IMHO.
> 
> The current behaviour is also a bit weird: when hovering a notification, everything _but_ the close button changes background colour, even though closing is what happens if you click it. I'd expect it the other way around: if you hover it, the close button gets highlighted.
> 
>  Kooha-04-28-2022-15-16-23.mp4

Yeah, this one would be a design decision - looking at the Figma, I followed the states illustrated. Do we want to reach out to design about this?

lloan:
> > Yeah, this one would be a design decision - looking at the Figma, I followed the states illustrated. Do we want to reach out to design about this?
> 
> Yeah I think that's a good idea, and you can add a link to [the preview deployment](https://deploy-preview-1854--fx-relay-demo.netlify.app/accounts/settings/?mockId=full) (saving the settings shows a toast) - I can imagine this not being intended, as it's always hard to predict how something will feels when it's actually interactive. Thanks!

Updated all states to match this now: 

image


L10n support added for the custom close button.

frontend/src/components/layout/Layout.tsx
```diff
@@ -1,4 +1,4 @@
-import { ReactNode } from "react";
+import React, { ReactElement, ReactNode } from "react";
```
Vinnl, line 1: No biggie, but this is an unused import now I think.

```suggestion
import { ReactElement, ReactNode } from "react";
```

Vinnl (approved):
Looking great @lloan! I'm still [curious to learn more about `:global`](https://github.com/mozilla/fx-private-relay/pull/1854#discussion_r860865813), but that needn't block merging this :)

frontend/src/components/layout/Layout.module.scss
Vinnl: (Reply to https://github.com/mozilla/fx-private-relay/pull/1854#issuecomment-1113340516)

Ah yes, that looks fine to me, and thanks for adding the comment to the stylelintrc. Since we're using toastify to do the toasts, I don't expect us to want to re-use the global `.toast` class anywhere, so I don't expect this to cause problems down the road. Nice find!

frontend/src/components/layout/Layout.module.scss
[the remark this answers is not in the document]
maxxcrawford: @Vinnl 100% agree about `!important` always needing an explanation. Thanks for adding, @lloan.

maxxcrawford:
> The current behaviour is also a bit weird: when hovering a notification, everything _but_ the close button changes background colour, even though closing is what happens if you click it. I'd expect it the other way around: if you hover it, the close button gets highlighted.

@lloan You'll know this better than us: How does VPN handle it? Is that how their toasts work? (With the hover states behaving as @Vinnl showed in the video) My larger org goal is to have them match. So whatever gets us all on the same page (and Eduardo and Peter are _also_ on the same page)!

lloan:
@Vinnl @maxxcrawford I applied the approved changes by Kaitlyn for MPP-1833 - Not sure if its just me on local, but when I sign out - the feature works as intended, but when I sign in, I just get a quick glance before I'm sent to the user dashboard. Is that not the case for you both?

frontend/src/components/layout/Layout.module.scss
```diff
@@ -7,17 +8,122 @@
...
+// Allows us to dynamically set the background color based on Toast container state
+$toastify-toast-close-button: var(--toast-close-button);
+
 .toast {
-  border-radius: $border-radius-md;
+  @include text-body-sm;
   text-align: center;
-  font-weight: 700;
+  font-weight: 400;
+  line-height: 21px;
```
maxxcrawford, line 34:
```suggestion
```

Suggestion: Looking at the rest of the styles (in `/fontend`) we do not set the line-height. If this can be avoided here, lets. If it does need to be set, use a [unit-less value](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#values) (1, 1.2, etc) so that it is relative to the font-size.

frontend/src/components/layout/Layout.module.scss
```diff
@@ -7,17 +8,122 @@
...
+  height: $layout-md;
```
maxxcrawford, line 35: Suggestion: I call this out in my general review comment, but if we can let the content (using padding, etc) dictate the toast height, let's. Esp for wordy messages here, on mobile, it'll need to break into two lines.

frontend/src/components/layout/Layout.tsx
```diff
@@ -86,6 +94,20 @@ export const Layout = (props: Props) => {
       <WhatsNewMenu profile={profiles.data[0]} />
     ) : null;
 
+  const closeToastButton = (closeToast: () => void): ReactElement => {
+    return (
+      <div className={styles["close-toast-button-container"]}>
+        <button className="Toastify__close-button Toastify__close-button--colored">
+          {CloseIcon({
+            alt: l10n.getString("toast-button-close-label"),
```
maxxcrawford, line 102: Praise: Nice alt tagging!

maxxcrawford (changes requested):
@lloan Excellent PR! Thanks for incorporating @codemist's PR as well. 

I found two small but blocking revisions: 

1. Can your make sure on smaller screens, the line-height/padding for messages is even? The the height of the toast can/should be defined by the contents rather than a static height, if possible. 

image

2. Can you add some padding to the toast on mobile, similiar to how it appears on desktop? Just a `$spacing-md` on top/right/left or something?

image

image

frontend/src/components/layout/Layout.module.scss
lloan: Removed

lloan:
> @lloan Excellent PR! Thanks for incorporating @codemist's PR as well.
> 
> I found two small but blocking revisions:
> 
> 1. Can your make sure on smaller screens, the line-height/padding for messages is even? The the height of the toast can/should be defined by the contents rather than a static height, if possible.
> 
> image
> 
> 2. Can you add some padding to the toast on mobile, similiar to how it appears on desktop? Just a `$spacing-md` on top/right/left or something?
> 
> image
> 
> image

I added padding and gave the close button a `min-width` as it was becoming smaller on smaller screens. 

image

frontend/src/components/layout/Layout.module.scss
```diff
@@ -31,8 +31,7 @@ $toastify-toast-close-button: var(--toast-close-button);
   @include text-body-sm;
   text-align: center;
   font-weight: 400;
-  line-height: 21px;
-  height: $layout-md;
+  min-height: $layout-md;
```
maxxcrawford, line 34: Nice solve.

frontend/src/components/layout/Layout.module.scss
```diff
@@ -47,6 +46,10 @@ $toastify-toast-close-button: var(--toast-close-button);
       }
     }
 
+    & > [class*="toast-body"] {
+      padding: 16px;
+    }
+
```
maxxcrawford, line 52: nit: use Protocol token `$spacing-md`

```suggestion
    & > [class*="toast-body"] {
      padding: $spacing-md;
    }

```

frontend/src/components/layout/Layout.module.scss
maxxcrawford, line 34: praise: Nice solve!

frontend/src/components/layout/Layout.module.scss
```diff
@@ -124,6 +127,7 @@ $toastify-toast-close-button: var(--toast-close-button);
 #close-toast-button-icon {
   width: 20px;
   height: 20px;
+  min-width: 20px;
```
maxxcrawford, line 130: question: Is this necessary? I imagine the static width attr does enough?

maxxcrawford (changes requested):
I'm still seeing the full-width toast on the smallest screen sizes: 

image

frontend/src/components/layout/Layout.module.scss
lloan: The icon keeps getting smaller the smaller your screen is. Regardless of width given.

frontend/src/components/layout/Layout.module.scss
lloan: done

lloan:
> I'm still seeing the full-width toast on the smallest screen sizes:
> 
> image

It's a bit rough to override the toast styling using css-modules. Even with the use of `:global` it can be hard. I took this approach and this is what it looks like: 


```
 @media screen and (max-width: $screen-sm) {
    width: calc(100% - $spacing-xl);
    margin-top: $spacing-md;
  }
```

image

maxxcrawford (approved):
Awesome work! Let's ship it. 🚀

PR #1855, opened 2022-04-27
MPP-1920: Detect SocialAccount without a token

### LINKED ISSUE #1850: A social token without an app leads to a 500 Server Error
If a `social_token` does not have an associated `app`, then `_get_oauth2_session` fails with ``AttributeError: /fxa-rp-events``:

https://github.com/mozilla/fx-private-relay/blob/f863c02a77d36ebaf232026bd55ad3ba2e3cd24c/privaterelay/views.py#L394-L395

This is seen on the dev server:

* Team: https://sentry.prod.mozaws.net/operations/fx-private-relay-dev/issues/19472768
* Public: https://sentry.prod.mozaws.net/share/issue/0288db5d2ad14a42965a57795f2b236d/

The data appears to be a team member who has a subscription update for `premium-relay`. This team member recently changed from a contractor to a full-time employee, so the auto-upgrade may be related. In addition, the social token may have been manually created.

jwhitlock:
Raise a new exception ``NoSocialToken`` when attempting to fetch a FxA user's profile with no associated token. ``SocialAccounts`` without a token are common on the development server (over half), but does not happen in production (confirmed by @moz-astults).

Fixes #1850 / MPP-1920.

There are no tests of the full `/fxa_rp_events` view, and it would take significant time to write them and fix existing issues. Since this is a dev-only issue, I held my nose and skipped the view tests (for now).

### DIFF
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -12,7 +12,7 @@ from django.urls import reverse
 from allauth.socialaccount.models import SocialToken
 from allauth.socialaccount.providers.fxa.views import FirefoxAccountsOAuth2Adapter
 
-from .views import _get_oauth2_session, update_social_token
+from .views import _get_oauth2_session, update_social_token, NoSocialToken
 
 metrics = markus.get_metrics("fx-private-relay")
 
@@ -39,7 +39,7 @@ class FxAToRequest:
                     FirefoxAccountsOAuth2Adapter.access_token_url
                 )
                 update_social_token(social_token, new_token)
-            except CustomOAuth2Error:
+            except (CustomOAuth2Error, NoSocialToken):
                 logout(request)
                 return redirect(reverse("home"))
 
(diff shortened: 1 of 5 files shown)


jwhitlock:
Force-push to rebase on main.

privaterelay/middleware.py
```diff
@@ -39,7 +39,7 @@ def __call__(self, request):
                     FirefoxAccountsOAuth2Adapter.access_token_url
                 )
                 update_social_token(social_token, new_token)
-            except CustomOAuth2Error:
+            except (CustomOAuth2Error, NoSocialToken):
```
say-yawn, line 42: Nice! Catching and handling for tokens that are invalid/dne during user interaction! TIL that middlewares are only applied when loading views (see this [doc](https://docs.djangoproject.com/en/3.2/topics/http/middleware/#middleware-order-and-layering)).

say-yawn (approved):
Looks good! Thanks for getting this fix in as it will help reduce the bad response we are sending to FxA.

PR #1859, opened 2022-04-27
Apply alias filter only when user clicks apply

### LINKED ISSUE #1811: Filter becomes active without using the "Apply" button
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox, latest release
Chrome, latest release

**Prerequisites:**
Have an account with active subscription.

**Steps to reproduce:**

2. Go to aliases list and click the filter button above it.
3. Select a few criteria and then click outside the filter menu.
4. Observe the aliases list.

**Actual result:**
The list is updated according to the selected criteria. The filter is active without clicking the "Apply" button.

**Expected result:**
The filter is active only when user clicks the "Apply" button.

**Notes:**
N/A

filter active without apply

flozia:
This PR changes the behaviour for filtering the list of aliases. Selection changes in `CategoryFilter` are now being applied only if the user explicitly clicks the ”apply“ button.

**Related issues**
- Fixes #1811

**How to test**


frontend/src/components/dashboard/aliases/CategoryFilter.tsx
```diff
@@ -32,6 +32,11 @@ export type Props = {
   onChange: (selected: SelectedFilters) => void;
 };
 
+type OnCloseParams = {
```
maxxcrawford, line 35: React newbie question: Why does this not have `export type` compared to the other two declarations above?

maxxcrawford (approved):
Works as intended! 

I did find one issue (may be regressive/may have already existed): 

_Testing note: I used `dev:mocked` to test this locally._ 

1. Open dashboard using `full` mockId (http://localhost:3000/accounts/profile/?mockId=full)
2. **Expected:** All three masks are collapsed
3. Select/apply any filter
4. Masks filtered. Still collapsed
5. Click "Reset" button under filter
6. **Expected:** Masks stay collapsed
7. **Actual:** Multiple masks are in expanded state

maxxcrawford:
@flozia Please let @Vinnl or @lloan also review this before merging.

frontend/src/components/dashboard/aliases/CategoryFilter.tsx
flozia: It’s not necessary to export type `OnCloseParams` since it’s only used within this component. If a `type` is exported it can be imported from another file using a named import. The type `SelectedFilters` for example is imported and used by the component `AliasList` as well: https://github.com/mozilla/fx-private-relay/blob/main/frontend/src/components/dashboard/aliases/AliasList.tsx#L9.

maxxcrawford: Thanks for the explanation! 🐣

flozia:
> Works as intended!
> 
> I did find one issue (may be regressive/may have already existed):
> 
> _Testing note: I used `dev:mocked` to test this locally._
> 
> 1. Open dashboard using `full` mockId (http://localhost:3000/accounts/profile/?mockId=full)
> 2. **Expected:** All three masks are collapsed
> 3. Select/apply any filter
> 4. Masks filtered. Still collapsed
> 5. Click "Reset" button under filter
> 6. **Expected:** Masks stay collapsed
> 7. **Actual:** Multiple masks are in expanded state

I can reproduce the issue and I suspect that it’s not introduced by this PR — I’m able to reproduce on `main` as well. In case it’s a regression: Is there an existing issue that should be reopened? Otherwise I’m happy to create a new one and tackle it in a follow-up.

Vinnl (approved):
Thanks for the PR @flozia, fix works as expected! Would you be up for writing an integration test for this in `profile.test.tsx`? I'm also happy to do it, or to pair on it, whatever you prefer :)

flozia:
Happy to add an integration test for the changes @Vinnl. I’ll give it a go and reach out in case there are any questions coming up.

flozia:
Added the following test cases to `profile.test.tsx` for the category filter:
1. Check if the category filter button is visible for users with premium
2. Applying filter selection to list of aliases
3. Discarding filter selection
4. Clearing filter selection

frontend/src/pages/accounts/profile.test.tsx
```diff
@@ -570,6 +571,136 @@ describe("The dashboard", () => {
...
+      name: "l10n string: [profile-label-apply], with vars: {}",
+    });
+    userEvent.click(categoryFilterApplyButton);
+
+    // The blocked label serves as a proxy for an alias card
+    const aliasCardBlockedLabels = screen.queryByText(
+      "l10n string: [profile-label-blocked], with vars: {}"
+    );
+
+    expect(aliasCardBlockedLabels).toBeInTheDocument();
```
Vinnl, line 621: While this does verify whether aliases that are set to blocking are still being displayed, it doesn't test whether the rest of the aliases are actually hidden.

Part of the challenge is that there's an error in the mocks, which I submitted a fix for in https://github.com/mozilla/fx-private-relay/pull/1879. With that fix, you should be able to do the following, not needing the "blocked" label as a proxy:

```suggestion
    const randomAlias1 = screen.queryByText(/address1/);
    const randomAlias2 = screen.queryByText(/address2/);
    expect(randomAlias1).not.toBeInTheDocument();
    expect(randomAlias2).not.toBeInTheDocument();
    const customAlias = screen.queryByText(/address3/);
    expect(customAlias).toBeInTheDocument();
```

Something similar could apply to the tests following this one.

Vinnl (approved):
The tests look great, and have us cross the 100 tests mark :tada: I added a small suggestion to also test the negative case in the test, but part of the challenge there (that I'm guessing you already ran into) is that there was a bug in the mocks - sorry about that. I submitted a fix for that in https://github.com/mozilla/fx-private-relay/pull/1879 (it has this branch as its target, so feel free to merge), and with that, it should be easier to match on aliases that are visible or not.

frontend/src/pages/accounts/profile.test.tsx
flozia: Thanks for the suggestion and the fix! I didn’t really realise that the mask addresses weren’t rendered. I applied your suggestion to the other tests as well.

Vinnl: Awesome, looking sharp! Feel free to merge :)

PR #1867, opened 2022-04-28
MPP-1917: Use Django settings for management commands `process_emails_from_sqs` and `check_health`

### LINKED ISSUE #1848: MPP-1917: Add defaults via settings for ``manage.py process_emails_from_sqs``
Configure the defaults for ``./manage.py process_emails_from_sqs`` via Django settings with environment overrides. This allows the task application settings to be defined in the environment, in the same way as the web application.

jwhitlock:
Use settings with environment overrides to set the defaults for `./manage.py process_emails_from_sqs` and `./manage.py check_heath`. This means that the environment can be used to configure the commands, instead of the command line parameters.

The new environment variables for `process_emails_from_sqs`:

* ``PROCESS_EMAIL_BATCH_SIZE``: How many SQS messages to process at a time, 1 through 10, default 10
* ``PROCESS_EMAIL_DELETE_FAILED_MESSAGES``: If SQS messages should be deleted on exceptions, true or false, default false
* ``PROCESS_EMAIL_MAX_SECONDS``: How long to run the process, integer, 0 to run forever, default forever
* ``PROCESS_EMAIL_VERBOSITY``: How much to log, 0 through 3, default 1
* ``PROCESS_EMAIL_VISIBILITY_SECONDS``: How long to reserve SQS messages, integer, default 120
* ``PROCESS_EMAIL_WAIT_SECONDS``: How long to wait for new SQS messages, Integer, default 5

This is used by both `process_emails_from_sqs` and `check_health`:

* ``PROCESS_EMAIL_HEALTHCHECK_PATH``: path to JSON file, no default

This is used by `check_health`:
* ``PROCESS_EMAIL_HEALTHCHECK_MAX_AGE``: Integer seconds, default 120.

Also fix some test and runtime issues:

* Override the environment variables so that the developer's settings do not affect the tests
* Check that `AWS_SQS_EMAIL_QUEUE_URL` is set and not the default of `None`
* Close the healthcheck file after reading in `./manage.py check_heath`
* Log message processing at ``INFO`` level. I think we'll want to see what is happening at this level for a while.

This PR fixes #1848 / MPP-1917

## Testing

I think code reading should be enough, but if you want to do local testing, first update reqs:

```
pip install -r requirements.txt
```

### Environment overrides

Run the help commands:

```
./manage.py process_emails_from_sqs --help
./manage.py check_health --help
```

This will display the current settings and the environment variable name, such as:

```
  --wait-seconds WAIT_SECONDS
                        Time to wait for messages with long polling. Defaults to 5 from settings.PROCESS_EMAIL_WAIT_SECONDS.
  --visibility-seconds VISIBILITY_SECONDS
                        Time to mark a message as reserved for this process. Defaults to 120 from settings.PROCESS_EMAIL_VISIBILITY_SECONDS.
```

You can then set this variable in your `.env` file to a different value, re-run the ``--help`` command, and see the text update.

If you are setup with a local SQS queue, you can set ``AWS_SQS_EMAIL_QUEUE_URL`` and run the command. The first log entry contains the values of the command parameters.

### Null SQS URL and invalid values tests

You can override the environment variables and see failures:

```
AWS_SQS_EMAIL_QUEUE_URL= ./manage.py process_emails_from_sqs
PROCESS_EMAIL_BATCH_SIZE=11 ./manage.py process_emails_from_sqs
PROCESS_EMAIL_DELETE_FAILED_MESSAGES=maybe ./manage.py process_emails_from_sqs
```

### Healthcheck closing

Run tests in `main` with `PYTHONWARNINGS=always pytest -v --capture=no emails/tests/mgmt_check_health_tests.py`. The warnings section will contain warnings like:

```
emails/tests/mgmt_check_health_tests.py::test_check_health_passed_no_logs
  /Users/john/src/fx-private-relay/emails/tests/mgmt_check_health_tests.py:34: ResourceWarning: unclosed file <_io.TextIOWrapper name='/private/var/folders/_y/72__kkbs2tqd5ch4gwrgnwtr0000gn/T/pytest-of-john/pytest-112/test_check_health_passed_no_lo0/healthcheck.json' mode='r' encoding='utf8'>
    call_command("check_health", str(path))
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
```

If you run `PYTHONWARNINGS=always pytest -v --capture=no emails/tests/mgmt_check_health_tests.py` in this branch, the warnings go away.


jwhitlock:
Rebased and added another commit to log each message processing at ``INFO``. This mirrors the current logs for SNS ``POST`` delivery, and I think we'll want to see how long successes are as well for a while.

emails/management/commands/check_health.py
```diff
@@ -24,20 +25,29 @@
 class Command(BaseCommand):
     help = "Check that a healthcheck JSON file exists and is recent."
 
-    DEFAULT_MAX_AGE_SECONDS = 120
+    def help_message(self, text, setting_name):
```
groovecoder, line 28: suggestion (non-blocking): can this be `help_message_for_setting` to make it more obvious what the 2nd argument is when calling this? E.g., `self.help_message_for_setting("Path to healthcheck JSON file.", "PROCESS_EMAIL_HEALTHCHECK_PATH")`

emails/management/commands/process_emails_from_sqs.py
```diff
@@ -36,23 +36,19 @@
...
-
     def __init__(
         self,
         *args,
         batch_size=None,
         wait_seconds=None,
         visibility_seconds=None,
         healthcheck_file=None,
-        delete_failed_messages=False,
+        delete_failed_messages=None,
```
groovecoder, line 46: suggestion (blocking): This value seems to be all of `None`-type (here), a boolean (in `test_settings`), or an integer (in the `@override_settings` of `test_command_setup_from_settings`? Are these all the same variable controlling the same logic defined by the same setting? If so, can we consolidate its type & arguments?

emails/management/commands/process_emails_from_sqs.py
```diff
@@ -117,50 +128,72 @@ def init_vars(
         )
         assert 0 <= self.verbosity <= 3
 
+    def help_message(self, text, setting_name):
```
groovecoder, line 131: suggestion (non-blocking): can this be `help_message_for_setting` to make it more obvious what the 2nd argument is when calling this? E.g., `self.help_message_for_setting("Path to healthcheck JSON file.", "PROCESS_EMAIL_BATCH_SIZE")`

privaterelay/settings.py
```diff
@@ -17,7 +17,7 @@
 # This needs to be before markus, which imports pytest
 IN_PYTEST = "pytest" in sys.modules
 
-from decouple import config, Csv
+from decouple import config, Choices, Csv
```
groovecoder, line 20: praise: TIL `decouple.Choices` !

groovecoder (changes requested):

emails/management/commands/process_emails_from_sqs.py
jwhitlock: In this case, `None` means "use the value from settings". If we set these to an integer or boolean, we'd be unable to detect that the parameter was not passed.

I've seen other code do something like this:

```python
class Unset:
    """The argument was unset""""
    pass

def __init__(
    self,
    *args,
    batch_size=Unset(),
    wait_seconds=Unset(),
    ...
):
```

Would you prefer something like that?

jwhitlock: Re-reading this, I think you were specifically calling out `delete_failed_messages`. I still think `None` is the right value here, and I've changed `test_command_setup_from_settings` from `1` to `True`. We can get more precise once mypy is available (PR #1831)

jwhitlock:
Force-push to rebase on main, and add two commits to address review feedback:

* Rename `help_message` to `help_message_for_setting` in both commands
* Change `PROCESS_EMAIL_DELETE_FAILED_MESSAGES` override from `1` to `True` in tests

jwhitlock:
Back to you @groovecoder. If you're thinking something bigger than the one-line change for ``--delete-failed-messages``, let's discuss.

emails/management/commands/process_emails_from_sqs.py
groovecoder: Ah, okay that makes more sense to have either `None` or `bool`. Thanks.

privaterelay/settings.py
```diff
@@ -708,3 +708,23 @@ def _get_initial_middleware():
...
     SILKY_PYTHON_PROFILER_BINARY=True
+
+# Settings for manage.py process_emails_from_sqs
+PROCESS_EMAIL_BATCH_SIZE = config(
+    "PROCESS_EMAIL_BATCH_SIZE", 10, cast=Choices(range(1, 11), cast=int)
+)
+PROCESS_EMAIL_DELETE_FAILED_MESSAGES = config(
+    "PROCESS_EMAIL_DELETE_FAILED_MESSAGES", False, cast=bool
+)
+PROCESS_EMAIL_HEALTHCHECK_PATH = config("PROCESS_EMAIL_HEALTHCHECK_PATH", "") or None
```
groovecoder, line 719: question & suggestion (blocking): `process_emails_from_sqs` treats the healthcheck file as optional. Should it?

Since `check_health` requires a `healthcheck_path`, let's either add a default value for it, or add a local dev value for it to `.env-dist`.

When I tried to run `python manage.py check_health` I got the error:
```
manage.py check_health: error: the following arguments are required: healthcheck_path
```

I added a value to my `PROCESS_EMAIL_HEALTHCHECK_PATH` env var ...
```bash
> echo $PROCESS_EMAIL_HEALTHCHECK_PATH
/tmp/privaterelay_process_emails_from_sqs_healthcheck
```

But I still got:
```
manage.py check_health: error: the following arguments are required: healthcheck_path
``````

Furthermore, even when I pass a healthcheck_path argument to `check_health` ...
```
python manage.py check_health /tmp/privaterelay_process_emails_from_sqs_healthcheck
```

I got: `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

So I dug into `process_emails_from_sqs.py` and saw that it only has a single call to `write_healthcheck` in the `except ClientError as e` block of code? Did we remove the "healthy" write call?

**Review by groovecoder — CHANGES_REQUESTED** [review]

File: privaterelay/settings.py (hunk 5 above)
Comment (jwhitlock) [reply privaterelay/settings.py:719]: Thanks for the in-depth review!

This is the [second time](https://github.com/mozilla/fx-private-relay/pull/1768#discussion_r847855939) you expected a default for the healthcheck file, so I'll add it.

I'm wondering if we should have any command line parameters for these functions, since you and @moz-astults prefer environment variables. That's a bigger change, but if we tackle it now, then we'll discover bugs while our attention is on the new code.

Good catch on the healthcheck call! I accidentally removed it in https://github.com/mozilla/fx-private-relay/commit/f8f6b737fa4e2b9ec12009f04ade83eb28df976d. I'll add a failing test and fix it.

I also think it makes sense to touch the file after processing each email. A successful call to SES in the dev deployment takes 100 ms. The auto-retry functionality could stretch this into seconds, but shouldn't cause the task to be killed.

> I got: `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

It is wrong that there is no file, but I intentionally did _not_ add a lot of error handling to `check_health`. This error from trying to read an empty file as JSON will exit with a non-zero code, and be (correctly) interpreted as a health failure.

**Comment by jwhitlock** [comment]
Re-based, and then a significant re-write:

* The healthcheck path is required, and defaults to ``tmp/healthcheck.json``
* The command-line parameters and manual overrides are removed. The only way to change configuration is through settings. A new class `CommandFromDjangoSettings` has the details, and is used by both the new commands.
* With the manual overrides gone, the tests now use `call_command` exclusively rather than call `process_queue` or other functions directly. This also requires looking at the logs for the summary information. The tests are similar, but almost completely re-written
* The healthcheck file is opened when writing to it. This avoids a bug where I was appending the healthcheck object rather than replacing it.

Tests pass, and I've manually tested locally. Back to you, @groovecoder!

File: emails/tests/mgmt_check_health_tests.py [hunk 6]
```diff
@@ -10,7 +10,15 @@
 from emails.management.commands.check_health import Command
 
 
-def write_healthcheck(folder, age=0):
+@pytest.fixture(autouse=True)
+def test_settings(settings, tmp_path):
+    """Override settings for tests"""
+    settings.PROCESS_EMAIL_HEALTHCHECK_PATH = tmp_path / "healthcheck.json"
```
groovecoder, line 16: praise: TIL `pathlib` lets you use `/` to concat!

emails/tests/mgmt_process_emails_from_sqs_tests.py
```diff
@@ -125,26 +138,91 @@ def make_client_error(message="Unknown", code="Unknown", operation_name="Unknown
...
+
+
+def summary_from_exit_log(caplog_fixture):
+    """Get the extra data from the final log message"""
+    last_log = caplog_fixture.records[-1]
+    assert last_log.message == f"Exiting {COMMAND_NAME}"
+    return log_extra(last_log)
+
+
+def test_no_messages(caplog, test_settings):
```
groovecoder, line 182: question (non-blocking): This test looks like it uses `log_extra` while the other tests use `summary_from_exit_log`. Can this test be written more concisely by using `summary_from_exit_log` too?

groovecoder (approved):
Spot-checks work and review comments are all non-blocking.

emails/management/command_from_django_settings.py
```diff
@@ -0,0 +1,79 @@
...
+                )
+            if local_name == "verbosity":
+                # The setting overrides the default, but use the command-line value
+                self.verbosity = verbosity
+            else:
+                setattr(self, local_name, getattr(settings, setting_key))
+
+    def handle(self, *args, **kwargs):
+        verbosity = kwargs.get("verbosity")
+        self.init_from_settings(verbosity)
```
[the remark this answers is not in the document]
jwhitlock: I was using `handle` so I could use `super().handle(*args, **kwargs)`, but that seems like a pretty weak reason actually. I think I'll move this code into the `handle` functions and get rid of this.

emails/tests/mgmt_process_emails_from_sqs_tests.py
jwhitlock: 👍 good idea! Then `log_extra` can go away entirely...

BTW, I'm aiming for the day when I can replace all this with `structlog`:

https://www.structlog.org/en/stable/testing.html

jwhitlock: Oh, I remember why. This is the only test that examines logs that aren't the final "Exiting process_emails_from_sqs_tests" message.

``log_extra`` takes a ``LogRecord``, a standard Python logging object, and strips out the items that weren't added by ``extra={}``

``summary_from_exit_log`` takes the ``caplog`` item, gets the last ``.record``, asserts that it is an ``Exiting... `` log entry, and calls ``log_extra`` on it.

So, ``test_no_messages`` needs access to ``log_extra``, because it is the one place that is looking at other log entries.

For completeness, I think I'll add a test to ``test_one_message`` that looks at the message processing log entry.

jwhitlock:
Rebase on current master, then add 3 commits:

* ``check_health`` was closing the JSON file in ``check_healthcheck``. After switching from ``argparse.FileType`` to a string in Django settings, it is better to close it in the ``handle`` method.
* Remove ``CommandFromDjangoSettings.handle``, and instead call ``self.init_from_settings`` in the derived class ``handle`` method. It is still one line of code in the derived class, 4 less lines in the base class.
* Add a test of the ``Message processed`` log entry, since it may change in the future, and to justify the existence of ``log_extra``

PR #1868, opened 2022-04-29
Use environment-specific Basket URLs

Vinnl:
[The request](https://mozilla.slack.com/archives/CUU5EC46L/p1651163577447129?thread_ts=1651096179.335249&cid=CUU5EC46L) was for our stage environment to point to the dev environment of Basket, whereas before we only did so when running locally. Therefore, this PR makes it so that we fetch the relevant Basket URL from the back-end, where it can be set with environment variables.

I also fixed a bug in which the country detected for a user would not be used, because we'd already set the value to the country derived from the user's Accept language. (Only uncovered on stage because that's where we have proper location-detection.)

How to test:

### DIFF
--- a/frontend/__mocks__/hooks/api/runtimeData.ts
+++ b/frontend/__mocks__/hooks/api/runtimeData.ts
@@ -13,8 +13,8 @@ const mockedUseRuntimeData = useRuntimeData as jest.MockedFunction<
 
 export function getMockRuntimeDataWithPremium(): RuntimeData {
   return {
-    FXA_ORIGIN: "https://example.com",
-    BASKET_ORIGIN: "https://example.com",
+    FXA_ORIGIN: "https://fxa-mock.com",
+    BASKET_ORIGIN: "https://basket-mock.com",
     GOOGLE_ANALYTICS_ID: "UA-123456789-0",
     PREMIUM_PRODUCT_ID: "prod_123456789",
     PREMIUM_PLANS: {
@@ -34,8 +34,8 @@ export function getMockRuntimeDataWithPremium(): RuntimeData {
 }
 export function getMockRuntimeDataWithoutPremium(): RuntimeData {
   return {
-    FXA_ORIGIN: "https://example.com",
-    BASKET_ORIGIN: "https://example.com",
+    FXA_ORIGIN: "https://fxa-mock.com",
+    BASKET_ORIGIN: "https://basket-mock.com",
     GOOGLE_ANALYTICS_ID: "UA-123456789-0",
     PREMIUM_PRODUCT_ID: "prod_123456789",
     PREMIUM_PLANS: {
--- a/frontend/__mocks__/hooks/api/runtimeData.ts
+++ b/frontend/__mocks__/hooks/api/runtimeData.ts
@@ -14,6 +14,7 @@ const mockedUseRuntimeData = useRuntimeData as jest.MockedFunction<
 export function getMockRuntimeDataWithPremium(): RuntimeData {
   return {
     FXA_ORIGIN: "https://example.com",
+    BASKET_ORIGIN: "https://basket.mozilla.org",
     GOOGLE_ANALYTICS_ID: "UA-123456789-0",
     PREMIUM_PRODUCT_ID: "prod_123456789",
     PREMIUM_PLANS: {
@@ -34,6 +35,7 @@ export function getMockRuntimeDataWithPremium(): RuntimeData {
 export function getMockRuntimeDataWithoutPremium(): RuntimeData {
   return {
     FXA_ORIGIN: "https://example.com",
+    BASKET_ORIGIN: "https://basket.mozilla.org",
     GOOGLE_ANALYTICS_ID: "UA-123456789-0",
     PREMIUM_PRODUCT_ID: "prod_123456789",
     PREMIUM_PLANS: {
--- a/frontend/__mocks__/hooks/api/runtimeData.ts
+++ b/frontend/__mocks__/hooks/api/runtimeData.ts
@@ -14,7 +14,7 @@ const mockedUseRuntimeData = useRuntimeData as jest.MockedFunction<
 export function getMockRuntimeDataWithPremium(): RuntimeData {
   return {
     FXA_ORIGIN: "https://example.com",
-    BASKET_ORIGIN: "https://basket.mozilla.org",
+    BASKET_ORIGIN: "https://example.com",
     GOOGLE_ANALYTICS_ID: "UA-123456789-0",
     PREMIUM_PRODUCT_ID: "prod_123456789",
     PREMIUM_PLANS: {
@@ -35,7 +35,7 @@ export function getMockRuntimeDataWithPremium(): RuntimeData {
 export function getMockRuntimeDataWithoutPremium(): RuntimeData {
   return {
     FXA_ORIGIN: "https://example.com",
-    BASKET_ORIGIN: "https://basket.mozilla.org",
+    BASKET_ORIGIN: "https://example.com",
     GOOGLE_ANALYTICS_ID: "UA-123456789-0",
     PREMIUM_PRODUCT_ID: "prod_123456789",
     PREMIUM_PLANS: {
--- a/frontend/src/apiMocks/mockData.ts
+++ b/frontend/src/apiMocks/mockData.ts
@@ -7,8 +7,8 @@ export const mockIds = ["empty", "onboarding", "some", "full"] as const;
 
 // This is the same for all mock users, at this time:
 export const runtimeData: RuntimeData = {
-  FXA_ORIGIN: "https://example.com",
-  BASKET_ORIGIN: "https://example.com",
+  FXA_ORIGIN: "https://fxa-mock.com",
+  BASKET_ORIGIN: "https://basket-mock.com",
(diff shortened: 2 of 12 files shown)


frontend/src/pages/premium/waitlist.page.tsx
```diff
@@ -55,8 +55,13 @@ const PremiumWaitlist: NextPage = () => {
...
     try {
       const response = await subscribe({
+        // The back-end should provide the URL to Basket, pointing to its dev environment
+        // if running on stage. If the back-end's response doesn't come in in time,
+        // we just assume production:
+        basketOrigin:
+          runtimeData.data?.BASKET_ORIGIN ?? "https://basket.mozilla.org",
         email: email,
-        countryCode: country,
+        countryCode: country ?? currentCountry ?? "US",
```
maxxcrawford, line 64: Question: Does this read that if there's no `currentCountry` set, fall back to the `US`? 

If so, does this weight any specific country metrics' analytics?

frontend/src/pages/premium/waitlist.test.tsx
```diff
@@ -1,16 +1,44 @@
...
 
 import PremiumWaitlist from "./waitlist.page";
 
 jest.mock("@fluent/react", () => mockFluentReact);
 jest.mock("next/router", () => mockNextRouter);
 jest.mock("../../config.ts", () => mockConfigModule);
 jest.mock("../../hooks/gaViewPing.ts");
+jest.mock("../../components/waitlist/countryPicker.tsx", () => ({
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  CountryPicker: (props: any) => <select {...props} />,
```
maxxcrawford, line 20: nit/question: Why the use of `any` here?

privaterelay/settings.py
```diff
@@ -569,6 +569,7 @@ def _get_initial_middleware():
 FXA_SETTINGS_URL = config('FXA_SETTINGS_URL', f'{FXA_BASE_ORIGIN}/settings')
 FXA_SUBSCRIPTIONS_URL = config('FXA_SUBSCRIPTIONS_URL', f'{FXA_BASE_ORIGIN}/subscriptions')
 FXA_SUPPORT_URL = config('FXA_SUPPORT_URL', f'{FXA_BASE_ORIGIN}/support/')
+BASKET_ORIGIN = config('BASKET_ORIGIN', 'https://basket.mozilla.org')
```
maxxcrawford, line 572: Question: Should this be the dev URL value (`https://basket-dev.allizom.org`) or do we want to fallback to production here?

frontend/__mocks__/hooks/api/runtimeData.ts
```diff
@@ -14,6 +14,7 @@ const mockedUseRuntimeData = useRuntimeData as jest.MockedFunction<
 export function getMockRuntimeDataWithPremium(): RuntimeData {
   return {
     FXA_ORIGIN: "https://example.com",
+    BASKET_ORIGIN: "https://basket.mozilla.org",
```
maxxcrawford, line 17: Question: Why are we using this URL instead of the development URL? `https://basket-dev.allizom.org`

maxxcrawford (commented):
General praise: I'm bookmarking this PR for a "How to" on adding/consuming an ENV from the Django app on the `/frontend`. 

I want @groovecoder to also review, so I'm leaving this as a abstain'd review (and the fact that all of my comments are questions).

frontend/src/pages/premium/waitlist.test.tsx
Vinnl: I love that you're asking that question! And in fact, when setting up the mocks I figured I'd be ignoring most of the usually-required props, but it's actually fairly easy to use the right props here, so I did: 0ae82355

privaterelay/settings.py
Vinnl: I figured I'd rather risk accidentally sending test data to production, than sending production data to the dev environment, hence this choice.

frontend/__mocks__/hooks/api/runtimeData.ts
Vinnl: That is a good question; we shouldn't. In fact, it's probably best to not use a realistic URL in the first place, since this is mock data, and we want to make sure we're hitting the values from the `runtime_data` API at the right moment, rather than a fallback value. Fixed in 2a6868af.

Vinnl:
> I'm bookmarking this PR for a "How to" on adding/consuming an ENV from the Django app on the /frontend.

While seeing actual code is indeed most useful, it can be accompanied by [the documentation we have on that](https://github.com/mozilla/fx-private-relay/blob/ecab15eb7f833130a4427b86078c9db58bcae1af/docs/frontend-architecture.md#addmodify-environment-specific-data) :)

frontend/__mocks__/hooks/api/runtimeData.ts
```diff
@@ -14,6 +14,7 @@ const mockedUseRuntimeData = useRuntimeData as jest.MockedFunction<
 export function getMockRuntimeDataWithPremium(): RuntimeData {
   return {
     FXA_ORIGIN: "https://example.com",
+    BASKET_ORIGIN: "https://example.com",
```
groovecoder, line 17: suggestion (blocking): make this a separate value from the mock `FXA_ORIGIN` so if we're trying to debug it's more obvious what is what?
```suggestion
    FXA_ORIGIN: "https://fxa-mock.com",
    BASKET_ORIGIN: "https://basket-mock.com",
```

frontend/__mocks__/hooks/api/runtimeData.ts
```diff
@@ -34,6 +35,7 @@ export function getMockRuntimeDataWithPremium(): RuntimeData {
 export function getMockRuntimeDataWithoutPremium(): RuntimeData {
   return {
     FXA_ORIGIN: "https://example.com",
+    BASKET_ORIGIN: "https://example.com",
```
groovecoder, line 38: suggestion (blocking): make this a separate value from the mock `FXA_ORIGIN` so if we're trying to debug it's more obvious what is what?
```suggestion
    FXA_ORIGIN: "https://fxa-mock.com",
    BASKET_ORIGIN: "https://basket-mock.com",
```

frontend/src/pages/premium/waitlist.test.tsx
```diff
@@ -1,16 +1,54 @@
...
 
 import PremiumWaitlist from "./waitlist.page";
 
 jest.mock("@fluent/react", () => mockFluentReact);
 jest.mock("next/router", () => mockNextRouter);
 jest.mock("../../config.ts", () => mockConfigModule);
 jest.mock("../../hooks/gaViewPing.ts");
+jest.mock("../../components/waitlist/countryPicker.tsx", () => ({
+  // We're mocking out the country picker because it dynamically imports the
+  // list of available countries, which would make the test pretty cumbersome
```
groovecoder, line 22: question (non-blocking): what does "pretty cumbersome" mean in this context? Lots of extra testing code? Slower testing runs? Both?

groovecoder (changes requested):

frontend/src/pages/premium/waitlist.test.tsx
```diff
@@ -25,4 +63,52 @@ describe("The waitlist", () => {
...
+
+    const submitButton = screen.getByRole("button", {
+      name: "l10n string: [waitlist-submit-label], with vars: {}",
+    });
+    userEvent.click(submitButton);
+
+    expect(global.fetch).toHaveBeenCalledTimes(1);
+    expect(global.fetch).toHaveBeenCalledWith(
+      expect.any(String),
+      expect.objectContaining({ body: expect.stringContaining("") })
```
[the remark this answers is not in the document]
Vinnl: Heh, so what happened here is that I just wanted to match on a string in general, because the body contains form data that needs to be parsed before being able to make assertions on it (I'm doing that in the lines below this one). I couldn't find a matcher for strings, so I went for this. Then I later thought of the right way to do that, then got distracted and forgot to implement it. I now did that in e591b670.

frontend/src/pages/premium/waitlist.test.tsx
Vinnl: Tried to clarify that in 12a89486.

frontend/src/apiMocks/mockData.ts
```diff
@@ -8,6 +8,7 @@ export const mockIds = ["empty", "onboarding", "some", "full"] as const;
 // This is the same for all mock users, at this time:
 export const runtimeData: RuntimeData = {
   FXA_ORIGIN: "https://example.com",
+  BASKET_ORIGIN: "https://example.com",
```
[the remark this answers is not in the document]
Vinnl: 3d127b5c

frontend/__mocks__/hooks/api/runtimeData.ts
Vinnl: 3d127b5c

frontend/__mocks__/hooks/api/runtimeData.ts
Vinnl: Ah good one: 3d127b5c

groovecoder (approved):

PR #1880, opened 2022-05-02
Re-implement FxA metrics flow calls

### LINKED ISSUE #1807: Send `utm_source` to FxA's `/metrics-flow`
This isn't a high priority but I noticed Relay isn't sending a `utm_source` parameter to FxA.  In [our docs](https://mozilla.github.io/ecosystem-platform/relying-parties/reference/metrics-for-relying-parties#metrics-related-query-parameters) the `utm_source` identifies an "Unambiguous identifier of site or browser UI element that linked to the page containing the beginning of the FxA sign-in/sign-up process".  I think this could be `relay.firefox.com` and would be added to [this function](https://github.com/mozilla/fx-private-relay/blob/main/static/js/analytics.js#L49).

Thanks!

Vinnl:
This essentially replicates the following PR in React:
https://github.com/mozilla/fx-private-relay/pull/348

In essence, the new `useFxaFlowTracker` hook does the same our `useGaViewPing` hook does, but in addition to sending an event to Google Analytics when the ref'ed element comes into view, it also sends a request to FxA's `/metrics-flow` to add an ID to the sign-in link.

This flow is not yet implemented for the "Sign up" button on the
homepage (not in the navigation), nor in the "Upgrade now" button
on the Premium promotional page (at /premium), but the same is true
for the old UI.

This PR also fixes #1807.

How to test: I don't think we can actually test this until it's on stage (possibly dev), since http://127.0.0.1:8000/ is not listed as an allowed origin in FxA's response headers. But you can at least verify that the requests are being made for the "sign up" and "sign in" button in the navigation, and that those buttons still work even if the parameters to add to them could not be retrieved.

- [ ] I've added a unit test to test for potential regressions of this bug. - Adding a test to mock out FxA's ID and verify that it's added to the link feels like not worth the effort, but happy to reconsider.

### DIFF
--- a/frontend/src/components/layout/Navigation.tsx
+++ b/frontend/src/components/layout/Navigation.tsx
@@ -5,7 +5,7 @@ import { event as gaEvent } from "react-ga";
 import styles from "./Navigation.module.scss";
 import { useIsLoggedIn } from "../../hooks/session";
 import { getRuntimeConfig } from "../../config";
-import { useGaViewPing } from "../../hooks/gaViewPing";
+import { useFxaFlowTracker } from "../../hooks/fxaFlowTracker";
 
 /** Switch between the different pages of the Relay website. */
 export const Navigation = () => {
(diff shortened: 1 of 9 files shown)


frontend/src/components/layout/Navigation.tsx
```diff
@@ -36,14 +57,35 @@ export const Navigation = () => {
...
+  if (signInFxaFlowTracker.flowData) {
+    signInUrl.searchParams.append(
+      "flowId",
+      signInFxaFlowTracker.flowData.flowId
+    );
+    signInUrl.searchParams.append(
+      "flowBeginTime",
+      signInFxaFlowTracker.flowData.flowBeginTime
+    );
+  }
```
groovecoder, line 84: suggestion (blocking): this code is all duplicated from the `relay-sign-up-heaader` code above? Consolidate into a function that can be called from each spot.

frontend/src/hooks/fxaFlowTracker.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  useEffect(() => {
+    if (args === null || !inView) {
+      return;
+    }
+    gaEvent({
+      ...args,
+      action: "View",
+      nonInteraction: true,
+    });
```
groovecoder, line 39: praise: thank you for preserving the event record into our own GA so we can continue to see these in our own GA reports too.

groovecoder (changes requested):

frontend/src/components/layout/Navigation.tsx
Vinnl: bf76ad70

I also added an API mock so we can emulate the behaviour, which actually helped me find a bug (I'd used `flowBeginType` rather than `flowBeginTime` - I've probably been corrupted by TS), so we can now actually see the parameters being added [here](https://deploy-preview-1880--fx-relay-demo.netlify.app/).

groovecoder (approved):
cool, thanks for the refactor. glad it helped catch a bug!

Vinnl:
Rebased to resolve conflicts; merging when checks succeed :)

PR #1889, opened 2022-05-04
Allow sending data to Basket in our Content Security Policy

Vinnl:
(And I added a mock endpoint in the mock environment.)

How to test: run the waitlist from the server (i.e. `npm run watch` rather than `npm run dev`) - subscribing via http://127.0.0.1:8000/premium/waitlist/ should be successful now.

- [x] I've added a unit test to test for potential regressions of this bug. N/A

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -90,6 +90,8 @@ CSP_CONNECT_SRC = (
     'https://www.google-analytics.com/',
     'https://accounts.firefox.com',
     'https://location.services.mozilla.com',
+    'https://basket-dev.allizom.org',
+    'https://basket.mozilla.org',
 )
 CSP_DEFAULT_SRC = ("'self'",)
 CSP_SCRIPT_SRC = (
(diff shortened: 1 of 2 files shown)


privaterelay/settings.py
```diff
@@ -90,6 +90,8 @@
     'https://www.google-analytics.com/',
     'https://accounts.firefox.com',
     'https://location.services.mozilla.com',
+    'https://basket-dev.allizom.org',
+    'https://basket.mozilla.org',
```
groovecoder, line 94: issue (blocking): Only add `basket-dev.allizom.org` to dev environments. Something like:
```python
BASKET_SRC = 'https://basket-dev.allizom.org' if DEBUG or ON_HEROKU else 'https://basket.mozilla.org'
```
?

That means our stage environment would use production `basket.mozilla.org` though. How will basket handle that? Will it add a bunch of stage/QA users to the production lists? (Probably fine.)

groovecoder (changes requested):

privaterelay/settings.py
groovecoder: Oh, duh - no that's perfect. Add `BASKET_ORIGIN` is exactly what we want to enable per-environment CSP connections.

Vinnl: Applied in 8c834a9d.

PR #1891, opened 2022-05-04
Remove non-React UI code

### LINKED ISSUE #1858: A user with an expired token gets a Server Error instead of redirect to `/`
The `FxAToRequest` middleware checks that the user's FxA token is still valid. If it is expired, it attempts to refresh it. If this fails, they are logged out and returned to the home page. However, after the React update, the URL "home" is no longer know, so they instead get a server error.

This is seen as `NoReverseMatch: Reverse for 'home' not found. 'home' is not a valid view function or pattern name.` when calling `/api/v1/profiles/{pk}/` in Sentry:

* Team link: https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/19469747
* Public link: https://sentry.prod.mozaws.net/share/issue/45467b22210747d6b8199ded2f33c89c/

The code:

https://github.com/mozilla/fx-private-relay/blob/b3d715119be4c5968a60aa71240871be5d476cf4/privaterelay/middleware.py#L37-L48

Vinnl:
This PR removes all code that no longer runs, because it was used to generate the non-React UI:

- Unused Django partials and views, i.e. everything but `wrapped_email.html`. That said, there are still empty templates for home and faq, since there is Python code referring to those routes to generate their URLs. My intuition would be that just referring to those URLs directly would be a better solution, but I'll defer to Django authorities on that.
- Gulp, Sass, ESLint (at the root, not in `/frontend/`) and stylelint (at the root, not in `/frontend`) are removed.
- References to `SERVE_REACT` are removed, and React is now served by default.
- The `package.json` at the root now refers to the one in `/frontend` as a [workspace](https://docs.npmjs.com/cli/v7/using-npm/workspaces). The file is still used to tell Heroku how to build, and this ensures that you can also just run `npm install` at the root. This does also result in a big update to the `package-lock.json` at the root, which is the sole reason that this PR has more additions than deletions.
- ~~Images are now co-located with the code that's using them, as requested in reviews of The Big One.~~ → extracted to https://github.com/mozilla/fx-private-relay/pull/1998.

There's also [an equivalent PR for the add-on](https://github.com/mozilla/fx-private-relay-add-on/pull/320), though it's independent from this one and needn't be merged at the same time.

How to test: basically, nothing should break anywhere!

This also fixes #1858 according to John.

- [x] I've added a unit test to test for potential regressions of this bug. N/A
- [x] I've added or updated relevant docs in the docs/ directory. (Onboarding doc still to be updated, after this is merged.)


frontend/src/components/landing/DemoPhone.tsx
```diff
@@ -34,12 +36,19 @@ export const DemoPhone = (props: Props) => {
       screenshot = PremiumScreenshotDe.src;
     }
   }
+  let foreground = FgImage.src;
+  if (lang === "fr") {
+    foreground = FgImageFr.src;
+  }
+  if (lang === "de") {
+    foreground = FgImageDe.src;
+  }
```
groovecoder, line 45: Suggestion (non-blocking) kill `let` with fire ...
```suggestion
  const foreground = lang === "de" ? FgImageDe.src : lang === "fr" ? FgImageFr.src : FgImage.src;
```

Though I think maybe in the past I asked to REMOVE a nested ternary expression? I guess I dislike `let` more than I dislike nested ternary expressions.

privaterelay/urls.py
```diff
@@ -44,19 +44,14 @@
     ),
     path('accounts/', include('allauth.urls')),
     path('api/', include('api.urls')),
+    # This route is still referred to in wrapped_email.html, so leave it in
+    # (even though the homepage is no longer rendered by Django):
```
groovecoder, line 48: 🦅 👁️ 
```suggestion
    # (even though the faq is no longer rendered by Django):
```

privaterelay/urls.py
```diff
@@ -44,19 +44,14 @@
...
-if not settings.SERVE_REACT:
-    urlpatterns += [
-        path('accounts/profile/', views.profile, name='profile'),
-        path('accounts/settings/', views.settings_view, name='settings'),
-        # This redirects users back to the homepage after updating settings
-        path('accounts/profile/settings_update', views.settings_update_view, name='settings-update'),
-        path('premium', views.premium_promo, name='premium-promo'),
-        path('', views.home, name='home'),
-    ]
-
```
groovecoder, line 59: suggestion (non-blocking): Looks like we can also remove these entire view functions from `privaterelay.views`: `profile`, `settings_view`, `settings_update_view`, `premium_promo`. Since they are not bound to any

groovecoder (commented):
Changes (mostly removals) in the python look good. Let's leave the url-referenced no-op views and empty templates in place just to minimize the risk of knock-on effects in this change (which is already really large).

I'd like @maxxcrawford, @codemist , and/or @lloan to check thru the SCSS changes too.

privaterelay/urls.py
```diff
@@ -44,19 +44,14 @@
     ),
     path('accounts/', include('allauth.urls')),
     path('api/', include('api.urls')),
+    # This route is still referred to in wrapped_email.html, so leave it in
+    # (even though the homepage is no longer rendered by Django):
     path('faq', views.faq, name='faq'),
+    # This route is still referred to in middleware.py, so leave it in
+    # (even though the homepage is no longer rendered by Django):
+    path('', views.home, name='home'),
```
jwhitlock, line 52: 👍 fixes #1858

frontend/src/components/dashboard/Onboarding.module.scss
```diff
@@ -38,7 +38,7 @@
 
       &::before {
         content: counter(step-counter);
-        background-image: url("../../../../static/images/dashboard-onboarding/shield.svg");
+        background-image: url("./images/shield.svg");
```
maxxcrawford, line 41: Praise: Now that's a shorter URL. 😎

maxxcrawford (commented):
This PR is a tiny bit bittersweet 🥲 

General Praise: Thanks for circling back on some requests from the Big One. That's some excellent callback! 

> As requested at:
> - https://github.com/mozilla/fx-private-relay/pull/1518#discussion_r799811585
> - https://github.com/mozilla/fx-private-relay/pull/1518#pullrequestreview-872231018

Note: This will be a two-part review. Having to stop at my EOD.

privaterelay/urls.py
Vinnl: 4517fb9b

privaterelay/urls.py
Vinnl: Ah yes, good one: 42131f01

package.json
```diff
@@ -1,49 +1,29 @@
...
-  "devDependencies": {
-    "stylelint": "^13.13.1",
-    "stylelint-config-recommended": "^5.0.0",
-    "stylelint-scss": "^3.19.0"
-  },
   "engines": {
     "node": "14.18.x",
     "npm": "8.1.x"
   },
-  "homepage": "https://github.com/mozilla/fx-private-relay#readme",
```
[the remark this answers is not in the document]
Vinnl: Ah, no strong reason, just thought it was part of some default autogenerated template, since there's also relay.firefox.com. Restored in b4bf17d4.

frontend/src/components/Banner.module.scss
```diff
@@ -79,6 +79,10 @@
 
   .icon {
     margin-right: $spacing-md;
+
+    .warning & {
+      color: $color-yellow-50;
+    }
```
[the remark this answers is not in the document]
Vinnl: That's because rather than having different assets in different colours in our repository, I've been trying to use [`currentColor`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) in SVGs, so that we can just use the right colour in the component we're using, while re-using the same single asset.

.eslintrc.js
```diff
@@ -1,50 +0,0 @@
...
-        ClipboardJS: "readonly"
-      },
-    },
-  ],
-  parserOptions: {
-    ecmaVersion: 2020,
-  },
-  root: true,
-  rules: {
-    indent: ["off", 2, { SwitchCase: 1 }],
```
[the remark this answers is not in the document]
Vinnl: Good question, which led me to check them one by one. There's a bunch of different cases, but none of them seem to require changes to our current ESLint config, I think:

- globals: :tada: no longer necessary, since we can now import modules.
- `indent`, `linebreak-style`, `quotes`, `semi` - handled automatically by Prettier
- - `no-console` - already in the recommended configs that we're extending
- `no-prototype-builtins` - no code that wants this off anymore, I think

frontend/src/components/Icons.tsx
```diff
@@ -82,6 +112,401 @@ export const NewTabIcon = (
...
+      }}
+      {...props}
+    >
+      <title>{alt}</title>
+      <path d="M7.625 1.75c3.446 0 6.25 2.804 6.25 6.25s-2.804 6.25-6.25 6.25-6.25-2.804-6.25-6.25 2.804-6.25 6.25-6.25m0-1.25a7.5 7.5 0 1 0 0 15 7.5 7.5 0 0 0 0-15z" />
+      <path d="M7.625 9.709A.625.625 0 0 1 7 9.084l0-.767c0-.527.37-.987.879-1.092.577-.12.996-.635.996-1.225a1.252 1.252 0 0 0-2.432-.409.626.626 0 0 1-1.181-.408A2.5 2.5 0 1 1 8.25 8.421l0 .663c0 .345-.28.625-.625.625z" />
+      <path d="m8 12-.75 0-.25-.25L7 11l.25-.25.75 0 .25.25 0 .75z" />
+    </svg>
+  );
+};
```
Vinnl, line 346: Once #1883 is merged, this can be replaced by [SupportIcon](https://github.com/mozilla/fx-private-relay/blob/33e03861d415b2e76f2e330016a569783b60ab78/frontend/src/components/Icons.tsx#L208-L232).

frontend/src/components/Icons.tsx
```diff
@@ -82,6 +112,401 @@ export const NewTabIcon = (
...
+    >
+      <title>{alt}</title>
+      <path
+        fillRule="evenodd"
+        clipRule="evenodd"
+        d="M2.423 0H13.577C14.9145 0.00165294 15.9983 1.0855 16 2.423V8.577C15.9983 9.9145 14.9145 10.9983 13.577 11H13V14C12.9999 14.4227 12.7341 14.7996 12.3361 14.9417C11.938 15.0837 11.4936 14.9601 11.226 14.633L8.26 11H2.423C1.0855 10.9983 0.00165294 9.9145 0 8.577V2.423C0.00165294 1.0855 1.0855 0.00165294 2.423 0ZM13.8761 8.87611C13.9554 8.79678 14 8.68919 14 8.577V2.423C14 2.18938 13.8106 2 13.577 2H2.423C2.18938 2 2 2.18938 2 2.423V8.577C2 8.68919 2.04457 8.79678 2.12389 8.87611C2.20322 8.95543 2.31081 9 2.423 9H8.734C9.03434 8.99975 9.31889 9.13449 9.509 9.367L11 11.194V10C11 9.44771 11.4477 9 12 9H13.577C13.6892 9 13.7968 8.95543 13.8761 8.87611ZM11.5 4H4.5C4.22386 4 4 4.22386 4 4.5C4 4.77614 4.22386 5 4.5 5H11.5C11.7761 5 12 4.77614 12 4.5C12 4.22386 11.7761 4 11.5 4ZM4.5 6H11.5C11.7761 6 12 6.22386 12 6.5C12 6.77614 11.7761 7 11.5 7H4.5C4.22386 7 4 6.77614 4 6.5C4 6.22386 4.22386 6 4.5 6Z"
+      />
+    </svg>
+  );
+};
```
Vinnl, line 318: Once #1883 is merged, this can be replaced by [ContactIcon](https://github.com/mozilla/fx-private-relay/blob/33e03861d415b2e76f2e330016a569783b60ab78/frontend/src/components/Icons.tsx#L256-L280).

frontend/src/components/layout/UserMenu.tsx
```diff
@@ -24,11 +24,13 @@ import { useLocalization } from "@fluent/react";
...
-import SettingsImage from "../../../../static/images/settings.svg";
-import ContactImage from "../../../../static/images/icon-message.svg";
-import HelpImage from "../../../../static/images/help.svg";
-import SignoutImage from "../../../../static/images/glocal-sign-out.svg";
-import { NewTabIcon } from "../Icons";
+import {
+  HelpIcon,
+  MessageIcon,
+  NewTabIcon,
+  SettingsIcon,
```
Vinnl, line 31: For consistency with #1883, this should actually use `Cogwheel`.

frontend/src/components/Button.module.scss
```diff
@@ -6,6 +6,7 @@
   display: inline-flex;
   align-items: center;
   justify-content: center;
+  gap: $spacing-xs;
```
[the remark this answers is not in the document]
maxxcrawford: This is 18 days later but I have been using flex/gap on EVERYTHING instead of margin/padding on element. So composable! 😎

maxxcrawford:
@Vinnl Bleh. I know there's a ton of conflicts because this has been open for two weeks. Could you please catch this up (whenever you think is the best time of week to do this) to review/approve? 

Thanks! 

@jwhitlock mentioned this in the retro: If you wanted to break this up into separate, more focused/exploded PRs (_not saying this one isn't_), I'm happy to review that way too. Whatever works for ya!

privaterelay/urls.py
jwhitlock: We still need any patterns that are called by name by `reverse`. The ones I see left in the code are `faq`, `home`, `profile` (privaterelay/views.py line 58), `fxa_login`, and `relayaddress-detail`. So we need `profile` back.

Vinnl: Sharp eye @jwhitlock! The `reverse` for the profile never gets called though, because the view it's in in turn only exists for the URL to the homepage, and never gets rendered. Possibly I could replace the remaining views by empty functions, i.e. removing this `reverse` call?

That said, it feels a bit weird to me to keep these views just to be able to call `reverse` on them to generate a URL - I think the idea behind `reverse` is to ensure links remain up-to-date even if URLs change? Which doesn't get achieved because the views don't represent "real" URLs. (Besides, [URLs should remain valid](https://www.w3.org/Provider/Style/URI.html) anyway.) How would you feel about replacing all `reverse` calls with direct references to the URLs they end up at?

jwhitlock: > How would you feel about replacing all reverse calls with direct references to the URLs they end up at?

I think that's a good plan, ship it!

Vinnl: https://github.com/mozilla/fx-private-relay/pull/1999

Vinnl:
No more conflicts, but rebased again to be sure.

Vinnl:
> No more conflicts, but rebased again to be sure.

emails/templates/emails/wrapped_email.html
```diff
@@ -48,8 +48,7 @@
                   {{ email_wrap_forwarded_from|safe }}
                   {% if has_attachment %}
                       {% ftlmsg 'nav-faq' as faq_text %}
-                      {% url 'faq' as faq_url %}
-                      {% with SITE_ORIGIN|add:faq_url as faq_url %}
+                      {% with SITE_ORIGIN|add:"/faq/" as faq_url %}
```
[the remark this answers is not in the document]
groovecoder: I think http://127.0.0.1:8000/emails/wrapped_email_test is enough to test this template properly.

groovecoder:
> I did find one file path that was not modified/removed from that PR:
> 
>     * https://github.com/mozilla/fx-private-relay/blob/main/.buildpacks.
> 
> 
> Not sure if its still needed for our React frontend, or if it can also be removed. Lmk!

Ooo, AFAIK we only needed the node engine on heroku for django-gulp. Could even streamline heroku builds to remove it.

jwhitlock (approved):
Thanks @Vinnl! ❤️

emails/templates/emails/wrapped_email.html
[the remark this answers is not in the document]
Vinnl: Just a refactor with the same end result, and just verified another time with the wrapped_email_test template that it does indeed render correctly, to be sure.

Vinnl:
The buildpack is still needed I think, as the frontend is still built on Heroku, requiring Node (and of course Python for the back-end).

PR #1900, opened 2022-05-05
Add metric for replies mpp 1146

say-yawn:
# About this PR
Since we offer premium feature to reply to relayed emails, we want to separately increment the number of replies made instead of grouping it with forwarded.

# Acceptance Criteria
- [ ] Client side code can get number of replied per Relay address
- [ ] Client side code can get number of replied per Domain address
- [ ] Client side can get total number of replied per profile
- [ ] Forward metric is NO longer incremented when a user responds to email

#
[MPP-1146](https://mozilla-hub.atlassian.net/browse/MPP-1146)


emails/models.py
```diff
@@ -708,6 +713,15 @@ def profile(self):
     def owner_has_premium(self):
         return self.profile.has_premium
 
+    def increment_num_replied(self):
+        address = self.relay_address or self.domain_address
+        address.num_replied += 1
+        address.last_used_at = datetime.now(timezone.utc)
```
jwhitlock, line 719: praise: I'm going to use this instead of `datetime.now(tz=timezone.utc)`, saves 3 characters!

jwhitlock (changes requested):
Thanks @say-yawn! I like the more precise metric.

Required change: When creating the `DeletedAddress` for a `RelayAddress` and a `DomainAddress`, please copy `num_replied` as well.

jwhitlock:
Hmm, which probably means adding this field to the `DeletedAddress` as well...

jwhitlock (changes requested):
Thanks @say-yawn! A few more changes are needed to get the profile API working.

emails/models.py
```diff
@@ -281,6 +282,14 @@ def emails_blocked(self):
             + sum(da.num_blocked for da in self.domain_addresses)
             + self.num_email_blocked_in_deleted_address
         )
+    
+    @property
+    def emails_replied(self):
+        return (
+            self.relay_addresses.aggregate(models.Sum('num_replied')) +
+            self.domain_addresses.aggregate(models.Sum('num_replied')) +
+            self.num_email_replied_in_deleted_address
```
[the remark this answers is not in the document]
say-yawn: Nice catch! I caught some additional issues with using the `aggregate` function. Thanks for recommending this fix.

api/serializers.py
```diff
@@ -109,6 +109,7 @@ class Meta:
             "next_email_try",
             "bounce_status",
             "api_token",
+            'emails_replied',
```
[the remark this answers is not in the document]
jwhitlock: Thanks, it is in the API now.

Do you plan to add `"num_replied"` to the `domainaddress` and `relayaddress` APIs as well?

emails/migrations/0043_add_num_replied_field_on_addresses.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.27 on 2022-05-05 15:10
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('emails', '0042_domainaddress_used_on'),
```
jwhitlock, line 9: The single quotes will be replaced by double-quotes w/ `black`

emails/tests/models_tests.py
```diff
@@ -797,6 +797,26 @@ def test_user_created_after_premium_release_server_storage_True(self):
...
+    def test_emails_replied_preimum_user_aggregates_sum_of_replies_from_all_addresses(self):
+        subdomain = "test"
+        user = make_premium_test_user()
+        user_profile = Profile.objects.get(user=user)
+        user_profile.subdomain = subdomain
+        user_profile.num_email_replied_in_deleted_address = 1
+        user_profile.save()
+        baker.make(RelayAddress, user=user, num_replied=3)
+        baker.make(DomainAddress, user=user, num_replied=5)
+
```
jwhitlock, line 809: `black` will clean up some whitespace in this section.

jwhitlock (approved):
Thanks @say-yawn! Before merging, can you run `black` to update the formatting?

I tried this in my local development environment, and it worked well:

* Sent two emails to two relay aliases. Forwarded count increments by two.
* Reply to the emails. http://127.0.0.1:8000/api/v1/profiles/ shows `"emails_replied": 1` then `2`.
* Delete one of the aliases. `"emails_replied"` remains at `2`. However, forwarded count drops (issue #1971).
* Continue the conversation on the remaining alias, getting `"emails_replied"` to `3`.

I think this is mergable before the UI to report it.

PR #1912, opened 2022-05-09
MPP-1935: Handle issues with In-Reply-To header

### LINKED ISSUE #1555: Handle InReplyToNotFound while handling replies in sns_message
# About this Issue
While verifying the email relaying from our user back to the external sender, we are not catching the `InReplyToNotFound` in the sns_message.

# Acceptance Criteria
Handle `InReplyToNotFound` exception and return a 4xx error or continue the logic in sns_message to process the email

#
See Sentry error [here](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/18254312/?query=is%3Aunresolved)

### LINKED ISSUE #1619: `_handle_reply` raises exception if the reply address does not exist
The call to ``_get_reply_record_from_lookup_keys`` raises an exception if the reply address does not exist:

https://github.com/mozilla/fx-private-relay/blob/c9f332e090b26620bf431c4a6fd1b3b823c028d1/emails/views.py#L582-L586

This is tracked in two Sentry issues, for the `emails/sns-inbound` endpoint:
* https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/18263433/
* https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/13581873/

When this happens, a 500 error is returned to SNS and it retries 50-100 times before giving up. Instead we should a 40x error and discard the message.

### LINKED ISSUE #1773: Relay fails to process emails to "replies" without "in-reply-to" header
When Relay gets an email to replies@mozmail.com, it looks at the "In-Reply-To" header to see if it is a Relay user. If this header is omitted, an exception ``InReplyToNotFound`` is raised, but not caught. A ``500 Internal Server`` error is returned to SNS, and it is retried until expired.

Instead, we should catch this error, maybe log some information to see if we're missing something, delete the message, and return a non-500 error to SNS.

Replies are processed here, where a `try / except` block could be added:

https://github.com/mozilla/fx-private-relay/blob/578802b9a89ac1dc01acecf02cd12980ad1f9940/emails/views.py#L382-L383

Sentry event first seen in v3.3.4:

https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/19461236/?query=is%3Aunresolved

### LINKED ISSUE #1888: Unknown reply token is a 500 Server Error or kills email processing task
When a reply email is processed, the headers are parsed to determine which email is being replied to. If there is no matching reply in the database, then an exception is raised by `_handle_reply` - `Reply.DoesNotExist`: `Reply matching query does not exist.`.

https://github.com/mozilla/fx-private-relay/blob/4bed7098d5d43359a387124403f57a3f92a4c3b7/emails/views.py#L389-L393

This is not handled in `_sns_response`, which results in a 500 Server Error being returned to the caller.

This can happen if the original email is over 3 months old and the reply record has been removed from the database. There could also be a bug, or a brute force attempt.

* Team: https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/19461758

Acceptance Criteria:
* Exception `Reply.DoesNotExist` is handled
* Exception `InReplyToNotFound` (raised by ` _get_keys_from_headers` when the header is not present or valid)
* Message is deleted from S3
* Message is deleted from the email queue

jwhitlock:
When a message is to "replies":
* Return a 400 if there is no ``In-Reply-To`` header
* Return a 404 if the ``Reply`` record is missing

In both cases, delete any S3-stored email, and remove it from the queue.

This issue has been filed many times!
 * fixes #1555
 * fixes #1619
 * fixes #1773
 * fixes #1888 
 * is tracked by MPP-1935

How to test:

### DIFF
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -22,6 +22,7 @@ from emails.models import (
     DomainAddress,
     Profile,
     RelayAddress,
+    Reply,
 )
 from emails.views import (
     _get_address,
@@ -29,6 +30,7 @@ from emails.views import (
     _sns_message,
     _sns_notification,
     validate_sns_header,
+    InReplyToNotFound,
 )
 
 from .models_tests import make_premium_test_user, upgrade_test_user_to_premium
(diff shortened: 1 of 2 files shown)


emails/views.py
```diff
@@ -598,12 +598,20 @@ def _reply_allowed(from_address, to_address, reply_record):
...
-    reply_record = _get_reply_record_from_lookup_key(lookup_key)
+    try:
+        (lookup_key, encryption_key) = _get_keys_from_headers(mail["headers"])
+    except InReplyToNotFound:
+        return HttpResponse("No In-Reply-To header", status=400)
+
+    try:
+        reply_record = _get_reply_record_from_lookup_key(lookup_key)
+    except Reply.DoesNotExist:
+        return HttpResponse("Unknown or stale In-Reply-To header", status=404)
```
say-yawn, line 609: Nit: adding a metric (via Markus, logger, Sentry, etc.) would be helpful to later make decision on if this is be abused/attacked.

say-yawn (changes requested):
One change suggested to measure the when the Reply object is not found. I really liked the comments on the test that better describes what is being tested. I'll have to do this in the future.

emails/views.py
jwhitlock: I'm adding a metric for now. I think this code will need to be refactored for the task backend (no more HTTP responses), and that will be a good time to tackle logs.

jwhitlock:
Rebase, and add a commit to increment a new metric `fx.private.relay.reply_email_header_error` when the `In-Reply-To` header is missing (tag `detail:no-header`) or has no matching `Reply` record (tag `detail:no-reply-record`).

emails/tests/views_tests.py
```diff
@@ -351,6 +354,83 @@ def test_noreply_email_in_s3_deleted(self, mocked_message_removed):
...
+        self, mocked_get_keys, mocked_message_removed
+    ):
+        """If replies@... email has no "In-Reply-To" header, delete email, return 400."""
+        mocked_get_keys.side_effect = InReplyToNotFound()
+
+        with MetricsMock() as mm:
+            response = _sns_notification(EMAIL_SNS_BODIES["s3_stored_replies"])
+        mm.assert_incr_once(
+            "fx.private.relay.reply_email_header_error", tags=["detail:no-header"]
+        )
```
say-yawn, line 370: praise: So glad we are using MetricsMock to test that increments work as expected! Will use this in the future, thanks for adding it.

say-yawn (approved):
Changes look good and very glad we are making the replies feature more robust.

PR #1913, opened 2022-05-10
for mpp-1813: add new_from_address flag

groovecoder:
This PR is for [MPP-1813](https://mozilla-hub.atlassian.net/jira/software/c/projects/MPP/boards/198?modal=detail&selectedIssue=MPP-1813&sprint=1357).

How to test:
* Run tests: `pytest`
* Test on dev server


privaterelay/settings.py
```diff
@@ -153,6 +153,7 @@
 AWS_SQS_QUEUE_URL = config("AWS_SQS_QUEUE_URL", None)
 
 RELAY_FROM_ADDRESS = config("RELAY_FROM_ADDRESS", None)
+NEW_RELAY_FROM_ADDRESS = config("NEW_RELAY_FROM_ADDRESS", None)
```
jwhitlock, line 156: Nit: I know we use this pattern a lot, but it seems the code is broken if `NEW_RELAY_FROM_ADDRESS` isn't in the environment. If we used this, then that would be a startup error `decouple.UndefinedValueError: NEW_RELAY_FROM_ADDRESS not found. Declare it as envvar or define a default value.` rather than a "why are things not working?" error:

 ```suggestion
NEW_RELAY_FROM_ADDRESS = config("NEW_RELAY_FROM_ADDRESS")
```

File: emails/tests/utils_tests.py [hunk 2]
```diff
@@ -91,6 +92,23 @@ def test_generate_relay_From_with_premium_user(self):
         )
         assert formatted_from_address == expected_formatted_from
 
+    def test_generate_relay_From_with_new_from_user(self):
+        free_user = make_free_test_user()
+        new_from_flag = Flag.objects.create(name="new_from_address")
+        new_from_flag.users.add(free_user)
+        original_from_address = '"foo bar" <foo@bar.com>'
+        formatted_from_address = generate_relay_From(
```
jwhitlock, line 100: I suggest an ``override_settings`` decorator to force ``RELAY_FROM_ADDRESS`` and ``NEW_RELAY_FROM_ADDRESS`` to different knows values

Failing locally before setting `NEW_RELAY_FROM_ADDRESS`: ``IndexError: string index out of range``

emails/tests/utils_tests.py
```diff
@@ -91,6 +92,23 @@ def test_generate_relay_From_with_premium_user(self):
...
+            original_from_address, free_user.profile_set.first()
+        )
+        expected_encoded_display_name = (
+            "=?utf-8?b?IiJmb28gYmFyIiA8Zm9vQGJhci5jb20+IFt2aWEgUmVsYXldIg==?="
+        )
+        expected_formatted_from = "%s <%s>" % (
+            expected_encoded_display_name,
+            settings.NEW_RELAY_FROM_ADDRESS,
+        )
+        assert formatted_from_address == expected_formatted_from
```
jwhitlock, line 110: Failing in CircleCI: 

```
>       assert formatted_from_address == expected_formatted_from
E       AssertionError: assert '=?utf-8?b?Ii...om@127.0.0.1>' == '=?utf-8?b?Ii...7.0.0.1:8000>'
E         Skipping 74 identical leading characters in diff, use -v to show
E         - @127.0.0.1:8000>
E         ?           -----
E         + @127.0.0.1>
```

emails/tests/utils_tests.py
```diff
@@ -91,6 +92,23 @@ def test_generate_relay_From_with_premium_user(self):
         )
         assert formatted_from_address == expected_formatted_from
 
+    def test_generate_relay_From_with_new_from_user(self):
```
jwhitlock, line 95: 😄 I love tests for new functionality! And bonus, the old functionality was already covered by tests!

emails/utils.py
```diff
@@ -218,12 +219,17 @@ def get_post_data_from_request(request):
 
 
 def generate_relay_From(original_from_address, user_profile=None):
+    _, relay_from_address = parseaddr(settings.RELAY_FROM_ADDRESS)
+    try:
+        new_from_flag = Flag.objects.get(name='new_from_address')
+        if user_profile and new_from_flag.is_active_for_user(user_profile.user):
```
jwhitlock, line 225: Nit: test for flag exists but not for that user. And maybe flag exists but `user_profile=None`

jwhitlock (changes requested):
Thanks @groovecoder! Changes are needed to get this working in CircleCI, and maybe to make it more reliable locally as well.

emails/utils.py
groovecoder: Finally got these tests working but I had to explicitly call `new_from_flag.users.remove(free_user)` at the end of this test too. I guess `pytest-django` doesn't clean up the `Flag` objects between tests the same way I expected?

jwhitlock: Ug, that one sounds hard to debug!

Looking at the [docs](https://waffle.readthedocs.io/en/stable/testing/automated.html#unit-tests), the supported method is the `override_flag` decorator, which potentially will clean up after itself, although I see a few [open issues](https://github.com/django-waffle/django-waffle/issues/344) about cache clearing as well.

It may be worth investigating if 1) the database row in the many-to-many table is deleted during test cleanup, and 2) if there is a cache entry that is created and 3) if the cache entry is cleaned up.

jwhitlock (approved):
Thanks @groovecoder, sorry the tests were a struggle with this one! I think it is worth spending some time figuring out how to test waffle consistently, since we're planning on using it more in the future. This one is working now, so let's merge!

privaterelay/settings.py
jwhitlock: @maxxcrawford I suspect you hit an issue with this on Thursday. Was there a useful error?

maxxcrawford: I did! I've ran across it a few times, so I knew what to expect.

I mentioned it in Slack because I had multiple other engineers reach out to me directly about debugging it.

PR #1916, opened 2022-05-10
Fix conditional label storage

Vinnl:
There's two parts to this fix:

- The API now scrubs labels when updating addresses if the user has disabled server storage
- The website now no longer sends them to the API if the user has disabled server storage

We'll probably also want a migration that scrubs labels for every account that has server storage disabled, but I think I'll need some help writing that.

How to test: when visiting the website with the add-on installed and server storage disabled, verify that `description` is currently not included in the PATCH request. (The PATCH request is currently empty; I can also add something to drop the entire request in case no fields are updated, if desired.)

Additionally, try modifying the PATCH request to change the `description` anyway; the back-end should not store it.

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -493,14 +493,17 @@ class RelayAddress(models.Model):
         return super(RelayAddress, self).delete(*args, **kwargs)
 
     def save(self, *args, **kwargs):
+        profile = self.user.profile_set.first()
         if self._state.adding:
             check_user_can_make_another_address(self.user)
             while True:
                 if valid_address(self.address, self.domain):
                     break
                 self.address = address_default()
-            profile = self.user.profile_set.first()
             profile.update_abuse_metric(address_created=True)
+        if (not profile.server_storage):
+            self.description = ""
+            self.generated_for = ""
         return super().save(*args, **kwargs)
 
     @property
@@ -585,11 +588,13 @@ class DomainAddress(models.Model):
         return self.address
 
     def save(self, *args, **kwargs):
+        user_profile = self.user.profile_set.first()
         if self._state.adding:
-            user_profile = self.user.profile_set.first()
             check_user_can_make_domain_address(user_profile)
             user_profile.update_abuse_metric(address_created=True)
         # TODO: validate user is premium to set block_list_emails
+        if (not user_profile.server_storage):
+            self.description = ""
         return super().save(*args, **kwargs)
 
     @property
--- a/emails/models.py
+++ b/emails/models.py
@@ -119,7 +119,7 @@ class Profile(models.Model):
         # appropriate server-stored Relay address data.
         if not self.server_storage:
             relay_addresses = RelayAddress.objects.filter(user=self.user)
-            relay_addresses.update(description="", generated_for="")
+            relay_addresses.update(description="", generated_for="", used_on="")
         return ret
 
     @property
@@ -504,6 +504,7 @@ class RelayAddress(models.Model):
         if (not profile.server_storage):
             self.description = ""
             self.generated_for = ""
+            self.used_on = ""
         return super().save(*args, **kwargs)
 
     @property
--- a/frontend/__mocks__/modules/fluent__react.tsx
+++ b/frontend/__mocks__/modules/fluent__react.tsx
@@ -6,7 +6,7 @@ export const mockFluentReact = {
       l10n: {
         getString: (id: string, vars?: Record<string, string>) =>
           `l10n string: [${id}], with vars: ${JSON.stringify(vars ?? {})}`,
-        bundles: [{ locales: "en-GB" }],
+        bundles: [{ locales: ["en-GB"] }],
       },
     };
   },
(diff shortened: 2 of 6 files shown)


emails/models.py
```diff
@@ -493,14 +493,17 @@ def delete(self, *args, **kwargs):
...
         if self._state.adding:
             check_user_can_make_another_address(self.user)
             while True:
                 if valid_address(self.address, self.domain):
                     break
                 self.address = address_default()
-            profile = self.user.profile_set.first()
             profile.update_abuse_metric(address_created=True)
+        if (not profile.server_storage):
+            self.description = ""
```
Vinnl, line 505: @groovecoder I was wondering if this should also clear `used_on` and, if so, to what value - `None` or `""` or something else?

(Also, is it correct that `generated_for` and `used_on` are not present on DomainAddresses? With the add-on starting to support the latter as well, feels like those fields might need to be present too?)

frontend/src/components/dashboard/aliases/AliasList.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+jest.mock("@fluent/react", () => mockFluentReact);
+jest.mock("../../../config.ts", () => mockConfigModule);
+setMockLocalLabels();
+
+describe("<AliasList>", () => {
+  it("sends a request to the back-end to update the label if server-side label storage is enabled", async () => {
+    const updateCallback = jest.fn();
+    const storeLocalLabelCallback = jest.fn();
+    setMockLocalLabelsOnce(getLocalLabelsWhenDisabled(storeLocalLabelCallback));
```
groovecoder, line 23: There is a LOT of redirection going on here, including some `import ... as` which made this harder to trace. The result of all this logic is ... `setMockLabelsOnce(getReturnValueWhenDisabled as getLocalLabelsWhenDisabled => [null, jest.fn()]` ? Is all that redirection necessary?

question: Also this test is for "server-side label storage is enabled" but it's using `getLocalLabelsWhenDisabled`?

groovecoder:
Spot-check with latest `main` add-on and this branch of server looks good!

emails/models.py
Vinnl: OK, then I took the liberty of also clearing that field when the user disabled `server_storage`. Unfortunately I'm a bit unsure about what will go in that field (I recall seeing a JSON array?), so the mock data in the test may be somewhat idiosyncratic. cb8a5282

emails/models.py
```diff
@@ -493,14 +493,17 @@ def delete(self, *args, **kwargs):
         return super(RelayAddress, self).delete(*args, **kwargs)
 
     def save(self, *args, **kwargs):
+        profile = self.user.profile_set.first()
```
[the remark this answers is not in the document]
Vinnl: Heh, that makes sense - I just moved this line out of the `if` statement. Unfortunately, if I replace it with this, I get four failed tests that I'm not sure how to resolve.

frontend/__mocks__/modules/fluent__react.tsx
```diff
@@ -6,7 +6,7 @@ export const mockFluentReact = {
       l10n: {
         getString: (id: string, vars?: Record<string, string>) =>
           `l10n string: [${id}], with vars: ${JSON.stringify(vars ?? {})}`,
-        bundles: [{ locales: "en-GB" }],
+        bundles: [{ locales: ["en-GB"] }],
```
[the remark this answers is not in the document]
Vinnl: Yeah, the tests didn't actually hit any code yet that actually tried to read this value, but the new test did - and hence found out that the mock was actually wrong.

frontend/src/components/dashboard/aliases/AliasList.test.tsx
Vinnl: Yeah, so far I've been going with names that make sense in a module's own context, but might be duplicated across modules, and thus renaming them at the call-site (similar: every component with its own `Props` type, rather than e.g. `WhatsNewContentProps`), and especially in the case of mocks, the function `getReturnValue*` by itself doesn't really make sense outside of the module's context. But maybe a pattern that might be easier to follow here is to alias the entire module, and then simply call the functions from that? I applied that in 9cd3b02c, which feels like an improvement.

What makes this extra confusing is that there's different, but similar, things that can be enabled or disabled - this test verifies behaviour if server-side label storage is enabled, which automatically means that _local_ storage of labels is _disabled_, hence the use of that function. Hopefully that also makes more sense with the module-wide alias?

(Thought: If you prefer that pattern as well, I'd like to see if we could find a lint rule that prevents aliasing of specific imports.)

maxxcrawford (approved):
Solid fix, @Vinnl! Esp for the extra tests on AliasList! 

Approving, with one extra question: 

Would it be valuable to create a local-labels only user for our mocked dataset? (Example: "Empty, No Sync")

Vinnl:
@maxxcrawford 

> Would it be valuable to create a local-labels only user for our mocked dataset? (Example: "Empty, No Sync")

That could be useful, but unfortunately, I don't think there's an easy way to make it work. Or at least, the effect would be that the label editor would just not be shown, since if sync is off, it's only shown if the add-on is present and taking care of local storage of the labels, which means you need a special build of the add-on explicitly targeting the location of the mocked website.

emails/models.py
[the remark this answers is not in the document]
groovecoder: No worries then.

frontend/src/components/dashboard/aliases/AliasList.test.tsx
groovecoder: Hrm ... yeah there's just a lot going on here I suppose. I'm not sure I can come up with anything better either.

groovecoder (approved):

PR #1928, opened 2022-05-10
Test in CircleCI with PostgreSQL, update psycopg2

jwhitlock:
Run Python tests in CircleCI against a PostgreSQL database, to verify that things still work against a stricter database than sqlite3.

Use this to validated the psycopg2 update from PR #1649


.circleci/config.yml
```diff
@@ -19,7 +19,7 @@ commands:
 executors:
   python:
     docker:
-      - image: cimg/python:3.9.10
+      - image: cimg/python:3.9.12
```
say-yawn, line 22: question: did you mean to update the Python version? If so, can we update the dockerfile too?
https://github.com/mozilla/fx-private-relay/blob/41a4d562504713593677b4564101fbd2727e9f67/Dockerfile#L14

say-yawn (changes requested):
Glad that we added additional testing for Postgres on CircleCI since our local instance is setup with SQLite. Also, finally got to use GitHub's cool "embedded code snippet" (see [doc](https://github.blog/2017-08-15-introducing-embedded-code-snippets/)). One question left on comments on using Python 3.9.12.

.circleci/config.yml
jwhitlock: yes (I meant to update Python) and yes (I will update the Dockerfile)!

say-yawn (approved):
Changes to great. Thanks!

PR #1932, opened 2022-05-10
MPP-1967: Fix logout page

### LINKED ISSUE #1931: `GET /accounts/logout` sometimes results in a 500 Server Error
With v3.4.10, calling `GET /accounts/logout` sometimes results in a 500 Server Error.

The actual exception is: ``ValueError: No bundle set for ftl - use ftlconf/withftl to set bundle`` when rendering a template, which may be an error template.

Sentry:
* Team: https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/19479502/
* Public: https://sentry.prod.mozaws.net/share/issue/498a8a1b7dff48c98073febce067e63b/

The React app does a `POST /accounts/logout`, which seems to always work, and 302 Redirects to the home page.

I was sometimes able to reproduce, and other times I could not. This seemed to work:

* Open a private browsing session
* Go to https://relay.firefox.com
* Sign in to an existing account, or create a new one
* Return to https://relay.firefox.com/accounts/profile/
* Go to https://relay.firefox.com/accounts/logout/

jwhitlock:
Base the logout page on ``base.html``, so that FTL strings are available.

This PR fixes #1931 / MPP-1967

How to test:

1. On current main, change this line in `privaterelay/settings.py`:

    https://github.com/mozilla/fx-private-relay/blob/94c3866f508c50609e6a597487b68f57cb8c7aa8/privaterelay/settings.py#L625

    to 

    ```python
    ACCOUNT_LOGOUT_ON_GET = False
    ```

    This will get you into production mode, but while still having the debug error pages.

2. Still on main, start your local server
3. Login or create an account. You should be on http://127.0.0.1:8000/accounts/profile/
4. Change to http://127.0.0.1:8000/accounts/logout/. You should get the Debug error screen, with the error "ValueError at /accounts/logout/" - "No bundle set for ftl - use ftlconf/withftl to set bundle"
5. Checkout this branch, making the settings change again if needed.
6. Run the server, login, return to http://127.0.0.1:8000/accounts/profile/
7. Change to http://127.0.0.1:8000/accounts/logout/. You should see this somewhat ugly but working page:

logout-working

8. Clicking "Sign Out" will sign you out.

In production the "Sign Out" action is usually a `POST`, which avoids this page.

### DIFF
--- a/privaterelay/templates/account/logout.html
+++ b/privaterelay/templates/account/logout.html
@@ -1,4 +1,4 @@
-{% extends "account/base.html" %}
+{% extends "base.html" %}
 
 {% load ftl %}
 {% ftlconf bundle='privaterelay.ftl_bundles.main' %}
--- a/privaterelay/templates/base.html
+++ b/privaterelay/templates/base.html
@@ -16,7 +16,7 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
     <meta name="description" content="{% ftlmsg 'meta-description-2' %}" />
-    <title>{% ftlmsg 'meta-title' %}</title>
+    <title>{% block head_title %}{% ftlmsg 'meta-title' %}{% endblock %}</title>
     <link rel="stylesheet" href="{% static 'css/app.css' %}">
     <link rel="icon" type="image/svg+xml" href="{% static 'images/logos/relay-logo-dark.svg' %}">
     <link rel="shortcut icon" href="{%  static 'images/favicon.ico' %}">


maxxcrawford (approved):
Works as expected from the testing steps. 

Q: Should we file a ticket to (lightly) style is page? What scenario would this be visible to the user?

jwhitlock:
Thanks! I filed https://github.com/mozilla/fx-private-relay/issues/1935 to style it or replace it with a React version.

Looking at the last 24 hours of logs:

* 150 logouts
  - 120 `POST`s (80%)
     - 119 got a 302. 113 had a /accounts/profile/ referrer, 5 had a /accounts/settings/ referrer, and 1 had /faq/
     - 1 got a 403, no referrer
  - 29 `GET`s (19%)
     - 22 got a 500. 12 had a referrer from /accounts/fxa/login/callback, 2 from Sentry, 8 had no referrer
     - 7 got a 302, none had a referrer
  - 1 `HEAD` got a 302, no referrer 
  
  My guess:
  * A user clicks logout
  * We (backend?) detect that they are not logged in, redirect to login
  * They are redirected to /accounts/logout
  
  Weirdness!

jwhitlock:
Filed https://github.com/mozilla/fx-private-relay/issues/1936 for the general weirdness of getting to logout directly after logging in to FxA.

PR #1954, opened 2022-05-17
Revert "Fix conditional label storage"

Vinnl:
Reverts mozilla/fx-private-relay#1916. I deployed it to stage, and for some reason I'm now seeing labels no longer getting saved even with server storage turned on. I haven't been able to reproduce this locally, and stepping through the code I can't figure out where this is going wrong (`<Alias>` is [calling the `onUpdate` callback provided to it by `<AliasList>`] and at that point has the correct value for `newLabel`, but then when I submit the parameter [passed to that callback](https://github.com/mozilla/fx-private-relay/blob/a02c00e4bd24b37f4f6bc07b8708f86b35a1b689/frontend/src/components/dashboard/aliases/AliasList.tsx#L68), all of a sudden it is an empty object).

While trying to figure out the cause, it's probably best to revert the previous PR, to avoid impacting the default experience in which server-side storage is on by default.


maxxcrawford (approved):
I'm seeing the same thing. Let's back it out.

Vinnl:
@maxxcrawford Aargh, and of course now I'm not seeing this anymore on stage :/ How about you Maxx? And if it's now working for you too, how shall we proceed?

Vinnl:
After discussion with @maxxcrawford on Slack, closing this since the issue might have been caused by something else.

PR #1968, opened 2022-05-19
fix #1967 MPP-1639: Add DomainAddress.used_on field

groovecoder:
This PR fixes #1967.

How to test:
* Create a domain address
* Go to http://127.0.0.1:8000/api/v1/domainaddresses/
  * [ ] Verify you see a new `used_on` field for the domain address

- [ ] I've added a unit test to test for potential regressions of this bug.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #1969, opened 2022-05-20
Ensure labels are saved even with add-on present

### LINKED ISSUE #1970: Labeling/naming masks is not working with Data Collection turned on
#### Build:
- Stage;

#### Affected Platforms:
- Windows 10;

#### Browsers tested on:
- Firefox, Release version;

#### Prerequisites:
- Be signed in into a Free Firefox Relay account;
- Have at least one mask generated;
- Have the latest Stage version of the Add-on provided by Maxx;
- Have the Data Collection option from Settings turned on;

#### Steps to reproduce:
1. Go to the Relay dashboard;
2. Click on the “Add account name” field of one of your masks;
3. Write a name/label and press enter or click outside the field;
4. Refresh the page and observe the name/label of the respective mask;

#### Expected result:
- Pressing enter after typing a name/label saves the label and displays the “Label saved” tooltip;
- After refreshing the page the mask still displays the label/name saved;

#### Actual result:
- The “Label saved” tooltip is not displayed;
- After refreshing the page the mask that was named/labeled is not displaying any name;

#### Notes:
- Happens for both Premium and Free users;
- Generating a mask on a site through the add-on saves the site origin and refreshing doesn’t interfere with such masks;
- This issue is not present when the Data Collection option is off;
- See attached media:
lblsvd

Vinnl:
This PR fixes #1954, or the bug that that was attempting to back out (now reported, this fixes #1970). It resulted in labels not getting saved even with server-side storage enabled - but only when the add-on is installed.

The mistake I made before was to think that `useLocalLabels` would
return `null` for the local labels if server-side storage of labels
was disabled. However, that was not the case: it would return
`null` if the user did not have the add-on installed. Otherwise, it
would simply return an empty array.

The fix was to only store labels locally (and avoid sending them to
the server) if the user has actually disabled server-side label
storage.

How to test: run locally on 127.0... (i.e. `npm run watch`), and run the extension locally (`npm run web-ext-run`). On an account with server-side label storage enabled, you should be able to see a non-empty PATCH request being sent to the API containing your label. Before this change, that request would be empty.

### DIFF
--- a/frontend/src/components/dashboard/aliases/AliasList.test.tsx
+++ b/frontend/src/components/dashboard/aliases/AliasList.test.tsx
@@ -16,7 +16,7 @@ describe("<AliasList>", () => {
     const updateCallback = jest.fn();
     const storeLocalLabelCallback = jest.fn();
     LocalLabelsMock.setMockLocalLabelsOnce(
-      LocalLabelsMock.getReturnValueWhenDisabled(storeLocalLabelCallback)
+      LocalLabelsMock.getReturnValueWithoutAddon(storeLocalLabelCallback)
     );
     render(
       <AliasList
@@ -43,7 +43,7 @@ describe("<AliasList>", () => {
     const updateCallback = jest.fn();
     const storeLocalLabelCallback = jest.fn();
     LocalLabelsMock.setMockLocalLabelsOnce(
-      LocalLabelsMock.getReturnValueWhenEnabled([], storeLocalLabelCallback)
+      LocalLabelsMock.getReturnValueWithAddon([], storeLocalLabelCallback)
     );
     render(
       <AliasList
@@ -69,9 +69,36 @@ describe("<AliasList>", () => {
     );
   });
 
+  it("sends a request to the back-end to update the label if server-side label storage is enabled, even if the add-on is present", async () => {
+    const updateCallback = jest.fn();
+    const storeLocalLabelCallback = jest.fn();
+    LocalLabelsMock.setMockLocalLabelsOnce(
+      LocalLabelsMock.getReturnValueWithAddon([], storeLocalLabelCallback)
+    );
+    render(
+      <AliasList
+        aliases={[getMockRandomAlias()]}
+        onUpdate={updateCallback}
+        onCreate={jest.fn()}
+        onDelete={jest.fn()}
+        profile={getMockProfileData({ server_storage: true })}
+        user={{ email: "arbitrary@example.com" }}
+      />
+    );
+
+    const labelField = screen.getByRole("textbox");
+    userEvent.type(labelField, "Some label");
+    userEvent.tab();
+
+    expect(updateCallback).toHaveBeenCalledWith(expect.anything(), {
+      description: "Some label",
+    });
+    expect(storeLocalLabelCallback).not.toHaveBeenCalled();
+  });
+
   it("does not provide the option to edit the label if server-side storage is disabled, and local storage is not available (i.e. the user does not have the add-on)", async () => {
     LocalLabelsMock.setMockLocalLabelsOnce(
-      LocalLabelsMock.getReturnValueWhenDisabled()
+      LocalLabelsMock.getReturnValueWithoutAddon()
     );
     render(
       <AliasList
--- a/frontend/src/components/dashboard/aliases/AliasList.tsx
+++ b/frontend/src/components/dashboard/aliases/AliasList.tsx
@@ -68,7 +68,8 @@ export const AliasList = (props: Props) => {
     const onUpdate = (updatedFields: Partial<AliasData>) => {
       if (
         localLabels !== null &&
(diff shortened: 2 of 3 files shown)


frontend/src/components/dashboard/aliases/AliasList.test.tsx
```diff
@@ -69,9 +69,36 @@ describe("<AliasList>", () => {
...
+    );
+
+    const labelField = screen.getByRole("textbox");
+    userEvent.type(labelField, "Some label");
+    userEvent.tab();
+
+    expect(updateCallback).toHaveBeenCalledWith(expect.anything(), {
+      description: "Some label",
+    });
+    expect(storeLocalLabelCallback).not.toHaveBeenCalled();
```
groovecoder, line 96: question: So this is asserting that we *DO* store the description in the local storage, but *NOT* call the function which sends the data to the server?

groovecoder (changes requested):

frontend/src/components/dashboard/aliases/AliasList.tsx
```diff
@@ -68,7 +68,8 @@ export const AliasList = (props: Props) => {
     const onUpdate = (updatedFields: Partial<AliasData>) => {
       if (
         localLabels !== null &&
-        typeof updatedFields.description === "string"
+        typeof updatedFields.description === "string" &&
+        props.profile.server_storage === false
       ) {
         storeLocalLabel(alias, updatedFields.description);
         delete updatedFields.description;
```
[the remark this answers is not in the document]
Vinnl: We could add similar `if`s for those, but since those can't be set via the website, it shouldn't be necessary.

frontend/src/components/dashboard/aliases/AliasList.test.tsx
Vinnl: Other way around: _DON'T_ store it in local storage (`storeLocalLabelCallback` is `.not.` called), _DO_ call the function which sends it to the server (`updateCallback` _is_ called).

groovecoder: ah, okay got it - `updateCallback` is the one that sends to server, and `storeLocalLabelCallback` is the one that stores locally. makes sense.

groovecoder (changes requested):
Spot-check fails:

1. run locally on 127.0... (i.e. `python manage.py runserver` & `npm run watch`)
2. run the extension locally (`npm run web-ext-run`).
3. On an account with server-side label storage enabled, you should be able to see a non-empty PATCH request being sent to the API containing your label. ✅ 
4. Go to http://127.0.0.1:8000/accounts/profile/
5. Add a label to a domain address
6. Go to http://127.0.0.1:8000/accounts/settings/
7. Disable server-side storage
8. Go back to http://127.0.0.1:8000/accounts/profile/
   * [ ] Label is still there (from add-on storage) on the domain address
9. Change the label value
10. Refresh the page
   * 💔 The label value is gone

Vinnl:
@groovecoder Oh! I just verified that behaviour, and that's already an issue in `main`. And in fact, as far as I can tell it's been an issue forever: I don't think adding labels for custom masks (domain addresses) ever worked! I created https://github.com/mozilla/fx-private-relay-add-on/issues/341 for that, and think that @maxxcrawford's recent work on supporting custom masks in the add-on might help in solving that. But I propose not blocking this PR on that, since this one fixes a regression, and I think it might require some significant rework in the add-on (which so far just assumed random masks everywhere).

maxxcrawford:
FWIW, with @groovecoder's PR (https://github.com/mozilla/fx-private-relay/pull/1968), there's now a `type` attribute that will make parsing/submitting to the correct API route really straightforward.

Vinnl:
@maxxcrawford While definitely a neat enhancement, the API isn't really involved in the problem here, I think. The problem is that, when the add-on is responsible for storing labels, it does so using `browser.storage.local.set( { relayAddresses } )` which, at least at this time I think, only stores random masks, not custom masks.

maxxcrawford:
Ahh. The addons local storage is one collection of both random and custom masks. We can sync next week and see what refactoring will need to take place.

groovecoder:
Hrm ... I thought I saw the behavior again with a random alias, but I tried many times to reproduce it just now and I couldn't.

groovecoder (approved):

maxxcrawford (commented):
Approving as it fixes users who are syncing labels entirely. However, I'm still running into a few issues, all for non-syncing users: 

- If you delete a mask, even if labels are saved, it resets all masks in the add-on to be without labels. 
- (Premium) If you have a custom mask, after the refresh event (triggered by deleting mask on dashboard), it is no longer listed in the add-on.

Vinnl:
Thanks @maxxcrawford. I think I've addressed those issues now in https://github.com/mozilla/fx-private-relay-add-on/pull/344.

PR #1979, opened 2022-05-23
Block release builds on untranslatable strings

Vinnl:
Rather than just logging this as a warning in debug mode, with this
change it always gets logged when some strings have not been merged
into the submodule yet, and builds are actually broken off if it's
a build in CircleCI triggered by a new Git tag (i.e. it's a build
that will go to stage and could eventually end up in production).

How to test: run `npm run build` with a string in `pendingTranslations.ftl` that has not yet been submitted to the l10n repository. (There's actually one in there right now that [I just submitted](https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/83).) It should only show a warning. Now try to do the same with Circle's env vars for a Git tag (e.g. `CIRCLE_TAG="v3.5.1" npm run build`) and see that it fails.

**Note:** this does mean that we cannot push to stage until all strings are merged, and additionally, it means that builds that succeed in `main` may suddenly start failing when we push a new tag, so I'm not yet entirely convinced that this is desirable. But we can discuss that as a team.

### DIFF
--- a/frontend/src/functions/getL10n.ts
+++ b/frontend/src/functions/getL10n.ts
@@ -58,22 +58,31 @@ export function getL10n() {
           pendingTranslations
         );
         bundle.addResource(pendingTranslationsResource);
-        if (process.env.NEXT_PUBLIC_DEBUG === "true") {
-          const pendingTranslationsIds = pendingTranslationsResource.body.map(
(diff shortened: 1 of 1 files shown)


Vinnl:
There appears to be no appetite for this, so closing for now.

PR #1980, opened 2022-05-23
Add `no-cache, public` header to the homepage

Vinnl:
This should allow Django to kick in and redirect the user to their
dashboard if they are logged in. It should enable https://github.com/mozilla/fx-private-relay-add-on/pull/332 (and fix https://github.com/mozilla/fx-private-relay/pull/1816 not doing anything on prod at the moment).

How to test: This was suggested by Aaron, but since the cache only affects prod, I've only been able to verify that the correct cache headers are added when visiting 127.0.0.1:8000 while not logged in, and aren't changed for other resources.

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -526,7 +526,11 @@ WHITENOISE_INDEX_FILE = True
 # so that the `RedirectRootIfLoggedIn` middleware can kick in for logged-in
 # users.
 def set_index_cache_control_headers(headers, path, url):
-    if path == os.path.abspath("./frontend/out/index.html"):
+    if DEBUG:
+        home_path = os.path.join(BASE_DIR, "frontend/out", "index.html")
+    else:
+        home_path = os.path.join(STATIC_ROOT, "index.html")
+    if path == home_path:
         headers["Cache-Control"] = "no-cache, public"
 
 
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -519,6 +519,20 @@ MEDIA_URL = None
 
 WHITENOISE_INDEX_FILE = True
 
+
+# See
+# https://whitenoise.evans.io/en/stable/django.html#WHITENOISE_ADD_HEADERS_FUNCTION
+# Intended to ensure that the homepage does not get cached in our CDN,
+# so that the `RedirectRootIfLoggedIn` middleware can kick in for logged-in
+# users.
+def set_index_cache_control_headers(headers, path, url):
+    if path == os.path.abspath("./frontend/out/index.html"):
+        print("SETTING HEADER")
+        headers["Cache-Control"] = "no-cache, public"
+
+
+WHITENOISE_ADD_HEADERS_FUNCTION = set_index_cache_control_headers
+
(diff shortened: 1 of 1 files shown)


privaterelay/settings.py
```diff
@@ -519,6 +519,19 @@ def _get_initial_middleware():
...
 WHITENOISE_INDEX_FILE = True
 
+
+# See
+# https://whitenoise.evans.io/en/stable/django.html#WHITENOISE_ADD_HEADERS_FUNCTION
+# Intended to ensure that the homepage does not get cached in our CDN,
+# so that the `RedirectRootIfLoggedIn` middleware can kick in for logged-in
+# users.
+def set_index_cache_control_headers(headers, path, url):
+    if path == os.path.abspath("./frontend/out/index.html"):
```
jwhitlock, line 529: This works for `DEBUG=True` but fails in production mode. Here's a version that works in both:

```python
    if DEBUG:
        home_path = os.path.join(BASE_DIR, "frontend/out", "index.html")
    else:
        home_path = os.path.join(STATIC_ROOT, "index.html")

    if path == home_path:
        headers["Cache-Control"] = "no-cache, public"
```

jwhitlock (changes requested):
Thanks @Vinnl, this worked locally for me in `DEBUG=True`. Some changes were needed in production mode with `DEBUG=False`.

I tested in the browser, and with `curl -I http://127.0.0.1:8000/`.

Are there other web pages that include a redirect if logged in?

Vinnl:
> Are there other web pages that include a redirect if logged in?

Ehm, I guess there are a couple of pages that redirect to the login page if the user is _not_ logged in, but those should not be cached by the CDN. Other than that I think it's only the root.

privaterelay/settings.py
jwhitlock: They are different folders:

```
STATIC_ROOT = os.path.join(BASE_DIR, "staticfiles")
STATICFILES_DIRS = [
    os.path.join(BASE_DIR, "frontend/out"),
    os.path.join(BASE_DIR, "static"),
]
```

In `DEBUG=True`, the file is read from the source at `[proj]/frontend/out/index.html`. When building for prod with `manage.py collectstatic`, they are collected from the `STATICFILES` directories into `STATIC_ROOT`, so it is at `[proj]/staticfiles/index.html`.

15 years ago, we would have collected them to `/var/www/fx-private-relay/static` and served with Apache. 10 years ago, `manage.py collectstatic` would have uploaded them to an S3 bucket, and `STATIC_URL` would be on the CDN domain. Let's see what we're doing in 2030...

Vinnl: Gotcha, cheers!

jwhitlock (approved):
Thanks @Vinnl, the changes work locally!

For `curl -I http://127.0.0.1:8000/accounts/profile/`, I get `Cache-Control: max-age=0, public`. I forget the subtleties of `no-cache` vs `max-age=0`, but [MDN remembers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#max-age_2).

PR #1994, opened 2022-05-24
add ensure_server_data_deleted command

groovecoder:
This PR adds a new `ensure_server_data_deleted` command which will help ensure if any future bugs accidentally store data in the server when it shouldn't be stored there, we delete the data on a regular (i.e., cron) job.

How to test:
Not sure yet ... still determining the best way to re-create the buggy server state.


jwhitlock (commented):
I love the idea, and I immediately want to make it more complicated:

* Change the management command to "cleanup-data", to make it more general
* Give the cleanup function a slugified "name", like "apply-no-server-storage", so that it can show up on the command line, in reports, etc.
* Add a `dry` mode, where it just returns a summary of what would change. In this case, it could be something like `{"profile": 10, "relay_address": 20, "domain_address": 5, "total": 35}`, returning how many rows would be updated. It could be used as a confidence measure when running against prod data for the first time. The non-dry mode could return what actually changed, and should match the "dry" return.
* Adjust the commands so that no work is done if there is no work to do. Look for addresses that have data in the fields as well as have a related profile with the flag. That way, the second run should return no rows to update, and if it does, we've got a bug somewhere.

As far as tests go, it seems like you can use the Django ORM to setup some data, run the command, and test the results. If the ORM doesn't work, some raw SQL should. I'd avoid Django's test fixtures - they seem like the right tool but cause so much heartache as things change.

groovecoder:
Changed the name of the command and added a `--delete` flag to make it actually delete. What enhancements do we want as blockers?

jwhitlock:
Sorry to leave you hanging for _months_. I think the next actions are:

* Omit and/or count profiles with server_storage=False and no server-stored data
* Use `self.stdout` or a logger for output
* If needed, move bulky logic into a utils module for reuse and testing
* Tests that:
  * Setup test profiles with server_storage (False/ True), and (with / without) server- stored data
  * Test dry-run reporting
  * Test ``--delete`` functionality
  
I see this is assigned to me. I can continue this work, unless you want to keep working on it.

jwhitlock:
I talked to @groovecoder IRL, he still wants this and is OK with me taking the next steps 😈

jwhitlock:
@groovecoder this is ready for review, but you can't review because you opened it. Should we start over with a new PR, or keep this one?

Alternatively, _I_ can still review it...

jwhitlock:
Re-opened as #2225

PR #1767, opened 2022-04-11
fix #1748: copy email-images to frontend/public/static

groovecoder:
Supersedes https://github.com/mozilla/fx-private-relay/pull/1754.

This PR fixes #1748 ([MPP-1837](https://mozilla-hub.atlassian.net/browse/MPP-1837)).

How to test:
1. Go to http://127.0.0.1:8000/emails/wrapped_email_test
   * [ ] The images in the email wrapper should show up

- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #1998, opened 2022-05-25
Co-locate images with their usage site

Vinnl:
This was extracted from https://github.com/mozilla/fx-private-relay/pull/1891.

This co-locates images with their site of use, where possible, rather than referring to them in `../../(etc)/static`, which was done to ease the transition. This was requested in reviews of The Big One.

How to test: nothing should have changed.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A


jwhitlock (commented):
I checked that wrapped emails continue to render with this change. It looks like the images were moved back in PR #1767, so this removes the duplicates in the original location.

frontend/src/components/dashboard/PremiumOnboarding.module.scss
```diff
@@ -69,6 +69,10 @@
       align-items: center;
       gap: $spacing-xs;
       padding-bottom: $spacing-xs;
+
+      svg {
+        color: $color-blue-50;
```
maxxcrawford, line 74: praise: Using `currentColor` will make this way more flexible.

frontend/src/components/dashboard/aliases/Alias.tsx
```diff
@@ -174,11 +173,9 @@ export const Alias = (props: Props) => {
                 onClick={copyAddressToClipboard}
               >
                 <samp className={styles.address}>{address}</samp>
-                <img
-                  src={copyIcon.src}
-                  alt=""
-                  className={styles["copy-icon"]}
-                />
+                <span className={styles["copy-icon"]}>
+                  <CopyIcon alt="" />
```
maxxcrawford, line 177: question (nit, non-blocking): Should this have an entry in the `alt` tag?

frontend/src/components/dashboard/subdomain/ConfirmationModal.module.scss
```diff
@@ -41,10 +41,7 @@
         align-items: center;
         justify-content: center;
         padding: $spacing-sm 0;
-
-        img {
-          margin-right: $spacing-sm;
-        }
+        gap: $spacing-sm;
```
maxxcrawford, line 44: praise: `gap` on the parent > `margin` on the child element forever.

maxxcrawford (approved):
Thanks for breaking this out into two PRs! Good to go.

frontend/src/components/dashboard/aliases/Alias.tsx
```diff
@@ -8,8 +8,7 @@ import {
   useTooltipTrigger,
 } from "react-aria";
 import styles from "./Alias.module.scss";
-import copyIcon from "../../../../../static/images/copy-to-clipboard.svg";
-import arrowDownIcon from "../../../../../static/images/arrowhead.svg";
+import { ArrowDownIcon, CopyIcon } from "../../Icons";
```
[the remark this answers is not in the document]
Vinnl: These are actually React components, rather than separate files (i.e. the full filename is `Icons.tsx`). The reason for that is that that allows us to inline the SVGs, which then allows us to use `currentcolor`, which is useful for symbolic items that have just a single colour that might differ depending on their context/hover state.

Vinnl:
Thanks for the review! I'll refrain from merging this until #1891 is merged, since it merges into that branch and therefore would become part of the PR anyway, removing the benefit of having them split out.

Vinnl:
Rebased on `main`. Quite a few conflicts, so I hope I didn't break anything :crossed_fingers:

PR #1999, opened 2022-05-25
Refer directly to non-Django paths

Vinnl:
As discussed in https://github.com/mozilla/fx-private-relay/pull/1891#discussion_r881689812.

`reverse` is useful to maintain links even if URLs change (which
they shouldn't!), but since we were maintaining views that never
get rendered (because they're replaced by static HTML, not rendered
by Django), that benefit no longer plays out.

How to test: Note that I don't know how to actually encounter those two instances of `reverse`, so I haven't been able to verify that this does what it looks like it's doing...

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -25,7 +25,6 @@ from django.db import transaction
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import redirect, render
 from django.template.loader import render_to_string
-from django.urls import reverse
 from django.views.decorators.csrf import csrf_exempt
 
 from .models import (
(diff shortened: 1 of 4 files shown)


jwhitlock (changes requested):
Thanks @Vinnl - I think this is a good way to move away from Django serving these URLs.

There's one more reference inside the HTML email wrapper that needs to change:

https://github.com/mozilla/fx-private-relay/blob/862704501f3c965b4af9bbbf45b01baf96ea4680/emails/templates/emails/wrapped_email.html#L51-L53

This worked for me:

```html
    {% with SITE_ORIGIN|add:'/faq/' as faq_url %}
        {% bold_violet_link faq_url faq_text as faq_link %}
```

groovecoder:
Note: you can check these links at http://127.0.0.1:8000/emails/wrapped_email_test. That will show you the FAQ link that @jwhitlock commented about.

Vinnl:
Thanks @jwhitlock, I wasn't aware of the `url` template tag - fixed that in [cff1e13](https://github.com/mozilla/fx-private-relay/pull/1999/commits/cff1e13ea0c543955255bedd643e6282836a7f36) and also checked for other instances of that (there were none), and this one I was actually able to verify with `wrapped_email_test` (thanks again for that Luke).

jwhitlock (approved):
Nicely done @Vinnl!

say-yawn:
Resolves #1858

Issue #2000, opened 2022-05-25
Users without FxA profile raise AttributeError when determining language, premium availability


jwhitlock:
When calling `user.profile.language` for a user without a related FxA account, an exception `AttributeError: 'NoneType' object has no attribute 'extra_data'` is raised. This happens rarely in production, 0-7 time a day, usually when forwarding HTML email.

* Team link: https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/19481485/
* Public link: https://sentry.prod.mozaws.net/share/issue/88850e95c98e486c87567f2d2e5ede60/

It is possible these are manually created accounts, or former users who have deleted their FxA account.

One option is to fall back to `en` in these cases, which is what we do when a user's FxA profile doesn't have the extra data with a language.

jwhitlock:
A similar error happens when determining if premium is available for the user.

PR #2004, opened 2022-05-26
MPP-2015: Fallback to "en", non-premium when a user has no FxA account

jwhitlock:
When a user doesn't have an associated FxA account, fallback to "en" (English) for the language, and that Premium is not available in their country. This happens rarely in production (0-7 times a day), but often in a fresh development environment.

This PR fixes issue #2000 / MPP-2015.

How to test:

* Install a local development environment, reset the database (delete or rename `db.sqlite3`), go through the setup steps, and add just a superuser.
* On `main`, start the server and go to http://127.0.0.1:8000/emails/wrapped_email_test . You should get the debug exception page, `AttributeError at /emails/wrapped_email_test`, `'NoneType' object has no attribute 'extra_data'`.
* Switch to this branch, restart the server to reload the template and caches.
* Go to http://127.0.0.1:8000/emails/wrapped_email_test, you should see the test email page in English, with no message "Upgrade to Firefox Relay Premium..."

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -124,7 +124,7 @@ class Profile(models.Model):
 
     @property
     def language(self):
-        if self.fxa.extra_data.get("locale"):
+        if self.fxa and self.fxa.extra_data.get("locale"):
             for accept_lang, _ in parse_accept_lang_header(
                 self.fxa.extra_data.get("locale")
             ):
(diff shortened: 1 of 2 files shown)


jwhitlock:
When developing this, I frequently hit the error `AttributeError: 'NoneType' object has no attribute 'extra_data'` filed as issue #2000. It may be useful to review and merge PR #2004 first.

groovecoder (approved):
Code and spot-check LGTM. I didn't even have to make a new user or sign in with it. Going to http://127.0.0.1:8000/emails/wrapped_email_test while signed-out triggered the error on `main` and it was fixed on this branch.

Issue #1618, opened 2022-03-08
Investigate FTL / Fluent exceptions


jwhitlock:
We're getting 500 - 3,000 FTL exceptions per hour in Sentry:

https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/11770350/

Investigate if there are common causes and if any of these are fixable.

The translations repo is at https://github.com/mozilla-l10n/fx-private-relay-l10n

Vinnl:
I think these shouldn't occur so much anymore now that most FTL strings are loaded in the front-end. (i.e. the remaining ones are probably in the email templates?).

jwhitlock:
Yes, much lower. I looked at 5 recent ones:

* locale [da], message 'forwarded-email-footer-2'
* locale [da], message 'forwarded-email-header-from'
* locale [es-ar], message 'nav-profile-sign-out'
* locale [es-ar], message 'nav-profile-sign-out-relay'
* locale [es-ar], message 'nav-profile-help'

I think it would be useful to parse these and generate a report.

jwhitlock:
I've created a BigQuery query to process Fluent error messages, and then some Python to make a report. I think this is a more useful way then Sentry to analyze these errors.

# Fluent Error Report

7446 errors reported between 2022-05-10 and 2022-05-11

## Supported Locales

* cs (288 errors)
    * `'forwarded-email-footer-2'`: 94 errors
    * `'forwarded-email-header-attachment'`: 6 errors
    * `'forwarded-email-header-from'`: 94 errors
    * `'logo-premium-alt'`: 94 errors
* de (201 errors)
    * `'landing-pricing-free-feature-free'`: 201 errors
* el (50 errors)
    * `'landing-pricing-free-feature-free'`: 50 errors
* en (223 errors)
    * `'landing-pricing-free-feature-free'`: 223 errors
* en-gb (201 errors)
    * `'landing-pricing-free-feature-free'`: 201 errors
* es-mx (165 errors)
    * `'landing-pricing-free-feature-free'`: 165 errors
* fr (202 errors)
    * `'landing-pricing-free-feature-free'`: 202 errors
* hu (108 errors)
    * `'landing-pricing-free-feature-free'`: 108 errors
* id (108 errors)
    * `'landing-pricing-free-feature-free'`: 108 errors
* it (186 errors)
    * `'landing-pricing-free-feature-free'`: 186 errors
* nl (168 errors)
[...]

PR #2005, opened 2022-05-26
MPP-1747: Ignore Fluent errors in Sentry, expand translations docs

jwhitlock:
Ignore `django_ftl.message_errors` logger, which Fluent uses when a string is unavailable in the requested language. These errors are not easy to parse as Sentry events, and we are ignoring them at the moment. Instead, we'll start parsing the logs into reports about recent errors. The report is Mozilla-only, so I've linked to it in internal docs rather than in the public repo.

This PR fixes issue #1618 / MPP-1747 by no longer sending these errors to Sentry. This is a similar fix to other "ignore logs" fixes, like the duplicate `request.summary` fix in PR #1620. I don't think it should be tested locally.

It also includes some new documentation about Translation and Localization, that I'd like reviewers to read for correctness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,11 +91,15 @@ Please refer to our [coding standards](docs/coding-standards.md) for code styles
     ```
 
 ### Working with translations
+The following docs will get you started with development, include creating new
+strings to translate.  See [Translation and Localization](docs/translations.md)
+for general information on Relay localization.
+
 #### Getting the latest translations
 We use a [git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules)
 for translated message files. The `--recurse-submodules` step of installation
 should bring the message files into your working directory already, but you may
-want also want to udpate the translations after install. The easiest way to do
+want also want to update the translations after install. The easiest way to do
 that is:
 
 * `git submodule update --remote`
@@ -116,7 +120,7 @@ You can then open a pull request from the `message-updates-yyyymmdd` branch to
 [the l10n repo](https://github.com/mozilla-l10n/fx-private-relay-l10n) `main` branch.
 
 If you're not yet ready to submit some strings for translation, you can
-tentatively add them to frontend/pendingTranslations.ftl. Strings in that file
+tentatively add them to `frontend/pendingTranslations.ftl`. Strings in that file
 will show up until strings with the same ID are added to the l10n repository.
 
 #### Commit translations for release
@@ -131,6 +135,9 @@ You can then commit and push to set the app repository to the updated version
 of the translations submodule:
 
 * `git push`
+
+An automated process updates the submodule daily, bringing in any new changes
+and translations from the Localization Team.
 
 ### Recommended: Enable Firefox Accounts authentication
 To enable Firefox Accounts authentication on your local server, you can use the
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ that is:
 
 * `git submodule update --remote`
 
+To update the submodule automatically when running `git pull` or other commands:
+
+* `git config --global submodule.recurse true`
+
 #### Add/update messages for translation
 The `privaterelay/locales` directory is a git repository like any other, so to
 make changes to the messages:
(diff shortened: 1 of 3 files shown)


docs/translations.md
```diff
@@ -0,0 +1,32 @@
...
+than 20 languages are supported.
+
+The translation repositories are included as submodules, such as
+`privaterelay/locales`. The submodule refers to a specific commit in the
+separate repository. A GitHub action periodically updates this to the latest
+commit, bringing in any new translations or other changes from the localization
+team.
+
+The translation bundles use the [Fluent format](https://projectfluent.org/).
+They are also included in the Docker image, and read by the email processing
```
jwhitlock, line 17:
```suggestion
They are included in the Docker image, and read by the email processing
```

README.md
```diff
@@ -91,11 +91,15 @@ Please refer to our [coding standards](docs/coding-standards.md) for code styles
...
+
 #### Getting the latest translations
 We use a [git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules)
 for translated message files. The `--recurse-submodules` step of installation
 should bring the message files into your working directory already, but you may
-want also want to udpate the translations after install. The easiest way to do
+want also want to update the translations after install. The easiest way to do
 that is:
 
 * `git submodule update --remote`
```
Vinnl, line 105: Unrelated, so not blocking, but this might be a good place to also [mention](https://mozilla.slack.com/archives/C013CSYEL5T/p1646304489835969) the Git config option to automatically do this, which has been working pretty well for me:

```sh
git config --global submodule.recurse true
```

docs/translations.md
Vinnl, line 17: Maybe add something like

```suggestion
They are also included in the Docker image that is deployed to the stage and production environments, and read by the email processing
```

just so people know what that Docker image is used for? (I only recently learned this, and before that might be confused as to why this is relevant.)

docs/translations.md
```diff
@@ -0,0 +1,32 @@
...
+apps at runtime. They are also embedded in the JavaScript during the build
+process, so that the website text is translated.
+
+The user's desired language is stored with their Firefox Account, which uses
+the `Accept-Language` header provided by the browser. If there is a matching
```
Vinnl, line 22: Hmm, this is partially true. When we forward an email to the user, we use the language that is stored in their Firefox account, which uses the Accept-Language provided by their browser _at the time they signed up for their Firefox account_, IIRC.

When they visit the website, however, we ignore that, and just use the Accept-Language that the user's browser is sending at that time, even if that differs from the language that it sent when they created their Firefox account.

Also, I'm not sure if it's worth mentioning, but there isn't always an exact match between the Accept-Language and the languages we have available, and we rely on [`@fluent/langneg`](https://github.com/projectfluent/fluent.js/tree/master/fluent-langneg) to provide us with a list of applicable locales - for example, `de-DE` if the user's Accept-Language is set to just `de`.

docs/translations.md
```diff
@@ -0,0 +1,32 @@
...
+supported language, it is used for the website, add-on, and in email. If there
+is no matching supported language, we fall back to English (`en`).
+
+The application logs an error when a translation is missing. These logs (with
```
Vinnl, line 26:
```suggestion
The application logs an error when a translation is missing in the back-end (i.e. when adding strings to forwarded emails). Missing translations in the website front-end are not logged, and neither are missing translations in the add-on. These logs (with
```

Vinnl (approved):
Docs mostly look good, added some suggestions for clarifications. It's also important to realise (also mentioned inline) that those logged errors only refer to a _very_ limited subset of all the strings we include.

docs/translations.md
jwhitlock: I think these are the kind of details that belong here, thanks!

jwhitlock:
Thanks for the review and the suggestions @Vinnl. I've re-based and incorporated the changes.

If anyone notices other documentation issues, please submit a PR and I'll be happy to review.

PR #2012, opened 2022-05-27
Add API key field and copy to clipboard logic on settings page

maxxcrawford:
Fixes: [MPP-2030](https://mozilla-hub.atlassian.net/browse/MPP-2030)

Notes: 
- I plan on writing a test. I found [this SO article](https://stackoverflow.com/questions/62351935/how-to-mock-navigator-clipboard-writetext-in-jest) to get me started. We also don't have one for the Alias page, so I'll add that as a chore. 

Testing steps: 
- Log in
- Go to /profile/settings page
- **Expected:** There should be a new row under "Privacy" called "API Key"
- Click the copy icon
- **Expected:** API key should be copied to clipboard

Screenshots: 

Screenshot 2022-05-27 at 16-26-59 Firefox Relay

Screenshot 2022-05-27 at 16-26-51 Firefox Relay

### DIFF
--- a/frontend/src/pages/accounts/settings.page.tsx
+++ b/frontend/src/pages/accounts/settings.page.tsx
@@ -1,12 +1,20 @@
 import { useLocalization } from "@fluent/react";
 import type { NextPage } from "next";
-import { FormEventHandler, useEffect, useReducer, useState } from "react";
+import {
+  FormEventHandler,
+  MouseEventHandler,
+  useEffect,
+  useReducer,
+  useRef,
+  useState,
+} from "react";
 import { toast } from "react-toastify";
 import styles from "./settings.module.scss";
 import { Layout } from "../../components/layout/Layout";
 import { Banner } from "../../components/Banner";
 import { useProfiles } from "../../hooks/api/profile";
 import messageIcon from "../../../../static/images/icon-message-purple.svg";
+import copyIcon from "../../../../static/images/copy-to-clipboard.svg";
 import helpIcon from "../../../../static/images/help-purple.svg";
 import performanceIcon from "../../../../static/images/performance-purple.svg";
 import infoTriangleIcon from "../../../../static/images/icon-orange-info-triangle.svg";
@@ -28,6 +36,8 @@ const Settings: NextPage = () => {
   const [labelCollectionEnabled, setLabelCollectionEnabled] = useState(
     profileData.data?.[0].server_storage
   );
+  const [justCopiedApiKey, setJustCopiedApiKey] = useState(false);
+  const apiKeyElementRef = useRef<HTMLInputElement>(null);
 
   const [
     labelCollectionDisabledWarningToggles,
@@ -125,6 +135,13 @@ const Settings: NextPage = () => {
     </li>
   ) : null;
 
+  const copyApiKeyToClipboard: MouseEventHandler<HTMLButtonElement> = () => {
+    navigator.clipboard.writeText(profile.api_token);
+    apiKeyElementRef.current?.select();
+    setJustCopiedApiKey(true);
+    setTimeout(() => setJustCopiedApiKey(false), 1000);
+  };
+
   return (
     <>
       <Layout>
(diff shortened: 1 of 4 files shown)


frontend/src/pages/accounts/settings.page.tsx
```diff
@@ -157,6 +164,43 @@ const Settings: NextPage = () => {
...
+                          />
+                        </button>
+                        {/* <span
+                          aria-hidden={!justCopied}
+                          className={`${styles["copied-confirmation"]} ${
+                            justCopied ? styles["is-shown"] : ""
+                          }`}
+                        >
+                          {l10n.getString("profile-label-copied")}
+                        </span> */}
```
lloan, line 199: question: is it intentional to leave this commented out?

frontend/src/pages/accounts/settings.page.tsx
```diff
@@ -157,6 +164,43 @@ const Settings: NextPage = () => {
...
+                          className={styles["copy-button"]}
+                          title={l10n.getString("settings-button-copy")}
+                          aria-label={l10n.getString("settings-button-copy")}
+                          onClick={copyApiKeyToClipboard}
+                        >
+                          <img
+                            src={copyIcon.src}
+                            alt=""
+                            className={styles["copy-icon"]}
+                          />
```
lloan, line 190: suggestion (non-blocking): We should come up with a uniform way (probably what we will talk about @ work week) to use icons uniformly. I don't know if this is a nebula icon, but it looks like an icon and I know we have an `icons.tsx` file. Just to make things more predictable in the long run.

frontend/src/pages/accounts/settings.page.tsx
```diff
@@ -157,6 +164,43 @@ const Settings: NextPage = () => {
                     {labelCollectionWarning}
                   </div>
                 </div>
+                <div className={styles.field}>
+                  <h2 className={styles["field-heading"]}>
+                    {l10n.getString("setting-label-api-key")}
```
lloan, line 169: question: we submitted this to the l10n repository right? I noticed this here: 
image

frontend/src/pages/accounts/settings.page.tsx
```diff
@@ -157,6 +164,43 @@ const Settings: NextPage = () => {
...
+                    <div className={styles["copy-api-key-display"]}>
+                      {profile.api_token}
+                    </div>
+                    <span className={styles["copy-controls"]}>
+                      <span className={styles["copy-button-wrapper"]}>
+                        <button
+                          className={styles["copy-button"]}
+                          title={l10n.getString("settings-button-copy")}
+                          aria-label={l10n.getString("settings-button-copy")}
+                          onClick={copyApiKeyToClipboard}
```
lloan, line 184: issue: This is making your test fail - we can probably use `MouseEvent<HTMLButtonElement>` from React

lloan (changes requested):
Pretty straightforward - awesome new feature.

frontend/src/pages/accounts/settings.page.tsx
Vinnl: Part of the trick here is to parse TypeScript's sometimes inscrutable error messages. In this case, when you hover over `onClick`'s red squiggly line, it says (among others):

```
Type '(e: MouseEvent) => void' is not assignable to type 'MouseEventHandler<HTMLButtonElement>'.
```

The former is the type of `copyApiKeyToClipboard`: it's a function with one parameter (`e` of type `MouseEvent`) that returns nothing (`void`). `onClick`, however, expects a `MouseEventHandler<HTMLButtonElement>`. You can tell TypeScript that it is of that type as follows:

```diff
-   const copyApiKeyToClipboard = (e: MouseEvent) => {
+   const copyApiKeyToClipboard: MouseEventHandler<HTMLButtonElement> = (e) => {
```

(The types are pretty similar, but this provides some extra info, such as that `e.currentTarget` is an `HTMLButtonElement`, and thus that you can access e.g. `e.currentTarget.form` inside the handler. Which, of course, is the case for an `onClick` handler of a `<button>`.)

frontend/src/pages/accounts/settings.page.tsx
```diff
@@ -125,6 +126,12 @@ const Settings: NextPage = () => {
     </li>
   ) : null;
 
+  const copyApiKeyToClipboard = (e: MouseEvent) => {
+    // This preventDefault was necessary to stop the form from submitting
+    e.preventDefault();
```
Vinnl, line 131: That's because a `<button>` inside a `<form>` [is of type `submit` by default](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type), and hence clicking it submits the form.

(Of course the button shouldn't be in the form in the first place, UX-wise, but :shrug: :slightly_smiling_face:  )

Anyway, you can fix that by adding `type="button"` to the `<button>` element.

frontend/src/pages/accounts/settings.page.tsx
Vinnl: We used to keep them all in `/static/` to ensure that the non-React and React website would use the same website. Now that we've converted to React, I think it makes sense to move them into `Icons.tsx` so that we can recolour icons as necessary - https://github.com/mozilla/fx-private-relay/pull/1891 actually does that, but of course, while it's awaiting a review, new PRs will arrive that don't yet have a consistent method to emulate.

frontend/src/pages/accounts/settings.page.tsx
Vinnl: I think that was in-progress work to display a confirmation after copying. I just finished that in 84db24ffdcb55ffa209dc79b54d0495524d49970.

frontend/src/pages/accounts/settings.page.tsx
Vinnl: I'm seeing it correctly; you might not be up-to-date with the l10n repo submodule yet? Here's a tip that might be interesting (I should submit a PR to the README with this): https://mozilla.slack.com/archives/C013CSYEL5T/p1646304489835969

Vinnl (approved):
This is ready to go as far as I'm concerned. I have a vague recollection of needing to merge this today, but since it's not really QA'able other than making sure that it's actually shown and copied to the clipboard, I'll hold off until US folks are back.

lloan (approved):
Looks good!

PR #2021, opened 2022-05-31
Fix filtering by local labels

### LINKED ISSUE #2008: With Data Collection turned off, masks cannot be searched on the dashboard by their label/name

Vinnl:
This PR fixes #2008. We used to apply the locally-stored labels to the mask data *after*
applying the filter. This flips that around.

How to test: disable server-side label storage. Add labels with the extension, then enter that label in the search field. (Make sure to run on 127.0.0.1, so that the extension becomes active.) As follows:

https://user-images.githubusercontent.com/4251/171151940-670520b0-82ff-42cd-8f2a-831459fd7012.mp4

I also fixed the one warning we used to get when running the front-end tests about updating an unmounted component in 5d481203ee895da3589067abf0d0384659c9999c.

### DIFF
--- a/frontend/src/components/dashboard/aliases/LabelEditor.tsx
+++ b/frontend/src/components/dashboard/aliases/LabelEditor.tsx
@@ -29,7 +29,8 @@ export const LabelEditor = (props: Props) => {
 
   useEffect(() => {
     setLabelJustUpdated(true);
-    setTimeout(() => setLabelJustUpdated(false), 1000);
+    const updateTimeout = setTimeout(() => setLabelJustUpdated(false), 1000);
+    return () => clearTimeout(updateTimeout);
   }, [props.label]);
 
   const onSubmit: FormEventHandler = (event) => {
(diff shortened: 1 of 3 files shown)


frontend/src/components/dashboard/aliases/LabelEditor.tsx
```diff
@@ -29,7 +29,8 @@ export const LabelEditor = (props: Props) => {
 
   useEffect(() => {
     setLabelJustUpdated(true);
-    setTimeout(() => setLabelJustUpdated(false), 1000);
+    const updateTimeout = setTimeout(() => setLabelJustUpdated(false), 1000);
+    return () => clearTimeout(updateTimeout);
```
lloan, line 33: question (non-blocking): For sake of learning on my part, why do you set a timeout for this cleanup function?

lloan (approved):
LGTM. Left a non-blocking question.

frontend/src/components/dashboard/aliases/LabelEditor.tsx
Vinnl: The timeout was already there, to hide the "Saved!" tooltip after a second passes. However, if the component gets unmounted before the timeout's callback executes (e.g. because the user navigates to another page within a second, or in tests), the `setLabelJustUpdated` call no longer makes sense. Thus, the change here is to drop that timeout as soon as the component unmounts.

image

PR #2027, opened 2022-06-01
Prevent Premium tooltip from expanding the slider

### LINKED ISSUE #2023: Mask status changes when the user clicks the link in Promotionals pop-up and gets back to previous page
**Build:**
Stage

**Affected platforms:**
Windows 10

**Browsers tested on:**
Firefox, latest release

**Preconditions:**
The user does not have an active subscription.
At least one mask is available in the list.

**Steps to reproduce:**
1. Go to Relay homepage and log in.
2. Expand one mask card and click the Promotionals (Premium only) option.
3. Click "Upgrade now"/ "Join the Relay Premium waitlist" link if the user is in a premium available country/ premium not available country.
4. Click the browser's "Back" button.
5. Observe the status of the card.

**Actual result:**
The status of the card is changed to Block all. The "All" circle is also highlighted when you repeat the steps one more time.

**Expected result:**
The status doesn't change.

**Notes:**
N/A

status changes to block all

Vinnl:
This PR fixes #2023.

When free users try to select promotional email blocking, we pop
up a tooltip that explains that they will need to upgrade to
Premium for that.

However, because the tooltip element was located inside the
slider's track, useSlider counted it as part of the slider, and
clicking it would change the slider value.

To solve this, I wrapped an element (.track-wrapper) around the
track, that contains both the track and the tooltip. Everything can
then be positioned relative to that wrapper instead of relative to
the track, without that track being considered the slider.

(Most of this complexity is due to the weird hybrid we have of both
a slider and a button that pretends that it is part of the slider,
but actually is not. Perhaps at some point we'll want to split up
<PromotionalTrackStop> into the "real" one that behaves as normal
for Premium users, and is simply hidden for non-Premium users, and
a button that *looks* like the real one, but actually is completely
separate from the slider.)

How to test: [with a free user](https://deploy-preview-2027--fx-relay-demo.netlify.app/?mockId=empty), expand a mask and click the track stop for promotional email blocking. This should open up a pop-up. Clicking anywhere on that pop-up should not affect the slider value (as opposed to [the current state](https://fx-relay-demo.netlify.app/?mockId=empty)).

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, UI bug.

### DIFF
--- a/frontend/src/components/dashboard/aliases/BlockLevelSlider.module.scss
+++ b/frontend/src/components/dashboard/aliases/BlockLevelSlider.module.scss
@@ -42,7 +42,7 @@ $trackLineHeight: 4px;
       display: inline-block;
     }
 
-    .track {
+    .track-wrapper {
       // To absolutely position the track in the middle of the control, it needs
       // to be offset from the top by:
       // 1. the height of the icon on top of it ($iconHeight), plus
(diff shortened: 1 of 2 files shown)


frontend/src/components/dashboard/aliases/BlockLevelSlider.module.scss
```diff
@@ -52,26 +52,59 @@ $trackLineHeight: 4px;
...
+        }
+
+        .thumb-container {
+          position: absolute;
+          transform: translateX(-50%);
+          cursor: pointer;
+          // The thumb has a z-index of 1 to overlap .track-stop-promotional,
+          // which is also absolutely positioned but occurs later in the DOM,
+          // and therefore would overlap the thumb.
+          z-index: 1;
```
Vinnl, line 86: This is new, and needed because `.track-stop-promotional` now occurs later than `.thumb-container` (because I had to move it out of the slider track).

frontend/src/components/dashboard/aliases/BlockLevelSlider.module.scss
```diff
@@ -143,6 +176,8 @@ $trackLineHeight: 4px;
           // offset it by half of the width of the track, minus half its own
           // width:
           left: calc(50% - var(--thumbDiameter) / 2);
+          top: $iconHeight;
+          cursor: pointer;
```
Vinnl, line 180: This is new, and needed because `.track-stop-promotional` no longer automatically got positioned at the right height by virtue of being part of the `.track`.

Additionally, since  the entire track is clickable (to move the slider), it already had `cursor: pointer`, so now that `.track-stop-promotional` is outside of the track, it needed its own.

frontend/src/components/dashboard/aliases/BlockLevelSlider.module.scss
```diff
@@ -193,6 +202,10 @@ $trackLineHeight: 4px;
           img {
             display: none;
           }
+
+          &.track-stop-promotional {
+            top: 0;
+          }
```
Vinnl, line 208: As above, this now needs to be correctly positioned by itself.

frontend/src/components/dashboard/aliases/BlockLevelSlider.tsx
```diff
@@ -103,14 +103,37 @@ export const BlockLevelSlider = (props: Props) => {
...
           {sliderSettings.label}
         </label>
-        <div {...trackProps} ref={trackRef} className={styles.track}>
-          <div className={styles["track-line"]} />
-          <div
-            className={getTrackStopClassNames(props.alias, sliderState, "none")}
-          >
-            <img src={UmbrellaClosedMobile.src} alt="" />
-            <p aria-hidden="true">{getLabelForBlockLevel("none", l10n)}</p>
+        <div className={styles["track-wrapper"]}>
```
Vinnl, line 106: Essentially, the only change here is the new element `.track-wrapper`, wrwapping `.track`, and that `<PromotionalTrackStop>` was moved to below `.track` (but inside `.track-wrapper`).

Vinnl (commented):
Some review guidance - it looks like a lot of changes, but it's not actually that much.

frontend/src/components/dashboard/aliases/BlockLevelSlider.module.scss
```diff
@@ -52,26 +52,59 @@ $trackLineHeight: 4px;
...
+          // which is also absolutely positioned but occurs later in the DOM,
+          // and therefore would overlap the thumb.
+          z-index: 1;
+
+          .thumb {
+            width: var(--thumbDiameter);
+            height: var(--thumbDiameter);
+            border-radius: 50%;
+            background-color: $color-violet-20;
+            border: 4px solid $color-violet-50;
```
lloan, line 93:
```suggestion
            border: $border-radius-sm solid $color-violet-50;
```

lloan (approved):
I added one non-blocking suggestion, but overall, LGTM. Tested - works.

frontend/src/components/dashboard/aliases/BlockLevelSlider.module.scss
Vinnl: Hmm, is that the right abstraction? As in, if Protocol changes the "small" border radius, do we want the width of the slider thumb's border to change as well?

(PRs are low-effort btw, so happy to do a follow-up PR if you think it is.)

PR #2028, opened 2022-06-01
DNM: v3.5.6 candidate

jwhitlock:
**DO NOT MERGE**: This is only a PR to get CircleCI to run tests before tagging, and will not be merged.

This is a candidate for the v3.5.6 tag. It includes the changes from v3.5.4 to main, but removes #1673, which needs a bugfix round.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,11 +91,15 @@ Please refer to our [coding standards](docs/coding-standards.md) for code styles
     ```
 
 ### Working with translations
+The following docs will get you started with development, include creating new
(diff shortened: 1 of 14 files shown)


jwhitlock:
Tagged as https://github.com/mozilla/fx-private-relay/releases/tag/v3.5.6

Vinnl:
@jwhitlock Shouldn't this (and the other tag) still be merged into `main`, so that there's at least a commit history flowing from current production to the next release? Just to make sure there aren't any commits in here that we'd remove again if we were to do another prod deploy?

PR #2029, opened 2022-06-01
Replace Premium Upsell banner to Loyalist List Firefox banner 

codemist:
# New feature description
On the top dashboard, replace the upsell banner to focus more on supporting Mozilla and the mission while subscribing to Relay.

l10n: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/86/files
# Screenshot (if applicable)

Preview:
image
image

# How to test

l10n pr : https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/86/files

On a free account, check that the premium banner has been replaced with the loyalist banner.

# Checklist

### DIFF
--- a/frontend/src/components/Banner.tsx
+++ b/frontend/src/components/Banner.tsx
@@ -18,6 +18,12 @@ export type BannerProps = {
     onClick?: () => void;
     gaViewPing?: Parameters<typeof useGaViewPing>[0];
   };
+  ctaLargeButton?: {
+    target: string;
+    content: string;
+    onClick?: () => void;
+    gaViewPing?: Parameters<typeof useGaViewPing>[0];
+  };
   /**
    * See {@link useLocalDismissal}; determines whether and for how long the user can dismiss this banner.
    */
@@ -70,6 +76,20 @@ export const Banner = (props: BannerProps) => {
     </div>
   ) : null;
 
+  const ctaLargeButton = props.ctaLargeButton ? (
+    <div className={styles["cta-large-button"]}>
+      <OutboundLink
+        to={props.ctaLargeButton.target}
+        eventLabel={props.ctaLargeButton.content}
+        target="_blank"
+        rel="noopener noreferrer"
+        onClick={props.ctaLargeButton.onClick}
+      >
+        <span ref={ctaRef}>{props.ctaLargeButton.content}</span>
+      </OutboundLink>
+    </div>
+  ) : null;
+
   const dismissButton =
     typeof props.dismissal !== "undefined" ? (
       <button
@@ -93,11 +113,12 @@ export const Banner = (props: BannerProps) => {
     >
       <div className={`${styles["highlight-wrapper"]}`}>
         {illustration}
-        <div>
+        <div className={`${styles["title-text"]}`}>
           {title}
           {props.children}
           {cta}
         </div>
+        {ctaLargeButton}
       </div>
       {dismissButton}
     </div>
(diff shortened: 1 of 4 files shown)


frontend/src/components/dashboard/ProfileBanners.tsx
```diff
@@ -194,26 +201,55 @@ const NoChromeExtensionBanner = () => {
...
+      illustration={<img src={FirefoxLogo.src} alt="" width={60} height={60} />}
+      ctaLargeButton={{
         target: getPremiumSubscribeLink(props.runtimeData),
-        content: l10n.getString("banner-upgrade-cta"),
+        content: l10n.getString("banner-upgrade-loyalist-cta"),
         onClick: () => trackPurchaseStart(),
         gaViewPing: {
           category: "Purchase Button",
-          label: "profile-banner-promo",
+          label: "profile-banner-loyalist-promo",
```
codemist, line 244: I've only updated the GA label to track the performance of this banner as opposed to the premium banner - is there anywhere else I should change from the old banner build? @Vinnl

frontend/src/components/dashboard/ProfileBanners.tsx
```diff
@@ -194,26 +198,55 @@ const NoChromeExtensionBanner = () => {
...
+//         gaViewPing: {
+//           category: "Purchase Button",
+//           label: "profile-banner-promo",
+//         },
+//       }}
+//     >
+//       <p>{l10n.getString("banner-upgrade-copy-2")}</p>
+//     </Banner>
+//   );
+// };
```
Vinnl, line 224: Obviously no biggie, but a small tip: you don't need to comment out this entire thing either. You can get rid of the ESLint complaint by prepending a single line, accompanied by a comment about why it's disabled:

```js
// Unused but left in for when we no longer want to use <LoyalistPremiumBanner>
// eslint-disable-next-line @typescript-eslint/no-unused-vars
```

VSCode will even tell you what the line is if you hit the :bulb: icon:

image

frontend/src/components/Banner.tsx
```diff
@@ -18,6 +18,12 @@ export type BannerProps = {
     onClick?: () => void;
     gaViewPing?: Parameters<typeof useGaViewPing>[0];
```
Vinnl, line 19: Haha, so funny that the regular CTA wasn't big enough :P 

The current approach is fine, but since it's interesting to see how different people approach problems, I'll share what my intuition was :) SInce large CTA is otherwise the same as the regular CTA, I'd just have added an optional property to that definition, and then use that to determine which class to add:

```suggestion
    gaViewPing?: Parameters<typeof useGaViewPing>[0];
    size?: "regular" | "large";
```

(i.e. that adds an optional `size` property that can be set to `"regular"` or `"large"`, and then you could do something like `className={props.cta.size === "large" ? styles[["cta-large-button"] : styles.cta }`)

Again, no need to change this, just sharing in case it's interesting :)

Vinnl: Oh actually, it also affects its position in the DOM. Well, never mind then - though I'm leaving this in because the thought process might still be interesting :P

frontend/src/components/dashboard/ProfileBanners.tsx
Vinnl: Nope, that's it!

Vinnl (approved):
Good to go, looking good!

frontend/src/components/Banner.tsx
codemist: Oh that is awesome to know in the future - I figured there was some kind of way to do it simply like that. Would I have had to do add the size property for all the other banners in ProfileBanners.tsx if that had been the case, or just the one?

frontend/src/components/dashboard/ProfileBanners.tsx
codemist: This is really good to know! May add this in the No Shame doc. I didn't have the VSCode prompts like you do when I hovered over it, is there a library I need to download?

frontend/src/components/Banner.tsx
Vinnl: Nope, the `?` marks it as optional (i.e. `size?: "regular" | "large"` is roughly the same as `size: "regular" | "large" | undefined`)!

PR #2035, opened 2022-06-03
change ON_HEROKU to RELAY_CHANNEL and add Bitwarden QA origin

groovecoder:
Bitwarden Web Vault integrates Relay. So, our production server allows
the Bitwarden production origin. Our stage server should also allow
the Bitwarden stage origin.

How to test:
The real test will come from the real Bitwarden stage origin, so we can't test that part.

We *should* test to make sure that none of our other channel/deployment/environment-specific logic is broken. E.g., on the dev server, we should still prepend aliases with `mail`.

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -48,7 +48,7 @@ TMP_DIR = os.path.join(BASE_DIR, "tmp")
 # defaulting to blank to be production-broken by default
 SECRET_KEY = config("SECRET_KEY", None, cast=str)
 
-ON_HEROKU = config("ON_HEROKU", False, cast=bool)
+RELAY_CHANNEL = config("RELAY_CHANNEL", "prod", cast=str)
 DEBUG = config("DEBUG", False, cast=bool)
 if DEBUG:
     INTERNAL_IPS = config("DJANGO_INTERNAL_IPS", default=[])
@@ -100,7 +100,7 @@ csp_style_values = ["'self'"]
 # Next.js dynamically inserts the relevant styles when switching pages,
 # by injecting them as inline styles. We need to explicitly allow those styles
 # in our Content Security Policy.
-if DEBUG:
+if RELAY_CHANNEL == "local":
     # When running locally, styles might get refreshed while the server is
     # running, so their hashes would get oudated. Hence, we just allow all of
     # them.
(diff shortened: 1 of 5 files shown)


privaterelay/settings.py
```diff
@@ -48,7 +48,7 @@
 # defaulting to blank to be production-broken by default
 SECRET_KEY = config("SECRET_KEY", None, cast=str)
 
-ON_HEROKU = config("ON_HEROKU", False, cast=bool)
+RELAY_CHANNEL = config("RELAY_CHANNEL", "prod", cast=str)
```
say-yawn, line 51: Rather than having a separate env variable for the different environment, we should use the `settings.SITE_ORIGIN` to set the `RELAY_CHANNEL`, this will remove the duplicate check on whether or not the environment is prod, stage, dev, or local.
```suggestion
ORIGIN_ENV_MAP = {"dev.fxprivaterelay.nonprod.cloudops.mozgcp.net": "dev", "stage.fxprivaterelay.nonprod.cloudops.mozgcp.net": "stage", "relay.firefox.com": "prod"}
RELAY_CHANNEL = ORIGIN_ENV_MAP[SITE_ORIGIN] if SITE_ORIGIN in ORIGIN_ENV_MAP else "local"
```

say-yawn (changes requested):
Requesting that we use SITE_ORIGIN to derive the channel rather than using another env variable.

privaterelay/settings.py
say-yawn: According to the infra document seems like the `https://` needs to be added: https://github.com/mozilla-services/cloudops-infra/blob/b9e9796555e0e05aaf4a1a2e5e60c8cbbdbe0ff8/projects/fxprivaterelay/k8s/values/app.yaml#L45

groovecoder: Oh nice! Happy to not add another env var. I'll update.

groovecoder: Updated.

jwhitlock: Something like:

```python
_DEFAULT_FOR_CHANNEL = {
    "prod": {
        "SITE_ORIGIN": "https://relay.firefox.com",
    },
    "stage": {
        "SITE_ORIGIN": "https://stage.fxprivaterelay.nonprod.cloudops.mozgcp.net",
    },
    "dev": {
        "SITE_ORIGIN": "https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net",
    },
    "local": {
        "SITE_ORIGIN": "http://127.0.0.1:8000",
    }
}

def config_with_channel_default(name, default=None, cast = None)
    if RELAY_CHANNEL in _DEFAULT_FOR_CHANNEL:
        default = _DEFAULT_FOR_CHANNEL[RELAY_CHANNEL][name]
    if cast is not None:
        return config(name, default, cast=cast)
    else:
        return config(name, default)

SITE_ORIGIN = config_with_channel_default("SITE_ORIGIN", None)
```

groovecoder: Can we pull `RELAY_CHANNEL` from `SITE_ORIGIN` For Now™ so we can merge + deploy this without needing SRE to do anything?

say-yawn (approved):
LGTM

PR #2036, opened 2022-06-03
Fix #2034  - Change Bento icon color to white

### LINKED ISSUE #2034: Firefox apps and services icon from the menu is black on homepage
**Build:**
Stage

**Affected platforms:**
Windows 10

**Browsers tested on:**
Firefox, latest release
Chrome, latest release

**Preconditions:**
N/A

**Steps to reproduce:**
1. Go to Relay homepage.
2. Observe the Firefox apps and services icon.
3. Sign in and observe the Firefox apps and services icon again.
4. Sign out and observe the Firefox apps and services icon again.

**Actual result:**
The Firefox apps and services icon is black on the homepage.

**Expected result:**
The Firefox apps and services icon is white and more easily visible.

**Notes:**
N/A

icon is black
FF apps and services black

codemist:
This PR fixes #2034

Preview:
image

How to test:
Check that the bento icon is white on the landing page, and on the free user's dashboard.

### DIFF
--- a/frontend/src/components/layout/navigation/AppPicker.module.scss
+++ b/frontend/src/components/layout/navigation/AppPicker.module.scss
@@ -18,6 +18,7 @@
   border-radius: $border-radius-sm;
 
   svg {
+    fill: $color-white;
     color: $color-white;
 
     &.premium {


frontend/src/components/layout/navigation/AppPicker.module.scss
```diff
@@ -18,6 +18,7 @@
   border-radius: $border-radius-sm;
 
   svg {
+    fill: $color-white;
```
Vinnl, line 21: Hmm, I don't understand why this would be needed - as far as I can see, `<AppPicker>` [uses](https://github.com/mozilla/fx-private-relay/blob/7bee0b40192f536b06c7b4aeac3a02467e28836f/frontend/src/components/layout/navigation/AppPicker.tsx#L269-L272) the `<BentoIcon>` from `Icons.tsx`, which as far as I can see [sets its `fill` to `currentColor`](https://github.com/mozilla/fx-private-relay/blob/7bee0b40192f536b06c7b4aeac3a02467e28836f/frontend/src/components/Icons.tsx#L74), which means `color` should already set the right colour.

But writing this I see what the error is: it sets `fill` to `currentcolor` rather than `currentColor` (i.e. there's a casing error). @codemist could you apply that change instead, for consistency with the other icons?

(The advantage of using `currentColor` is that we can set the colour of an icon regardless of whether that colour is defined using `fill` or `stroke` or a combination of the two.)

Vinnl (changes requested):

frontend/src/components/layout/navigation/AppPicker.module.scss
codemist: Did you mean to switch all of the `fill` properties in `Icons.tsx` to `currentColor`? Weirdly that did not fix the issue with the Bento showing as black instead of white.

Vinnl: Oh hmm, looks like I misremembered: `currentcolor` (uncapitalised) [appears to be](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) the correct spelling :/ @lloan @maxxcrawford do you have any idea why that does not result in a white bento icon when logged out?

Interestingly, I'm also only seeing this issue [on stage](https://stage.fxprivaterelay.nonprod.cloudops.mozgcp.net/); in [the demo site](https://fx-relay-demo.netlify.app/), the Bento icon is white.

It looks like the inline `style` attribute gets removed on stage:

image

where it is just present in the demo site:

image

That might imply that it gets removed in the build, but when I run the production build locally (i.e. `npm run watch`), I _do_ see the `style` attribute :/

If nobody knows the cause, I suppose we can go with this fix (it's not _too_ important), but maybe you could add a comment with a link here, to explain that it's a workaround? Because for all I can see, the current implementation _should_ work.

Vinnl: Hmm OK, I _can_ reproduce this locally when I checkout v3.5.10 (i.e. what's currently on `stage`) and then run `npm run watch`, but I _can't_ reproduce it with latest `main`. So I guess this is already fixed? (Though it's not clear to me what got it fixed.)

maxxcrawford: I def. see the `fill:currentcolor` being removed from `main` on stage.

maxxcrawford:
@Vinnl Note: I'm merging this with your requested changes. Let's sync tomorrow to see if this needs to be resolved further. I want to get the fix out as Stage is still broken. (Trying to deploy [v3.5.11](https://github.com/mozilla/fx-private-relay/releases/tag/v3.5.11)

PR #2059, opened 2022-06-08
Add API Key settings page context

codemist:
# New feature description
This adds some context to the API key on the settings page.

l10n: https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/88/files

# Screenshot (if applicable)
image

# How to test

# Checklist

### DIFF
--- a/frontend/src/pages/accounts/settings.module.scss
+++ b/frontend/src/pages/accounts/settings.module.scss
@@ -117,6 +117,7 @@ $field-gap: $spacing-lg;
 
 .copy-api-key-content {
   display: flex;
+  flex-wrap: wrap;
   flex-direction: row;
   gap: $spacing-md;
   align-items: center;
@@ -131,6 +132,22 @@ $field-gap: $spacing-lg;
   white-space: nowrap;
   overflow: hidden;
   font-family: monospace;
+  max-width: 200px;
+  text-overflow: ellipsis;
+
+  @media screen and #{$mq-md} {
+    max-width: 350px;
+  }
+}
+
+.settings-api-key-wrapper {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+}
+
+.settings-api-key-description {
+  width: 100%;
 }
 
 .copy-button-wrapper {
@@ -141,6 +158,7 @@ $field-gap: $spacing-lg;
 
   .copy-button {
     display: block;
+    padding-left: $spacing-md;
     appearance: none;
     border: 0;
     background-color: transparent;
--- a/frontend/src/pages/accounts/settings.module.scss
+++ b/frontend/src/pages/accounts/settings.module.scss
@@ -146,7 +146,7 @@ $field-gap: $spacing-lg;
   align-items: center;
 }
 
-.settings-api-key-description {
+.settings-api-key-copy {
   width: 100%;
 }
 
--- a/frontend/src/pages/accounts/settings.module.scss
+++ b/frontend/src/pages/accounts/settings.module.scss
@@ -144,6 +144,7 @@ $field-gap: $spacing-lg;
   display: flex;
   flex-wrap: wrap;
   align-items: center;
+  gap: $spacing-sm;
 }
 
 .settings-api-key-copy {
@@ -158,7 +159,6 @@ $field-gap: $spacing-lg;
 
   .copy-button {
     display: block;
-    padding-left: $spacing-md;
     appearance: none;
     border: 0;
     background-color: transparent;
--- a/frontend/src/pages/accounts/settings.page.tsx
+++ b/frontend/src/pages/accounts/settings.page.tsx
@@ -1,4 +1,4 @@
-import { useLocalization } from "@fluent/react";
+import { Localized, useLocalization } from "@fluent/react";
 import type { NextPage } from "next";
 import {
   FormEventHandler,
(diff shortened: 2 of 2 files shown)


frontend/src/pages/accounts/settings.module.scss
```diff
@@ -131,6 +132,22 @@ $field-gap: $spacing-lg;
   white-space: nowrap;
   overflow: hidden;
   font-family: monospace;
+  max-width: 200px;
```
codemist, line 135: Truncate the api key on smaller device sizes.

frontend/src/pages/accounts/settings.module.scss
```diff
@@ -141,6 +158,7 @@ $field-gap: $spacing-lg;
 
   .copy-button {
     display: block;
+    padding-left: $spacing-md;
```
Vinnl, line 161: Could you instead add this as a `gap` on `.settings-api-key-wrapper`? To have the parent determine the layout of its children, rather than children assuming the elements they're being used in.

frontend/src/pages/accounts/settings.page.tsx
```diff
@@ -183,38 +183,47 @@ const Settings: NextPage = () => {
...
+                    <div className={styles["settings-api-key-wrapper"]}>
+                      <input
+                        id="api-key"
+                        ref={apiKeyElementRef}
+                        className={styles["copy-api-key-display"]}
+                        value={profile.api_token}
+                        size={profile.api_token.length}
+                        readOnly={true}
+                      />
+                      <span className={styles["copy-controls"]}>
```
Vinnl, line 195: Heh, I think this is a leftover of my own code, but I think this `<span>` is actually redundant, is it not? (Since the class isn't used, and it just contains a single element.) In case it's easy for you to check, but never mind if it's more than 10s work :)

frontend/src/pages/accounts/settings.page.tsx
```diff
@@ -183,38 +183,47 @@ const Settings: NextPage = () => {
...
                       </span>
-                    </span>
+                    </div>
+                    <div className={styles["settings-api-key-copy"]}>
+                      {l10n.getString("settings-api-key-description")}
+                      <b>
+                        {" "}
+                        {l10n.getString("settings-api-key-description-bolded")}
+                      </b>
+                    </div>
```
Vinnl, line 226: I think this probably isn't problematic, but you may want to check with Peying whether she expects the order of these two sentences to change in any language.

If so, you can instead allow the localisers to use the `<b>` tag like this:

```suggestion
                    <Localized id="settings-api-key-description" elems={{ b: <b/> }}>
                      <div className={styles["settings-api-key-copy"]} />
                    </Localized>
```

(see other instances of `<Localized>` for an example.)

Vinnl (approved):
Looking good! Added the non-blocking (I think - the last one might be) suggestions.

frontend/src/pages/accounts/settings.page.tsx
Vinnl: Yep, I'll definitely defer to her then!

Vinnl: Oh, although I see that she just said that it should be in bold, not that it should be separated...

frontend/src/pages/accounts/settings.page.tsx
```diff
@@ -183,38 +183,49 @@ const Settings: NextPage = () => {
...
-                    </span>
+                    </div>
+                    <div className={styles["settings-api-key-copy"]}>
+                      {l10n.getString("settings-api-key-description")}{" "}
+                      <Localized
+                        id="settings-api-key-description-bolded"
+                        elems={{ b: <b /> }}
+                      >
+                        <span className={styles["settings-api-key-copy"]} />
+                      </Localized>
```
codemist, line 227: @Vinnl Did I do this right? It doesn't show bolded for me:

image

Vinnl: @codemist Yes, that looks right, but the localiser can then use `<b>` in their strings - so you'll have to update `pendingTranslations.ttl` (and the l10n PR) as well.

frontend/src/pages/accounts/settings.page.tsx
codemist: Peiying confirmed that this should be fine.

frontend/src/pages/accounts/settings.page.tsx
codemist: @Vinnl Andrei confirmed that he still isn't seeing the bolded string, is updating `pendingTranslations` necessary on stage?

Vinnl: @codemist Hmm, I see this in `app.ftl`:

```
settings-api-key-description = This is your { -brand-name-relay } API Key. It allows other apps to automatically access your { -brand-name-relay } account.
settings-api-key-description-bolded = Only use this with apps that you trust to access your { -brand-name-relay } account for you.
```

I'd have expected something like this?

```
settings-api-key-description-bolded = This is your { -brand-name-relay } API Key. It allows other apps to automatically access your { -brand-name-relay } account. <b>Only use this with apps that you trust to access your { -brand-name-relay } account for you.<b/>
```

(And no, if the string is already in the l10n repository, that should override whatever is in `pendingTranslations.ftl`, so updating the latter shouldn't be necessary.)

PR #2066, opened 2022-06-10
Clarify a workaround

Vinnl:
Since the workaround appears not to be needed, and nothing appears
broken when it's removed, I've added a comment that clarifies how
to reproduce the error, so future readers know why the workaround
is there and when it can be removed.

This is a followup to https://github.com/mozilla/fx-private-relay/pull/2036#discussion_r893270078, since it does appear to fix the issue (at least for the Bento icon - the mobile menu icons are still the wrong colour).


maxxcrawford (approved):
Comments for future travelers always.

PR #2086, opened 2022-06-16
Validate custom addresses in the UI

Vinnl:
This is the front-end counterpart to https://github.com/mozilla/fx-private-relay/pull/2085. It uses the browser's native validation messages to report about invalid characters.

How to test: [in the custom mask dialog](https://deploy-preview-2086--fx-relay-demo.netlify.app/?mockId=full), try to enter a space, a capital letter, a hyphen at the start or end, or any other non-alphanumeric characters. You should see a warning message (with a custom one for spaces).

- [ ] l10n changes have been submitted to the l10n repository, if any. - Not yet, will do at the same time as tracker removal messages. But even untranslated, this will be an improvement over the current situation, and people are probably unlikely to try these anyway.

### DIFF
--- a/frontend/src/components/dashboard/aliases/AddressPickerModal.test.tsx
+++ b/frontend/src/components/dashboard/aliases/AddressPickerModal.test.tsx
@@ -1,0 +1,85 @@
+import { getAddressValidationMessage } from "./AddressPickerModal";
+
+describe("getAddressValidationMessage", () => {
+  it("returns `null` for valid addresses", () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const mockL10n = { getString: jest.fn() } as any;
+    expect(getAddressValidationMessage("valid-address", mockL10n)).toBeNull();
+  });
+
+  it("returns `null` for valid single-character addresses", () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const mockL10n = { getString: jest.fn() } as any;
+    expect(getAddressValidationMessage("a", mockL10n)).toBeNull();
+  });
+
+  it("returns a message specific to spaces when the address includes a space", () => {
+    const mockL10n = {
+      getString: jest.fn((id: string) => `String ID: ${id}`),
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any;
+    expect(getAddressValidationMessage("invalid address", mockL10n)).toBe(
+      "String ID: modal-custom-alias-picker-form-prefix-spaces-warning"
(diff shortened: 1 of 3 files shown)


frontend/src/components/dashboard/aliases/AddressPickerModal.tsx
```diff
@@ -1,12 +1,14 @@
...
+  FocusEventHandler,
   FormEventHandler,
   ReactElement,
   ReactNode,
   useRef,
   useState,
 } from "react";
 import Link from "next/link";
-import { useLocalization } from "@fluent/react";
+import { ReactLocalization, useLocalization } from "@fluent/react";
```
groovecoder, line 11: question (non-blocking): Am I understanding correctly that we import the `ReactLocalization` class to use in a function rather than using the `useLocalization` hook like we do in the component?

If so, that makes me think that `getAddressValidationMessage` function belongs in `frontend/src/functions` rather than here in `frontend/src/components`?

groovecoder (approved):

frontend/src/components/dashboard/aliases/AddressPickerModal.test.tsx
```diff
@@ -0,0 +1,85 @@
+import { getAddressValidationMessage } from "./AddressPickerModal";
+
+describe("getAddressValidationMessage", () => {
+  it("returns `null` for valid addresses", () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const mockL10n = { getString: jest.fn() } as any;
+    expect(getAddressValidationMessage("valid-address", mockL10n)).toBeNull();
+  });
+
+  it("returns `null` for valid single-character addresses", () => {
```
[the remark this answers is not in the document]
Vinnl: Hehe, I added this test because I initially made a mistake in the regex that disallowed valid single-character addresses, so this is to avoid a regression of that mistake in case someone in the future things they can simply the regex :)

frontend/src/components/dashboard/aliases/AddressPickerModal.tsx
Vinnl: There's a small nuance there in that it's technically not a class I'm importing, but that class's type (i.e. you won't see it in the final JS bundle), but yes, it's used to indicate that the function expects an instance of that class because hooks only work in components.

So yeah, there's certainly something to be said for moving it into functions. The reason I kept it in this file is that I didn't expect the function to be re-used in different files; the only reason I extracted it into a separate function in the first place is so that I could isolate the interesting behaviour for a unit test.

Issue #2045, opened 2022-06-06
Premium user has issues setting masks to the “Promotionals” status


AOiegas:
#### Build:
- Stage;

#### Affected Platforms:
- Windows 10;

#### Browsers tested on:
- Firefox, Release version;
- Chrome, Release version;

#### Prerequisites:
- Be signed in into a Premium Firefox Relay account;
- Have a Subdomain set;

#### Steps to reproduce:
1. Go to the Relay dashboard;
2. Generate a new Random mask and a new Custom Subdomain mask;
3. In the open cards of the two freshly generated masks, switch their statuses to “Promotionals”;
4. Continue switching between the statuses;
5. Observe the results;

#### Expected result:
- The masks have no issues switching through the statuses;

#### Actual result:
- Sometimes the masks do not switch to the “Promotionals” status;

#### Notes:
- This issue is constant, but sometimes switching to the “Promotionals” status successfully. From what I’ve seen this may happen when switching faster than usual;
- See attached media:
![switch](https://user-images.githubusercontent.com/85748693/172121202-934272e1-0b61-4c69-83c7-ab76ce261954.gif)

PR #2090, opened 2022-06-18
#2045 ⁃ Hamburger menu is not fixed on the page on mobile

lloan:
This PR fixes MPP-2045

# New feature description
This makes it so the mobile menu follows you when in mobile menu (under 768px actually)

# Screenshot (if applicable)

image

# How to test
1. [Visit any page](https://deploy-preview-2090--fx-relay-demo.netlify.app/accounts/profile/) in mobile mode 
2. Scroll/Drag the page - menu should follow you.

# Checklist

### DIFF
--- a/frontend/src/components/layout/Layout.module.scss
+++ b/frontend/src/components/layout/Layout.module.scss
@@ -2,6 +2,8 @@
 @import "~@mozilla-protocol/core/protocol/css/includes/lib";
 
 .wrapper {
+  // Mobile menu is fixed so this padding offsets that spacing
+  padding-top: 75px;
   display: flex;
   flex-direction: column;
   height: 100%;
@@ -22,6 +24,11 @@
 
   --toastify-toast-min-height: $layout-md;
   --toastify-toast-max-height: $layout-md;
+
+  @media screen and #{$mq-md} {
+    // We set padding to zero as we set position for menu to sticky for non-mobile users.
+    padding-top: 0;
+  }
 }
 
 // Allows us to dynamically set the background color based on Toast container state
@@ -141,7 +148,7 @@ $toastify-toast-close-button: var(--toast-close-button);
 }
 
 .header {
-  position: sticky;
+  position: fixed;
   left: 0;
   top: 0;
   // The sticky position causes a new stacking context,
@@ -161,6 +168,7 @@ $toastify-toast-close-button: var(--toast-close-button);
   box-shadow: 0px 4px 12px rgba(0, 0, 0, 0.06);
 
   @media screen and #{$mq-md} {
+    position: sticky;
     grid-template-areas: "logoWrapper navWrapper appPickerWrapper userMenuWrapper";
   }
 
--- a/frontend/src/components/layout/Layout.module.scss
+++ b/frontend/src/components/layout/Layout.module.scss
@@ -2,7 +2,7 @@
 @import "~@mozilla-protocol/core/protocol/css/includes/lib";
 
 .wrapper {
-  // Mobile menu is fixed so this padding offsets that spacing
+  // Mobile menu is fixed. This padding offsets the height of the menu and pushes content down (so its not hidden).
   padding-top: 75px;
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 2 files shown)


Vinnl (commented):
Hmm, it doesn't look quite right for me at the moment; the menu appears from underneath the header when collapsed, and has a gap when it's not:

https://user-images.githubusercontent.com/4251/174564136-251c3d91-0416-4be8-80ab-c76688f5d58d.mp4

This might be due to the mock warning, but then I'm worried about seeing the same when e.g. showing the CSAT survey at the top of the page...

lloan:
> Hmm, it doesn't look quite right for me at the moment; the menu appears from underneath the header when collapsed, and has a gap when it's not:
> 
>  Kooha-06-20-2022-10-52-44.mp4 
> This might be due to the mock warning, but then I'm worried about seeing the same when e.g. showing the CSAT survey at the top of the page...

Yeah, its that offset. That's going to be fun. Well, back to the drawing board. Thanks @Vinnl !

lloan:
@Vinnl You can use [this ](https://deploy-preview-2090--fx-relay-demo.netlify.app/accounts/profile/)link to test.

frontend/src/components/layout/navigation/MobileNavigation.module.scss
```diff
@@ -3,7 +3,9 @@
 
 .mobile-menu {
   width: 100%;
-  position: relative;
+  position: sticky;
+  top: 76px; // threshold for mobile menu - height of header
```
Vinnl, line 7: One worry I have with this is that this "happens" to be the width of the header, but we're never explicitly setting it to this value. In other words, if there's anything that might affect the height of the header, the menu will suddenly be out of place.

[Reading up on sticky positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky) a bit, I'm wondering if a different solution might be possible - one that potentially also fixes the issue with the `height: 100%` on Layout's `.wrapper`. Specifically:

> Note that a sticky element "sticks" to its nearest ancestor that has a "scrolling mechanism" (created when overflow is hidden, scroll, auto, or overlay), even if that ancestor isn't the nearest actually scrolling ancestor.

So, given that we don't want the menu to be positioned below the header, but also to overlap the content, I'm wondering if it might work if in `<Layout>`, we wrap `<MobileNavigation>`, `.content` and `footer.footer` (or possibly not even the latter) in a div with `height: 100%; overflow: auto`? As I understand it, that div will then always be below the header (so without needing `top` with a magic number), while `<MobileNavigation>` will be sticky to the top of that div? Not sure if it will work, but might be worth a try.

frontend/src/components/layout/navigation/MobileNavigation.module.scss
```diff
@@ -3,7 +3,9 @@
...
+  height: 0; // prevents element from pushing main elements down
```
Vinnl, line 8: This is a good comment; it answers something I'd definitely wonder about if it wasn't there :)

Vinnl (commented):
Ugh, positioning is such a pain to get right 😣 I dropped one idea for a potential alternative strategy in here, and otherwise it's up to Eduardo to decide whether this is worth a regression on other pages.

frontend/src/components/layout/Layout.module.scss
```diff
@@ -4,7 +4,7 @@
 .wrapper {
   display: flex;
   flex-direction: column;
-  height: 100%;
```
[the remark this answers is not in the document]
lloan: I think we have to solve it to look as it currently does. It would be silly to leave it as is - I did not know there was a dependency here. Thanks for pointing that out. I'll take a look at the suggestion below and see if that affects any other parts of the site (if it works).

frontend/src/components/layout/navigation/MobileNavigation.module.scss
lloan: Taking another look to see if we can solve this issue - thanks!

lloan:
> I read through @Vinnl's notes. I agree/+1 to all of his comments. The less strict/more flexible code, the better, when it comes to setting header heights, etc. I would prefer to avoid any magic numbers (`76px`) where we can.

Yeah, I've tried a couple of approaches, including his - not working yet. Going to try one more approach before I call the front-end team to assemble, 😆.

lloan:
Closing in favor of #2177

PR #2078, opened 2022-06-14
Add profile field for email tracker behavior

say-yawn:
# New feature description
`remove_level_one_email_trackers` default to false, but when set to true relayed emails will be stripped of level one email trackers

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist

- [ ] l10n dependencies have been merged, if any.
- [ ] All acceptance criteria are met.
- [ ] I've added or updated relevant docs in the docs/ directory.
- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/static/scss/libs/protocol/css/includes/tokens/dist/index.scss`).
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #2110, opened 2022-06-22
MPP-2060: Allow new field to be null

jwhitlock:
New fields need to allow `null` so that existing code will continue to work during a deployment. Otherwise, when deploying the new version, any operations that write to the Profile table will fail with an integrity error.

Once the new code is deployed, we can add two new migrations:

1) A data migration to set any `null` values to `False`
2) A schema migration to disallow `null`.

This is follow-on work for merged PR #2078


say-yawn (approved):
LGTM. Would like to look into ways to not manually create migrations for unsetting the `null=True` but the TODO does a great job to remind the changes should happen and this will prevent errors like [this Sentry events](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/19494922/) from happening during duration.

PR #2116, opened 2022-06-23
MPP-2104 Put tracker removal UI behind a flag.

lloan:
This PR fixes #2115   
 

# New feature description
We place the UI for tracker removal setting behind a flag using new `flagIsActive` function (kudos to @groovecoder) 

# Screenshot (if applicable)

Not applicable.

# How to test

Toggling `tracker_removal` flag on and off dictates whether this UI is rendered or not.  

# Checklist

- [ ] l10n dependencies have been merged, if any.
- [ ] All acceptance criteria are met.
- [ ] I've added or updated relevant docs in the docs/ directory.
- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/static/scss/libs/protocol/css/includes/tokens/dist/index.scss`).
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #2117, opened 2022-06-23
Add flag logic to Tracker Removal item in Whats New list

maxxcrawford:
Related to #2116

### DIFF
--- a/frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
+++ b/frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
@@ -310,10 +310,7 @@ export const WhatsNewMenu = (props: Props) => {
     },
   };
   // Only show its announcement if tracker removal is live:
-  if (
-    typeof props.profile.remove_level_one_email_trackers === "boolean" &&
-    flagIsActive(runtimeData.data, "tracker_removal")
-  ) {
+  if (flagIsActive(runtimeData.data, "tracker_removal")) {
     entries.push(trackerRemoval);
   }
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
```diff
@@ -307,7 +310,10 @@ export const WhatsNewMenu = (props: Props) => {
     },
   };
   // Only show its announcement if tracker removal is live:
-  if (typeof props.profile.remove_level_one_email_trackers === "boolean") {
+  if (
+    typeof props.profile.remove_level_one_email_trackers === "boolean" &&
```
say-yawn, line 314:
```suggestion
```
We do not need to check the profile to see if user has enabled remove email tracker since this is news.

say-yawn (changes requested):
Change requested

frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
maxxcrawford: Updated.

say-yawn (approved):
Code makes sense, local env is not working so testing in Stage.

frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
Vinnl: This wasn't a check to see whether the user has it enabled (it doesn't check for true or false), but just if the back-end provides the value in general (i.e. whether it's of type `boolean` rather than `undefined`) :slightly_smiling_face:

PR #2120, opened 2022-06-24
Revise flagIsActive arguments to use runtimeData that is already included in the component

maxxcrawford:


Vinnl (approved):
Nailed it!

lloan:
@Vinnl @maxxcrawford does this one get taken care of by #2136 ?

frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
```diff
@@ -310,7 +308,7 @@ export const WhatsNewMenu = (props: Props) => {
     },
   };
   // Only show its announcement if tracker removal is live:
-  if (flagIsActive(runtimeData.data, "tracker_removal")) {
+  if (flagIsActive(props.runtimeData, "tracker_removal")) {
```
Vinnl, line 311: Oh, good one @lloan! It doesn't, but @maxxcrawford, you'll get a conflict here, which you'll want to resolve to this:

```suggestion
  if (flagIsActive(props.runtimeData, "tracker_removal")) {
```
```suggestion
  if (isFlagActive(props.runtimeData, "tracker_removal")) {
```

PR #1987, opened 2022-05-23
MPP-2008: Drop second checkout from heroku postbuild

jwhitlock:
Heroku automatically checks out the submodule as one of the first steps of building after a new push, so the removed command fails every time.>

This PR fixes #1984 / MPP-2008

How to test:

* See a recent build, verify that it works like #1984.
* If needed, push this branch to Heroku, and see that the build is successful.

PR #2140, opened 2022-06-29
Fix Heroku build

jwhitlock:
The Heroku build was broken after removing the gulp tools. This fixes the Heroku build by allowing the `postinstall` script to continue when dev dependency `msw` is removed (it runs successfully during the `npm ci` step).

This PR has further Heroku changes:

* Skip adding the locales submodule, since Heroku takes care of installing it
* Skip installing `husky` in Heroku, since the source code does not include a `git` repo, required for husky
* Add a `version.json` file, so that https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/__version__ works (and so we can add it to https://whatsdeployed.io/s/xLS/mozilla/fx-private-relay)

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -15,7 +15,7 @@
     "licensecheck": "license-checker --excludePrivatePackages --production --onlyAllow 'MPL; MIT; BSD-3-Clause; BSD-2-Clause; 0BSD; Apache-2.0; ISC; Zlib; CC0; Unlicense; WTFPL; Unicode-DFS-2016;'",
     "test": "jest",
     "postinstall": "msw init public/ --save",
-    "prepare": "cd .. && husky install frontend/.husky"
+    "prepare": "if [ \"$ON_HEROKU\" != \"True\" ]; then cd .. && husky install frontend/.husky; fi"
   },
   "dependencies": {
     "@fluent/bundle": "^0.17.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -24,7 +24,8 @@
     "frontend"
   ],
   "scripts": {
-    "heroku-postbuild": "git clone https://github.com/mozilla-l10n/fx-private-relay-l10n.git privaterelay/locales; cd frontend; NODE_ENV=\"development\" npm ci; npm run build"
```
Vinnl, line 27: Huh, I don't recall resolving a conflict here, but looks like Git reverted https://github.com/mozilla/fx-private-relay/pull/1987 when I rebased https://github.com/mozilla/fx-private-relay/pull/1891 :/ I'll chalk that up under the downsides of having PRs open for too long....

frontend/package.json
```diff
@@ -14,8 +14,8 @@
     "lint": "stylelint '**/*.scss' && prettier --check './src' && next lint",
     "licensecheck": "license-checker --excludePrivatePackages --production --onlyAllow 'MPL; MIT; BSD-3-Clause; BSD-2-Clause; 0BSD; Apache-2.0; ISC; Zlib; CC0; Unlicense; WTFPL; Unicode-DFS-2016;'",
     "test": "jest",
-    "postinstall": "msw init public/ --save",
-    "prepare": "cd .. && husky install frontend/.husky"
+    "postinstall": "if type msw > /dev/null; then msw init public/ --save; else echo 'msw not installed'; fi",
```
Vinnl, line 17: Hmm, I don't understand why this would be necessary, as msw is just listed as a dependency :/ Yes, it's just a devDependency, but the `NODE_ENV=\"development\"` in the root `package.json`'s `heroku-postbuild` script [should take care of that](https://devcenter.heroku.com/articles/nodejs-support#skip-pruning). And I especially don't understand why this would be necessary _now_, i.e. how would the removal of Gulp affect this?

But msw shouldn't be necessary on dev anyway, so if this makes it work, that's good enough I guess :shrug:

Vinnl (approved):

PR #2177, opened 2022-07-09
Make the mobile menu stick when scrolling

### LINKED ISSUE #2047: Hamburger menu is not fixed on the page on mobile
**Build:**
Stage

**Affected platforms:**
Android 12
iOS 15.4

**Browsers tested on:**
Firefox, latest release

**Preconditions:**
There are some masks already in the list.

**Steps to reproduce:**
1. Go to Relay homepage and log in using a free/ premium account.
2. Scroll down the page and then tap on the Hamburger menu in the upper right hand corner.
3. Observe the menu.

**Actual result:**
The Hamburger menu is not visible/ fully visible when user is lower in the page. The user needs to scroll to be able to see it.

**Expected result:**
The Hamburger menu is fixed.

**Notes:**
Also when you scroll to the bottom of the page the Firefox Relay header is not visible anymore.

https://user-images.githubusercontent.com/79697848/172144679-531c9e76-832b-454d-85c6-9615b7752fc4.mp4

https://user-images.githubusercontent.com/79697848/172145089-ce2b46a0-e0ce-4795-945a-0705996dc09b.mp4

Vinnl:
This PR fixes #2047 (and is an alternate approach to / closes #2090). It makes the expanded mobile menu stick to the top of the screen even when scrolling, without [leaving white space at the bottom of short pages](https://github.com/mozilla/fx-private-relay/pull/2090#discussion_r909584056). Of course, there might be other layout effects I didn't think of, so be sure to try everything you can see that might break :)

The trick here is that elements with `position: sticky`

>  "sticks" to its nearest ancestor that has a "scrolling mechanism" (created when overflow is hidden, scroll, auto, or overlay), even if that ancestor isn't the nearest actually scrolling ancestor.

https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky

Thus, I added a wrapper that contained everything but the things the menu should be positioned below (i.e. the header bar, mock data notification, CSAT survey, notifications) but included the mobile menu, and gave that an `overflow: auto`. Then I could give the mobile menu `position: sticky`, and that appears to have worked.

https://user-images.githubusercontent.com/4251/178084295-c2f702dc-4587-466b-a7e8-1d0137b32ccd.mp4

How to test: on a small screen, expand the mobile menu and scroll down. The menu should remain in view. Also, no other layout things should break if it wasn't broken before - a good stress test won't hurt :)

### DIFF
--- a/frontend/src/components/layout/Layout.module.scss
+++ b/frontend/src/components/layout/Layout.module.scss
@@ -242,6 +242,23 @@ $toastify-toast-close-button: var(--toast-close-button);
   }
 }
 
+.non-header-wrapper {
+  // This `overflow` ensures that the `position: sticky` on `.mobile-menu`
+  // in <MobileNavigation> is sticky relative to this wrapper, i.e. will always
+  // stay below the header bar.
+  // For more info, see
+  // https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky,
+  // and specifically:
+  // > Note that a sticky element "sticks" to its nearest ancestor that has a
+  // > "scrolling mechanism" (created when overflow is hidden, scroll, auto, or
+  // > overlay), even if that ancestor isn't the nearest actually scrolling
+  // > ancestor.
+  overflow: auto;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+}
+
 .content {
   flex-grow: 2;
   display: flex;
--- a/frontend/src/components/layout/Layout.tsx
+++ b/frontend/src/components/layout/Layout.tsx
@@ -138,14 +138,6 @@ export const Layout = (props: Props) => {
           </div>
         </header>
 
-        <MobileNavigation
-          mobileMenuExpanded={mobileMenuExpanded}
-          hasPremium={hasPremium}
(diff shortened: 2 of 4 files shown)


lloan (changes requested):
It looks like it causes an issue here (dashboard page on mobile): 

image

Vinnl:
Ah thanks, there we go :) Looking into it now, some notes in case I get interrupted:

- It only happens in Chrome (or at least, doesn't happen in Firefox).
- It only happens when the Tips area is visible. There's some funky positioning going on there too, so it's likely that that's related.
- It also happens on wider screens.

Vinnl:
Thanks for the keen eye @lloan! Fixed that in [84944d1](https://github.com/mozilla/fx-private-relay/pull/2177/commits/84944d1b9b872b30e58de27d9fb5145cb84ab446), and also added [026157b](https://github.com/mozilla/fx-private-relay/pull/2177/commits/026157b85d9d537d32545e5dbea712eb3b5605f2) because I realised that the menu not being scrollable might cause issues if the entire menu doesn't fit on the screen.

lloan (approved):
Tested and it looks like it works!

PR #2219, opened 2022-07-19
use PHONES_ENABLED env var to better separate phones app

groovecoder:
The dev server is throwing a `RuntimeError` because `api.urls` is importing `api.views` which is importing `phones.models` while the phones app is disabled.

This PR:
1. Adds a new single `PHONES_ENABLED` env var that can be set independent of `TWILIO_*` env vars
2. Moves phones API code into a number of specific modules:
   * `api.serializers.phones`
   * `api.views.phones`
   * `api.tests.phones_views_tests`
3. Uses the changes in 1. and 2. to make a cleaner on/off state for the phones API in the `api.urls` code

How to test:
1. Set `PHONES_ENABLED=False`
2. Run `pytest`
   * [x] Should pass fine (all phone tests should skip)
3. Check out http://127.0.0.1:8000/api/v1/docs/
   * [x] You should NOT see any of the phone APIs
4. Spot check http://127.0.0.1:8000/
   * [x] Everything should work as expected like it does now
5. Set `PHONES_ENABLED=True`
6. Run `pytest`
   * [x] Should pass fine (all phone tests should run)
7. Check out http://127.0.0.1:8000/api/v1/docs/
   * [x] You SHOULD see the phone APIs
8. Spot check http://127.0.0.1:8000/
   * [x] Everything should work as expected like it does now


api/views/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+            "created_at",
+            "last_modified_at",
+            "last_used_at",
+            "num_forwarded",
+            "num_blocked",
+            "num_spam",
+        ]
+
+
+class RelayAddressViewSet(SaveToRequestUser, viewsets.ModelViewSet):
```
jwhitlock, line 86: Nit / Future Work: I'd prefer this stuff in `api/views/emails.py` rather than `api/views/__init__.py`

groovecoder: Yeah, I want to come back thru and add a bunch of docs to the other API endpoints too like I did with the phones endpoints.

PR #2171, opened 2022-07-06
Feature: Phone Masking (Frontend Onboarding MVP) 

maxxcrawford:
Based/blocked by #2026 

## Summary

This feature adds the MVP on-boarding for phone masking in Relay. Uses cans verify their true phone number and register a suggested Relay number. 

TODO: 
- [ ] Add lint/error states for phone inputs
- [ ] Add confirmation modal on final step when selecting Relay number
- [ ] Move translations to L10N repo
- [ ] Remove/account for current TODOs

## Testing: 

TBD

PR #2256, opened 2022-07-27
Add API mocks for the /realphone/ API

Vinnl:
# New feature description

To familiarise myself with the phones API, and to make it easier for me to test if I don't manage to set up the full end-to-end phone infra, I thought I'd initialise the phone API mocks.

I've set them up so that the `onboarding` user has just purchased `phones`, but has otherwise not set it up yet. Any verification code will be accepted, and as soon as you submit a real number, it will pretend the verification code has been sent 4:40 m ago (so if you want to test what happens when it expires, you'll only need to wait 20 seconds).

The `some` and `full` users will just have one confirmed real number.

Additionally, I enabled the flag in the mocked `runtimeData`.

# Screenshot (if applicable)

Not applicable.

# How to test

Run `npm run dev:mocked`, and test the phone UI with the `empty` user (should see prompt to upgrade), the `onboarding` user (should be able to go through the in-progress onboarding flow) and the `full` user (should see the TODO for the phone dashboard).

# Checklist

### DIFF
--- a/frontend/src/apiMocks/handlers.ts
+++ b/frontend/src/apiMocks/handlers.ts
@@ -1,5 +1,6 @@
 import { rest, RestHandler, RestRequest } from "msw";
 import { CustomAliasData, RandomAliasData } from "../hooks/api/aliases";
+import { UnverifiedPhone, VerifiedPhone } from "../hooks/api/phone";
 import { ProfileData } from "../hooks/api/profile";
 import {
   mockIds,
@@ -8,6 +9,7 @@ import {
   mockedRelayaddresses,
   mockedRuntimeData,
   mockedUsers,
+  mockedRealphones,
 } from "./mockData";
 
 export function getHandlers(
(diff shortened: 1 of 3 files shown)


frontend/src/apiMocks/mockData.ts
```diff
@@ -228,3 +236,28 @@ export const domainaddresses: Record<
     },
   ],
 };
+
+export const mockedRealphones: Record<typeof mockIds[number], RealPhoneData> = {
+  empty: [],
+  onboarding: [],
+  some: [
+    {
+      id: 0,
```
groovecoder, line 245: `id`s should never be 0.

frontend/src/apiMocks/handlers.ts
```diff
@@ -285,6 +287,94 @@ export function getHandlers(
...
+
+  addGetHandler("/api/v1/realphone/", (req, res, ctx) => {
+    const mockId = getMockId(req);
+    if (mockId === null) {
+      return res(ctx.status(400));
+    }
+
+    return res(ctx.status(200), ctx.json(mockedRealphones[mockId]));
+  });
+  addPostHandler("/api/v1/realphone/", (req, res, ctx) => {
```
groovecoder, line 299: Both you and @maxxcrawford - would this code be a lot simpler if I removed the ability to `POST` the verification code? So the only kind of request that would be a "verification" request is a `PATCH`?

groovecoder (changes requested):

frontend/src/apiMocks/mockData.ts
Vinnl: Is that problematic in mocks? I'm not particularly opposed to changing it, but it does complicate the `POST` mocks a bit (right now I can just use the number of existing mocks as the new ID).

(Actually, I'd be interested in learning why that's problematic in actual back-ends too, but only if you have time :smile: )

frontend/src/apiMocks/handlers.ts
Vinnl: I think indeed that having just one way to verify phone numbers would be simpler, _but_ IIRC Maxx's current code is using `POST` to verify. It's not necessarily much work to change, but given that there's already quite a bit on his plate, maybe we should keep that as a future enhancement.

Vinnl:
Rebased to align with the rebasing that happened in #2171.

frontend/src/apiMocks/mockData.ts
groovecoder: Oh ... it's not a technical problem - more of a "technically impossible" problem. It should never BE ABLE to be 0, since it's an auto-increment integer id field. So less of a "don't do this" and more of a "the DB will never do this."

groovecoder (approved):
`0` in mock is fine. Also approved to merge with `POST` for verifying the number, unless Maxx wants to change it to `PATCH`.

frontend/src/apiMocks/handlers.ts
maxxcrawford: Heh. I'm using [them](https://github.com/mozilla/fx-private-relay/blob/start-phone-api-react/frontend/src/hooks/api/phone.ts#L140) [both](https://github.com/mozilla/fx-private-relay/blob/start-phone-api-react/frontend/src/hooks/api/phone.ts#L124). 😅

frontend/src/hooks/api/phone.ts
```diff
@@ -109,8 +109,9 @@ export function useRealPhonesData(): SWRResponse<RealPhoneData, unknown> & {
   // const realphone: SWRResponse<RealPhoneData, unknown> =
   //   useApiV1("/realphone/");
 
-  const realphone: SWRResponse<RealPhoneData, unknown> = useApiV1("/realphone");
-  //     "/realphone",
+  const realphone: SWRResponse<RealPhoneData, unknown> =
```
maxxcrawford, line 112: nit: white space change? Is this necessary for this PR?

maxxcrawford (approved):
I added one nit:question! Thanks for doing this. I walked through it and have a better grasp of mockData now. 

@Vinnl Would you have done this part (`mockData.ts`) before jumping into the actual UI usage (`PhoneOnboarding.tsx`)?

frontend/src/hooks/api/phone.ts
Vinnl: I think the added `/` (which was necessary in that I also used the `/` in the mocks, because all other mocks/API hooks do so as well - the other option would've been to remove it or make it optional there) pushed it over Prettier's line length limit, and therefore introduced the whitespace change.

Vinnl:
@maxxcrawford I probably would've implemented every step against the real API first (to make sure I understand what it expects and returns), and implement the mock API after that, to be able to get to the next steps more quickly.

groovecoder:
Looks like this needs conflict resolution and front-end test fixes still?

Vinnl:
Yeah, there was another rebase in https://github.com/mozilla/fx-private-relay/pull/2171 since my update this morning. Will rebase again and merge in tomorrow. (The tests are already failing in the target branch, so those can be fixed there.)

PR #2232, opened 2022-07-21
Mpp 2062 return trackers found removed in email header context

say-yawn:
This PR fixes #<issue ID>.

How to test:

- [ ] l10n changes have been submitted to the l10n repository, if any.
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] I've added or updated relevant docs in the docs/ directory.
- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

# New feature description

# Screenshot (if applicable)

Not applicable.

# How to test

# Checklist

- [ ] l10n changes have been submitted to the l10n repository, if any.
- [ ] All acceptance criteria are met.
- [ ] I've added or updated relevant docs in the docs/ directory.
- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

Issue #470, opened 2020-06-23
New tab on upgrade — is the waitlist open?


hackel:
This has happened to me on 2 different machines now.  I have the Firefox Relay extension installed in anticipation of the public beta opening.  After the most recent automatic update, a `moz-extension://../first-run.html` tab automatically opens (annoying) with a link prompting me to "Join the waitlist."  Clicking on this link simply takes me to https://relay.firefox.com/, where there is another link: "Get on the waitlist with a Firefox account".  Clicking on this link prompts me to sign in to Firefox Private Relay, after which I just see the message "Were you invited to join the Private Relay beta? Visit the link in your email invitation to complete your registration. Be sure you are using the same email address you used to join the waitlist." This doesn't sound like I'm actually on any waitlist, but I can't be sure.

I have no problem waiting patiently for the waitlist/beta to open, but if it's not open, the extension shouldn't be opening a new tab automatically inviting people to join.  It's confusing.

Sim59121:
Hello,

I have exactly the same problem.
I have test with several devices.

Regard,

groovecoder:
The wait-list should be open for you to join now. Can you try again?

Sim59121:
I have retry and the problem is still present.

Other times, I log in and it takes me back to the Firefos Private Relay home page and asks me to log in to join the waiting list.

groovecoder:
Okay, we'll look into this today. Thanks for reporting.

Sim5912:
Thanks ! 

I remain at your disposal if you need me.

say-yawn:
Hi all, wait-list is open again. As for (1) stop displaying the first-run tab after it's first displayed and (2) making the experience better on enlisting on wait-list better from add-on, there are fixes coming later today.

hackel:
The first-run tab just opened again and I was able to follow all the links to join the waitlist successfully this time.

PR #2259, opened 2022-07-27
for MPP-2223: add RelayNumber.enabled field

groovecoder:
# New feature description
Adds a `PATCH /api/v1/relaynumber/{id}` endpoint to enable/disable the new `RelayNumber.enabled` field, which controls forwarding/blocking all calls and messages.

# Screenshot (if applicable)
Screen Shot 2022-07-27 at 5 30 19 PM

# How to test
1. Use the new endpoint at http://127.0.0.1:8000/api/v1/docs/ to set a relay number to `enabled: true`.
2. Send a text to the relay number
   * [x] It should be forwarded
3. Use the new endpoint at http://127.0.0.1:8000/api/v1/docs/ to set a relay number to `enabled: false`.
4. Send a text to the relay number
   * [x] It should NOT be forwarded

# Checklist


api/tests/phones_views_tests.py
```diff
@@ -326,11 +326,55 @@ def test_realphone_patch_invalid_verification_code(phone_user, mocked_twilio_cli
...
 
 
-def test_relaynumber_suggestions_bad_request_for_user_without_real_phone(phone_user):
-    real_phone = "+12223334444"
-    RealPhone.objects.create(user=phone_user, verified=True, number=real_phone)
+def test_relaynumber_post_with_existing_returns_error(phone_user, mocked_twilio_client):
+    number = "+12223334444"
     relay_number = "+19998887777"
-    RelayNumber.objects.create(user=phone_user, number=relay_number)
+    RealPhone.objects.create(user=phone_user, number=number, verified=True)
```
groovecoder, line 332: This change looks weird. Basically, I saw that these 2 tests (`test_relaynumber_suggestions_bad_request_for_user_without_real_phone` and `test_relaynumber_post_with_existing_returns_error`) had the wrong names. So I switched the names so the test names would match what the test is actually doing. But this is how git interpreted that change. 🤷

phones/models.py
```diff
@@ -155,13 +155,16 @@ class RelayNumber(models.Model):
     vcard_lookup_key = models.CharField(
         max_length=6, default=vcard_lookup_key_default, unique=True
     )
+    enabled = models.BooleanField(default=True)
```
jwhitlock, line 158: All new fields either need `nullable=True`, or we need to set a server default in the migration. The `nullable=True` has been a real pain, let's try to figure out the server default...

groovecoder: Hrm ... when I ran the migration on my local it didn't complain. Maybe I didn't have any existing records when I ran it? Or does django know enough that if `default=True` is in the field definition, it sets all existing records to `True` ?

jwhitlock: The problem is not running the migration, the problem is when we deploy it. The Kubernetes rollout preserves the number of web pods by replacing old ones with new ones, one at a time, so the old pods will run against the new database schema. You can replicate this with:

1. Check out this PR
2. Run migration 0017
3. Check out tag `v3.6.1` (or `main`, close enough)
4. Go through the steps to create a Relay number (register a real phone, setup a relay number)

Steps 4 is hard, but you'll end up where the old code is running:

```sql
INSERT INTO "phones_relaynumber"
   ("id", "number", "location", "user_id", "vcard_lookup_key")
   VALUES (123, '+1234567890', 'location', 456, 'acb123');
```

resulting in the error:

```
django.db.utils.IntegrityError: phones_relaynumber.enabled NOT NULL constraint failed
```

It was a little easier to see with PR #2232, where a integer column was added, and you could use the UI to fail to create a new alias.

So, we'll get database errors and data loss during the rollout. There's two ways to deal with this. One is to add `null=True`, the other is a database default.

The deploy sequence with `null=True`:

1. Deploy One:
    - A migration adds a column with `NULL` allowed.
    - Old code omits the new column on inserts and updates, and new rows get a `null` value.
    - The new code needs to expect a `null` value and interpret it as the default (in this case `True`).
2. Deploy Two:
    - A data migration replaces all `NULL` values with `True`, and a schema migration changes `NULL` to `NOT NULL`. 
    - The old code continues to set to `true` or `false`, and checks for `null` even though is won't happen.
    -  The new code can drop the `null` checks and assume only `true` or `false`.

The deploy sequence with a database default:

1. Deploy One:
   - A migration adds a column with a database default
   - Old code omits the new column on inserts and updates, and new rows get the database default
   - The new code gets either `true` or `false`, never `null`
2. Deploy Two:
   - A migration drops the database default
   - The old code and the new code are identical, always setting the column to `true` or `false`
   
The database default is a lot cleaner - the code doesn't change, and no mucking about with `null` values. The problem is that Django doesn't support database defaults ([issue #470](https://code.djangoproject.com/ticket/470)), so you can't do it without writing your own migrations. And _that_ is hard because we have to support both SQLite3 and PostgreSQL.

You can see what Django does with `./manage.py sqlmigrate phone 0017`:

```
BEGIN;
--
-- Add field enabled to relaynumber
--
CREATE TABLE "new__phones_relaynumber" ("id" integer NOT NULL PRIMARY KEY AUTOINCREMENT, "enabled" bool NOT NULL, "number" varchar(15) NOT NULL, "location" varchar(255) NOT NULL, "user_id" integer NOT NULL REFERENCES "auth_user" ("id") DEFERRABLE INITIALLY DEFERRED, "vcard_lookup_key" varchar(6) NOT NULL UNIQUE);
INSERT INTO "new__phones_relaynumber" ("id", "number", "location", "user_id", "vcard_lookup_key", "enabled") SELECT "id", "number", "location", "user_id", "vcard_lookup_key", 1 FROM "phones_relaynumber";
DROP TABLE "phones_relaynumber";
ALTER TABLE "new__phones_relaynumber" RENAME TO "phones_relaynumber";
CREATE INDEX "phones_relaynumber_number_742e5d6b" ON "phones_relaynumber" ("number");
CREATE INDEX "phones_relaynumber_user_id_62c65ede" ON "phones_relaynumber" ("user_id");
COMMIT;
```

The sequence is:
* Create a new table `new__phones_relaynumber` with the `enable` column
* Copy all the data from `phones_relaynumber`, setting the `enable` column to `1` (true)
* Drop `phones_relaynumber`
* Rename `new__phones_relaynumber` to `phones_relaynumber`
* Re-add the indexes

This is what the Django docs say for [sqlite3 migrations](https://docs.djangoproject.com/en/3.2/topics/migrations/#sqlite):

> SQLite has very little built-in schema alteration support, and so Django attempts to emulate it by:
>
> * Creating a new table with the new schema
> * Copying the data across
> * Dropping the old table
> * Renaming the new table to match the original name

I _think_ the SQL for PostgreSQL looks like:

```
ALTER TABLE "phones_relaynumber" ADD COLUMN "enabled" NOT NULL DEFAULT "true";
ALTER TABLE "phones_relaynumber" ALTER COLUMN "enabled" DROP DEFAULT;
```

but it is hard to get PostgreSQL running locally. I'll need to do that for this PR.

There's a note for [PostgreSQL](https://docs.djangoproject.com/en/3.2/topics/migrations/#postgresql):

> The only caveat is that prior to PostgreSQL 11, adding columns with default values causes a full rewrite of the table, for a time proportional to its size. For this reason, it’s recommended you always create new columns with null=True, as this way they will be added immediately.

We're running PostgreSQL 11.15.

api/serializers/phones.py
```diff
@@ -29,9 +29,12 @@ class RelayNumberSerializer(serializers.ModelSerializer):
     class Meta:
         model = RelayNumber
         fields = [
+            "id",
```
jwhitlock, line 32: Did you mean to expose the `id`?

api/views/phones.py
```diff
@@ -348,6 +364,8 @@ def inbound_sms(request):
         real_phone = RealPhone.objects.get(user=relay_number.user, verified=True)
     except ObjectDoesNotExist:
         raise exceptions.ValidationError("Could not find relay number.")
+    if not relay_number.enabled:
+        return response.Response(status=200, data={"message": "Number is not accepting messages."})
```
jwhitlock, line 368: Is a `200` the right response code?

phones/migrations/0017_relaynumber_enabled.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.14 on 2022-07-27 21:24
+
+from django.db import migrations, models
```
jwhitlock, line 3: I'm suggesting a larger change in PR #2261, to ensure that there is a database-level default for `enabled`. It probably is not a large issue for this change, since there are little or no phones being created in production, but I want us to get better about adding columns while the stakes are low.

jwhitlock (changes requested):
Thanks @groovecoder, I'm glad we're adding enabling / disabling this early in the feature!

Please run `black` on the codebase, and consider merging PR #2261 into this PR.

api/serializers/phones.py
jwhitlock: It looks like "yes", you need to know the ID so you can `PATCH` it later

api/views/phones.py
jwhitlock: It looks like 404 works as well, but a bigger issue might be that we're returning HTML rather than a JSON response. Twilio error logs are showing "12200 - Schema Validation Warning" even for correct responses. It looks like we're supposed to response with [Twilio Markup Language](https://www.twilio.com/docs/glossary/what-is-twilio-markup-language-twiml)

api/serializers/phones.py
groovecoder: Yeah, exactly - I'm returning the `id` so it can be sent in the `PATCH`.

api/views/phones.py
groovecoder: Oh dang, good catch! I worried that a `400` response code might cause Twilio to retry. But it looks like they don't do that? (FXA does, and sometimes AWS does too.)

I'll also fix this to return XML like the `inbound_call` does. Which reminds me - I need to add this `enabled` check to `inbound_call` too!!

phones/migrations/0017_relaynumber_enabled.py
groovecoder: praise: Thanks for doing that! I'll include it and now we can refer back to it for future columns.

api/views/phones.py
groovecoder: Okay, I added https://github.com/mozilla/fx-private-relay/pull/2259/commits/7cd0dfd780507e41094e3fbd5c84a6a63fd32921 to return 400's and valid TwiML.

groovecoder:
@jwhitlock this is ready for a final(?) review.

jwhitlock (approved):
Thanks @groovecoder! I checked it out, it looks good.

I think that instead of calling `client.messages.create` in `api.views.phones.inbound_sms`, we could respond with a [Message](https://www.twilio.com/docs/messaging/twiml/message#) verb:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Response>
    <Message to="+user_real_num" from="+mask_num">
        [Relay 📲 +incoming_num] Message
    </Message>
</Response>
```
but that's a different experiment.

groovecoder:
I filed https://mozilla-hub.atlassian.net/browse/MPP-2232 for us to look into returning TwiML to forward messages.

PR #2261, opened 2022-07-28
Add a database default for relaynumber.enabled

jwhitlock:
With a database-enabled default value, the existing v3.6.1 code will continue to work during the period when this migration is run and the new release is not fully deployed.

Tested locally with sqlite3 3.37.0 and PostgreSQL 14.4. CircleCI will run tests against both.

### DIFF
--- a/phones/migrations/0017_relaynumber_enabled.py
+++ b/phones/migrations/0017_relaynumber_enabled.py
@@ -3,16 +3,66 @@
 from django.db import migrations, models
 
 
+def add_db_default_forward_func(apps, schema_editor):
+    """
+    Add a database default of TRUE for enabled, for PostgreSQL and SQLite3
+
+    Using `./manage.py sqlmigrate` for the SQL, and the technique from:
+    https://stackoverflow.com/a/45232678/10612
+    """
+    if schema_editor.connection.vendor.startswith("postgres"):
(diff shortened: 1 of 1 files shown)


jwhitlock:
See https://github.com/mozilla/fx-private-relay/pull/2259/files#r932432801 for more context

phones/migrations/0017_relaynumber_enabled.py
```diff
@@ -3,16 +3,66 @@
...
-            name='enabled',
+            model_name="relaynumber",
+            name="enabled",
             field=models.BooleanField(default=True),
         ),
+        migrations.RunPython(
+            code=add_db_default_forward_func,
+            reverse_code=migrations.RunPython.noop,
+            elidable=True,
+        ),
```
groovecoder, line 67: Does it matter if this runs before/after the `migrations.AddField`?

groovecoder (approved):
approved with just 1 Q to make sure I understand and don't mess up the merge.

phones/migrations/0017_relaynumber_enabled.py
jwhitlock: yes for the PostgreSQL variant, no for the SQLite variant, so yes

PR #2264, opened 2022-07-28
MPP-79: Break the build on Python formating and typing errors

jwhitlock:
Fail the CircleCI build if `black` or `mypy` return errors. Fix the `black` error in `main`.

I've also added `black` and `mypy` to `frontend/.husky/pre-commit`, so if you are using that hook, the commands will run as well, and halt the commit on formatting errors.


jwhitlock:
I've used https://pre-commit.com/ in the past, I think it would potentially run our tools better than this method. I'm not sure if I'd run it from the husky pre-commit, or if it would replace husky.

frontend/.husky/pre-commit
```diff
@@ -3,3 +3,9 @@
 
 cd frontend
 npx --no-install lint-staged
+
+cd ..
+echo "Running black..."
+black --check .
+echo "Running mypy..."
+mypy .
```
Vinnl, line 11: Ideally you'd add these to [`.lintstagedrc`](https://github.com/mozilla/fx-private-relay/blob/main/frontend/.lintstagedrc.js) to use [lint-staged](https://www.npmjs.com/package/lint-staged), so they only run against staged changes. You'd add something like this:

```js
  "*.py": "black",
  "*.py": "mypy",
```

Vinnl (commented):
See my inline comment :)

One disadvantage of installing the hook using Husky, is that it only gets added when you run `npm install`; I'm not sure if that's something all back-end engineers do? Though I guess you'll need to anyway if you want to test the website with the UI.

I'd also be interested in learning what pre-commit does different that would run our tools better :)

jwhitlock:
Thanks @Vinnl! I didn't know much about `lint-staged`, this all makes a lot more sense now.

To get it working, I had to move `.lintstagedrc` to the root of the project, and run `lint-staged` from there. This is because `mypy` needs to run in the root of the project, and there did not appear to be support for running a single command from a different folder. I was able to test `black` by making a change to a file that broke formatting, _and_ changed something about the code. If you just break formatting, then `black` fixes it, and `lint-staged` refuses to create an empty commit.

I'm running `prettier --write` on all `.md` files, which will include the `docs` directory. That may be a mistake. Otherwise, I've restricted JS linters to the `frontend` folder. **You may need to check that this did not break anything**.

Something like this works to test the Python code:

Add this line to `emails/apps.py`:

```python
test: str   =   3.14
```

Run `git add emails/apps.py; git commit -m "Test precommit". It fails with:

```
✔ Preparing lint-staged...
❯ Running tasks for staged files...
  ❯ .lintstagedrc.js — 1 file
    ↓ frontend/*.{scss,css} — no files [SKIPPED]
    ↓ frontend/*.{ts,tsx,js,jsx,scss,css} — no files [SKIPPED]
    ↓ frontend/*.{ts,tsx,js,jsx} — no files [SKIPPED]
    ↓ *.md — no files [SKIPPED]
    ❯ *.py — 1 file
      ✔ black
      ✖ mypy [FAILED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

✖ mypy:
emails/apps.py:15: error: Incompatible types in assignment (expression has type "float", variable has type "str")  [assignment]
Found 1 error in 1 file (checked 1 source file)
husky - pre-commit hook exited with code 1 (error)
```

Change the line to:

```python
test: float   =   3.14
```

(no `mypy` error, but black doesn't like formatting)

Run `git add emails/apps.py; git commit -m "Test precommit". It succeeds with:

```
✔ Preparing lint-staged...
✔ Running tasks for staged files...
✔ Applying modifications from tasks...
✔ Cleaning up temporary files...
[add-python-precommit eed488d7] test precommit
 1 file changed, 2 insertions(+)
 ```
 
 Load `emails.apps.py`, and the line has changed to:
 
```python
 test: float = 3.14
```

Running `git show` will show the commit changed as well.

.lintstagedrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  "frontend/*.{scss,css}": "stylelint --fix",
+  "frontend/*.{ts,tsx,js,jsx,scss,css}": "prettier --write",
+  "frontend/*.{ts,tsx,js,jsx}": (filenames) =>
+    `next lint --fix --file ${filenames
+      .map((file) => file.split(process.cwd())[1])
+      .join(" --file ")}`,
```
Vinnl, line 7: Sorry for taking a while to get back to this - had a pile of other work too.

> I'm running prettier --write on all .md files, which will include the docs directory. That may be a mistake. Otherwise, I've restricted JS linters to the frontend folder. You may need to check that this did not break anything.

Hmm, I tried a bunch of things, but I can't seem to get it to work. The first cause was that these patterns didn't match staged files - I had to change them from `frontend/*.` to `frontend/**/*.`.

After that, lint-stage at least gets trigged if I modify a frontend file (e.g. replacing `"` by `'` in an import in `frontent/src/components/Banner.tsx`). Then Prettier appears to run correctly, but it hangs on linting:

https://user-images.githubusercontent.com/4251/186926483-5e7c99ab-432e-4c4c-91cc-8ec290dbece5.mp4

At first I thought that might be related to it not running in `/frontend`, but then Prettier running successfully makes me reject that thought. So not sure what's going on :/ Might have to revert back to just running in Husky's pre-commit again.

Running Prettier on all Markdown files seems fine by me.

jwhitlock: Thanks for the example that will trigger `prettier`. I couldn't find any simple edits that `next lint` would fix, however.

I updated the pattern to `frontend/**/*`. It took me a while to figure out how to even run `next lint` manually. When I ran `npx next lint` in the root project folder, not `frontend`, it prompted:

```
Need to install the following packages:
  next
Ok to proceed? (y) 
```

I suspect that is why the command was hanging - it was waiting for someone to type `y`, but the output was swallowed by `lint-staged`.

I updated `next lint --fix ...` to `next lint frontend --fix ...` to set the base directory, and that seemed to work.

jwhitlock: ... and I rebased on `main`, which had a handful of `black` and `mypy` issues

Vinnl: Huh. I have no idea why `next lint frontend` would work (I can't find any documentation of `next` that specifies the ability to pass `frontend`, and I also don't know why it wouldn't ask to install `next` when run with that parameter), but it does seem to work.

(I tested `next lint` fixes by replacing `===` with `==` in the comparison with `typeof` on line 315 of `/frontend/src/apiMocks/handlers.ts`.)

Vinnl (approved):
As mentioned, I don't know why (which I hate), but it does work for me!

.lintstagedrc.js
jwhitlock: > Huh. I have no idea why next lint frontend would work (I can't find any documentation of next that specifies the ability to pass frontend, and I also don't know why it wouldn't ask to install next when run with that parameter), but it does seem to work.

I can't find the docs online. I used:

```
npx next lint --help

      Description
        Run ESLint on every file in specified directories. 
        If not configured, ESLint will be set up for the first time.

      Usage
        $ next lint <baseDir> [options]      

      <baseDir> represents the directory of the Next.js application.
      If no directory is provided, the current directory will be used.
      
      ...
```

jwhitlock:
@Vinnl thanks for sticking with a 1 month review process!

I've gotten some practical experience with `husky` + `lint-staged` with this PR, and I think it works just as well as `pre-commit` did on the project I used it on, so I don't think we need to investigate. Now that we've figured out how to run linters for Python, JS, and Markdown, I think it can be trivial to add more linters in the future.

.lintstagedrc.js
Vinnl: Oh neat, didn't think of that!

PR #2297, opened 2022-08-03
let twilio exceptions raise up thru our code

groovecoder:
To help us debug twilio issues in dev or stage environments, let Twilio
exceptions raise up thru our code, so they can be logged in sentry.


jwhitlock (approved):
Thanks @groovecoder, let's debug! 🐛

PR #2303, opened 2022-08-05
Add `store_phone_log` to `Profile` mpp 2250

groovecoder:
This PR fixes MPP-2250.

How to test:
1. Check out this code
2. Run `python manage.py migrate`
3. Run `python manage.py runserver`
4. Sign in to http://127.0.0.1:8000
5. Go to http://127.0.0.1:8000/api/v1/profiles/
   * [x] there should be a `store_phone_log` item (defaults to `true`)
   * Note the `id` value
6. Go to http://127.0.0.1:8000/api/v1/docs/
7. Scroll down to the "profiles" endpoints
8. Expand the `PATCH /profiles/{id}/` item
9. Click "Try it out"
10. Change `  "store_phone_log": false,`
11. Enter the `id` value from step 5
12. Click "Execute"
13. Refresh http://127.0.0.1:8000/api/v1/profiles/
    * [x] The `store_phone_log` item should now show `false`

- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #2316, opened 2022-08-08
Add contacts model mpp 2259

groovecoder:
This PR fixes MPP-2259. It builds on https://github.com/mozilla/fx-private-relay/pull/2303, so let's review & merge that first.
# New feature description
* Adds a new `InboundContact` model and API endpoints that will power the "Caller and Sender log" screen.
   * GET `/api/v1/inboundcontact/` shows all inbound contacts for the `RelayNumber` of the signed-in user
   * PATCH `/api/v1/inboundcontact/{id}` to set the `blocked:True|False` field of a contact
* Changes `inbound_sms` and `inbound_call` to:
   * Check if the profile has `store_phone_log:True|False`
   * If `True`, Get or create an `InboundContact` model for the caller/texter
   * If contact is blocked, increment the `num_calls_blocked` or `num_texts_blocked` field
   * If contact is allowed, increment the `num_calls` or `num_texts` field

# Screenshot (if applicable)
image

How to test:
1. Make sure you've set up [end-to-end local phone development](https://github.com/mozilla/fx-private-relay/blob/main/docs/end-to-end-local-phone-dev.md)
2. Check out this branch and `python manage.py migrate`
3. Go to http://127.0.0.1:8000/api/v1/docs/
4. Run the `GET /api/v1/inboundcontact/` endpoint
   * [x] It should show 0 results
5. Send a text to your local Relay number
6. Run the `GET /api/v1/inboundcontact/` endpoint
   * [x] It should show 1 result
   * [x] Note the `last_inbound_date` value
7. Wait a little bit
8. Send a text to your local Relay number
9. Run the `GET /api/v1/inboundcontact/` endpoint
   * [x] It should show 1 result, and the `last_inbound_date` should be updated
10. Send a text to your local Relay number **FROM A DIFFERENT NUMBER**
11. Run the `GET /api/v1/inboundcontact/` endpoint
    * [x] It should show 2 results
12. Run the `PATCH /api/v1/profiles/{id}` endpoint to update your profile to `store_phone_log:false`
13. Send a text to your local Relay number
14. Run the `GET /api/v1/inboundcontact/` endpoint
    * [x] It should show only the 2 results from step 11


emails/migrations/0050_profile_store_phone_log.py
```diff
@@ -0,0 +1,127 @@
...
+            ' ("id" integer NOT NULL PRIMARY KEY AUTOINCREMENT,'
+            ' "api_token" char(32) NOT NULL,'
+            ' "user_id" integer NOT NULL REFERENCES "auth_user" ("id") DEFERRABLE INITIALLY DEFERRED,'
+            ' "address_last_deleted" datetime NULL,'
+            ' "num_address_deleted" integer unsigned NOT NULL CHECK ("num_address_deleted" >= 0),'
+            ' "last_hard_bounce" datetime NULL,'
+            ' "last_soft_bounce" datetime NULL,'
+            ' "subdomain" varchar(63) NULL UNIQUE,'
+            ' "server_storage" bool NOT NULL,'
+            ' "store_phone_log" bool NOT NULL,'
```
jwhitlock, line 30: Nice job with this messiness! One change, add the database-level default of `true`:

```suggestion
            ' "store_phone_log" bool NOT NULL DEFAULT 1,'
```

emails/models.py
```diff
@@ -126,6 +129,8 @@ def save(self, *args, **kwargs):
         if not self.server_storage:
             relay_addresses = RelayAddress.objects.filter(user=self.user)
             relay_addresses.update(description="", generated_for="", used_on="")
+        # TODO: any time a profile sets store_phone_log to False, delete the
```
jwhitlock, line 132: TODO: Write an issue to make sure we do this task. Bonus: reference the issue number in the comment.

phones/models.py
```diff
@@ -204,6 +204,23 @@ def relaynumber_post_save(sender, instance, created, **kwargs):
...
 
+def last_inbound_date_default():
+    return datetime.now(timezone.utc)
+
+
+class InboundContact(models.Model):
+    relay_number = models.ForeignKey(RelayNumber, on_delete=models.CASCADE)
+    inbound_number = models.CharField(max_length=15)
+    last_inbound_date = models.DateTimeField(
+        default=last_inbound_date_default
```
jwhitlock, line 215: Do we want to use [auto_now_add](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.DateField.auto_now_add) instead? I think it does the same thing, except you may need some extra code in tests if you want to initialize to a different value than now.

phones/models.py
```diff
@@ -204,6 +204,23 @@ def relaynumber_post_save(sender, instance, created, **kwargs):
         )
 
 
+def last_inbound_date_default():
+    return datetime.now(timezone.utc)
+
+
+class InboundContact(models.Model):
+    relay_number = models.ForeignKey(RelayNumber, on_delete=models.CASCADE)
+    inbound_number = models.CharField(max_length=15)
```
jwhitlock, line 213: Suggestion: an index on `(relay_number, inbound_number)` for the query in `_get_phone_objects`. It's a light suggestion, since the index on `relay_number` should be OK for most users, and I expect incoming numbers in the dozens rather than the thousands, so a serial scan should be OK.

jwhitlock (changes requested):
Thanks @groovecoder, I think per-contact blocking is useful (and we'll be wanting it for emails someday...)

Blockers:
* Add the database default for `store_phone_log` for sqlite
* Run `black`

api/serializers/phones.py
```diff
@@ -38,3 +38,27 @@ class Meta:
...
+class InboundContactSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = InboundContact
+        fields = [
+            "id",
+            "relay_number",
+            "inbound_number",
+            "last_inbound_date",
+            "num_calls",
+            "num_calls_blocked",
```
jwhitlock, line 52: Do we want `num_calls_blocked` and `num_texts_blocked` to be API writable?

api/views/phones.py
```diff
@@ -396,19 +415,55 @@
...
 
+    _, real_phone, inbound_contact = _get_phone_objects(
+        inbound_to, inbound_from, "calls"
+    )
+    _check_and_update_contact(inbound_contact, "calls")
+
+    # Note: TwilioInboundCallXMLRenderer will render this as TwiML
+    return response.Response(
+        status=201,
+        data={"inbound_from": inbound_from, "real_number": real_phone.number},
```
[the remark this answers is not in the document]
jwhitlock: I'm OK dismissing this as invalid or won't fix. We validate that the incoming message is from Twillio, and then return the response to Twillio, so I don't consider it a true user-provided value.

jwhitlock:
I ran the test procedures. I might have hit some limit with AT&T - Twilio is saying some messages were sent, but the carrier didn't immediately respond that it was delivered. They were then delivered about the same time, one 5 minutes later and the other 12 minutes later.

For the last manual test, I disabled `store_phone_log`:

* For a message from a new number, a new entry is added for http://127.0.0.1:8000/api/v1/inboundcontact
* For a message from an existing number, `last_inbound_date` and `num_texts` were updated.

Is this expected?

phones/models.py
groovecoder: I got scared off `auto_now_add` by this note in the docs:

> The auto_now and auto_now_add options will always use the date in the [default timezone](https://docs.djangoproject.com/en/3.2/topics/i18n/timezones/#default-current-time-zone) at the moment of creation **or update**. If you need something different, you may want to consider using your own callable default or overriding save() instead of using auto_now or auto_now_add; or using a DateTimeField instead of a DateField and deciding how to handle the conversion from datetime to date at display time.

Which I took to mean anytime we update the `blocked` field of the contact, it would ALSO update the `last_inbound_date`?

api/serializers/phones.py
groovecoder: Not right now - there's no reason for an external client to change these values. Only the incoming call or incoming text logic should update these values.

phones/models.py
jwhitlock: The note is poorly worded. `auto_now_add` works at creation, and `auto_add` at update (. I tried `auto_now_add` and what you suggested - sending a `PATCH` with `blocked` does _not_ update `last_inbound_date`.

Now that I think of it, shouldn't we have a bunch of `created` and `modified` fields for all these models?

groovecoder:
> For the last manual test, I disabled `store_phone_log`:
> 
>     * For a message from a new number, a new entry is added for http://127.0.0.1:8000/api/v1/inboundcontact
> 
>     * For a message from an existing number, `last_inbound_date` and `num_texts` were updated.
> 
> 
> Is this expected?

Whoops I had un-committed changes on my local. I pushed them to this branch. Try again?

groovecoder:
@jwhitlock this should be ready for final review!

api/serializers/phones.py
jwhitlock: >> Do we want `num_calls_blocked` and `num_texts_blocked` to be API writable?

> Not right now - there's no reason for an external client to change these values. Only the incoming call or incoming text logic should update these values.

OK, please add `num_calls_blocked` and `num_texts_blocked` to `read_only_fields`

jwhitlock (changes requested):
Thanks @groovecoder! One last change to the API before merging, please.

jwhitlock (approved):
🚢 SHIP IT

PR #2324, opened 2022-08-09
MPP-79: Add .flake8 configuration, fix a few issues

jwhitlock:
As part of MPP-79, add a `.flake8` configuration that is compatible with `black`, and fix a few of the detected issues.

[Flake8](https://flake8.pycqa.org/en/latest/) is a static analyzer used on many Python projects to enforce [PEP8](https://peps.python.org/pep-0008/) and detect unused variables and imports. It comes with some good analyzers by default, and others (like docstring checkers) can be added. I've found it much faster than `pylint` for similar benefits.

I wasn't planning on adding `flake8` itself - we're still getting used to `black` - but I've found it useful in my own development work, and run it on each save in Vim (through [scrooloose/syntastic](https://github.com/vim-syntastic/syntastic)). 

This PR also fixes some issues in the files I'm working on in another PR - mostly long lines, but some unused imports and variables. After these changes, there are 119 issues remaining, which would take about an hour to clean up.

# How to test

* Checkout `main` branch
* Run `pip install flake8`
* Run `flake8 emails/models.py --statistics`
    - [ ] There are over 40 warnings for `E501 line too long`
    - [ ] There is one error `F841 local variable 'address_contains_badword' is assigned to but never used`
* Checkout this branch `start-flake8-mpp-79`
* Run `flake8 emails/models.py --statistics`
    - [ ] There is no output, because there are no remaining issues.
* Run `flake8 . --statistics`
    - [ ] There are 100 - 150 warnings and errors, and about 8 unique issues

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,12 @@
+[flake8]
+
+# Recommend matching the black line length (default 88),
+# rather than using the flake8 default of 79:
+max-line-length = 88
+extend-ignore =
+    # See https://github.com/PyCQA/pycodestyle/issues/373
+    E203,
+exclude =
+    .git,
+    __pycache__,
+    env
--- a/emails/apps.py
+++ b/emails/apps.py
@@ -15,7 +15,8 @@ logger = logging.getLogger("events")
 def get_trackers(category="Email"):
     # email tracker lists from shavar-prod-list as per agreed use under license:
     resp = requests.get(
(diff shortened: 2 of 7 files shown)


jwhitlock:
I'm going to continue using this, but I think we're still working on incorporating other tools into our toolbox, so I'm not ready to make it a team requirement.

PR #2330, opened 2022-08-10
fix MPP-2261: delete InboundContact records

groovecoder:
When Profile.store_phone_log is set to False, delete all the
InboundContact records.

This PR fixes MPP-2261.

How to test:
1. Run `pytest`
   * [x] All tests should pass
2. Send some text messages to your local Relay number
3. Go to http://127.0.0.1:8000/api/v1/docs/
4. Execute the `GET /inboundcontact/` endpoint
   * [x] You should see an inbound contact record for the sender of the text message
5. Go to http://127.0.0.1:8000/accounts/settings/
6. Un-check "Allow ⁨Firefox Relay⁩ to keep a log of your callers and text senders."
7. Click Save
   * [ ] You should see "Your settings have been saved."
8. Go back to http://127.0.0.1:8000/api/v1/docs/
9. Execute the `GET /inboundcontact/` endpoint
   * [x] You should NOT see any inbound contact records

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ RUN ln --symbolic /app/privaterelay/locales/pt-BR/ privaterelay/locales/pt
 RUN ln --symbolic /app/privaterelay/locales/es-ES/ privaterelay/locales/es
 COPY --chown=app .env-dist /app/.env
 
+ARG PHONES_ENABLED
 RUN mkdir -p /app/staticfiles && \
     python manage.py collectstatic --no-input -v 2
 
--- a/phones/tests/models_tests.py
+++ b/phones/tests/models_tests.py
@@ -12,6 +12,7 @@ from allauth.socialaccount.models import SocialAccount, SocialToken
 from model_bakery import baker
 
 from emails.models import Profile
+from phones.models import InboundContact
 
 if settings.PHONES_ENABLED:
     from ..models import (
@@ -45,7 +46,6 @@ def mocked_twilio_client():
 def make_phone_test_user():
     phone_user = baker.make(User)
     phone_user_profile = Profile.objects.get(user=phone_user)
-    phone_user_profile.server_storage = True
     phone_user_profile.date_subscribed = datetime.now(tz=timezone.utc)
     phone_user_profile.save()
     upgrade_test_user_to_phone(phone_user)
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -30,6 +30,7 @@ RUN ln --symbolic /app/privaterelay/locales/pt-BR/ privaterelay/locales/pt
 RUN ln --symbolic /app/privaterelay/locales/es-ES/ privaterelay/locales/es
 COPY --chown=app .env-dist /app/.env
 
+ARG PHONES_ENABLED
```
jwhitlock, line 33: Nit: I wonder if we should force-enable phones (and for that matter emails) for the `Dockerfile`, since it is needed to build the assets:

```
ENV PHONES_ENABLED=1
```

Then we wouldn't need to pass it as a build arg.

However, this works, so no need to change anything for this PR.

privaterelay/settings.py
```diff
@@ -230,7 +230,7 @@
         "emails.apps.EmailsConfig",
     ]
 
-if PHONES_ENABLED and TWILIO_ACCOUNT_SID and TWILIO_AUTH_TOKEN:
+if PHONES_ENABLED:
```
jwhitlock, line 233: Thanks, I'm much more comfortable with `PHONES_ENABLED` as the on/off switch, and leaving secrets out of it.

phones/tests/models_tests.py
```diff
@@ -319,3 +319,43 @@ def test_area_code_numbers(mocked_twilio_client):
...
+    user = make_phone_test_user()
+    profile = Profile.objects.get(user=user)
+    baker.make(RealPhone, user=user, verified=True)
+    relay_number = baker.make(RelayNumber, user=user)
+    inbound_contact = baker.make(InboundContact, relay_number=relay_number)
+    profile.store_phone_log = False
+    profile.save()
+
+    with pytest.raises(InboundContact.DoesNotExist):
+        inbound_contact.refresh_from_db()
```
jwhitlock, line 361: TIL: `refresh_from_db` can raise `DoesNotExist`

jwhitlock (approved):
Thanks @groovecoder!

The test instructions did not work for me, because my settings page does not include "Allow ⁨Firefox Relay⁩ to keep a log of your callers and text senders.". I don't think that is a switch, I suspect you are using UI from another PR. I instead directly switched `Profile.store_phone_log`. It correctly deleted my existing records, and did not create new ones.

PR #2346, opened 2022-08-15
fix MPP-2288: add InboundContact.last_inbound_type

groovecoder:
# New feature description
[The design for the "Callers and SMS Senders"](https://www.figma.com/file/g829jRfj82xF2x9dHKbpTk/Phone-Number-Mask?node-id=394%3A8827) "Latest Activity" column shows a different icon if the last activity was a call or a text.
image

# Screenshot (if applicable)
This code adds a new `InboundContact.last_inbound_type` field that shows up on the `GET /api/v1/inboundcontact/[{id}]` endpoints:
image

# How to test

1. `pytest`
   * [ ] All tests pass
3. Make sure your profile is set to "Allow ⁨Firefox Relay⁩ to keep a log of your callers and text senders."
4. Send a text to your relay number
5. Check http://127.0.0.1:8000/api/v1/inboundcontact/
   * [ ] Check that the `last_inbound_type` field shows up and is accurate
6. Use `PATCH http://127.0.0.1:8000/api/v1/inboundcontact/{id}` to set `blocked: true`
7. Call your relay number
   * [ ] Check that your call is rejected (still ugly "Application Error" message for now)
8. Check http://127.0.0.1:8000/api/v1/inboundcontact/ again
   * [ ] Check that the `last_inbound_type` is still `text` (from step 3)

# Checklist


api/views/phones.py
```diff
@@ -458,6 +458,8 @@ def _check_and_update_contact(inbound_contact, contact_type):
         raise exceptions.ValidationError(f"Number is not accepting {contact_type}.")
 
     inbound_contact.last_inbound_date = datetime.now(timezone.utc)
+    # strip trailing "s": InboundContact.last_inbound_type is max_length 4
+    inbound_contact.last_inbound_type = contact_type.strip("s")
```
say-yawn, line 462:
```suggestion
    inbound_contact.last_inbound_type = contact_type[:-1]
```
to drop the last character in case we deal with future types like `"spams"` which the current logic will return `"pam"` (unsure if this is a real contact type but was a delicious salty example).

api/tests/phones_views_tests.py
```diff
@@ -1,3 +1,4 @@
+from datetime import timezone
```
say-yawn, line 1:
```suggestion
```
nit: imported by unused

phones/models.py
```diff
@@ -212,6 +216,9 @@ class InboundContact(models.Model):
     relay_number = models.ForeignKey(RelayNumber, on_delete=models.CASCADE)
     inbound_number = models.CharField(max_length=15)
     last_inbound_date = models.DateTimeField(default=last_inbound_date_default)
+    last_inbound_type = models.CharField(
+        max_length=4, choices=LAST_CONTACT_TYPE_CHOICES, default="text"
```
say-yawn, line 220: question: any reasons on why max length 4?

say-yawn (changes requested):
First time reviewing migrations with `add_db_default_forward_func` was a bit intimidated at first but after talking with @groovecoder I did a quick check in the code that all the fields are present and the newly added field is being populated with `text`. Some suggested changes and questions.

phones/models.py
groovecoder: Not any good one. Since it's storing a finite set of options and they're both 4-characters long, I thought we might as well save DB storage space.

groovecoder:
Merged reviewed suggestions.

say-yawn (approved):

PR #2360, opened 2022-08-20
fix MPP-2299: pull TopicArn and Type from JSON body

groovecoder:
This PR fixes MPP-2299 by pulling the `TopicArn` and `Type` values from the verified JSON body rather than the HTTP headers.

How to test:
1. `pytest` should run with no error nor failures
2. Mail processing should work exactly as before.

### DIFF
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from datetime import datetime, timezone
 from email.message import EmailMessage
 from unittest.mock import patch, Mock
@@ -33,7 +34,7 @@ from emails.views import (
     _record_receipt_verdicts,
     _sns_message,
     _sns_notification,
-    validate_sns_header,
+    validate_sns_arn_and_type,
     wrapped_email_test,
     InReplyToNotFound,
 )
@@ -870,13 +871,13 @@ TEST_AWS_SNS_TOPIC2 = TEST_AWS_SNS_TOPIC + "-alt"
 
 
 @override_settings(AWS_SNS_TOPIC={TEST_AWS_SNS_TOPIC, TEST_AWS_SNS_TOPIC2})
-class ValidateSnsHeaderTests(SimpleTestCase):
-    def test_valid_headers(self):
-        ret = validate_sns_header(TEST_AWS_SNS_TOPIC, "SubscriptionConfirmation")
+class ValidateSnsArnTypeTests(SimpleTestCase):
+    def test_valid_arn_and_type(self):
+        ret = validate_sns_arn_and_type(TEST_AWS_SNS_TOPIC, "SubscriptionConfirmation")
         assert ret is None
 
     def test_no_topic_arn(self):
-        ret = validate_sns_header(None, "Notification")
+        ret = validate_sns_arn_and_type(None, "Notification")
         assert ret == {
             "error": "Received SNS request without Topic ARN.",
             "received_topic_arn": "''",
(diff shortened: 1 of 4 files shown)


emails/views.py
```diff
@@ -307,23 +307,22 @@ def _index_DELETE(request_data, user_profile):
...
+        logger.error("validate_sns_arn_and_type_error", extra=error_details)
         return HttpResponse(error_details["error"], status=400)
 
-    json_body = json.loads(request.body)
-    verified_json_body = verify_from_sns(json_body)
     return _sns_inbound_logic(topic_arn, message_type, verified_json_body)
 
 
-def validate_sns_header(topic_arn, message_type):
+def validate_sns_arn_and_type(topic_arn, message_type):
```
jwhitlock, line 325: Nit: Another opportunity to add some typing. The function signature will be easier than the return type:
```suggestion
def validate_sns_arn_and_type(topic_arn: Optional[str], message_type: Optional[str]) -> Optional[dict[str, Union[str, list[str]]]]
```

emails/views.py
```diff
@@ -307,23 +307,22 @@ def _index_DELETE(request_data, user_profile):
...
-    # Grabs message information for validation
-    topic_arn = request.headers.get("X-Amz-Sns-Topic-Arn", None)
-    message_type = request.headers.get("X-Amz-Sns-Message-Type", None)
+    # First thing we do is verify the signature
+    json_body = json.loads(request.body)
+    verified_json_body = verify_from_sns(json_body)
 
-    # Validates header
-    error_details = validate_sns_header(topic_arn, message_type)
+    # Validate ARN and message type
```
jwhitlock, line 314: Future work: It seems to me that this code could be in `_sns_inbound_logic` (which would just take `verified_json_body` or in `verify_from_sns` (which currently just checks the crypto signing). But, there are other changes to make to fully migrate from a push endpoint to the queue pulling task, we can do it then.

jwhitlock (approved):
Thanks @groovecoder! There are a few ignorable nits, depending on your appetite for refactoring and adding types for this PR.

emails/tests/views_tests.py
```diff
@@ -937,8 +937,9 @@ def test_valid_message(self):
         )
         assert ret.status_code == 200
 
-    def test_no_topic_arn_header(self):
-        self.mock_verify_from_sns.side_effect = FAIL_TEST_IF_CALLED
+    def test_no_topic_arn(self):
+        pre_test_topic_arn = self.valid_message["TopicArn"]
```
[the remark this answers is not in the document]
groovecoder: Love it, thanks!

emails/views.py
groovecoder: I haven't found a technique I like for prioritizing when and how to add types. And before you chalk it up to my laziness ...

I wanted to see how the function understandability improved with these type declarations. It changed the intelli-sense pop-up from this:

image

to this:
image

Hot-takes:

* I *really* like the type hints on the function parameters. `topic_arn: str, message_type: str` is way more helpful than `topic_arn: Unknown, message_type: Unknown`
* I *really* like doc-blocks that explain the parameters & return values of the function. 
* Like you say - the complex return type is harder, and I'm not sure if `dict [str, str | list[str]] | None` is more helpful than just `dict [str, Unknown] | None`? In either case, I'm likely going to have to inspect the function to really know how to handle what it returns.

On that last point - for this particular function, it seems like `pywright` was able to automatically determine `dict [str, ...] | None` part of the return type, but then it couldn't determine the types of the values of the dictionary? And yet, when I hover over each of the values in the returned dictionary, `pywright` **IS** able to determine each individual type? The possible value types are:

* `error: str | None`
* `received_topic_arn: str`
* `supported_topic_arn: list[SupportsRichComparisonT@sorted | _T@sorted]`
* `received_sns_type: str`
* `supported_sns_types: list[str]`

So, I tried creating a `dataclass` for it:
```python
@dataclass
class SNSValidationErrorDetails:
    error: str
    received_topic_arn: str
    supported_topic_arn: list
    received_sns_type: str
    supported_sns_types: list[str]


def validate_sns_arn_and_type(topic_arn: str, message_type: str) -> Optional[SNSValidationErrorDetails]:
```

But that made the intellisense even **less** helpful (IMO):

image

Just based on this example, I think I like:

* Add type hints to function parameters
* Add doc blocks explaining the parameters & return values of functions
* Add **basic** return types to functions

So what I ended up with is:
```python
def validate_sns_arn_and_type(topic_arn: str, message_type: str) -> Optional[dict[str, Any]]:
```

Which generates this IntelliSense:

image

emails/tests/views_tests.py
```diff
@@ -870,13 +870,13 @@ def test_attachment_no_filename(self):
 
 
 @override_settings(AWS_SNS_TOPIC={TEST_AWS_SNS_TOPIC, TEST_AWS_SNS_TOPIC2})
-class ValidateSnsHeaderTests(SimpleTestCase):
-    def test_valid_headers(self):
-        ret = validate_sns_header(TEST_AWS_SNS_TOPIC, "SubscriptionConfirmation")
+class ValidateSnsArnTypeTests(SimpleTestCase):
+    def test_valid_arn_and_type(self):
```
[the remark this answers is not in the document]
groovecoder: `-> None` on test functions doesn't seem very helpful, since the only code calling it will be the test runner. Skipping this one.

PR #2377, opened 2022-08-22
fix MPP-2298: remove the un-used /emails/ url

groovecoder:
This PR fixes MPP-2298.

How to test:

1. `pytest` should run with no errors nor failures
2. Create a random mask
   * [ ] It should still work
3. Create a domain mask
   * [ ] It should still work
4. Run the add-on (`web-ext run -s src/`)
5. Create a new mask with the add-on
   * [ ] It should still work

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -20,12 +20,10 @@ from waffle import sample_is_active
 from waffle.models import Flag
 
 from django.conf import settings
-from django.contrib import messages
 from django.contrib.auth.models import User
-from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
+from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
-from django.http import HttpResponse, JsonResponse
-from django.shortcuts import redirect, render
+from django.http import HttpResponse
 from django.template.loader import render_to_string
 from django.utils.html import escape
 from django.views.decorators.csrf import csrf_exempt
@@ -47,7 +45,6 @@ from .utils import (
     remove_trackers,
     count_all_trackers,
     get_message_content_from_s3,
-    get_post_data_from_request,
     incr_if_enabled,
     histogram_if_enabled,
     ses_relay_email,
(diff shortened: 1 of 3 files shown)


say-yawn (commented):
While testing that when the `emails_index` is enabled the `/emails` path still works I found out:
- when the waffle is created and switched on with an instance that is already running, you need to restart the server/instance for the switch to be used
- enabling `emails_index` switch and making a GET request on `/emails` results in the `Reverse for 'profile' not found. 'profile' is not a valid view function or pattern name.` error which was seen 7 days ago on prod ([sentry error](https://sentry.prod.mozaws.net/operations/fx-private-relay-prod/issues/19469691/?query=is%3Aunresolved%20transaction%3A%2Femails%2F))

All this to say I am supportive of us deleting the index view on the email app altogether. cc @jwhitlock

groovecoder:
Updated to remove the url completely.

jwhitlock (approved):
Do you want to remove the backing code from `emails/views.py` as well, or do that later? If you do it all in one commit, it is easy to revert in the future if we decide we want it back.

If you don't, I _love_ deleting code, I can write the follow-up PR.

groovecoder:
I removed it now. That should increase our code coverage too, right? 😄

emails/views.py
```diff
@@ -47,7 +47,6 @@
     remove_trackers,
     count_all_trackers,
     get_message_content_from_s3,
-    get_post_data_from_request,
```
jwhitlock, line 50: A few more imports can be removed now:

```
emails/views.py:23:1: F401 'django.contrib.messages' imported but unused
emails/views.py:25:1: F401 'django.core.exceptions.PermissionDenied' imported but unused
emails/views.py:27:1: F401 'django.http.JsonResponse' imported but unused
emails/views.py:28:1: F401 'django.shortcuts.redirect' imported but unused
emails/views.py:28:1: F401 'django.shortcuts.render' imported but unused
```

jwhitlock (changes requested):
Yep, removing completely untested code should increase our code coverage 😁 

Thanks! Please remove a few more imports!

emails/views.py
groovecoder: removed!

jwhitlock (approved):
❤️ Thanks @groovecoder!

PR #2379, opened 2022-08-22
Fix MPP-2307 - Refine Copy Number UI

codemist:
This PR fixes (from Jira) MPP-2307

How to test:

Check that the Copy button on the Phone dashboard contains the alt text "Click to Copy" and that the confirmation UI not overlap over the phone number on mobile.

https://user-images.githubusercontent.com/13066134/186009030-58ea140b-57f2-49d0-8f85-0d11183776c5.mov

### DIFF
--- a/frontend/src/components/phones/dashboard/PhoneDashboard.module.scss
+++ b/frontend/src/components/phones/dashboard/PhoneDashboard.module.scss
@@ -39,16 +39,18 @@
   .senders-cta {
     display: flex;
     flex-direction: row;
-    padding: $spacing-sm;
-    position: relative;
-    left: $spacing-sm;
     border-radius: $border-radius-lg;
     align-items: center;
-    gap: $spacing-xs;
-    margin-left: $spacing-sm;
+    justify-content: center;
 
-    &:hover {
-      background: $color-light-gray-10;
+    @media screen and #{$mq-md} {
+      padding: $spacing-sm;
+      gap: $spacing-xs;
+      margin-left: $spacing-sm; // Cancel out extra space from padding
+
+      &:hover {
+        background: $color-light-gray-10;
+      }
     }
 
     span {
@@ -56,7 +58,6 @@
 
       @media screen and #{$mq-md} {
         display: block;
-        left: 0;
       }
     }
   }
@@ -351,7 +352,6 @@
         background-color: $color-violet-30;
         color: $color-white;
         border-radius: $border-radius-md;
-        position: static;
         right: 0;
 
         &.is-shown {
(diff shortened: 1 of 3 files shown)


frontend/src/components/phones/dashboard/PhoneDashboard.module.scss
```diff
@@ -39,7 +39,9 @@
   .senders-cta {
     display: flex;
     flex-direction: row;
-    padding: $spacing-xs $spacing-sm;
+    padding: $spacing-sm;
+    position: relative;
+    left: $spacing-sm;
```
Vinnl, line 44: Hmm, I'm not really following why we're using relative positioning here (and what the `left` and `position: static` elsewhere are for) - as far as I can see the "Copied!" notice will just be displayed next to the button always? Feels like all those rules could be removed?

codemist: For the `copied-confirmation`, I just brought in the same class from the original masks page, but taking a look to prune what I can now (was sort of in a rush to get this dashboard in! lol)

codemist: I just looked over this PR again and pruned as much of the random positioning, I just left out the `left` with a comment about why I added that in (mainly to cancel out the padding left by the item)

frontend/src/components/phones/dashboard/PhoneDashboard.module.scss
```diff
@@ -39,14 +39,18 @@
...
-    gap: $spacing-xs;
-    margin-left: $spacing-sm;
 
-    &:hover {
-      background: $color-light-gray-10;
+    @media screen and #{$mq-md} {
+      padding: $spacing-sm;
+      gap: $spacing-xs;
+      left: $spacing-sm; // Cancel out extra space from padding
+      position: relative;
```
Vinnl, line 49: Does this not make more sense, rather than messing around with positioning?

```suggestion
      padding: $spacing-sm;
      padding-right: 0;
      gap: $spacing-xs;
```

Or even:

```suggestion
      padding-block: $spacing-sm;
      gap: $spacing-xs;
```

With potentially a `gap` on the parent, if needed.

Vinnl (approved):
Yep, that's easier to follow, although there's still one more `position` I'd like to remove as well, if possible. :+1:

frontend/src/components/phones/dashboard/PhoneDashboard.module.scss
codemist: I did that because I wanted consistent padding around the button, with your implementation it looks like this on hover because of the omitted `padding-right` 
image

codemist: (that is the hover state on Figma :\ I can talk to Eduardo about keeping hover states simplier so we don't need to do wacky CSS on it)

PR #2383, opened 2022-08-23
Remove moment.js

Vinnl:
# New feature description

In response to @lloan's question [here](https://github.com/mozilla/fx-private-relay/pull/2344#pullrequestreview-1077866078):

> LGTM - only comment I have is about the process of picking up new libraries, such as moment. Do we have a process on how we vet and approve libraries @groovecoder @Vinnl? I only ask because I know we're in the security and privacy field and want to make sure we don't start accumulating a load of dependencies on 3rd party libraries if we can. If we have to, it would be great to know that we did our due diligence in vetting out libraries and that we are cognizant that we have to keep them updated.

I'm not aware of a particular process we have. However, Moment.js wasn't originally intended for use in the browser and is really big. But more important is [its project status](https://momentjs.com/docs/#/-project-status/):

> We now generally consider Moment to be a legacy project in maintenance mode. (...) In most cases, you should not choose Moment for new projects.

Thus, I think it's pretty safe to safe that regardless of what process we'd have, we wouldn't end up picking Moment.js :) Common alternatives are [`date-fns`](https://www.npmjs.com/package/date-fns) and [`dayjs`](https://www.npmjs.com/package/dayjs). However, since we're already not supporting IE, I think we're safe just using the native [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat). The only thing that can't quite do, is list dates as "Yesterday at 13:37" ([`Intl.RelativeTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) is close, but not quite that). I expect that we can do without that for a better experience on mobile (faster page loads), but otherwise we could look into using either of those libraries for that anyway.

(I also removed some ESLint exceptions that were no longer necessary.)

# Screenshot (if applicable)

Not applicable.

# How to test

The "date created" on the phone dashboard and the "latest activity" on the call and SMS log should still be displayed properly.

# Checklist

- [ ] All acceptance criteria are met. - Yesterday and today's date are now formatted differently.


frontend/src/apiMocks/mockData.ts
```diff
@@ -329,5 +329,31 @@ export const mockedInboundContacts: Record<
       num_texts_blocked: 18,
       blocked: false,
     },
+    {
```
Vinnl, line 332: I added these to show the "Yesterday at 13:37" lines. (Which are no longer shown as such after this PR.)

codemist (approved):
Rip Moment :')

Displays dates fine for me, just in a different format than what was in the design spec. 

image

v. 

image


Since this is MVP it probably doesn't matter.

codemist:
I had no idea `moment` was a legacy project in maintenance mode :( Looking into `date-fns` now as an alternative.

Vinnl:
@codemist Thanks. Alternatively, since we were just doing it for yesterday and today, we can also just manually detect those two cases and add two new strings (`Yesterday at { $time }` and `Today at { $time }`), and insert the formatted time strings into those.

PR #2384, opened 2022-08-23
Use a table for tabular log data

Vinnl:
Otherwise screen readers will announce the entire &lt;li&gt; as a single
item, disconnected from the headings. (If you're on MacOS, try enabling VoiceOver, and compare how it reads the log data before this change and after this change.)

How to test: the call and SMS log should be roughly the same, but when inspecting it through a screen reader, the separate cells should be read separately, with the associated table headings announced as well.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A

### DIFF
--- a/frontend/src/components/phones/dashboard/PhoneDashboard.module.scss
+++ b/frontend/src/components/phones/dashboard/PhoneDashboard.module.scss
@@ -135,6 +135,26 @@
       margin-top: $spacing-2xl;
     }
 
+    thead {
+      // On mobile, the table heading only exists for screen readers.
+      // See https://webaim.org/techniques/css/invisiblecontent/#offscreen.
+      position: absolute;
+      width: 1px;
+      height: 1px;
+      overflow: hidden;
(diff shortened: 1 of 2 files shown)


codemist (changes requested):
I just tested this on my local and the mobile view of the panel disappeared. May be missing some of the media 

https://user-images.githubusercontent.com/13066134/186183350-79520f0d-a2f5-44a4-9ba5-2834c2073ea1.mov

Vinnl:
Hmm, looks like something went wrong when I tried to split this out from my other work - rebasing on `main` fixed it.

codemist:
I checked out this branch again, rebased and I'm still getting a blank mobile view: https://github.com/mozilla/fx-private-relay/pull/2384

Vinnl:
So weird @codemist! How about now? There was a `display: none` that inadvertently also applied to the first element in the log, so maybe your log just included a single line?

codemist:
Works for me now! 
image

codemist (approved):

PR #2385, opened 2022-08-23
Improvements for phone masking 

lloan:
This PR fixes MPP-2280  

# New feature description

MPP-2279 actually deals with the drop-down for city/name when choosing a Relay Number - that is not handled here. I mistakenly worked on 2280 + a number of items caught during that work - listed below:

- Real phone number input takes only numbers now 
- Real phone number input can take Canadian numbers and US numbers (limited to these countries)
- Formatting for error message logic added into `formatPhone` (validates length and deals with those numbers)
- Simple test added for `formatPhone`
- New library added in older PR to format phones used to deal with phone formatting. 
- We limit the max length of this input
- Relay number suggestions step was showing blank when running out of numbers. Now it pulls a new list once we run out. 

# Screenshot (if applicable)

Canadian Number:
image

US Number: 
image

# How to test
Go to the /phone page. 
Try entering a CA number +1250xxxxxxx
Try entering a US number +1909xxxxxxx
Try entering letters or non-digit characters
Confirm a real number and get to the relay number step - spam click the `get more options` button - once it runs out, it should load more suggestions. 

# Checklist


frontend/src/components/phones/onboarding/RealPhoneSetup.module.scss
```diff
@@ -274,3 +260,162 @@
     text-align: center;
   }
 }
+
+// 3rd party custom input for verifying true phone number
+/* stylelint-disable selector-class-pattern */
+.step :global {
+  .PhoneInput {
+    /* This is done to stretch the contents of this component. */
```
lloan, line 268: @Vinnl this CSS comes from a 3rd party package we are using. We can't modify the CSS classes directly. Same issue we had with the Toast component. Let me know if you want to approach this a different way. We might want to revisit some of these variables.

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
```diff
@@ -120,17 +119,26 @@ const RealPhoneForm = (props: RealPhoneFormProps) => {
...
+          className={`${styles["PhoneInput"]} ${
+            phoneNumberError ? "is-error" : ""
+          }`}
           placeholder={l10n.getString(
             "phone-onboarding-step2-input-placeholder"
           )}
-          ref={phoneNumberField}
+          countries={["US", "CA"]}
+          flags={flags}
+          defaultCountry="US"
```
lloan, line 131: @groovecoder do we want the default country to be US?

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
```diff
@@ -120,17 +119,26 @@ const RealPhoneForm = (props: RealPhoneFormProps) => {
...
-          className={`${styles["c-verify-phone-input"]}`}
+        <PhoneInput
+          className={`${styles["PhoneInput"]} ${
+            phoneNumberError ? "is-error" : ""
+          }`}
           placeholder={l10n.getString(
             "phone-onboarding-step2-input-placeholder"
           )}
-          ref={phoneNumberField}
+          countries={["US", "CA"]}
```
lloan, line 129: @Vinnl @groovecoder I have not tested using a number for another country.

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
```diff
@@ -90,39 +100,38 @@ const RelayNumberSelection = (props: RelayNumberSelectionProps) => {
...
+    _getRelayNumberSuggestions();
+  }, []);
 
-      if (respObject.same_area_options) {
-        for (const sameAreaOption of respObject.same_area_options) {
-          suggestedNumbers.push(sameAreaOption.phone_number);
-        }
-      }
+  // GET request to get relay number suggestions
+  const _getRelayNumberSuggestions = () => {
```
lloan, line 110: @Vinnl @groovecoder I'm not great at names... let me know if you have something better.

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
Vinnl: I don't think numbers from other countries are supposed to work... With that in mind, would it be possible to remove the "International" option from the drop-down?

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
```diff
@@ -171,7 +179,7 @@ const RealPhoneVerification = (props: RealPhoneVerificationProps) => {
     useState<boolean>();
 
   const onChange: ChangeEventHandler<HTMLInputElement> = (event) => {
-    setVerificationCode(event.target.value);
+    setVerificationCode(event.currentTarget.value.replace(/[^0-9]/gi, ""));
```
Vinnl, line 182: Regexes are always somewhat complex to parse, so could you help the reader by adding a comment describing what this does? If I'm following it correctly, my guess would be that it removes everything that's not a number from the string? (Also, I think the `i` flag is not necessary, since numbers are case-insensitive?)

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
```diff
@@ -279,9 +287,10 @@ const RealPhoneVerification = (props: RealPhoneVerificationProps) => {
           placeholder={codeEntryPlaceholder}
           required={true}
           maxLength={6}
-          onChange={onChange}
           className={isVerifiedSuccessfully === false ? styles["is-error"] : ""}
+          onChange={onChange}
```
Vinnl, line 291: If we're stripping out everything that's not a number anyway, might as well guide the user by popping up a number pad on mobile:

```suggestion
          onChange={onChange}
          inputMode="numeric"
```

See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode.

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
Vinnl: That said, I think generally it's considered bad UX practice to modify the user's input while they're inputting it. For example, what sometimes happens for me is that I'm practically blindly typing something, realise I'm mistyping something, and pressing Backspace right away to correct it. With the current setup, that would result in my entering an invalid value.

IIRC, better is to just provide inline validation hints. In this case, I expect that setting `minLength={6}` in addition to the `maxLength` will probably be enough to ensure that the user enters the correct code.

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
```diff
@@ -73,12 +77,7 @@ const RealPhoneForm = (props: RealPhoneFormProps) => {
 
   const [phoneNumber, setPhoneNumber] = useState("");
   const [phoneNumberSubmitted, setPhoneNumberSubmitted] = useState("");
-  const errorMessage = useRef<HTMLDivElement>(null);
-  const phoneNumberField = useRef<HTMLInputElement>(null);
```
Vinnl, line 77: Nice job removing some unnecessary refs :clap:

frontend/src/components/phones/onboarding/RealPhoneSetup.module.scss
```diff
@@ -274,3 +260,162 @@
     text-align: center;
   }
 }
+
+// 3rd party custom input for verifying true phone number
+/* stylelint-disable selector-class-pattern */
+.step :global {
```
Vinnl, line 266: Are all these styles just copy-pasted from the component? If so, could we not just do, like mentioned in the README:

```js
import "react-phone-number-input/style.css";
```

?

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
```diff
@@ -120,17 +119,26 @@ const RealPhoneForm = (props: RealPhoneFormProps) => {
...
+          defaultCountry="US"
           type="tel"
+          international
+          autoComplete="tel"
+          limitMaxLength
+          value={phoneNumber}
           required={true}
-          onChange={onChange}
           autoFocus={true}
+          onChange={(number: E164Number) => setPhoneNumber(number ?? "")}
```
Vinnl, line 139: Doesn't this _always_ get a string?

```suggestion
          onChange={(number: E164Number) => setPhoneNumber(number)}
```

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
```diff
@@ -81,6 +82,15 @@ const RelayNumberIntro = (props: RelayNumberIntroProps) => {
...
   registerRelayNumber: (phoneNumber: string) => Promise<Response>;
 };
+type RelayNumberOption = {
+  friendly_name: string;
+  iso_country: string;
+  locality: string;
+  phone_number: string;
+  postal_code: string;
+  region: string;
+};
```
Vinnl, line 92: This structure is determined directly by the API, right? If so, I'd co-locate it with everything API-related we have, i.e. in `/frontend/src/hooks/api/relayNumber.ts`.

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
Vinnl: I think what I'd do here is replace the original `getRelayNumberSuggestions` in `/frontend/src/hooks/api/relayNumber.ts` with an SWR hook just like we do for other API endpoints, e.g. something like

```ts
export function useRelayNumberSuggestions(): SWRResponse<RelayNumberOption[], unknown> {
  const relayNumberSuggestions = useApiV1("/relaynumber/suggestions/");
  return relayNumberSuggestions;
}
```

Then you can simply include that hook in this components, and it will make the request as soon as it renders, without you having to manually keep track of whether it's loading and all that, nor having to include a wrapper that does roughly the same and hence has a very similar name. Should save a bit of the awkwardness here, I think.

frontend/src/functions/formatPhone.test.ts
```diff
@@ -0,0 +1,13 @@
+import { formatPhone } from "./formatPhone";
```
Vinnl, line 1: Unit tests!!!!111ONEONE :tada:

frontend/src/functions/formatPhone.test.ts
```diff
@@ -0,0 +1,13 @@
...
+
+it("returns a formatted number even if not a valid length", () => {
+  const phoneNumber = "+122244";
+
+  expect(formatPhone(phoneNumber)).toStrictEqual("+1 (222) 44-");
```
Vinnl, line 6: [`toStrictEqual`](https://jestjs.io/docs/expect#tostrictequalvalue) is intended to compare objects, making sure that both sides of the comparison have the same properties. Since this is just a string, [`toBe`](https://jestjs.io/docs/expect#tobevalue) should be sufficient. It's not a biggie in a single case (it's mostly a performance concern, to avoid our unit tests from running too long), but it's good to get in the habit of using the right matchers in the right case, especially since people might copy-paste your unit tests to base their own future tests on :)

```suggestion
  expect(formatPhone(phoneNumber)).toBe("+1 (222) 44-");
```

(Also for the other test below.)

frontend/src/functions/formatPhone.ts
```diff
@@ -18,6 +18,17 @@ export type PhoneString = string;
 export function formatPhone(phoneNumber: PhoneString) {
   const parsedNumber = parsePhoneNumber(phoneNumber);
 
+  // This only checks for valid length of the phone number.
+  // Ex: ('+12223333333') === true
+  if (!parsedNumber?.isPossible()) {
+    // Bug: Any country code with two digits will break
```
Vinnl, line 24: Could be a helpful clarification here to indicate why this bug isn't fixed yet, and when it should be:

```suggestion
    // Bug: Any country code with two digits will break
    //         This is fine as long as we only support `+1`.
```

frontend/src/functions/formatPhone.ts
```diff
@@ -18,6 +18,17 @@ export type PhoneString = string;
 export function formatPhone(phoneNumber: PhoneString) {
   const parsedNumber = parsePhoneNumber(phoneNumber);
 
+  // This only checks for valid length of the phone number.
+  // Ex: ('+12223333333') === true
+  if (!parsedNumber?.isPossible()) {
```
Vinnl, line 23: I'm not following why we need this `if` block - if the phone number is not a valid phone number, why would be bother trying to format it? And if it's not valid, is it not possible that the input is something that would break the contents of the `if`?

Vinnl (commented):
I'm still kinda wrapping my head around everything what happens here, and am finding it hard to identify which of my comments are blockers or not, sorry. I guess that means you can take a look at my comments and use your own judgment, then re-request a review and then I'll probably defer to that :)

frontend/src/components/phones/onboarding/RealPhoneSetup.module.scss
```diff
@@ -274,3 +260,162 @@
...
+    margin-right: var(--PhoneInputCountrySelect-marginRight);
+  }
+
+  .PhoneInputCountrySelect {
+    position: absolute;
+    top: 0;
+    left: 0;
+    height: 100%;
+    width: 100%;
+    z-index: 1;
```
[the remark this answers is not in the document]
lloan: @Vinnl as I mentioned above: this CSS comes from a 3rd party package we are using. We can't modify the CSS classes directly. Same issue we had with the Toast component since we're using CSS modules. I didn't write the CSS inside the `:global` - do you have a better idea on how we want to pull it from the package to use it here?

frontend/src/components/phones/onboarding/RealPhoneSetup.module.scss
```diff
@@ -274,3 +260,162 @@
...
+    border: 0;
+    opacity: 0;
+    cursor: pointer;
+  }
+
+  .PhoneInputCountrySelect[disabled],
+  .PhoneInputCountrySelect[readonly] {
```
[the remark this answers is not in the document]
lloan: @Vinnl as I mentioned above: this CSS comes from a 3rd party package we are using. I didn't write the CSS inside the `:global` - do you have a better idea on how we want to pull it from the package to use it here?

frontend/src/components/phones/onboarding/RealPhoneSetup.module.scss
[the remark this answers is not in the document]
Vinnl: Yeah I guessed something like that later - see https://github.com/mozilla/fx-private-relay/pull/2385#discussion_r953681101.

frontend/src/functions/formatPhone.ts
lloan: @Vinnl the function checks for length, that's it. If we know its an invalid number, we still want to format the pieces we do have. Do you have a different approach you would like us to try?

image

frontend/src/functions/formatPhone.ts
lloan: comment updated

frontend/src/functions/formatPhone.ts
Vinnl: Ah gotcha, thanks for the screenshot! No that makes sense, I didn't realise that we'd also show invalid phone numbers in the UI (in the error message stating that it is so - duh :sweat_smile: ).

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
lloan: Weird - it was complaining before about it being potentially undefined - updated.

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
lloan: Spoke to Luke about this one. He decided to have us just cycle through the initial batch we get from the API. So I'm taking the hook route. Still keeping the spinner though, with Maxx's implementation - I like the implementation - we remove the spinner exactly when we want it to go away. However, if you want me to change it, I can open another PR.

frontend/src/components/phones/onboarding/RealPhoneSetup.module.scss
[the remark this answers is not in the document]
lloan: done - pretty sure I missed that - feels a lot cleaner, at least now we don't have to see it in our styling, lol.

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
lloan: image
image

I added minLength, pattern, onInvalid and title. Feels a bit better. This gives us a tooltip for when user hovers over the input. This also makes it so that when the user submits this and it doesn't match the pattern, it lets them know "Enter 6-digit code".

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
```diff
@@ -120,17 +119,26 @@ const RealPhoneForm = (props: RealPhoneFormProps) => {
...
+        <PhoneInput
+          className={`${styles["PhoneInput"]} ${
+            phoneNumberError ? "is-error" : ""
+          }`}
           placeholder={l10n.getString(
             "phone-onboarding-step2-input-placeholder"
           )}
-          ref={phoneNumberField}
+          countries={["US", "CA"]}
+          flags={flags}
```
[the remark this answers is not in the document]
lloan: I went ahead and added 2 flags within our /public folder as /public/images/countryFlags then passed that URL structure into the component's `flagUrl` prop. Added a comment. Component works just as well.

frontend/src/components/phones/onboarding/RealPhoneSetup.module.scss
```diff
@@ -274,3 +260,162 @@
...
+
+  .PhoneInputCountry {
+    position: relative;
+    align-self: stretch;
+    display: flex;
+    align-items: center;
+    margin-right: var(--PhoneInputCountrySelect-marginRight);
+  }
+
+  .PhoneInputCountrySelect {
```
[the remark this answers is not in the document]
lloan: Not tabbed into country picker: 
image

Tabbed into country picker: 
image

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
```diff
@@ -103,35 +95,26 @@ const RelayNumberSelection = (props: RelayNumberSelectionProps) => {
   const [relayNumberIndex, setRelayNumberIndex] = useState(0);
 
   useEffect(() => {
```
Vinnl, line 97: Now that the side effect (the GET request) is contained in a hook, there's no longer a need to do this in `useEffect` anymore.

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
```diff
@@ -103,35 +95,26 @@ const RelayNumberSelection = (props: RelayNumberSelectionProps) => {
...
 
-        setRelayNumberSuggestions(
-          suggestedNumbers.map(
-            (suggestion: RelayNumberOption) => suggestion.phone_number
-          )
-        );
+      // set the relay number suggestions - an array of strings
+      setRelayNumberSuggestions(
+        suggestedNumbers.map((suggestion) => suggestion.phone_number)
+      );
```
Vinnl, line 112: You also don't need a separate state variable for this if it's outside of the effect; you can just directly do:

```ts
const relayNumberSuggestions = `relayNumberSuggestionsData.data?.map(/* ... */) ?? [];
```

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
```diff
@@ -103,35 +95,26 @@ const RelayNumberSelection = (props: RelayNumberSelectionProps) => {
...
 
-        setIsLoading(false);
-      })
-      .catch((error) => {
-        // Do nothing, just keep the default values
-      });
-  };
+      // remove the loading spinner
+      setIsLoading(false);
```
Vinnl, line 115: Additionally, the `isLoading` state variable is no longer needed either - just replace every instance of `isLoading` with `!relayNumberSuggestionsData.data`.

An additional advantage is that if you do that in an `if`, TypeScript can infer that `.data` is _not_ `undefined` in the `else` branch (or after a `return` in that `if`).

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
Vinnl: Ah, I didn't see this remark because I was using the "Changes since last review" view, but again, really clean solution!

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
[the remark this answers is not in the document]
Vinnl: Yep, awesome, great job :)

Vinnl (approved):
Looking good! Some small inline comments, and the following:

I think the images in `/frontend/src/components/phones/onboarding/images/countryFlags/` are redundant now, no? (Since the same images are present in `/frontend/public/...`.)

Additionally, and I'm not sure if this just has to do with the mock data, but sometimes the number suggestions appear empty for me:

image

I thought that might have to do with the mock data returning an empty element, but as you can see, the `input[value]` does show a number (`+16245678910`) - it's just not listed in the `<label>`. So I'm guessing maybe `formatPhone` is returning an empty string for some reason?

Anyway, all fairly straightforward suggestions, so approving already. I've also [asked Eduardo](https://mozilla.slack.com/archives/C013CSYEL5T/p1661420541618729) about focus styles for the flag selector, so we (e.g. I :stuck_out_tongue:) can create a followup PR.

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
lloan: Done - yes, this is much cleaner. I think I was getting a react minification error 301/321 at some point, but I think it was my placement - I like this a lot better.

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
lloan: Done - I went ahead and implemented something very similar.

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
lloan: Cleaned up

frontend/src/hooks/api/relayNumber.ts
```diff
@@ -76,3 +76,30 @@ export async function getRelayNumberSuggestions(): Promise<Response> {
   });
   return response;
 }
```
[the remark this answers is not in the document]
lloan: removed

PR #2387, opened 2022-08-23
only import phones models if PHONES_ENABLED

groovecoder:
This PR fixes #<issue ID>.

Recreate the prod issue:
1. Check out `2022.08.16`
2. Set `PHONES_EANBLED=False`
3. `runserver`
   * [ ] You'll see the same error as prod: `RuntimeError: Model class phones.models.RealPhone doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.`

How to test the fix:

1. Check out `stagefix-2022.08.16.1`
2. Set `PHONES_ENABLED=False`
5. `runserver`
   * It should work fine

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -21,8 +21,6 @@ from django.utils.translation.trans_real import (
 
 from rest_framework.authtoken.models import Token
 
-from phones.models import InboundContact, RelayNumber
-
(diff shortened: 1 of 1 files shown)


say-yawn (approved):
👍 note to self to put imports behind flags!

PR #2388, opened 2022-08-23
only import phones.models if PHONES_ENABLED

groovecoder:
This PR fixes SVCE-743.

How to re-create prod canary error:
1. `git checkout 2022.08.16`
2. Set `PHONES_ENABLED=False`
3. `python manage.py runserver`
   * [ ] You should see the same canary error: `RuntimeError: Model class phones.models.RealPhone doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.`

How to test this fix:
1. `git checkout stage-fix-2022.08.16`
2. Set `PHONES_ENABLED=False`
3. `python manage.py runserver`
   * [ ] It should work fine

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -21,7 +21,6 @@ from django.utils.translation.trans_real import (
 
 from rest_framework.authtoken.models import Token
 
-from phones.models import InboundContact, RelayNumber
 
 emails_config = apps.get_app_config("emails")
 logger = logging.getLogger("events")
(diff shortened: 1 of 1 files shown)


say-yawn (approved):
Change LGTM. Having a separate conversation about have a pre-prod environment to test these kind of scenarios and also get QA testing to deploy changes/configs that will be on a prod. @jwhitlock is also creating a separate CI action that will test for when `PHONES_ENABLED` flag is `False`

PR #2392, opened 2022-08-24
Add Phone Masking What's New entry

codemist:
# New feature description

This adds the new What's New Entry for Phone Masking. 

# Screenshot (if applicable)

https://user-images.githubusercontent.com/13066134/186756783-a69d748c-5a58-424f-a91d-364bdf068e82.mov

image

Non-premium (w/o upgrade CTA):
image

Premium (w/ upgrade CTA)
image

# How to test

1. With a non-premium account, check that a new news item appears in the What's New module. The Phone Masking news item should show with an upgrade button.
2. With a premium account, run through the same step as 1. and check that the upgrade button does not show.

# Checklist


frontend/src/components/layout/navigation/whatsnew/WhatsNewContent.tsx
```diff
@@ -1,10 +1,25 @@
...
+import { useGaViewPing } from "../../../../hooks/gaViewPing";
 import styles from "./WhatsNewContent.module.scss";
 
 export type Props = {
   heading: string;
   description: string;
   image: string;
   videos?: Record<string, string>;
+  cta?: {
+    type?: "upgrade" | undefined;
```
Vinnl, line 17: The `?` roughly means ` | undefined`, so the latter is redundant :)

```suggestion
    type?: "upgrade";
```

frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
```diff
@@ -321,6 +325,40 @@ export const WhatsNewMenu = (props: Props) => {
...
+    dismissal: useLocalDismissal(`whatsnew-feature_phone_${props.profile.id}`),
+    announcementDate: {
+      year: 2022,
+      month: 10,
+      day: 11,
+    },
+  };
+
+  // Only show its announcement if phone masking is live:
+  if (isFlagActive(props.runtimeData, "phones")) {
```
Vinnl, line 358: I think we'll also probably only want to show this if the user hasn't already purchased the phones plan?

```suggestion
  if (isFlagActive(props.runtimeData, "phones") && !props.profile.has_phone) {
```

(Not sure if this is the correct check for that, but I presume you know which is if it's not :) )

frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
```diff
@@ -321,6 +325,40 @@ export const WhatsNewMenu = (props: Props) => {
...
+    snippet: l10n.getString("whatsnew-feature-phone-snippet"),
+    content: (
+      <WhatsNewContent
+        description={l10n.getString("whatsnew-feature-phone-description")}
+        heading={l10n.getString("whatsnew-feature-phone-header")}
+        image={PhoneMaskingHero.src}
+        cta={{
+          type: "upgrade",
+          content: l10n.getString("whatsnew-feature-phone-upgrade-cta"),
+          onClick: () => trackPurchaseStart(),
```
Vinnl, line 339: I'm wondering if we're doing this elsewhere for phones as well, but `trackPurchaseStart` sets a cookie `clicked-purchase`, so that we can [send an event to GA](https://github.com/mozilla/fx-private-relay/blob/1c5ee036cef7a6b702f1e57dacf28797d66fdc64/frontend/src/pages/accounts/profile.page.tsx#L70-L80) when someone completes the purchase of Premium.

However, that checks for `profile.has_premium`, so I think (?) it won't necessarily fire for phones, and if it does, it doesn't distinguish between regular Premium purchases and phones purchases. So maybe we should rename that `trackPremiumPurchaseStart` and add a new `trackPhonePurchaseStart` that works in a similar way, but for phones?

(Or remove this line altogether and create a separate ticket for that work - I don't know how important tracking these is.)

frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
Vinnl: Probably related, because I'm using a Premium account: now if I do see the News announcement, there's no CTA:

image

Though seeing that now, I guess maybe it's intentional to show it anyway? Doesn't look too weird.

frontend/pendingTranslations.ftl
```diff
@@ -191,6 +191,12 @@ phone-banner-resend-welcome-sms-cta = Resend welcome SMS
 phone-banner-resend-welcome-sms-title = Quick Tip
 phone-banner-resend-welcome-sms-body =  Remember to save the contact we shared with you by SMS to help you identify messages forwarded by { -brand-name-relay }. Can’t find it?  
 
+# Phone What's New
+whatsnew-feature-phone-header = Introducing phone number masking
+whatsnew-feature-phone-snippet = With { -brand-name-relay-premium }, you can now create a phone number mask that helps you get...
+whatsnew-feature-phone-description = With { -brand-name-relay-premium }, you can now create a phone number mask that helps you get texts and calls without revealing your true number.
```
Vinnl, line 197: It's a bit confusing that this says that Premium offers phone masks, even though the regular Premium plan doesn't.

frontend/src/components/layout/navigation/whatsnew/WhatsNewContent.tsx
Vinnl: That said, I feel like `<WhatsNewContent>` knows too much about the content of the CTA here; intuitively, I'd say the Call To Action could be anything, e.g. an upgrade link, but also just a link to an updated FAQ or something.

Wouldn't it be possible to move these conditions (`hasPremium`, `isPremiumAvailableInCountry`, etc.) into the `WhatsNewEntry` for the phone announcement, and have that use those conditions to determine what values it passes to the `cta` property?

```suggestion
```

frontend/src/components/layout/navigation/whatsnew/WhatsNewContent.tsx
```diff
@@ -1,10 +1,25 @@
...
+    target?: string;
+    content: string;
+    onClick?: () => void;
+    gaViewPing?: Parameters<typeof useGaViewPing>[0];
```
Vinnl, line 21: `gaViewPing` doesn't seem to be used at the moment? (Though maybe it should be?)

Vinnl (commented):
I think maybe content needs to take another look, since the wording (referring to regular Premium) is a bit confusing right now. I'd also `<WhatsNewContent>` to have fewer responsibilities that are dependent on the actual content passed into it - see my inline comment there.

frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
codemist: I decided to just remove it for now and create a separate ticket for that work as you've mentioned.

frontend/pendingTranslations.ftl
codemist: Hmm, I got a bit confused too. I thought upgrading to Premium would upgrade you to Phones as well. Thanks for the clarification.

frontend/src/components/layout/navigation/whatsnew/WhatsNewContent.tsx
codemist: Just got rid of it for now.

frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
```diff
@@ -84,6 +92,22 @@ export const WhatsNewMenu = (props: Props) => {
...
+    !props.profile.has_premium &&
+    isPremiumAvailableInCountry(props.runtimeData) ? (
+      <OutboundLink
+        to={getPhoneSubscribeLink(props.runtimeData)}
+        eventLabel={l10n.getString("whatsnew-feature-phone-upgrade-cta")}
+        className={styles.cta}
+      >
+        <span>{l10n.getString("whatsnew-feature-phone-upgrade-cta")}</span>
+      </OutboundLink>
+    ) : null;
```
codemist, line 109: I refactored this such that this element is getting passed to `WhatsNewContent` from `WhatsNewEntry`

frontend/pendingTranslations.ftl
codemist: Checking with Chris that the strings are correct. Since everything's in `pendingTranslations` it'll be easy to update though.

frontend/src/components/layout/navigation/whatsnew/WhatsNewMenu.tsx
Vinnl: OK, passing the whole component is even more generic, but works for me.

Vinnl (approved):
:+1: Don't forget to add Chris's updated string, the rest looks good.

PR #2393, opened 2022-08-24
fix MPP-2308: Reply to last sender

groovecoder:
This PR fixes MPP-2308.

How to test:
1. `PHONES_ENABLED=0 pytest` should run with no errors nor failures
2. `PHONES_ENABLED=1 pytest` should run with no errors nor failures
3. Set up [end-to-end local phone dev](https://github.com/mozilla/fx-private-relay/blob/main/docs/end-to-end-local-phone-dev.md)
4. Send your local relay number a text message
   * [x] You should receive the text message to your real phone
6. Reply to the text message sent by relay
   * [x] The original sender should receive the reply from the relay number
7. As the original sender, reply to the reply
   * [x] You should receive the reply to the reply to your real phone
8. Repeat as desired


jwhitlock (commented):
When I checked [PR 2330](https://github.com/mozilla/fx-private-relay/pull/2330), I would send a message to my Relay number from my real number (555-867-5309 in this example), and would get a response:

```
[Relay 📲 +15558675309]
Message to check PR 2330
```

With this PR, I get:

```
You can only reply if you allow
Firefox Relay to keep a log of
your callers and text senders
```

There is then a link to relay.firefox.com

* Do I need a second phone number to test relay now?
* Should the link be to my server instead of relay.firefox.com?

groovecoder:
> When I checked [PR 2330](https://github.com/mozilla/fx-private-relay/pull/2330), I would send a message to my Relay number from my real number (555-867-5309 in this example), and would get a response:
> 
> ```
> [Relay 📲 +15558675309]
> Message to check PR 2330
> ```
> 
> With this PR, I get:
> 
> ```
> You can only reply if you allow
> Firefox Relay to keep a log of
> your callers and text senders
> ```
> 
> There is then a link to relay.firefox.com
> 
>     * Do I need a second phone number to test relay now?

Yes. Any message coming **FROM** a Relay customer's real number is now treated as a reply message. So you'll need to send message to your Relay Number from some other number. I use a Google Voice number for it. 

> 
>     * Should the link be to my server instead of relay.firefox.com?

Updated.

jwhitlock (commented):
Thanks! I used a second phone number, and got the forwarded message. However, when I replied to my relay mask number, the message "You can only reply if you allow Firefox Relay to keep a log of your callers and text senders." was not delivered. 

Twillio reports the message (with my site URL) was delivered, but dropped by the carrier with [Error 30007](https://www.twilio.com/docs/errors/30007)
 Your message content was flagged as going against carrier guidelines.
 
I use [AT&T ActiveArmor](https://www.att.com/security/). It is possible that it allowed the message with `relay.monitor.com` but blocked it with `my-app.ngrok.io` I have "Fraud Risk" set to "Block". I changed it to "Flag" to try again. I also find that it tends to break my internet connection, so I restarted my phone, which seems to fix it. Great app! However, changing to "Flag" in ActiveArmor did not deliver the message.

When I dropped the URL entirely, the message was delivered.

I'm curious if this is a jwhitlock issue, an AT&T issue, or AT&T + ActiveArmor issue. I'd like to see if others are blocked with an `ngrok.io` URL, and if it works in dev or stage.

Maybe include the URL if in production, but just say "Visit the Relay dashboard" otherwise?

api/tests/phones_views_tests.py
```diff
@@ -48,6 +55,16 @@ def mocked_twilio_validator():
         yield mock_twilio_validator
```
jwhitlock, line 55: Nit: Add a default `validate` returning `True` in the fixture?

```python
mocked_twilio_validator.validate = Mock(return_value=True)
```

You could then delete that line from tests using `mocked_twilio_validator`. For the tests returning `False`, you'd need:

```python
mocked_twilio_validator.validate.return_value = False
```

api/tests/phones_views_tests.py
```diff
@@ -752,10 +678,76 @@ def test_inbound_sms_valid_twilio_signature_blocked_contact(
     assert inbound_contact.last_inbound_date == pre_block_contact_date
 
 
+def test_inbound_sms_reply_not_storing_phone_log(phone_user, mocked_twilio_client):
+    mocked_twilio_validator.validate = Mock(return_value=True)
```
jwhitlock, line 682: This test doesn't use `mocked_twilio_validator`, so I think this is unneeded

api/tests/phones_views_tests.py
```diff
@@ -752,10 +678,76 @@ def test_inbound_sms_valid_twilio_signature_blocked_contact(
...
+    mocked_twilio_client.messages.create.assert_called_once()
+    call_kwargs = mocked_twilio_client.messages.create.call_args.kwargs
+    assert call_kwargs["to"] == real_phone.number
+    assert call_kwargs["from_"] == relay_number.number
+    assert "You can only reply" in call_kwargs["body"]
+    assert f"{settings.SITE_ORIGIN}" in call_kwargs["body"]
+
+
+def test_inbound_sms_reply_no_previous_sender(phone_user, mocked_twilio_client):
+    mocked_twilio_validator.validate = Mock(return_value=True)
```
jwhitlock, line 707: Same, ` mocked_twilio_validator` is not passed as a fixture

api/tests/phones_views_tests.py
```diff
@@ -752,10 +678,76 @@ def test_inbound_sms_valid_twilio_signature_blocked_contact(
...
+    assert "Could Not Find A Previous Text Sender" in decoded_content
+    mocked_twilio_client.messages.create.assert_called_once()
+    call_kwargs = mocked_twilio_client.messages.create.call_args.kwargs
+    assert call_kwargs["to"] == real_phone.number
+    assert call_kwargs["from_"] == relay_number.number
+    assert "Could not find a previous text sender" in call_kwargs["body"]
+
+
+def test_inbound_sms_reply(phone_user, mocked_twilio_client):
+    mocked_twilio_validator.validate = Mock(return_value=True)
```
jwhitlock, line 728: and again.

Unless the test is supposed to use the `mocked_twilio_validator` fixture, and there is some weirdness going on...

api/views/phones.py
```diff
@@ -464,28 +476,62 @@ def inbound_call(request):
...
+        body=inbound_body,
+        to=last_text_sender.inbound_number,
+    )
+
+
+def _check_disabled(relay_number, contact_type):
     # Check if RelayNumber is disabled
     if not relay_number.enabled:
-        raise exceptions.ValidationError(f"Number is not accepting {contact_type}.")
+        raise exceptions.ValidationError(f"Number is not accepting {contact_type}s.")
```
jwhitlock, line 521: Good catch!

emails/models.py
```diff
@@ -415,6 +415,11 @@ def is_flagged(self):
         return True
 
 
+def get_storing_phone_log(relay_number):
+    profile = Profile.objects.get(user=relay_number.user)
+    return profile.store_phone_log
```
jwhitlock, line 420: Nit: change to a one-liner?:

```suggestion
    return relay_number.user.profile_set.get().store_phone_log
```

phones/models.py
```diff
@@ -19,6 +20,8 @@
     ("text", "text"),
 ]
 
+logger = logging.getLogger("events")
```
jwhitlock, line 23: The `logger` appears unused - remove it?

jwhitlock (approved):
Thanks @groovecoder! The steps worked for me, once I removed the URL from the "You can't reply" message. I'll leave it up to you if you want to do something about that now, or see if it replicates on the dev server.

There's a few nits and code cleanup suggestions, but they are in tests, not the application code. Feel free to re-ping me if you want a second review after cleanup.

phones/models.py
```diff
@@ -76,6 +79,16 @@ def get_valid_realphone_verification_record(user, number, verification_code):
     ).first()
 
 
+def get_last_text_sender(relay_number):
+    return (
+        InboundContact.objects.filter(
+            relay_number=relay_number, last_inbound_type="text"
+        )
+        .order_by("-last_inbound_date")
+        .first()
```
jwhitlock, line 88: Maybe use `latest`?

```suggestion
        .latest("last_inbound_date")
```

groovecoder: Aha! That's why we should have non-null fields!! :trollface:

api/tests/phones_views_tests.py
```diff
@@ -48,6 +55,16 @@ def mocked_twilio_validator():
         yield mock_twilio_validator
 
 
+def _make_real_phone(phone_user, **kwargs):
```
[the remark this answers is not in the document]
groovecoder: maybe easier after `RealPhone` is always included? got this when I started down this route:
image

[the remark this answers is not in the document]
groovecoder: because we have `if settings.PHONES_ENABLED: from phones.models import RealPhone`

groovecoder:
@jwhitlock I think this is ready for final review!

jwhitlock (approved):
Thanks for the changes @groovecoder! Please run `black` and then merge!

PR #2395, opened 2022-08-26
MPP-2310 Remove ability to resubmit existing number for verification

lloan:
This PR fixes MPP-2310

# New feature description

A user was able to enter an already used phone number on the “Verify your true phone number” screen

# Screenshot (if applicable)

image

# How to test

- Go to the Relay dashboard;- 
- Click on the “Phone Number” button from the site header;
- Click on the “Upgrade Now” button;
- Go through the purchase flow;
- Return to the “Phone Number” tab;
- Enter the already used phone number from the first account from the prerequisites and proceed;
- Observe the results;

#### Prerequisites:

Have a Premium Relay account that has a Phone Number Mask set;

Have a Premium Relay account that does NOT have the Phone Number subscription purchased;

Be signed in to the second Relay account;

#### Expected result:

The user is displayed an error message 

#### Actual result:

The user is sent the verification code and redirected to the “Verify your true phone number” screen;

# Checklist

### DIFF
--- a/frontend/src/components/phones/onboarding/PhoneOnboarding.tsx
+++ b/frontend/src/components/phones/onboarding/PhoneOnboarding.tsx
@@ -23,7 +23,7 @@ export const PhoneOnboarding = (props: Props) => {
 
   // Make sure profile and runtime data are available
   if (profiles.data?.[0] === undefined || !runtimeData.data) {
-    return null;
+    return <>TODO: Profile Loading/Error</>;
   }
 
   // Show Upgrade Prompt - User has not yet purchased phone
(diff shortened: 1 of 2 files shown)


lloan:
https://github.com/mozilla/fx-private-relay/blob/862e99371bb6730e0d31df43e3a500983e8e6754/frontend/src/components/phones/onboarding/RealPhoneSetup.tsx#L43

@Vinnl this and `!isEnteringNumber || phonesPendingVerification.length === 0` (triggered when we remove the phone number from a user, but they're not entering a number) is most likely why you might see a quick back and forth. What do you suggest? I might have to show you so you get a better idea.

Vinnl (approved):
The changes look good.

Thanks for the clarification @lloan. I think the reason you're seeing the error with the new number is because `phoneNumberError` is still `true` while waiting for the server's response, and in between `phoneNumber` already gets updated to the new value entered by the user. I think you should be able to deal with that by resetting it to `false` before sending the request with the new number (i.e. add `setPhoneNumberError(false)` to `onSubmit` before calling `requestPhoneVerification` and `requestPhoneRemoval`).

PR #2390, opened 2022-08-23
Add Email/Phone Mask Duo Mobile Nav

codemist:
# New feature description

Adds a Mask/Phone header on mobile. 

# Screenshot (if applicable)

https://user-images.githubusercontent.com/13066134/186454650-2e44e3cf-ba2e-43c8-bd72-e8102eb33863.mov

# How to test

Check that the mobile menu includes a mask/phone duo header that links to the masks and the phones page respectively.

# Checklist

- [x] l10n changes have been submitted to the l10n repository, if any.
- [x] All acceptance criteria are met.
- [x] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #2396, opened 2022-08-26
Make the dashboard switcher non-sticky

Vinnl:
This PR (that targets #2390) tackles two things at once:

- Ensures the Duo Mobile Nav is not visible on the settings, FAQ and interstitial pages when the user is logged in
- Makes the Duo Mobile Nav non-sticky again, to preserve screen real estate on mobile (discussed with Eduardo)

@codemist This PR will merge into your PR, so up to you if you're OK with this approach, would like to try another one, or even want to push back on whether the changes are necessary in the first place :)

### DIFF
--- a/frontend/pendingTranslations.ftl
+++ b/frontend/pendingTranslations.ftl
@@ -11,8 +11,9 @@ nav-settings = Settings
 nav-support = Help and Support
 nav-sign-out = Sign Out
 nav-contact = Contact Us
-nav-duo-email-mask-alt = Go to Email Masks Dashboard
-nav-duo-phone-mask-alt = Go to Phone Masks Dashboard
+nav-duo-description = Switch dashboards
+nav-duo-email-mask-alt = Email masks
+nav-duo-phone-mask-alt = Phone masks
 
 landing-pricing-free-feature-3 = Email tracker removal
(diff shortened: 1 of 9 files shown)


codemist:
LGTM - but just wondering if the dashboard switcher (dig that name btw!) should stay sticky?

https://user-images.githubusercontent.com/13066134/187132117-5d41fc39-4034-42fb-bdf8-c548d2aec302.mov

Vinnl:
@codemist Yeah, I discussed that with Eduardo, but it's taking up quite a bit of screen real estate, which is a scarce resource on small screens :) Ideally he'd have wanted it to show up sticky, but only when scrolling up, but I pushed back on that a bit because that's hard to do without negatively impacting performance (also a scarce resource on many mobile devices!), so he then said that non-sticky would be better:

image

(I brought it up separately from your PR because I didn't want to slow down that PR in case he didn't want it to change.)

codemist:
@Vinnl Works for me - we should probably then distinguish between 'phone' and 'email' dashboards in the regular Mobile navigation.

codemist (approved):

Vinnl:
@codemist Oh yeah, good point! And maybe

PR #2406, opened 2022-08-29
Dashboard banner to announce the end of introductory pricing

Vinnl:
# New feature description

This adds a banner to announce the end of the introductory pricing rate

# Screenshot (if applicable)

image

# How to test

Log in with an account that does not have Premium yet, while Premium is available, and with the `intro_pricing_countdown` flag enabled. You should see a countdown on the email mask dashboard, counting down to September 27th, 9AM PT.

# Checklist

- [ ] l10n changes have been submitted to the l10n repository, if any. - Not yet, the banner will also appear on a couple of other pages that should be quick to add, so I'll submit the strings after that.
- [ ] All acceptance criteria are met. - The upgrade button is filled, rather than outlined as in the design, and the date is localised by the browser and hence also includes the year (both changes [are approved](https://mozilla.slack.com/archives/C013CSYEL5T/p1661529448639169?thread_ts=1661527618.018519&cid=C013CSYEL5T)).

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -52,6 +52,7 @@
     "eslint-config-next": "^12.2.5",
     "eslint-plugin-jest-dom": "^4.0.2",
     "eslint-plugin-testing-library": "^5.6.0",
+    "fast-check": "^3.1.2",
     "husky": "^8.0.1",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^28.1.3",
--- a/frontend/src/components/Banner.tsx
+++ b/frontend/src/components/Banner.tsx
@@ -90,14 +90,16 @@ export const Banner = (props: BannerProps) => {
         <div className={`${styles["highlight-wrapper"]}`}>
           {illustration}
           {infoIcon}
-          <div className={`${styles["title-text"]}`}>
-            {title}
-            {props.children}
-            {/* A regular-sized button is shown close to the content… */}
-            {props.cta?.size !== "large" ? cta : null}
+          <div className={styles["title-and-large-cta-wrapper"]}>
+            <div className={`${styles["title-text"]}`}>
+              {title}
+              {props.children}
+              {/* A regular-sized button is shown close to the content… */}
+              {props.cta?.size !== "large" ? cta : null}
+            </div>
+            {/* …whereas a large button can stand on its own. */}
+            {props.cta?.size === "large" ? cta : null}
           </div>
-          {/* …whereas a large button can stand on its own. */}
-          {props.cta?.size === "large" ? cta : null}
         </div>
         {dismissButton}
       </div>
--- a/frontend/src/components/Banner.tsx
+++ b/frontend/src/components/Banner.tsx
@@ -16,12 +16,7 @@ export type BannerProps = {
     content: string;
     onClick?: () => void;
     gaViewPing?: Parameters<typeof useGaViewPing>[0];
-  };
-  ctaLargeButton?: {
-    target: string;
-    content: string;
-    onClick?: () => void;
-    gaViewPing?: Parameters<typeof useGaViewPing>[0];
+    size?: "medium" | "large";
   };
   btn?: {
     content: string;
@@ -77,7 +72,11 @@ export const Banner = (props: BannerProps) => {
   ) : null;
 
   const cta = props.cta ? (
-    <div className={styles.cta}>
+    <div
+      className={
+        props.cta.size === "large" ? styles["cta-large-button"] : styles.cta
+      }
+    >
       <OutboundLink
         to={props.cta.target}
         eventLabel={props.cta.content}
@@ -86,20 +85,6 @@ export const Banner = (props: BannerProps) => {
         onClick={props.cta.onClick}
       >
         <span ref={ctaRef}>{props.cta.content}</span>
-      </OutboundLink>
-    </div>
-  ) : null;
-
-  const ctaLargeButton = props.ctaLargeButton ? (
-    <div className={styles["cta-large-button"]}>
-      <OutboundLink
-        to={props.ctaLargeButton.target}
-        eventLabel={props.ctaLargeButton.content}
-        target="_blank"
-        rel="noopener noreferrer"
-        onClick={props.ctaLargeButton.onClick}
-      >
-        <span ref={ctaRef}>{props.ctaLargeButton.content}</span>
       </OutboundLink>
     </div>
   ) : null;
@@ -142,10 +127,10 @@ export const Banner = (props: BannerProps) => {
           <div className={`${styles["title-text"]}`}>
             {title}
             {props.children}
-            {cta}
+            {props.cta?.size !== "large" ? cta : null}
             {btn}
           </div>
-          {ctaLargeButton}
+          {props.cta?.size === "large" ? cta : null}
         </div>
         {dismissButton}
       </div>
--- a/frontend/src/components/Banner.tsx
+++ b/frontend/src/components/Banner.tsx
@@ -52,7 +52,9 @@ export const Banner = (props: BannerProps) => {
             {props.title}
           </h2>
         )) ||
-        (type === "info" && <h2 className={styles.title}>{props.title}</h2>)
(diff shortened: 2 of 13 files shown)


frontend/package.json
```diff
@@ -52,6 +52,7 @@
     "eslint-config-next": "^12.2.5",
     "eslint-plugin-jest-dom": "^4.0.2",
     "eslint-plugin-testing-library": "^5.6.0",
+    "fast-check": "^3.1.2",
```
Vinnl, line 55: This is a new dependency. I'll comment elsewhere what I'm using it for, but some reasons why I think it's a safe addition:

- It's only a `devDependency` that is only run during tests (which can still inspect and modify env vars, but shouldn't otherwise impact our shipped code).
- It only has a single transitive dependency.
- It has existed for quite a while already and has been quite stable in that time (I've also used this [at my previous employer](https://github.com/inrupt/solid-client-js/blob/24043d1bf3a09fe88c98a2224f4bbbf9075500b3/package.json#L177) without major issues.
- I've followed [the author on Twitter](https://twitter.com/ndubien/) for a while, who appears to be a real and trustworthy person :)

frontend/src/components/Banner.tsx
```diff
@@ -11,23 +11,7 @@ export type BannerProps = {
...
-    content: string;
-    onClick?: () => void;
-    gaViewPing?: Parameters<typeof useGaViewPing>[0];
-  };
-  btn?: {
-    content: string;
-    onClick?: () => void;
-    gaViewPing?: Parameters<typeof useGaViewPing>[0];
-  };
+  cta?: BannerCtaProps;
```
Vinnl, line 14: I refactored the props to `<Banner>` a bit to unify the `cta`, `ctaLarge` and `btn` props into a single `cta` prop.

- To get the `ctaLargeButton` behaviour, pass the same properties to `cta`, and additionally `size: "large"`.
- To get the `btn` behaviour, pass the same properties to `cta`. The lack of a `target` turns it into a `<button>` rather than an `<a>`.

frontend/src/components/Banner.tsx
```diff
@@ -69,48 +52,16 @@ export const Banner = (props: BannerProps) => {
             {props.title}
           </h2>
         )) ||
-        (type === "info" && <h2 className={styles.title}>{props.title}</h2>)
+        (["info", "promo"].includes(type) && (
```
Vinnl, line 55: Looks like we forgot to display the title for `promo` banners as well - this fixes that.

frontend/src/components/CountdownTimer.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+      )
+    ).toStrictEqual({
+      remainingDays: 5,
+      remainingHours: 4,
+      remainingMinutes: 3,
+      remainingSeconds: 59,
+    });
+  });
+
+  it("never returns a negative number of days", () => {
```
Vinnl, line 146: From this test onwards I'm using the new `fast-check` dependency. The reason I added them, was because I happened to be testing the `<CountdownTimer>` component at a point in time in which some of the numbers were negative numbers (see tests above), which got me nervous that other points in time might lead to similar issues.

`fast-check` is a library for "property-based testing". What it does is this: it generates lots of inputs (in this case, numbers between 0 and the number of milliseconds that are in 90 days `fc.nat({ max: 90 * 24 * 60 * 60 * 1000 })`), and verifies that certain properties hold (e.g. in this case, that the calculated number of remaining days is always a positive number). With these tests succeeding, I'm confident that we won't suddenly find out on September 26th that the timer is all messed up.

frontend/package.json
Vinnl, line 55: This is a new dependency. I'll comment elsewhere what I'm using it for, but some reasons why I think it's a safe addition:

- It's only a `devDependency` that is only run during tests (which can still inspect and modify env vars, but shouldn't otherwise impact our shipped code).
- The tests that it is used on can always be removed if we no longer want to move forward with this dependency.
- It only has a single transitive dependency (see `package-lock.json`).
- It has existed for quite a while already and has been quite stable in that time (I've also used this [at my previous employer](https://github.com/inrupt/solid-client-js/blob/24043d1bf3a09fe88c98a2224f4bbbf9075500b3/package.json#L177) without major issues.
- I've followed [the author on Twitter](https://twitter.com/ndubien/) for a while, who appears to be a real and trustworthy person :)

frontend/src/components/Banner.tsx
Vinnl, line 14: I refactored the props to `<Banner>` a bit to unify the `cta`, `ctaLarge` and `btn` props into a single `cta` prop.

- To get the `ctaLargeButton` behaviour, pass the same properties to `cta`, and additionally `size: "large"`.
- To get the `btn` behaviour, pass the same properties to `cta`. The lack of a `target` turns it into a `<button>` rather than an `<a>`.

frontend/src/components/Banner.tsx
```diff
@@ -154,3 +106,46 @@ export const Banner = (props: BannerProps) => {
...
   return null;
 };
+
+type BannerCtaProps = {
+  target?: string;
+  content: string;
+  onClick?: () => void;
+  gaViewPing?: Parameters<typeof useGaViewPing>[0];
+  size?: "medium" | "large";
+} & ({ target: string } | { onClick: () => void }); // At least one of `target` and `onClick` is required:
```
codemist, line 116: The condition of having either one of `target`/`onClick` is clever!

frontend/src/components/Banner.tsx
```diff
@@ -154,3 +106,46 @@ export const Banner = (props: BannerProps) => {
...
+        eventLabel={props.content}
+        target="_blank"
+        rel="noopener noreferrer"
+        onClick={props.onClick}
+      >
+        <span ref={ctaRef}>{props.content}</span>
+      </OutboundLink>
+    </div>
+  );
+};
```
codemist, line 151: non-blocking: I'd add some comments for this logic just to understand whats conditions are being checked to display which view.

frontend/src/components/CountdownTimer.module.scss
```diff
@@ -0,0 +1,40 @@
...
+  dl {
+    display: flex;
+    gap: $spacing-sm;
+
+    > div {
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+      gap: $spacing-xs;
+      width: 60px;
```
codemist, line 14: Could we use `$layout-lg` instead? It's `64px`.

frontend/src/components/CountdownTimer.module.scss
```diff
@@ -0,0 +1,40 @@
...
+      }
+
+      dd {
+        @include text-title-sm;
+        font-weight: 700;
+      }
+
+      @media (prefers-reduced-motion) {
+        &.remaining-seconds {
+          display: none;
```
codemist, line 35: 🤌

frontend/src/components/CountdownTimer.test.tsx
codemist: Woah, never seen a timer done like this before. Would you recommend I use this same approach for the addon?

codemist (approved):
Looks really good, mostly non-blocking comments.

frontend/src/components/Banner.tsx
codemist: Personally I'd leave this in a comment

lloan:
Removing self as Kaitlyn already reviewed :)

frontend/src/components/Banner.tsx
Vinnl: Yeap good callout. How about 2b3af820?

frontend/src/components/Banner.tsx
Vinnl: Good point, would help understanding what to change when in the future: a22e53b9

frontend/src/components/CountdownTimer.module.scss
Vinnl: Oh that's right, I'd forgotten about that one! 1410c3aa

PR #2407, opened 2022-08-29
Check number of emails forwarded per day

say-yawn:
When a user receives more than X number of emails, the user will get flagged for abuse and will pause for 24hrs before being able to:
- make aliases
- make replies
- receive emails

⚠️ If the user gets flagged for abuse from generating too much aliases or sending replies, they will also get paused from getting emails forwarded too.

How to test:

## Email received more than the limit results in account being flagged
1. Setup the env var `MAX_FORWARDED_PER_DAY` to some small positive integer like `5`
2. Run the app and send 5 emails to an alias
3. When the limit for emails relayed is hit, check that there is an`INFO` level log for `abusemetrics` that looks like:
```json
{
  "Timestamp": 1661815672630869000,
  "Type": "abusemetrics",
  "Logger": "fx-private-relay",
  "Severity": 6,
  "Pid": 4,
  "Fields": {
    "uid": "UUID-0912349523",
    "flagged": 1661815672.626743,
    "replies": 0,
    "addresses": 4,
    "forwarded": 5,
    "msg": "Abuse flagged"
  }
}
```
4. Check that the user is returned when running the following psql query:
```sql
SELECT * FROM emails_profile WHERE last_account_flagged IS NOT NULL
```
5. Check that you no longer can receive relayed emails, send replies, or create aliases.


emails/migrations/0051_abusemetrics_num_email_forwarded_per_day.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("emails", "0050_profile_store_phone_log"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="abusemetrics",
+            name="num_email_forwarded_per_day",
+            field=models.PositiveSmallIntegerField(default=0),
```
jwhitlock, line 16: You'll need to allow `null` or add a database-level default to avoid errors when you go to production. There are CircleCI tests for this, but they were not present in `2022.08.16.1`.

emails/models.py
```diff
@@ -355,7 +356,9 @@ def add_subdomain(self, subdomain):
         RegisteredSubdomain.objects.create(subdomain_hash=hash_subdomain(subdomain))
         return subdomain
 
-    def update_abuse_metric(self, address_created=False, replied=False):
+    def update_abuse_metric(
```
jwhitlock, line 359: Can you add tests for the new abuse parameter?

jwhitlock (changes requested):
Thanks @say-yawn! I'm not sure if we should be aiming for the current production tag or the one that is going out tomorrow. I'd feel better with tests and if we ran it on stage first.

PR #2409, opened 2022-08-30
2022.08.24 hotfix

say-yawn:
When a user receives more than X number of emails, the user will get flagged for abuse and will pause for 24hrs before being able to:
- make aliases
- make replies
- receive emails

⚠️ If the user gets flagged for abuse from generating too much aliases or sending replies, they will also get paused from getting emails forwarded too.

How to test:

## Email received more than the limit results in account being flagged
1. Setup the env var `MAX_FORWARDED_PER_DAY` to some small positive integer like `5`
2. Run the app and send 5 emails to an alias
3. When the limit for emails relayed is hit, check that there is an`INFO` level log for `abusemetrics` that looks like:
```json
{
  "Timestamp": 1661815672630869000,
  "Type": "abusemetrics",
  "Logger": "fx-private-relay",
  "Severity": 6,
  "Pid": 4,
  "Fields": {
    "uid": "UUID-0912349523",
    "flagged": 1661815672.626743,
    "replies": 0,
    "addresses": 4,
    "forwarded": 5,
    "forwarded_size_in_bytes": <some positive int value>
    "msg": "Abuse flagged"
  }
}
```
4. Check that the user is returned when running the following psql query:
```sql
SELECT * FROM emails_profile WHERE last_account_flagged IS NOT NULL
```
5. Check that you no longer can receive relayed emails, send replies, or create aliases.


emails/tests/models_tests.py
```diff
@@ -928,6 +929,40 @@ def test_profile_created_does_not_emit_metric_and_logs_from_measure_feature_usag
...
+    @override_settings(MAX_FORWARDED_PER_DAY=5)
+    def test_update_abuse_metric_flags_profile_when_abuse_threshold_met(
+        self, mocked_datetime, mocked_abuse_info
+    ):
+        expected_now = datetime.now(timezone.utc)
+        mocked_datetime.combine.return_value = datetime.combine(
+            datetime.now(timezone.utc).date(), datetime.min.time()
+        )
+        mocked_datetime.now.return_value = expected_now
+        mocked_datetime.side_effect = lambda *args, **kw: datetime(*args, **kw)
```
say-yawn, line 943: Used the example on Python documentation here to mock `datetime.now` without getting an error! https://docs.python.org/3/library/unittest.mock-examples.html#partial-mocking

say-yawn:
@jwhitlock the branch itself is based off of `2022.08.24` which is what is on Stage right now. It seems the merge to `main` will be a clean one but if I need to rebase the changes onto `main` I'm happy to do that too. Let me know!

emails/migrations/0051_abusemetrics_num_email_forwarded_per_day.py
```diff
@@ -0,0 +1,82 @@
...
+    if schema_editor.connection.vendor.startswith("postgres"):
+        schema_editor.execute(
+            'ALTER TABLE "emails_abusemetrics"'
+            ' ALTER COLUMN "num_email_forwarded_per_day" SET DEFAULT 0;'
+        )
+    elif schema_editor.connection.vendor.startswith("sqlite"):
+        schema_editor.execute(
+            'CREATE TABLE "new__emails_abusemetrics"'
+            ' ("id" integer NOT NULL PRIMARY KEY AUTOINCREMENT,'
+            ' "num_email_forwarded_per_day" smallint unsigned NOT NULL CHECK ("num_email_forwarded_per_day" >= 0),'
```
jwhitlock, line 22: This needs a default:

```suggestion
            ' "num_email_forwarded_per_day" smallint unsigned NOT NULL DEFAULT 0 CHECK ("num_email_forwarded_per_day" >= 0),'
```

jwhitlock (changes requested):
Thank you @say-yawn. I haven't tried this locally yet, but the code looks good.

I know this targets the number of forwarded emails, but I wonder if we can target the volume of emails by size. For example, if you send 1,000 1K emails, you're probably using as many resources as a single 1 MB email. It would be nice to limit to 50 - 100 MB a day or similar, regardless of how many emails that includes.

emails/views.py
```diff
@@ -534,6 +534,10 @@ def _sns_message(message_json):
         # an external sender to a relay user
         pass
 
+    # if account flagged for abuse, early return
+    if user_profile.is_flagged:
+        return HttpResponse("Address is temporarily disabled.")
+
```
groovecoder, line 540: suggestion (non-blocking): This should move above the check for reply from external sender to Relay user. If this is a malicious actor, they will be able to bypass this prevention by changing from plain emails to replies.

emails/utils.py
```diff
@@ -111,6 +117,8 @@ def ses_send_raw_email(
             ConfigurationSetName=settings.AWS_SES_CONFIGSET,
         )
         incr_if_enabled("ses_send_raw_email", 1)
+        profile = Profile.objects.filter(user__email=to_address).first()
+        profile.update_abuse_metric(email_forwarded=True)
```
groovecoder, line 121: suggestion (non-blocking): this adds an extra DB query to every email send operation. is there a way we can move this code to a place where we already have the `profile` object?

groovecoder (approved):
Code LGTM, but I would treat @jwhitlock 's review as more authoritative.

emails/views.py
say-yawn: Good suggestion! There is already a check to see if the user is flagged before a reply is sent:
https://github.com/mozilla/fx-private-relay/blob/1b4a66baabc0775a7d31490ce80bb7854b93652c/emails/views.py#L598-L602
Was added when I first added the abuse metrics.

emails/utils.py
say-yawn: Definitely possible, it just requires more code in different places that could call the `ses_send_raw_email`. I can do this if you think this should be a blocker.

say-yawn: John and I looked at the code and also noted that the `update_abuse_metrics` was double counting for replies too so we decided this was a blocking change and made the update on [5a4a83a](https://github.com/mozilla/fx-private-relay/pull/2409/commits/5a4a83a8913680e99eba00fbff093bc105fdb34e)

emails/models.py
```diff
@@ -375,27 +382,46 @@ def update_abuse_metric(self, address_created=False, replied=False):
...
+            abuse_metric.num_email_forwarded_per_day += 1
+        if forwarded_email_size > 0:
+            abuse_metric.forwarded_email_size_per_day += forwarded_email_size
         abuse_metric.last_recorded = datetime.now(timezone.utc)
         abuse_metric.save()
 
         # check user should be flagged for abuse
         hit_max_create = False
         hit_max_replies = False
+        hit_max_forwarded = False
```
jwhitlock, line 395: Nit: We're missing `hit_max_forwarded_email_size`, but I don't think we need any of the initializers

emails/tests/models_tests.py
```diff
@@ -928,6 +948,64 @@ def test_profile_created_does_not_emit_metric_and_logs_from_measure_feature_usag
         mocked_incr.assert_not_called()
         mocked_events_info.assert_not_called()
 
+    @patch("emails.models.abuse_logger.info")
+    @override_settings(MAX_FORWARDED_PER_DAY=5)
+    def test_update_abuse_metric_flags_profile_when_emails_forwarded_abuse_threshold_met(
```
jwhitlock, line 953: `AbuseMetrics` is tested! 🎉

emails/views.py
```diff
@@ -932,12 +942,13 @@ def _get_text_html_attachments(message_json):
...
         message_content = get_message_content_from_s3(bucket, object_key)
-        histogram_if_enabled("relayed_email.size", len(message_content))
-
+    email_size = len(message_content)
+    histogram_if_enabled("relayed_email.size", email_size)
     bytes_email_message = message_from_bytes(message_content, policy=policy.default)
 
     text_content, html_content, attachments = _get_all_contents(bytes_email_message)
-    return text_content, html_content, attachments
+    # TODO add logs on the entire size of the email and the time it takes to download/process
```
jwhitlock, line 950: Note: I asked for this `TODO`. We have a metric for how long it took to get the content from S3, and a histogram for the size, but we'd need to log both together to model how long it takes to download from S3 based on the object size. We're also curious if message size is important for email processing in general. It seemed out of scope for this hotfix, so we're deferring it.

privaterelay/settings.py
```diff
@@ -437,6 +437,8 @@ def _get_initial_middleware() -> list[str]:
 
 MAX_ADDRESS_CREATION_PER_DAY = config("MAX_ADDRESS_CREATION_PER_DAY", 100, cast=int)
 MAX_REPLIES_PER_DAY = config("MAX_REPLIES_PER_DAY", 100, cast=int)
+MAX_FORWARDED_PER_DAY = config("MAX_FORWARDED_PER_DAY", 1000, cast=int)
+MAX_FORWARDED_EMAIL_SIZE_PER_DAY = config("MAX_FORWARDED_EMAIL_SIZE_PER_DAY", 1000_000_000, cast=int)
```
jwhitlock, line 441: Nit: once more underscore
```suggestion
MAX_FORWARDED_EMAIL_SIZE_PER_DAY = config("MAX_FORWARDED_EMAIL_SIZE_PER_DAY", 1_000_000_000, cast=int)
```

jwhitlock (approved):
Thanks @say-yawn! I paired for this, so I approve. Looking in the GitHib diff, there are a handful of ignorable nits.

emails/models.py
say-yawn: Completed in: [81c5f02](https://github.com/mozilla/fx-private-relay/pull/2409/commits/81c5f0206cea99a1ed43e3eb70058473ba19f526)

jwhitlock (approved):

PR #2410, opened 2022-08-30
MPP-2279 Add area code search for relay number suggestions. 

lloan:
This PR fixes MPP-2279 

# New feature description

This PR adds the functionality of searching for numbers based on area code. Updated UI to reflect new changes in Figma. 

# Screenshot (if applicable)

image
                                                                           
# How to test

Get to the part in the onboarding process where you start choosing your Relay Number. Try using different area codes or city names in the top-most input. Submitting that input's information should update the numbers suggested. 

# Checklist


frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
```diff
@@ -82,29 +84,76 @@ const RelayNumberIntro = (props: RelayNumberIntroProps) => {
...
+    setIsSearching(true);
+
+    // submit string for search to our API
+    props
+      .search(searchValue)
+      .then((res) => res?.json())
+      .then((data) => {
+        if (data?.length !== 0) {
+          // empty out current relay number suggestions
+          relayNumberSuggestions.length = 0;
```
Vinnl, line 139: Blocking question: Woah, TIL that you can write to `Array.length`! That said, generally you don't want to directly write to state variables, but instead use the setter (`setRelayNumberSuggestions`) - otherwise the component won't re-render with the updated state. Is there a reason you're doing this? Feels like it should be directly overwritten by the next line. Maybe you just need a ` ?? []` after the `.map` to ensure that at least the current suggestions get removed if no suggestions were found?

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
```diff
@@ -82,29 +84,76 @@ const RelayNumberIntro = (props: RelayNumberIntroProps) => {
...
+    if (isSearching) return;
+
+    // set search to true as we begin searching
+    setIsSearching(true);
+
+    // submit string for search to our API
+    props
+      .search(searchValue)
+      .then((res) => res?.json())
+      .then((data) => {
```
Vinnl, line 136: Non-blocking suggestion: `data` is untyped here (because `.json()` resolves to `any`), which is a bit of a shame. Repercussions are limited, since it's only used here, but ideally, I'd like to see `search` to call `.json()` by itself and define a type representing that data, rather than returning an opaque `Response`.  That way, if we ever find we need to do validation on the response value, we can do so in one place and have TypeScript tell us if that breaks anywhere that `.search` is used, rather than having to do a search across the codebase and manually read and understand all usage sites.

frontend/src/hooks/api/relayNumber.ts
```diff
@@ -83,14 +83,44 @@ export type RelayNumberSuggestionsData = {
...
+    const searchParameter = !isNaN(+search)
+      ? `?area_code=${search}`
+      : `?location=${search}`;
+
+    // use api to search for relay number suggestions based on search body
+    const response = await apiFetch(`/relaynumber/search/${searchParameter}`, {
+      method: "GET",
+    });
+
+    relayNumberSuggestions.mutate();
```
Vinnl, line 120: I don't think the call to `/relaynummber/search/` would affect what the return value of `/relaynumber/suggestions` will be, right? If so, then there's no need to call `.mutate` on it, which in turn means that this function can just stand on its own rather than needing to be attached to the hook.

Vinnl (commented):
This looks good, with the one caveat that we shouldn't write directly to state variables. I'm hoping my suggested quick fix resolves the issue that was added for, otherwise we'll have to investigate some more what's going on and how to fix it in a way that won't break.

Also, let me bring up a pattern that is absolutely not a suggestion and therefore should not be taken as something to change, but rather as an attempt at knowledge sharing - it might come in useful in the future. Feel free to not even read everything from here on :smile: 

So: an alternative way to approach this would have been to, instead of manually calling the async `search` function, to instead also turn that in an SWR hook, and then create a `<SearchResults>` component that would take a search query as a prop, and would pass that into that hook. It might be interesting to give [the SWR Getting Started guide](https://swr.vercel.app/docs/getting-started) a quick look; it's not that long, and explains some of the ideas behind it.

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
lloan: Yeah, that's a remnant of something else I was trying - I was making it so that I could keep using a `const` within an array. I wanted to see if I could still mutate the value and you can definitely clear the array and keep pushing new values, lol. But I just removed that line and implemented all the other feedback. I'm still not sure what approach we are taking with the actual data for the drop-down. For now, I have a .json file. I'll implement it and see if we have a different source we want to use.

frontend/src/hooks/api/relayNumber.ts
```diff
@@ -94,3 +90,39 @@ export function useRelayNumberSuggestions(): SWRResponse<
...
+  // use api to search for relay number suggestions based on search body
+  const response = await apiFetch(`/relaynumber/search/${searchParameter}`, {
+    method: "GET",
+  });
+
+  if (!response.ok) {
+    throw new FetchError(response);
+  }
+
+  const data: RelayNumberSuggestion[] = await response.json();
```
Vinnl, line 125: Oh neat, I didn't realise this API just returned data in the same format as the suggestions API - that makes things easier :)

Vinnl (approved):
Yeah, the data source is still an issue - I guess it's a question for Tony if we want to pay for an up-to-date data source to support this feature, and for legal if there's a way to incorporate that in an open source project.

PR #2430, opened 2022-08-30
Use BadRequest instead of SuspiciousOperation for 400 errors

jwhitlock:
Use [BadRequest](https://docs.djangoproject.com/en/3.2/ref/exceptions/#badrequest), added in Django 3.2, to return a `400 Bad Request` response when the user can not perform the requested action. Previously, a `SuspiciousOperation` was used, which also sent an unneeded exception to Sentry.

Also, ignore `SuspiciousFileOperation` in Heroku, which is triggered by security scanners looking for interesting endpoints.

How to test:

- [ ] All tests run with `pytest .`

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -9,7 +9,7 @@ import uuid
 from django.apps import apps
 from django.conf import settings
 from django.contrib.auth.models import User
-from django.core.exceptions import SuspiciousOperation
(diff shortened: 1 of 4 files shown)


groovecoder (approved):
nice. thanks for this clean-up!

PR #2431, opened 2022-08-30
for MPP-2174: add text stats

groovecoder:
This PR fixes half (i.e., the "texts" stats) of MPP-2173.
image

How to test:
1. Make sure you have set up [local end-to-end phone dev](https://github.com/mozilla/fx-private-relay/blob/main/docs/end-to-end-local-phone-dev.md).
2. Run `python manage.py migrate` to add the new DB fields for stats
3. Go to http://127.0.0.1:8000/phone/
   * [x] Verify the stats show 50 minutes and 75 texts remaining
5. Send a text message to your relay number
   * [x] Verify the stats show 74 texts remaining, 1 calls and texts forwarded
6. Toggle your relay number to blocking
7. Send a text message to your relay number
   * [x] Verify the stats show 74 texts remaining, 1 calls and texts forwarded, 1 calls and texts forwarded

- [x] TODO: I've added a unit test to test for potential regressions of this bug.


api/views/phones.py
```diff
@@ -442,14 +442,17 @@ def inbound_sms(request):
...
-        _check_and_update_contact(inbound_contact, "texts")
+        _check_and_update_contact(inbound_contact, "texts", relay_number)
 
     client = twilio_client()
     client.messages.create(
         from_=relay_number.number,
         body=f"[Relay 📲 {inbound_from}] {inbound_body}",
         to=real_phone.number,
     )
+    relay_number.remaining_texts -= 1
```
groovecoder, line 453: TODO: (either here or somewhere else) add logic to:
1. check if `remaining_texts` need to be reset to the limit because we're in a new billing period since the customer's subscription started.
2. check if `remaining_texts` is 0 - don't forward the text and/or text the user that this is their last text for the period

privaterelay/settings.py
```diff
@@ -183,6 +183,8 @@
 MAX_MINUTES_TO_VERIFY_REAL_PHONE = config(
     "MAX_MINUTES_TO_VERIFY_REAL_PHONE", 5, cast=int
 )
+MAX_TEXTS_PER_BILLING_CYCLE = config("MAX_TEXTS_PER_BILLING_CYCLE", 75, cast=int)
+MAX_MINUTES_PER_BILLING_CYCLE = config("MAX_MINUTES_PER_BILLING_CYCLE", 50, cast=int)
```
groovecoder, line 187: Note: this may need to be "per month" since there will be annual billing cycles soon.

api/views/phones.py
jwhitlock: 3. and if `relay_number.remaining_texts` is `None`, reset to limit. Same for other places where `null` is allowed

privaterelay/views.py
```diff
@@ -275,10 +277,21 @@ def _update_all_data(social_account, extra_data, new_email):
             newly_premium = not had_premium and now_has_premium
             no_longer_premium = had_premium and not now_has_premium
             if newly_premium:
+                incr_if_enabled("user_purchased_premium", 1)
```
groovecoder, line 280: note: this fixes a long-standing issue where these aren't showing up in grafana.

api/views/phones.py
```diff
@@ -513,14 +522,44 @@ def _handle_sms_reply(relay_number, real_phone, inbound_body):
...
 def _check_disabled(relay_number, contact_type):
     # Check if RelayNumber is disabled
     if not relay_number.enabled:
+        attr = f"{contact_type}_blocked"
+        setattr(relay_number, attr, getattr(relay_number, attr) + 1)
+        relay_number.save()
         raise exceptions.ValidationError(f"Number is not accepting {contact_type}s.")
 
 
+def _check_remaining(relay_number, resource_type):
```
groovecoder, line 539: This is basically the logic here:
image

groovecoder: Oh, except step 1.5 in the top right: set last-checked date.

groovecoder: The top of this is moved into the new `update_phone_remaining_stats.py` command.

privaterelay/views.py
```diff
@@ -275,10 +277,20 @@ def _update_all_data(social_account, extra_data, new_email):
...
                 profile.date_subscribed = datetime.now(timezone.utc)
                 profile.save()
             if no_longer_premium:
                 incr_if_enabled("user_has_downgraded", 1)
+            now_has_phone = profile.has_phone
+            newly_phone = not had_phone and now_has_phone
+            no_longer_phone = had_phone and not now_has_phone
+            if newly_phone:
+                incr_if_enabled("user_purchased_phone", 1)
```
groovecoder, line 289: TODO?: Re-set the relay number remaining stats? So when someone purchases phones AGAIN after letting their phone subscription lapse, they immediately get texts & minutes back?

(Is my logic here right?)

phones/models.py
```diff
@@ -171,6 +171,24 @@ class RelayNumber(models.Model):
...
+        blank=True, null=True, default=settings.MAX_TEXTS_PER_BILLING_CYCLE
+    )
+    calls_forwarded = models.IntegerField(blank=True, null=True, default=0)
+    calls_blocked = models.IntegerField(blank=True, null=True, default=0)
+    texts_forwarded = models.IntegerField(blank=True, null=True, default=0)
+    texts_blocked = models.IntegerField(blank=True, null=True, default=0)
+
+    @property
+    def calls_and_texts_forwarded(self):
+        return self.calls_forwarded + self.texts_forwarded
```
jwhitlock, line 187: I appreciate you trying `null=True`, but it requires making the code that uses the nullable columns more careful. In `calls_and_texts_forwarded`, `self.calls_forwarded` can be `null`, and `self.texts_forwarded` can be null. `mypy` can catch this, if you add an annotation so it will check it rather than skip the function:

```python
    @property
    def calls_and_texts_forwarded(self) -> int:
        return self.calls_forwarded + self.texts_forwarded
```

which will result in the errors:

```
phones/models.py:187: error: Unsupported operand types for + ("int" and "None")  [operator]
phones/models.py:187: error: Unsupported operand types for + ("None" and "int")  [operator]
phones/models.py:187: error: Unsupported left operand type for + ("None")  [operator]
phones/models.py:187: note: Both left and right operands are unions
```

The fix:

```python
    @property
    def calls_and_texts_forwarded(self) -> int:
        return (self.calls_forwarded or 0) + (self.texts_forwarded or 0)
```

You'll need to add similar annotations and fixes to all the places that use ``calls_forwarded``, ``calls_blocked``, ``texts_forwarded``, ``texts_blocked``. I'm not sure if annotations will help with the API however, and some `null` values might get sent to the frontend.

Alternatively, set `null=False` and move the ugliness to the migrations.

jwhitlock (commented):
Thanks @groovecoder! I'd like to decide on `null=True` or column defaults before continuing with my review.

I hit an issue with test step 3, "Go to http://127.0.0.1:8000/phone/". If I went to that URL, I was immediately redirected to `/premium/`. I found the (undocumented) waffle flag `phones`, but this didn't solve my issue. Clicking on "Phone Number" in the top menu of the website did work. I reported this on Slack, the frontend devs are aware of it, and it should be fixed soon.

api/tests/phones_views_tests.py
```diff
@@ -608,10 +610,34 @@ def test_inbound_sms_valid_twilio_signature_disabled_number(
...
+    path = "/api/v1/inbound_sms"
+    data = {"From": "+15556660000", "To": relay_number.number, "Body": "test body"}
+    response = client.post(path, data, HTTP_X_TWILIO_SIGNATURE="valid")
+
+    assert response.status_code == 400
+    decoded_content = response.content.decode()
+    assert decoded_content.startswith("<?xml")
+    relay_number.refresh_from_db()
+    assert "Number Is Out Of Texts" in decoded_content
+    mocked_twilio_client.messages.create.assert_not_called()
```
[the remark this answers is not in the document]
groovecoder: No, it shouldn't. This is just texts blocked because the customer WANTED us to block them.

phones/models.py
groovecoder: Okay I updated the models to drop the blank/null values and squashed it into new `phones/migrations/0021_add_relaynumber_stats_20220913_1959.py` migration with default values in the migration too.

Be sure to migrate BACKWARD before you try this.

emails/migrations/0052_20220807_add_phone_sub_fields_to_profile.py
```diff
@@ -0,0 +1,26 @@
...
+            "emails",
+            "0051_add_email_size_per_day_and_num_email_forwarded_per_day_fields",
+        ),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="profile",
+            name="date_phone_subscription_checked",
+            field=models.DateTimeField(blank=True, null=True),
```
jwhitlock, line 19: Did you want to convert this one to column-level defaults as well?

groovecoder: I thougth about it briefly ... I *think* this should work, right?

1. The `RelayNumber` stats fields will get their default values at the time the migration runs
2. This `date_phone_subscription_checked` gets a value set to the day that the migration runs

1 and 2 work out well together?

privaterelay/management/commands/update_phone_remaining_stats.py
```diff
@@ -0,0 +1,59 @@
...
+        # subscription?
+        if (
+            not profile.date_phone_subscription_checked
+            or profile.date_phone_subscription_checked
+            <= datetime.now(timezone.utc) - timedelta(30)
+        ):
+            # Re-fetch all FXA data into profile so we can see if the user still has
+            # a phone subscription.
+            update_fxa(social_account)
+            profile.date_phone_subscription_checked = datetime.now(timezone.utc)
```
jwhitlock, line 35: I had to think about this, in the context of the test below.

* User has paid for phones, or is otherwise eligible (foxfooder, etc.)
* They haven't setup a Relay phone yet.
* We set that we've checked their profile, don't look again for 30 days.
* 28 days later, they get a Relay number, initialized to the `MAX_x_PER_BILLING_CYCLE`, but there is only 2 days left in the cycle
* 2 days later, this runs again, they are reset to `MAX_x_PER_BILLING_CYCLE`

That seems OK to me.

privaterelay/tests/mgmt_update_phone_remaining_stats_tests.py
```diff
@@ -0,0 +1,96 @@
...
+        "privaterelay.management.commands.update_phone_remaining_stats.update_fxa"
+    ):
+        updated_profiles = update_phone_remaining_stats()
+
+    profile.refresh_from_db()
+    relay_number.refresh_from_db()
+    assert (
+        profile.date_phone_subscription_checked.date()
+        == datetime.now(timezone.utc).today().date()
+    )
```
jwhitlock, line 92: Nit: I think you are trying to assert that `date_phone_subscription_checked` is recent. This test can fail around midnight. Are you willing to take that risk!

api/views/phones.py
```diff
@@ -534,11 +555,16 @@ def _get_inbound_contact(relay_number, inbound_from):
     return inbound_contact
 
 
-def _check_and_update_contact(inbound_contact, contact_type):
+def _check_and_update_contact(inbound_contact, contact_type, relay_number):
     if inbound_contact.blocked:
-        attr = f"num_{contact_type}_blocked"
-        setattr(inbound_contact, attr, getattr(inbound_contact, attr) + 1)
+        contact_attr = f"num_{contact_type}_blocked"
```
jwhitlock, line 560: 😍 I love using `contact_attr` and `relay_attr` rather than re-using `attr`

emails/migrations/0052_20220807_add_phone_sub_fields_to_profile.py
jwhitlock: makes sense 👍

api/views/phones.py
```diff
@@ -464,16 +467,21 @@ def inbound_call(request):
...
     _check_disabled(relay_number, "calls")
+    _check_remaining(relay_number, "minutes")
     inbound_contact = _get_inbound_contact(relay_number, inbound_from)
     if inbound_contact:
-        _check_and_update_contact(inbound_contact, "calls")
+        _check_and_update_contact(inbound_contact, "calls", relay_number)
 
     # Note: TwilioInboundCallXMLRenderer will render this as TwiML
+    relay_number.calls_forwarded += 1
+    relay_number.save()
```
jwhitlock, line 479: Nit: move these lines above the line `# Note: TwilioInboundCallXMLRenderer...`, so that comment remains close to `return response.Response`

api/views/phones.py
```diff
@@ -464,16 +467,21 @@ def inbound_call(request):
...
     return response.Response(
         status=201,
         data={"inbound_from": inbound_from, "real_number": real_phone.number},
     )
+    # TODO: update relay_number.remaining_minutes
```
jwhitlock, line 484: Nit: Move this before `relay_number.save()`, where the new code will go.

api/views/phones.py
```diff
@@ -464,16 +467,21 @@ def inbound_call(request):
...
+
     _check_disabled(relay_number, "calls")
+    _check_remaining(relay_number, "minutes")
     inbound_contact = _get_inbound_contact(relay_number, inbound_from)
     if inbound_contact:
-        _check_and_update_contact(inbound_contact, "calls")
+        _check_and_update_contact(inbound_contact, "calls", relay_number)
 
     # Note: TwilioInboundCallXMLRenderer will render this as TwiML
+    relay_number.calls_forwarded += 1
```
jwhitlock, line 478: Can you update a test to check the value of `relay_number.calls_forwarded`?

api/views/phones.py
```diff
@@ -439,17 +439,20 @@ def inbound_sms(request):
...
-        _check_and_update_contact(inbound_contact, "texts")
+        _check_and_update_contact(inbound_contact, "texts", relay_number)
 
     client = twilio_client()
     client.messages.create(
         from_=relay_number.number,
         body=f"[Relay 📲 {inbound_from}] {inbound_body}",
         to=real_phone.number,
     )
+    relay_number.remaining_texts -= 1
```
jwhitlock, line 453: Can you update an incoming SMS test to check the value of `relay_number.remaining_texts`?

api/views/phones.py
```diff
@@ -513,14 +521,27 @@ def _handle_sms_reply(relay_number, real_phone, inbound_body):
         body=inbound_body,
         to=last_text_sender.inbound_number,
     )
+    relay_number.remaining_texts -= 1
```
jwhitlock, line 524: Can you update a reply SMS test to check the value of `relay_number.remaining_texts`?

phones/migrations/0021_add_relaynumber_stats_20220913_1959.py
```diff
@@ -0,0 +1,107 @@
+# Generated by Django 3.2.15 on 2022-09-13 19:59
+
+from django.db import migrations, models
+
+
+def add_db_default_forward_func(apps, schema_editor):
```
jwhitlock, line 6: Thanks! this is complex but looks correct to me!

privaterelay/views.py
jwhitlock: They do not get their minutes back until 30 days plus `profile.date_phone_subscription_checked`. So, if it is a short pause, it may be a few days to reset, and if it is a longer (31+ days) pause, it will be when `update_phone_remaining_stats.py` runs again. If they had an old balance, they can use their old minutes / texts until then.

`date_subscribed_phone` has nothing to do with remaining minutes / texts. I think that is OK?

jwhitlock (changes requested):
Thanks @groovecoder! The test steps worked for me, and the logic looks good.

I see two required changes:

1) Move the 30-day billing period into a constant, something like:

```
DAYS_PER_BILLING_CYCLE = 30
```
or:

```
BILLING_CYCLE=timedelta(days=30)
```

2) Fix the `handle()` function in the management command. Try `./manage.py update_phone_remaining_stats` to see the issue.

The rest of the comments are style nits, testing suggestion, and praise. Treat accordingly.

privaterelay/tests/mgmt_update_phone_remaining_stats_tests.py
groovecoder: Changed to:
```python
     assert (
         profile.date_phone_subscription_checked
         >= pre_update_datetime
     )
```

phones/tests/models_tests.py
```diff
@@ -53,7 +53,9 @@ def make_phone_test_user():
 
 
 def upgrade_test_user_to_phone(user):
-    random_sub = random.choice(settings.SUBSCRIPTIONS_WITH_PHONE.split(","))
+    random_sub = random.choice(
+        list(filter(None, settings.SUBSCRIPTIONS_WITH_PHONE.split(",")))
+    )
```
jwhitlock, line 58: This is some lisp wizard stuff 🧙‍♂️ . It works, but I'll continue to use my version:

```suggestion
    random_sub = random.choice(
        [s for s in settings.SUBSCRIPTIONS_WITH_PHONE.split(",") if s])
    )
```

jwhitlock (approved):
Thanks for the changes @groovecoder! `./manage.py update_phone_remaining_stats` works for me now, and the phone dashboard works. 🚢 SHIP IT!

PR #2433, opened 2022-08-31
Update Dockerfile weekly

jwhitlock:
Enable Dependabot updates for our `Dockerfile` Python 3.9.13 is the current release, so I don't expect any updates for a while.


Vinnl (approved):
Huh, missed this option when setting up Dependabot, nice one.

groovecoder (approved):

PR #2434, opened 2022-08-31
MPP-1928 - Replace CTAs on Interstitial Page to Join Waitlist for non-premium countries

codemist:
This PR fixes #MPP-1928

image

This replaces 'Upgrade CTA' to 'Join the Waitlist' on the interstitial page for users in non-premium countries.

How to test:
Set your browser's language to a non-premium language, such as Romana, and the main CTA in the hero of the interstitial page should display 'Join the Waitlist`, which takes you to the premium waitlist sign up page.

### DIFF
--- a/frontend/src/pages/premium.page.tsx
+++ b/frontend/src/pages/premium.page.tsx
@@ -216,12 +216,12 @@ const PremiumPromo: NextPage = () => {
       )}
     </LinkButton>
   ) : (
-    <Button
-      disabled={true}
+    <LinkButton
+      href="/premium/waitlist"
       title={l10n.getString("premium-promo-availability-warning-2")}
     >
-      {l10n.getString("premium-promo-hero-cta")}
-    </Button>
+      {l10n.getString("waitlist-submit-label")}
+    </LinkButton>
   );
 
   const getPerkCta = (label: keyof typeof perkCtaRefs) => {
--- a/frontend/src/pages/premium.page.tsx
+++ b/frontend/src/pages/premium.page.tsx
@@ -19,7 +19,7 @@ import OnTheGoPhone from "../components/landing/carousel/images/onthego-illustra
 import OnTheGoReceipt from "../components/landing/carousel/images/onthego-illustration-receipts.svg";
 import { Layout } from "../components/layout/Layout";
 import { useGaViewPing } from "../hooks/gaViewPing";
-import { Button, LinkButton } from "../components/Button";
+import { LinkButton } from "../components/Button";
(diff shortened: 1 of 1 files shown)


frontend/src/pages/premium.page.tsx
```diff
@@ -216,12 +216,12 @@ const PremiumPromo: NextPage = () => {
       )}
     </LinkButton>
   ) : (
-    <Button
-      disabled={true}
+    <LinkButton
```
Vinnl, line 219: There's a linting error that you didn't remove the `Button` import at the top :)

frontend/src/pages/premium.page.tsx
```diff
@@ -216,12 +216,12 @@ const PremiumPromo: NextPage = () => {
...
+      href="/premium/waitlist"
       title={l10n.getString("premium-promo-availability-warning-2")}
     >
-      {l10n.getString("premium-promo-hero-cta")}
-    </Button>
+      {l10n.getString("waitlist-submit-label")}
```
Vinnl, line 223: I think @peiying2's guidance usually is to use unique strings for every instance of a string, even if the text content in English happens to be the same. For example, while the string in English is "Join the Waitlist", a translation of "Join" might be different depending on the context (e.g. imply submission of a form, which is appropriate in its original context, but not here). And re-using an earlier translation should be easy in Pontoon, as the previous translation will be shown and can be accepted with a single click.

Vinnl (approved):
The implementation looks good, but there's a lint error to fix, and a new string to be added to the l10n repo (sorry, I know that's a hassle...). But I don't think those changes need another review, so approving in lieu of those changes with that caveat.

frontend/src/pages/premium.page.tsx
codemist: I hear that, although there was no design proposed for this, just that Tony requested the CTA strings to be the same throughout the interstitial, therefore I found it reasonable to just copy/paste the strings from the `Plans` component.
image

frontend/src/pages/premium.page.tsx
codemist: Oops, thanks for the heads up.

frontend/src/pages/premium.page.tsx
Vinnl: Yeah, it's fine that the string value is the same, but I think Peiying requested separate IDs as good practice. Anyway, up to you - I'm just relaying info I heard before, don't know how important it is :)

PR #2447, opened 2022-09-01
Add `STATSD_DEBUG` env variable, update `STATSD_PREFIX`, add docs

jwhitlock:
Setting `STATSD_DEBUG` will enable metrics as well as the the [logging backend](https://markus.readthedocs.io/en/latest/backends.html#logging-metrics) for markus. Setting `DJANGO_STATSD_ENABLED` is still required to enable the [Datadog backend](https://markus.readthedocs.io/en/latest/backends.html#datadog-metrics) (statsd plus tags).

This can be used to debug metrics for the webserver and for the email processing task. For example, it helped determine that timing metrics are in milliseconds.

How to test:

* Run `nc -lu localhost 8125` to listen for UDP packets on `localhost:8125`
* In a new terminal, run `./manage.py runserver`, visit http://127.0.0.1:8000. 
    * [ ] There are 3-6 log entries of `"Type": "request.summary"`, depending on if you are logged in or not.
    * [ ] The `nc` command shows no output.
* Add `STATSD_DEBUG=True` to `.env`, restart `./manage.py runserver`, and reload http://127.0.0.1:8000. 
    * [ ] There are about 50 log entries. There are the 3-6 log entries of `"Type": "request.summary"`, and the new ones are `"Type": "markus"`, mostly `timing` for the static assets (`status:304,view:<unknown_view>,method:GET"`).
    * [ ] The `nc` command shows no output.
* In `.env`, set `STATSD_DEBUG=False`, and `DJANGO_STATSD_ENABLED=True`. Restart `./manage.py runserver`, and reload http://127.0.0.1:8000. 
    * [ ] The server output is back to 3-6 log entries of `"Type": "request.summary"`, with no `"Type": "markus"` entries
    * [ ] The `nc` command shows around 50 metrics received
* In `.env`, set `STATSD_DEBUG=True`, and `DJANGO_STATSD_ENABLED=True`. Restart `./manage.py runserver`, **restart `nc -lu localhost 8125`**, and reload http://127.0.0.1:8000.  
    * [ ] The server output has about 50 log entries, both `"Type": "request.summary"` and `"Type": "markus"` entries
    * [ ] The `nc` command shows around 50 metrics received


say-yawn:
Created a table for testing the above configs:
| STATSD_DEBUG  | DJANGO_STATSD_ENABLED | Logs | nc Command |
| ------------- | --------------------- | ---- | ---------- |
| False | False | Few "Type": "request.summary" logs | 
| True | False  | Lots of "Type": "markus" logs mostly for timing | No output |
| False | True | Few "Type": "request.summary" logs with no "Type": "markus" entires | Some output |
| True | True | Some "Type": "request.summary" and lots of "Type": "markus" entries | Some output |

When `STATSD_DEBUG=True` while `DJANGO_STATSD_ENABLED=False` I'm still seeing few `"Type": "request.summary"` logs. Is that still ok? It was the same style of output as when both `STATSD_DEBUG` and `DJANGO_STATSD_ENABLED` was `False` the difference being both set to `True` returned output for nc command.. 

If we can add the behavior of the logs in the food , that would be helpful :)

jwhitlock:
> When `STATSD_DEBUG=True` while `DJANGO_STATSD_ENABLED=False` I'm still seeing few `"Type": "request.summary"` logs. Is that still ok? It was the same style of output as when both `STATSD_DEBUG` and `DJANGO_STATSD_ENABLED` was `False` the difference being both set to `True` returned output for nc command.. 

Yes, the `request.summary` logs should appear in all the modes. I've clarified the checklists in the PR description.

> If we can add the behavior of the logs in the README, that would be helpful :)

👍

say-yawn (approved):
Changes LGTM. This will help us better utilize logging. Thank you for the PR!

jwhitlock:
I've added some documentation in `/docs/metrics.md`. I wanted to also document the metrics themselves, but I think that should wait until we have some automation, or it will get out of date quickly.

When writing the documentation, it became clearer that `STATSD_PREFIX` of `fx-private-relay` was being converted to `fx.private.relay`. I changed the default to this converted value, to reduce any confusion. Underscores would be allowed (`fx_private_relay`), but we have over a year of metrics on the old value, so I don't advocate making that change.

@say-yawn still good?

PR #2466, opened 2022-09-06
Add environment to Sentry

jwhitlock:
Tell Sentry which environment generated the event. The logic is similar to `RELAY_CHANNEL`, except If the `SITE_URL` is unknown, use `local` instead of `prod`.


say-yawn (approved):
LGTM

PR #2467, opened 2022-09-07
Do not retry requests on server errors

Vinnl:
# New feature description

When the server returns an error response, SWR would automatically retry the request with exponential back-off. Since this might play a role in the user getting rate-limited (even if just by APIs that we're calling on the server), this disables that mechanism on server errors. If e.g. the user's connection drops out, the default retry mechanism still applies.

Note that requests to `/users/` and `/profiles/` still only retry on 403 errors, because they already had an exception for just that. I left that in just in case we revert this PR later - we'll still want to preserve that.

# How to test

Make the server return an error response from the API. (For example, when running with `npm run dev:mocked`, I added the following line to a handler in `/frontend/src/apiMocks/handlers.ts`: `return res(ctx.status(403));`.) You should now see a single failed request, and no more requests after that, whereas in `main`, after a while you should see another request, and then a while longer another one, etc.

# Checklist


groovecoder (approved):
Spot-check looks good. I wasn't able to verify that this fixes MPP-2330 for sure, because I couldn't reproduce that bug on `main`, but this doesn't hurt, and it should certainly help avoid unintended rate-throttling of the UI code.

PR #2375, opened 2022-08-22
Test branch migrations on production

jwhitlock:
Test if there will be database errors when the current production code runs against the new migrations during the next deploy. 

This is done with two new CircleCI jobs with an extra step that:
* Parses the production tag from https://relay.firefox.com/__version__
* Fetches that tag from the repository
* Checks out that tag
* Checks out the migrations from the current branch

`pytest` / Django will then run the migrations from the current branch, but run the tests from the production tag. This will simulate what happens during the next deployment, when the old code handles traffic while running against the newly migrated database.

How to test:

- [x] See [failing CircleCI results](https://app.circleci.com/pipelines/github/mozilla/fx-private-relay/7076/workflows/7d569dff-8eff-40cd-a225-12e9e2d5df3d/jobs/24481/steps) for running this step with sqlite3 (like developer environment)
- [x] See [passing CircleCI results](https://app.circleci.com/pipelines/github/mozilla/fx-private-relay/7077/workflows/e7a531b9-6d25-4d79-a41c-53377c735bd2/jobs/24495) for SQLite3 once a default is added
- [x] See [passing CircleCI results](https://app.circleci.com/pipelines/github/mozilla/fx-private-relay/7077/workflows/e7a531b9-6d25-4d79-a41c-53377c735bd2/jobs/24493) for running this step with postgres (like production environment)
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #2469, opened 2022-09-07
Fix CircleCI job failing when pushing tags

Vinnl:
The CircleCI job that tests migrations would try to compare against the branch that CircleCI was running against. However, when pushing a tag, there is no known branch (i.e. $CIRCLE_BRANCH is unset, whereas $CIRCLE_TAG is). To check out the actual relevant commit regardless of whether we're running for a branch of a tag, checking $CIRCLE_SHA1 should work.

The main thing I'm not so sure about is why this didn't happen with last week's stage deploy, since this change made it in before that (https://github.com/mozilla/fx-private-relay/pull/2375). But I'm having trouble navigation CircleCI's UI and finding that pipeline, so maybe it did fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -333,7 +333,7 @@ jobs:
                   echo "# Production tag is ${PROD_TAG}"
                   git fetch --force origin tag ${PROD_TAG}
                   git checkout ${PROD_TAG}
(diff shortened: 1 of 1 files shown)


jwhitlock (approved):
Thanks @Vinnl!

PR #2475, opened 2022-09-09
Redirect should happen only if flag is not turned on

lloan:
The check should return true if the flag is NOT active. Only then should we be redirected.


Vinnl (approved):
Ha, whoops.

frontend/src/pages/phone.page.tsx
```diff
@@ -42,7 +42,7 @@ const Phone: NextPage = () => {
   }, [isInOnboarding, relayNumberData]);
 
   // check if phone flag is active - return to premium page if not.
-  if (isFlagActive(runtimeData.data, "phones")) {
+  if (!isFlagActive(runtimeData.data, "phones")) {
     router.push("/premium");
   }
```
Vinnl, line 47: Ah, build is failing, due to the flag not being active when prerendering, when there's no route yet:

```suggestion
  useEffect(() => {
    if (!isFlagActive(runtimeData.data, "phones")) {
      router.push("/premium");
    }
  }, [runtimeData.data]);
```

Vinnl:
Ah, and for the failing tests, we'll want to mock the router, see e.g. how that's done here: https://github.com/mozilla/fx-private-relay/blob/8fd61f8788a75672d9fac16d7e4cf15f57248e35/frontend/src/pages/index.test.tsx#L12

codemist:
Is this good to merge?

PR #2477, opened 2022-09-09
Redirect user to /premium if they can't buy phones

Vinnl:
This builds on @lloan's work to redirect the user to /premium if `phones` is disabled, adding a check for whether phones is even available in their country.

I had to re-order the components in phone.page.tsx a bit; the reason for this is that if a user has purchased their phone plan in another country, but currently is in a country where it's not available, they should still be able to see their dashboard. Hence, I moved the dashboard to the top, and moved the
`isPhonesAvailableInCountry()` check to after that.

One caveat is that the link to phones in the navigation won't be visible if it's not available in the user's country. We could also add a check there (and everywhere else where that function is called) to see whether they're already all set up, but that does mean we'd add another API request that won't be needed 99.9% of the time to every other page. I expect that it's a bit of an edge case anyway, so we can always change that if US/CA'ians start complaining that they can't find the phone dashboard while abroad.

How to test: with the phones flag enabled, but from a country in which phones is not available (e.g. by setting your browser lang), visit `/phones`. If your account didn't go through the phones onboarding yet, you should be redirected to `/premium`. If it did go through the onboarding before, you should see the dashboard.


frontend/src/pages/phone.page.tsx
```diff
@@ -30,13 +31,32 @@ const Phone: NextPage = () => {
   const relayNumberData = useRelayNumber();
   const [isInOnboarding, setIsInOnboarding] = useState<boolean>();
   const realPhoneData = useRealPhonesData();
+  // The user hasn't completed the onboarding yet if...
+  const isNotSetup =
+    // ...they haven't purchased the phone plan yet, or...
+    profile?.has_phone === false ||
```
codemist, line 37: non-blocking: is `profile?.has_phone` a boolean value? Maybe this can just be `!profile?.has_phone`

frontend/src/pages/phone.page.tsx
```diff
@@ -30,13 +31,32 @@ const Phone: NextPage = () => {
...
-    if (!isFlagActive(runtimeData.data, "phones")) {
+    if (!runtimeData.data) {
+      return;
+    }
+    if (
+      // Send the user to /premium if the phones flag is disabled...
+      !isFlagActive(runtimeData.data, "phones") ||
+      // ...or if a phone subscription is not available in the current country,
+      // and the user has not set it up before (possibly in another country):
+      (!isPhonesAvailableInCountry(runtimeData.data) && isNotSetup)
```
codemist, line 55: Out of curiousity, is `!isPhoneAvailabeInCountry` not a sufficient enough condition to redirect to `/premium`?

codemist (commented):
Looks good, although we should probably have a way call out that Phones aren't available to users outside of US & CAN - like the 'Join the Waitlist' CTA and form.

frontend/src/pages/phone.page.tsx
Vinnl: It is, but `!profile?.has_phone` is also `true` if `profile?.has_phone` is `undefined`, which it is if `profile` is `undefined`. Which means that that would result in the user getting redirected when we're still loading the profile.

frontend/src/pages/phone.page.tsx
Vinnl: It could be, but I was thinking that it might be annoying if someone has purchased phones, then goes abroad and then all of a sudden isn't able to manage their phones. Though we'll still have that problem in the sense that the menu items will be gone and all that.

frontend/src/pages/phone.page.tsx
codemist: Ahh.. gotcha.

frontend/src/pages/phone.page.tsx
codemist: Good call out

PR #2480, opened 2022-09-11
MPP-2343

lloan:
This PR fixes MPP-2343
 
# New feature description

On the “Verify your true phone number” page, if a user types a phone number starting with 1 the input field’s formatting ends up broken.

# Screenshot (if applicable)

image

# How to test

Go to the phone page, without having verified a real phone number. Try typing a number that starts with 1, it should not break the input format. The input component should let you type a zip code and the 7-digit number without issue. 

# Checklist


frontend/src/pages/phone.page.tsx
```diff
@@ -33,7 +33,7 @@ const Phone: NextPage = () => {
 
   useEffect(() => {
     // check if phone flag is active - return to premium page if not.
-    if (!isFlagActive(runtimeData.data, "phones")) {
+    if (runtimeData.data && !isFlagActive(runtimeData.data, "phones")) {
```
lloan, line 36: This prevents this check from failing before `runtimeData` is available.

Vinnl: Ha, I had to add this check as well in https://github.com/mozilla/fx-private-relay/pull/2477/files#diff-b3d14fbc23006061636ccf3480d5f020d9b91671989a66ead4d6650c26c5bc81R47-R52 :)

Am I correct in thinking that this is a separate fix from the main thing in this PR (invalid phone number input)? If so, would you mind removing it from this one? Since it also gets fixed in my PR, which also moves a couple of other things around, so it would cause a conflict with that one.

frontend/src/functions/formatPhone.ts
```diff
@@ -1,39 +1,31 @@
...
 /**
  * Make sure that when we format phone numbers in a consistent way.
+ * phoneNumber: The phone number to format.
+ * withCountryCode: Whether to include the country code in the formatted number.
  */
-export function formatPhone(phoneNumber: PhoneString) {
-  const parsedNumber = parsePhoneNumber(phoneNumber);
+export function formatPhone(
+  phoneNumber: string,
+  withCountryCode = false
```
Vinnl, line 8: Not too big of an issue with a single function, but a good habit to get into: positional optional parameters can be hard to read at the calling site if you're not familiar with the function, and also make refactoring more work if you want to add additional optional arguments later.  You can avoid such problems by adding an `options` parameter, so that it is called like `formatPhone("+1 (250) 555 - 1234", { withCountryCode: true })` instead of `formatPhone("+1 (250) 555 - 1234", true)`:

```suggestion
  options: { withCountryCode?: boolean } = {}
```

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
```diff
@@ -1,8 +1,6 @@
 import { ChangeEventHandler, FormEventHandler, useState } from "react";
 import { Localized, useLocalization } from "@fluent/react";
 import { isPossiblePhoneNumber } from "react-phone-number-input";
-import PhoneInput from "react-phone-number-input/input";
-import { E164Number } from "libphonenumber-js/min";
```
Vinnl, line 5: Nice, that's almost a complete dependency removed :broom: As far as I can see, its only remaining use is via `isPossiblePhoneNumber`, which actually is re-exported from libphonenumber-js. So you could remove `react-phone-number-input` completely and replace it by just that dependency. But according to the docs that function only verifies phone number length, so maybe we could even rip that entire dependency out and just check that ourselves?

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
```diff
@@ -127,6 +129,16 @@ const RealPhoneForm = (props: RealPhoneFormProps) => {
     </div>
   );
 
+  const onChange: ChangeEventHandler<HTMLInputElement> = (e) => {
+    // remove all non-digit characters
+    const phone = e.target.value.replace(/\D/g, "");
+
+    // set the new value if under 10 digits
+    if (phone.length <= 10) {
```
Vinnl, line 137: Question: why wouldn't we want to update the state if the input length is less than 10 digits?

Vinnl (commented):
Nice job with the added tests and less dependence on external dependencies. It's working pretty well I think, but two things are still pretty confusing:

- I can't paste into the phone number field for some reason?
- If I _do_ type `+1` in the field, the `+` gets removed but the 1 stays in. If you type in your phone number on autopilot, that means that the last digit of the phone number will be removed, and that there's an additional `1` at the front. Possibly a fix could be to remove all non-numeric characters _except_ a `+` at the start?

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
lloan: Setting max/min length to 16 to include formatting characters "(234) 342 - 2342".length === 16

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
lloan: I did - removed import, now we just have to clean up package.json, I'm hesitant to update it because last time I recall there was an issue due to not using Volta?

lloan:
> Nice job with the added tests and less dependence on external dependencies. It's working pretty well I think, but two things are still pretty confusing:
> 
> * I can't paste into the phone number field for some reason?
> * If I _do_ type `+1` in the field, the `+` gets removed but the 1 stays in. If you type in your phone number on autopilot, that means that the last digit of the phone number will be removed, and that there's an additional `1` at the front. Possibly a fix could be to remove all non-numeric characters _except_ a `+` at the start?

You don't need to type in +1 in the phone number input field. Doesn't the design assume we have already prepended a +1 for the user with the static +1 in the input box? As the user types, we provide visual formatting as well. I'm able to paste, maybe its a non-windows issue? We should get @codemist or @groovecoder to try copy/pasting

frontend/src/pages/phone.page.tsx
lloan: Removed

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
Vinnl: That should be resolved now (Kaitlyn now installed Volta as well, and I believe you already had it?), and if not, we'll want to resolve it rather than never changing our dependencies anymore :P So feel free to remove it, if something goes wrong we can deal with that then.

Vinnl (approved):
> You don't need to type in +1 in the phone number input field. Doesn't the design assume we have already prepended a +1 for the user with the static +1 in the input box? 

Heh, that is true, but if there's one thing I know of users, it's that they don't read instructions or check what they're doing (source: I use things :sweat_smile: ). So some leniency there might save a bunch of user frustration.

>  I'm able to paste, maybe its a non-windows issue? We should get @codemist or @groovecoder to try copy/pasting

Hmm OK, would be nice if we could get that tested, and otherwise hopefully QA will find it.

And again, it's awesome seeing new unit tests with every change to `formatPhone`, which is exactly the type of code for which unit tests are most valuable :sunglasses:

PR #2504, opened 2022-09-14
Workaround for a bug in a new React Aria version

Vinnl:
So, there's [a new version of React Aria](https://github.com/mozilla/fx-private-relay/pull/2420). There's a couple of changes (e353fd3 and a3b37f8) that are fairly straightforward, but also one pretty annoying bug for which I don't know the cause. I reported it upstream at https://github.com/adobe/react-spectrum/issues/3517, but since it only occurs in combination with Next.js, I'm not holding my breath for an easy fix.

I came up with a workaround, but it feels ugly and possibly fragile, so I wanted to get some extra eyes on it. The relevant commit is 834976d, and it basically does the following: whenever a menu opens, it increments a counter, which is referenced in an invisible element (`<Fragment>`). The updated counter value will trigger React to re-render the component, which gives `useOverlayPosition` the opportunity to align the overlay to the button, rather than remaining invisble.

Ideally, someone else comes up with a better solution than this :sweat_smile: (Or even better, the excellent React Aria folks know an easy fix.) Otherwise, this should unblock the upgrade and keep everything working as it did. I left links to the upstream issue and isolated the workaround to a single commit, so if an easy fix does arise, we should hopefully be able to easily revert this.

How to test: before the workaround (but with the updated dependencies installed), things like the News menu or filter drop-down do not work. After the workaround, they should.

- [ ] I've added a unit test to test for potential regressions of this bug. - Feels N/A, given that code external to our codebase is not behaving like I'd expect? Or the bug is in our codebase, but without knowing the source, it's hard to write a test...


lloan (approved):
Only question I have is about the name, are we calling this `useOverlayBugWorkaround` on purpose so we don't forget to update it in the future once a fix is available? Or is there a better name we can use?

Vinnl:
OK, looks like the issue [also affects other people](https://github.com/adobe/react-spectrum/issues/3490) and is not limited to Next.js, so that slightly improves the odds of this getting fixed upstream (now that it's clear that "upstream" is React Aria). Will mark as Do Not Merge for now and keep an eye on that issue.

As for your question @lloan, yeah, naming is hard and I'm not sure if this is the best one, but I did indeed want to make it clear that ideally, this code doesn't linger around the codebase indefinitely :)

PR #2506, opened 2022-09-14
Insert final plan IDs for new Premium and Bundle

Vinnl:
These should be the final plan and product IDs for the new Premium
and Bundle subscriptions in production, as well as prices for the new Premium subscriptions.

image

The actual prices for phones and bundle, as well as the product and plan IDs for the
phone plans, are yet to come.

This also makes the plan ID for the bundle in the US and Canada
configurable via an environment variable, so we can test the
subscription flow on stage.

# How to test

At `api/v1/runtime_data`, the updated Plan IDs should be visible under PERIODICAL_PREMIUM_PLANS and BUNDLE_PLANS. Additionally, if you add the stage plan IDs to your `.env` as follows, you should see those under `BUNDLE_PLANS`:

```env
BUNDLE_PROD_ID=prod_MQ9Zf1cyI81XS2
BUNDLE_PLAN_ID_US=price_1LhJGWKb9q6OnNsLuCurigPk
```

# Checklist


privaterelay/settings.py
```diff
@@ -660,62 +651,44 @@ def _get_initial_middleware() -> list[str]:
     "usd": {
         "en": {
             "monthly": {
-                "id": "price_1LXUenJNcmPzuWtRw3rhjQNP",
+                "id": "TBD",
                 "price": "$?.??",
```
groovecoder, line 655: I think the ones in [the spreadsheet](https://docs.google.com/spreadsheets/d/1RD4xgEH9bHw5pCuIcwU2FFyy4IrCBxGIioaxYdgT9RQ/edit#gid=0) are final now?

```suggestion
                "id": "price_1Li0w8JNcmPzuWtR2rGU80P3",
                "price": "$4.99",
```

privaterelay/settings.py
```diff
@@ -660,62 +651,44 @@ def _get_initial_middleware() -> list[str]:
...
             },
             "yearly": {
-                "id": "price_1LXUhcJNcmPzuWtRHUFHVk12",
+                "id": "TBD",
                 "price": "$?.??",
```
groovecoder, line 659:
```suggestion
                "id": "price_1Li15WJNcmPzuWtRIh0F4VwP",
                "price": "$3.99",
```

privaterelay/settings.py
```diff
@@ -660,62 +651,44 @@ def _get_initial_middleware() -> list[str]:
...
             },
         },
```
groovecoder, line 661:
```suggestion
        },
        "ca": {
            "monthly": {
                "id": "price_1Li0w8JNcmPzuWtR2rGU80P3",
                "price": "$4.99",
            },
            "yearly": {
                "id": "price_1Li15WJNcmPzuWtRIh0F4VwP",
                "price": "$3.99",
            }
        },
```

groovecoder (changes requested):

.env-dist
```diff
@@ -34,8 +34,8 @@ TWILIO_AUTH_TOKEN=
 TWILIO_SMS_APPLICATION_SID=
 TWILIO_MAIN_NUMBER=
 PREMIUM_PROD_ID=prod_K29ULZL9pUR9Fr
-PERIODICAL_PREMIUM_PROD_ID=price_1LXUenJNcmPzuWtRw3rhjQNP
-PHONE_PROD_ID=prod_LgQiSgNi4xL7dq
+PERIODICAL_PREMIUM_PROD_ID=prod_KGizMiBqUJdYoY
```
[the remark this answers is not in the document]
Vinnl: Production. (That's the one we want in here, right?)

privaterelay/settings.py
Vinnl: Yep, they are [now](https://mozilla.slack.com/archives/CUU5EC46L/p1663254765204039?thread_ts=1663244037.591459&cid=CUU5EC46L): 2301276c

privaterelay/settings.py
Vinnl: 2301276c

privaterelay/settings.py
Vinnl: Hmm, I just mapped `ca` directly to the US entry, since [that's what the current mapping does](https://github.com/mozilla/fx-private-relay/blob/1a624ea0c303cc63890a511bad4cc8f362f8bf76/privaterelay/settings.py#L431-L433) and they're explicitly the same plan, but I might have misunderstood when to give items their own entry in the PLAN_ID_MATRIX and when to map them to existing entries in PLAN_COUNTRY_LANG_MAPPING?

privaterelay/settings.py
```diff
@@ -660,62 +663,44 @@ def _get_initial_middleware() -> list[str]:
...
     "usd": {
         "en": {
             "yearly": {
-                "id": "price_1La3d7JNcmPzuWtRn0cg2EyH",
+                # To allow testing the subscription flow on stage, we can set
+                # a custom plan ID via an environment variable:
+                "id": BUNDLE_PLAN_ID_US,
                 "price": "$?.??",
             },
         },
```
codemist, line 694: Do we have a monthly price? It's in the add on:
image

Vinnl: @codemist "Monthly" and "yearly" refer to the billing period, but the value for `price` is the price per month. (So the actual price is 12 times that.)

Btw, note that in [this follow-up PR](https://github.com/mozilla/fx-private-relay/pull/2507), the price will be changed to a number, so this would look like:

```js
            "yearly": {
                "id": "price_1La3d7JNcmPzuWtRn0cg2EyH",
                # To allow testing the subscription flow on stage, we can set
                # a custom plan ID via an environment variable:
                "id": BUNDLE_PLAN_ID_US,
                "price": 1337,
                "currency": "USD",
            },
```

This can then be formatted using [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat).

codemist: Oh okay! Got it, thanks.

codemist: For a second I was confused about `1337` and then I googled it, lol.

privaterelay/settings.py
```diff
@@ -660,62 +663,44 @@ def _get_initial_middleware() -> list[str]:
...
 PHONE_PLAN_COUNTRY_LANG_MAPPING = {
     "us": {
         "en": PHONE_PLAN_ID_MATRIX["usd"]["en"],
     },
-    # Commented out; no plan ID known yet
-    # "ca": {
-    #     "en": PHONE_PLAN_ID_MATRIX["usd"]["ca"],
-    # },
+    "ca": {
+        "en": PHONE_PLAN_ID_MATRIX["usd"]["en"],
```
codemist, line 681: Should this be `"cad"`?

privaterelay/settings.py
```diff
@@ -660,62 +663,44 @@ def _get_initial_middleware() -> list[str]:
...
 BUNDLE_PLAN_COUNTRY_LANG_MAPPING = {
     "us": {
         "en": BUNDLE_PLAN_ID_MATRIX["usd"]["en"],
     },
-    # Commented out; no plan ID known yet
-    # "ca": {
-    #     "en": BUNDLE_PLAN_ID_MATRIX["usd"]["ca"],
-    # },
+    "ca": {
+        "en": BUNDLE_PLAN_ID_MATRIX["usd"]["en"],
```
codemist, line 702: `"cad"` instead?

privaterelay/settings.py
Vinnl: Unfortunately no, plans will be in USD for Canadians.

privaterelay/settings.py
Vinnl: Haha sorry, should've realised that maybe that wasn't a thing anymore in 2022 :P

privaterelay/settings.py
Vinnl: What's even weirder is that people in the UK have to deal with this too! (But to be fair, at least they had the option of adopting the Euro :P)

.env-dist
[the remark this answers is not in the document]
groovecoder: Doesn't matter too much - we'll have to override in local, dev, and stage. This is "safest" in case the `.env-dist` value is put into prod it'll still work.

privaterelay/settings.py
groovecoder: No that seems fine. Let's go with it.

groovecoder (approved):

PR #2507, opened 2022-09-14
Locale-specific price formatting

Vinnl:
For the current Premium plan, the price formatting is hardcoded by the back-end. However, correct formatting is locale-dependent: some use comma's to separate decimal numbers, some periods, some put the currency symbol at the start, some at the end, etc. When we're switching to newly added endpoint data, we might as well outsource the formatting of pricess to the browser (via its Intl.NumberFormat API).

The only thing that makes me slightly nervous is that we're dealing with money using floating-point numbers, but that's what Intl.NumberFormat expects, and as long as we don't do math with it, we should be fine.

The main change here is to have the API return `{ price: number; currency: string }` instead of `{ price: string }`; the rest is updating the UI to pass those values to `Intl.NumberFormat`.

# How to test

The prices of new plans should be formatted according to your browser locale.

# Checklist


frontend/src/functions/getPlan.ts
```diff
@@ -45,8 +59,16 @@ export const getPhoneSubscribeLink = (
   return `${runtimeData.FXA_ORIGIN}/subscriptions/products/${runtimeData.PHONE_PRODUCT_ID}?plan=${plan.id}`;
 };
 
-export const getBundlePrice = (runtimeData: RuntimeDataWithBundleAvailable) => {
-  return getPlan(runtimeData.BUNDLE_PLANS, "yearly").price;
+export const getBundlePrice = (
+  runtimeData: RuntimeDataWithBundleAvailable,
+  l10n: ReactLocalization
```
lloan, line 64: question (non-blocking): For learning on my end, why do we have to pass `l10n` in every time we call this function - we can't call it from `getPlan`?

lloan (approved):
LGTM.

frontend/src/functions/getPlan.ts
Vinnl: Detecting the user's language happens at runtime, and the correct localisation strings are then passed into the React app [via a Context Provider](https://github.com/mozilla/fx-private-relay/blob/adf9164f5f19e4dd5dfc65d2ec9f26a22ad48569/frontend/src/pages/_app.page.tsx#L84). React components within that provider can then access it via the `useLocalization` hook, but outside of components (e.g. in the `getPlan` function), there's no way to get to that. (Or at least, I don't think there is - it might be possible to call `getL10n` outside of the Provider and get access to the same instance, but even if it is possible, I like to keep functions [pure](https://en.wikipedia.org/wiki/Pure_function) if possible, to make it easier to grasp their scope and make them easier to test if needed.)

Vinnl:
Thanks Lloan. @groovecoder, could I get a review from you on the back-end part? I'm wondering if there's a particular reason the initial setup was done as it was, or if that's just the way it was.

frontend/src/functions/getPlan.ts
```diff
@@ -15,9 +17,15 @@ const getPlan = <P extends Partial<PlanData>>(
 
 export const getPeriodicalPremiumPrice = (
   runtimeData: RuntimeDataWithPeriodicalPremiumAvailable,
-  billingPeriod: keyof PlanData
+  billingPeriod: keyof PlanData,
+  l10n: ReactLocalization
```
groovecoder, line 21: suggestion (non-blocking): can we save ourselves this extra parameter & argument on all the calls if we

`import { ReactLocalization, useLocalization } from "@fluent/react";`

and then

`const { l10n } = useLocalization();`

in these functions?

groovecoder (approved):

groovecoder:
Oh, that's just the way it was, so this is a good enhancement. Thanks.

frontend/src/functions/getPlan.ts
Vinnl: [@lloan had exactly the same question](https://github.com/mozilla/fx-private-relay/pull/2507#discussion_r971484739) :)

privaterelay/settings.py
```diff
@@ -677,7 +707,9 @@ def _get_initial_middleware() -> list[str]:
                 # To allow testing the subscription flow on stage, we can set
                 # a custom plan ID via an environment variable:
                 "id": BUNDLE_PLAN_ID_US,
-                "price": "$?.??",
+                # TODO: Insert correct price
+                "price": -1337,
```
codemist, line 711: Just to confirm - "Monthly" and "yearly" refer to the billing period **only** for Bundle pricing, not for Phone?

Vinnl: No, it's the billing period for everything - so the prices for the yearly plans for regular Premium and phones are monthy prices as well, but they're billed yearly.

codemist: Got it, thanks for clearing up the confusion

codemist (approved):

PR #2514, opened 2022-09-15
Prevent lockfile getting out-of-date

Vinnl:
~~While CircleCI [supposedly](https://circleci.com/developer/orbs/orb/circleci/node#usage-npm_install) uses `npm ci` to install dependencies, which should warn if the lockfile does not match what is specified in package.json, a mismatch did slip in without failing the build. My hypothesis is that this was due to dependencies being cached from a previous run.~~

Well, that did not work. Apparently `npm ci` only cares if necessary dependencies _aren't_ in the lockfile, but if there's too many, it's totally fine with it. Thus, this PR is now just updating the lockfile, and at least ensuring that CircleCI's cache isn't re-used if our lockfile has changed :/

~~Furthermore, the Netlify command did not use `npm ci` at all. It does now.~~ I reverted this, because it's fine if this fails just in CircleCI - having the preview deployment can still be useful then.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
           command: npm run build
           working_directory: ./frontend/
       - save_cache:
-          key: v1-frontend-build-{{ .Branch }}-{{ .Revision }}
+          key: v1-frontend-build-{{ .Branch }}-{{ .Revision }}-{{ checksum "frontend/package-lock.json" }}
           paths:
             - ./frontend/out/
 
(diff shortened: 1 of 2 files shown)


jwhitlock (approved):
This looks fine to me - I'm assuming the `package-lock.json` changes are the important ones. 

The Circle CI cache is weird to me. It makes some sense for `npm` download caches, not as much for the output of a build process. The cache key is based on the branch name, the commit hash, and now the checksum of `package-lock.json`. It seems like you are saying "make sure that the files we load in `build_test_backend` job are the files created in `build_backend` job". I think [workspaces](https://circleci.com/docs/workspaces) are the tool to use, and I'm not sure what we gain by using cache and possibly loading cross-build outputs.

Vinnl:
> The Circle CI cache is weird to me.

Oh yeah, never mind - with fresh eyes this change doesn't really make sense either. The lockfile already cannot change when the commit hash is the same, so adding it to the cache key is redundant. I just got lost in the weeds with npm's weird inability to warn us when dependencies haven't been removed from the lockfile. The lockfile changes were already taken care of by Dependabot while updating other dependencies, so I'll close this PR.

PR #2537, opened 2022-09-20
Use a toggle button for phone forwarding/blocking

Vinnl:
This PR fixes MPP-2304.

The main concern here was making sure that the tooltip when hovering over the forwarding/blocking button showed up when hovering the button, not just the image, but that quickly lead to a couple of other fixes:
- the hover text is now localisable,
- there now no longer are buttons that do nothing (i.e. the one representing the current state),
- forwarding/blocking state is now derived from the API, rather than the toggle state, so it only represents what's actually the case on the server, and
- there's now a mock API handler for updating a Relay number.

How to test: open the phone dashboard. Hover over the toggle control; the tooltip should now be visible regardless of your position. Also, if you tab to it with a screen reader, there should no longer be two buttons of which ones does nothing.

- [ ] l10n changes have been submitted to the l10n repository, if any. - No, just like the rest of phones.
- [ ] I've added a unit test to test for potential regressions of this bug. - No, layout bug.


frontend/src/components/phones/dashboard/PhoneDashboard.module.scss
```diff
@@ -403,47 +403,48 @@
...
+            color: $color-white;
           }
-        }
 
-        .active-button {
-          background-color: $color-violet-70;
-        }
+          &:hover {
+            border-color: $color-violet-70;
+          }
```
codemist, line 437: I don't think it's necessary to have a hover state when there's already one for the individual button. I checked with Eduardo about this too.

https://user-images.githubusercontent.com/13066134/191796045-129e5725-1638-4a8f-915a-df51969a8741.mov

frontend/src/components/phones/dashboard/PhoneDashboard.module.scss
```diff
@@ -403,47 +403,48 @@
...
-        .block-icon {
-          color: $color-white;
-          justify-items: center;
-          align-items: center;
+            &.is-forwarding .block-icon,
+            &.is-blocking .forward-icon {
+              color: $color-blue-50;
+              background-color: $color-purple-05;
+            }
+          }
```
codemist, line 447: Can this hover state show only when hovering over the inactive button, and not the container of the toggle?

frontend/src/components/phones/dashboard/PhoneDashboard.module.scss
Vinnl: Hmm, so I had it just for the individual button at first, but it got kinda confusing because the entire thing is the click target, not just the individual button - so you'd hover the currently-active "button", and then the other button would change background colour. What do you think? (I can also quickly push the change and you can see experience yourself if you want, see what you like better?)

codemist: I think it's fine to just show a hover state for the inactive button since it's still some sort of interactive feedback for the user. Ideally this would be a literal toggle lol
image

Vinnl: OK, removed the border in 1c8de44c941bc712aa0ac23f69024edf57867da0

I did change the blue icon colour to only appear on hover and not on focus, because otherwise there was no difference between the hover and focus states (which would be confusing when you click it, then move your mouse away from the toggle - it would still be focused, so there would be no observable difference).

[Screencast from 23-09-22 16:57:37.webm](https://user-images.githubusercontent.com/4251/191990687-11050d29-5d32-4be0-99bb-41fec1137b23.webm)

PR #2544, opened 2022-09-20
fix MPP-2332: add relay numbers to messaging service

groovecoder:
This PR fixes MPP-2332.
image

How to test:
1. You'll need @groovecoder to help you test by looking at the Twilio console (?)
2. Go thru the registration flow to get a relay number
   * Hint: to avoid buying another number, remove your existing relay number in the local admin UI, and then use the http://127.0.0.1:8000/api/v1/docs/ `POST /relaynumber/` endpoint to give the same number right back to yourself
3. Check the Twilio console (or maybe use Twilio API in a python REPL?) to verify that the number is in the "Firefox Relay" messaging service


phones/tests/models_tests.py
```diff
@@ -1,4 +1,5 @@
 from datetime import datetime, timedelta, timezone
+from types import SimpleNamespace
```
jwhitlock, line 2: 🧑‍🎓 TIL about [SimpleNamespace](https://docs.python.org/3/library/types.html#types.SimpleNamespace)!

jwhitlock (approved):
Thanks @groovecoder, looks good!

I hit issue #2548 when I was testing this, but I was unable to replicate the issue through the web interface, so I think our users won't hit it moving forward.

I did get Twillio errors when adding a number to the messaging service that already exists:

```
HTTP Error Your request was:

POST /Services/<settings.TWILIO_MESSAGING_SERVICE_SID>/PhoneNumbers

Twilio returned the following information:

Unable to create record: Phone Number or Short Code is already in the Messaging Service.

More information may be available here:

https://www.twilio.com/docs/errors/21710
```

We may need to handle that error in the future.

The [documentation](https://www.twilio.com/docs/messaging/services/api/phonenumber-resource#create-a-phonenumber-resource-add-a-phone-number-to-a-messaging-service) doesn't talk about errors, but it does have this warning:

> Each Service can have no more than 400 phone numbers by default. If you think you might need a higher limit, [contact Twilio Support](https://www.twilio.com/console/support/tickets/create) about a Messaging Service number limit increase, and include an explanation of your use case.

So, keep a 400 number limit in the back of your mind.

PR #2545, opened 2022-09-21
Use correct country code for Sweden

Vinnl:
We accidentally used the language code (sv) rather than the country code (se) for Sweden. This did not cause any problems, because the mapping from country and language to a plan made the same mistake, so it evened out. Still, probably a good idea to use the correct code.

To locally verify that it still works, set your Accept-Language to sv-SE, visit the locally running website, and verify that the upgrade links still point to the correct plan.

Thanks @birdsarah for spotting this!

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -379,7 +379,7 @@ PREMIUM_PLAN_ID_MATRIX = {
             "id": "price_1JmRCQJNcmPzuWtRprMnmtax",
             "price": "0,99 €",
         },
-        "sv": {
+        "se": {
             "id": "price_1KQc1PJNcmPzuWtRsEfb6inB",
             "price": "0,99 €",
         },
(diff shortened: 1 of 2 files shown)


groovecoder (approved):
On this branch, http://127.0.0.1:8000/premium/ shows the same [purchase link](https://accounts.stage.mozaws.net/subscriptions/products/prod_K29ULZL9pUR9Fr?plan=price_1KQc1PJNcmPzuWtRsEfb6inB) for Swedish (sv-se) as the `main` branch does. But they don't match what's in the plan ID column of the spreadsheet?

Vinnl:
@groovecoder What does that link look like? Are you seeing the new plan matrix (which has monthly and yearly subscribe links - the ones from the sheet), or the current page, which still has the old, monthly plan ID?

groovecoder:
image

groovecoder:
It's the former, so looks like this is working as expected.

PR #2287, opened 2022-08-01
Delete data from twilio mpp 2231

groovecoder:
This PR fixes https://mozilla-hub.atlassian.net/browse/MPP-2231

How to spot-test:
1. Set up your Twilio account & run `ngrok`
2. Send a text message to a local RelayNumber
3. Inspect the inbound request in `ngrok` to get the message's `SmsSid`
4. Go to your [Twilio console messaging logs](https://console.twilio.com/us1/monitor/logs/sms?frameUrl=%2Fconsole%2Fsms%2Flogs%3Fx-target-region%3Dus1&currentFrameUrl=%2Fconsole%2Fsms%2Flogs%3F__override_layout__%3Dembed%26bifrost%3Dtrue%26x-target-region%3Dus1)
5. Search for the sms sid
   * [ ] Verify the message isn't in the logs anymore
6. In your Twilio Console, purchase a new phone number
7. Set its TwiML app to your local Relay Twilio app
8. Send a text message to the phone number
   * Note: while the phone number is assigned to your app, it is NOT assigned to a Relay user or RealPhone number, so this will trigger the `ObjectDoesNotExist` error & code
9. Inspect the inbound request in `ngrok` to get the message's `SmsSid` (note: the request should have a 400 response)
10. Go to your [Twilio console messaging logs](https://console.twilio.com/us1/monitor/logs/sms?frameUrl=%2Fconsole%2Fsms%2Flogs%3Fx-target-region%3Dus1&currentFrameUrl=%2Fconsole%2Fsms%2Flogs%3F__override_layout__%3Dembed%26bifrost%3Dtrue%26x-target-region%3Dus1)
11. Search for the sms sid
    * [ ] Verify the message isn't in the logs anymore

- [x] I've added a unit test to test for potential regressions of this bug.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #2549, opened 2022-09-21
fix MPP-2351: add voice_status callback endpoint

groovecoder:
/voice_status receives Twilio call status callback requests and  updates RelayNumber.remaining_seconds

This PR fixes MPP-2350.

How to test:
1. Set up your local end to end phone development
2. Register a relay number on your local app
3. Make a call to your local relay number
4. Refresh your dashboard
   * [ ] The "Remaining minutes" box should be reduced by the number of minutes of your call


api/views/phones.py
```diff
@@ -475,7 +475,6 @@ def inbound_call(request):
         _check_and_update_contact(inbound_contact, "calls", relay_number)
 
     relay_number.calls_forwarded += 1
-    # TODO: update relay_number.remaining_minutes
```
jwhitlock, line 478: Good job catching the TODO!

api/views/phones.py
```diff
@@ -485,6 +484,24 @@ def inbound_call(request):
...
+    if called is None or call_status is None:
+        raise exceptions.ValidationError("Call data missing Called, CallStatus")
+    if call_status != "completed":
+        return response.Response(status=200)
+    call_duration = request.data.get("CallDuration", None)
+    if call_duration is None:
+        raise exceptions.ValidationError("completed call data missing CallDuration")
+    relay_number, _ = _get_phone_objects(called)
+    relay_number.remaining_seconds = relay_number.remaining_seconds - int(call_duration)
+    relay_number.save()
```
jwhitlock, line 501: small chance of a race condition here, if multiple `voice_status` requests come in at the same time. For example, Twillio could queue up some status calls due to some pause on their side, and send them all at once. However, it would be in the user's favor - less seconds would be deducted than planned. 👍

Also, any errors result in a server error, but what else should we do, other than send the error to Sentry? I tested, and Twillio does not seem to retry on a server error.

jwhitlock (approved):
Nice find @groovecoder! The steps worked for me.

There's a status callback for SMS messages as well, it may work as another approach for PR #2287

PR #2558, opened 2022-09-23
MPP-2359 - Toast notification after resending welcome text

codemist:
# New feature description

This adds a toast notification and dismissal to the resend welcome text banner after the CTA is clicked.

# Screenshot (if applicable)

https://user-images.githubusercontent.com/13066134/192007943-7a985ba4-b877-4c3e-a8e4-d51462229c46.mov

# How to test
On the phones dashboard page, check that the clicking the "Resend welcome text" CTA closes the banner, and displays a toast notification thereafter. Make sure that the contact card is still relayed normally to your real phone number.

# Checklist


frontend/src/pages/phone.page.tsx
```diff
@@ -43,6 +45,38 @@ const Phone: NextPage = () => {
...
+        title={l10n.getString("phone-banner-resend-welcome-sms-title")}
+        type="info"
+        cta={{
+          content: l10n.getString("phone-banner-resend-welcome-sms-cta"),
+          onClick: () => {
+            realPhoneData.resendWelcomeSMS();
+            toast(l10n.getString("phone-banner-resend-welcome-sms-toast-msg"), {
+              type: "success",
+            }),
+              resendWelcomeSMSDismissal.dismiss();
```
Vinnl, line 64: I think this might technically work, but it's a bit confusing and I'm assuming it is a typo :)

```suggestion
            });
            resendWelcomeSMSDismissal.dismiss();
```

frontend/src/pages/phone.page.tsx
```diff
@@ -43,6 +45,38 @@ const Phone: NextPage = () => {
     // ...there was a list of Relay numbers, but it was empty:
     relayNumberData.data?.length === 0;
 
+  const resendWelcomeSMSDismissal = useLocalDismissal(
+    `resend-sms-banner-${profile?.id}` ?? "unused"
```
Vinnl, line 49: If `profile.id` is undefined, this string will become `resend-sms-banner-undefined`, so the `?? "unused"` will never be executed:

```suggestion
    `resend-sms-banner-${profile?.id}`
```

The above would mean that there's a split second in which the key would be `resend-sms-banner-undefined`, which I'd personally be fine with. The below is provide just for educational value :)

If you really wanted to consciously provide a fallback value, it would either have to be like this:

```suggestion
    typeof profile?.id === "number" ? `resend-sms-banner-${profile?.id}` : "unused"
```

Alternatively, you can just provide a fallback for `undefined`:

```suggestion
    `resend-sms-banner-${profile?.id ?? "unknown-id"}`
```

Vinnl (approved):
Looks good, works as-is. One small suggestion I'd have is to create a separate variable for the dismissal key, and pass that to both `useLocalDismissal` and the `<Banner>` to clarify that they're both the same dismissal.

PR #2576, opened 2022-09-28
Add `relaynumber.remaining_minutes` back

jwhitlock:
The column needs to exist until the next deploy.

PR #2577, opened 2022-09-28
Change vpn download  link

birdsarah:
If a VPN client has come from the client to purchase the bundle then if we send the user to vpn.mozilla.org first, guardian can detect this and can redirect the user back into the client flow. If the user has not come from the VPN client, guardian will redirect correctly to the link that this was pointing to before this patch.


birdsarah:
@lloan can you help me get this to pass CI. I think it might just need a re-run but I don't have permissions.

jwhitlock:
@birdsarah the test failure was fixed in PR #2576, rebasing to get ✅ and to merge for stage

Vinnl (approved):

PR #2598, opened 2022-10-04
Load all l10n files

Vinnl:
This PR complements https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/103.

Previously, we'd only load `app.ftl`. With this change, all files in a localisation directory will be loaded. This enables a couple of things:

- Module-specific localisation files (e.g. separate files for the phones and bundle modules, which at the moment only need localisation to French and Spanish).
- Shared definitions, specifically, a brands.ftl file that can be referred to from the other Fluent files.

The main restriction at the moment is that I only implemented this change for the frontend - Django still only loads app.ftl. I figured this was fine, since the only strings added by Django are the ones in wrapped_email.html, which don't change often and are likely to remain in app.ftl. We can always update that to e.g. an emails.ftl file, but we'd want to figure out how to avoid losing attribution of specific strings to localisers.

This change should be backwards-compatible: since it loads _all_ `.ftl` files in a locale's directory, it will continue loading `app.ftl` as long as that's the only `.ftl` file.

How to test: check out the branch `partial-translation` from the l10n submodule. Verify that brand names, which in that branch are moved into a separate file, are still shown, and that string modifications in your locale still show up.

# Checklist

- [x] l10n changes have been submitted to the l10n repository, if any. - https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/103
- [x] I've added or updated relevant docs in the docs/ directory. (The docs already didn't refer to `app.ftl`.)


lloan (approved):
Works as intended. For a second I forgot to checkout this branch as I checked out the other l10n branch (sub module) and thought it was broken 😏. Looks good!

frontend/src/functions/getL10n.ts
```diff
@@ -16,16 +16,16 @@ export function getL10n() {
...
-      "./".length,
-      fileName.length - "/app.ftl".length
-    );
+    // Filenames are formatted as `./<locale>/<module>.ftl`.
+    // Example: ./en/bundle.ftl
+    const locale = fileName.split("/")[1];
 
     if (locale) {
-      RESOURCES[locale] = new FluentResource(translationsContext(fileName));
+      RESOURCES[locale] ??= [];
```
lloan, line 27: comment: never seen anyone actually use this `??= ` - awesome :)

Vinnl: Haha, I, too, was very excited about a chance to use it :P

PR #2540, opened 2022-09-20
change _validate_number to accept CA numbers

groovecoder:
This PR fixes an issue where Canadian numbers were not accepted.

How to test:
1. Spin up your local relay app
2. Go to http://127.0.0.1:8000/phone/
3. Try to register a Candian number (Maybe ask a Canadian colleague if you can use theirs?)
   * [ ] The number should receive the 6-digit confirmation code and allow you to register

- [x] I've added a unit test to test for potential regressions of this bug.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #2600, opened 2022-10-04
MPP-2409: Enable phones in canada

jwhitlock:
For MPP-2409, enable phone numbers in Canada. This builds on PR #2540.

A new setting `TWILIO_ALLOWED_COUNTRY_CODES` sets the allowed countries, currently `"US"` and `"CA"` by default.

This PR adds new model fields / columns `RealPhone.country_code` and `RelayNumber.country_code`, default to `"US"`, to store the number's country. These fields have been added as read-only fields on the API, and to the front-end interface and mocks.

The `RealPhone.country_code` is set from the Twilio lookup API when created from the API. The country code is then used when looking up suggested numbers in Twilio, or searching by area code or location, so only phone numbers in the same country as the user are returned.
 
When saving a new `RelayNumber`, the `RelayNumber.country_code` is set to the `RealPhone.country_code`, which is a guess since only that country's numbers were available through the API. I'm open to using the Twilio lookup API to get the country code, or dropping the column as unneeded.

For testing, we may be able to push this PR to dev before merging. To test locally:

* Setup your local development environment for [end-to-end phone testing](https://github.com/mozilla/fx-private-relay/blob/main/docs/end-to-end-local-phone-dev.md).
* Run migrations to get the new columns
* If you have an existing `RelayNumber`, take note of the phone number (so you can restore it later) and use the Django Admin to delete it.
* As a premium user, setup a `RealNumber` if you haven't already.
* Go to http://127.0.0.1:8000/phone/
    * [ ] The page shows "Congratulations! You’re now ready to choose your phone number mask."
* Click "Search for phone number masks"
    * [ ] Phone numbers are shown for your country
* In the search box, enter an area code in your country, and hit "Enter" 
    * [ ] Phone numbers are shown for that area code
* In the search box, enter an area code *not* in your country (918 is in the US, 249 is in Canada), and hit "Enter"
    * [ ] No results are found
* In the Django Admin, find your "RealPhone" record:
   * Un-check `Verified`
   * Click button "Save and continue editing"
   * Change the `Number` to a fake number in the other country (for example, `+12155551234` for US, `+12495551234` for Canada)
   * Change the `Country code` to the other country (`"US"` or `"CA"`)
   * Check `Verified`
   * Select "Save"
* As the premium user, go to http://127.0.0.1:8000/phone/
    * [ ] The page shows "Congratulations! You’re now ready to choose your phone number mask."
* Click "Search for phone number masks"
    * [ ] Phone numbers are shown for the new country
* In the search box, enter an area code for the new country, and hit "Enter" 
    * [ ] Phone numbers are shown for that area code if available
* In the search box, enter an area code *not* in the new country (918 is in the US, 249 is in Canada), and hit "Enter"
    * [ ] No results are found
* Restore your `RealPhone` and `RelayNumber` as desired.


phones/tests/models_tests.py
```diff
@@ -260,6 +260,41 @@ def test_create_relaynumber_creates_twilio_incoming_number_and_sends_welcome(
     assert relay_number_obj.vcard_lookup_key in call_kwargs["media_url"][0]
 
 
+def test_create_relaynumber_canade(phone_user, mocked_twilio_client):
```
say-yawn, line 263: Nit:
```suggestion
def test_create_relaynumber_canada(phone_user, mocked_twilio_client):
```

phones/models.py
```diff
@@ -201,7 +203,8 @@ def calls_and_texts_blocked(self):
         return self.calls_blocked + self.texts_blocked
 
     def save(self, *args, **kwargs):
-        if not get_verified_realphone_records(self.user):
+        realphone = get_verified_realphone_records(self.user).first()
```
say-yawn, line 206: Nit:
```suggestion
        realphone = get_verified_realphone_record_qs(self.user).first()
```
This function is way too close to the `get_verified_realphone_record` even though they receive different parameter. My recommendation is to rename the function to better indicate that we are returning a query set or combine this function to existing `get_verified_realphone_record`. However, this is a nice to have refactor.

say-yawn (approved):
Put some nits. Not yet tested on local but willing to merge this and test on dev and/or stage so that we can especially get peers from Canada to test this for us.

phones/models.py
jwhitlock: I agree. I think `records` implies a list or queryset, and I was confused about the different parameter type of `record`. I'd go with:

```
get_verified_realphone_records_for_user(User)
get_verified_realphone_record_by_number(number)
```

I also think this is a good opportunity for typing annotations.

However, I'm going to save the refactoring for a future PR, and focus this one on the blocking feature.

PR #2611, opened 2022-10-05
MPP 2402 area code random

maxxcrawford:
- Add random numbers to suggested relay numbers
- Add frontend code to add random number list into suggested number array


jwhitlock:
@maxxcrawford and I paired on this, and we're happy with each other's code. I'll merge when we CircleCI says ✅ , unless someone speaks up.

PR #2612, opened 2022-10-05
React aria workaround

Vinnl:
I'm resubmitting https://github.com/mozilla/fx-private-relay/pull/2504 here, which Dependabot closed (rude!). It looks like a new version of React Aria was released without a fix for [the bug](https://github.com/adobe/react-spectrum/issues/3490), and it has been labelled "has workaround" - so I don't expect it to get fixed upstream any time soon. Might be time for us to merge this :/


lloan (approved):
Looks good!

PR #2613, opened 2022-10-05
MPP-2332: Handle dupe US A2P 10DLC registration

jwhitlock:
Ignore error 21710 when a US relay number is already registered with the US A2P 10DLC campaign. At this point, this issue probably only affects developers with access to the Django admin or the database.

How to test:

- Set up [end-to-end local phone dev](https://github.com/mozilla/fx-private-relay/blob/main/docs/end-to-end-local-phone-dev.md)
- Register a real phone and a relay number.
- As an admin use, load the Django admin
- Load the `RealPhone`. Change the user to a different user, such as the admin user, and click "Save"
- Load the `RelayNumber`. Ensure the `country_code` is `US`. Change the user to the different user, and select "Save"

On `main`, an exception is raised:

```
HTTP Error Your request was:

POST /Services/<TWILIO_MESSAGING_SERVICE_SID>/PhoneNumbers

Twilio returned the following information:

Unable to create record: Phone Number or Short Code is already in the Messaging Service.

More information may be available here:

https://www.twilio.com/docs/errors/21710
```

With this PR:

* [ ] The RelayNumber is saved.

After testing, you can use the Django admin to return the `RealNumber` and `RelayNumber` to the original user.

### DIFF
--- a/phones/models.py
+++ b/phones/models.py
@@ -13,6 +13,8 @@ from django.db.models.signals import post_save
 from django.dispatch.dispatcher import receiver
 from django.urls import reverse
 
+from twilio.base.exceptions import TwilioRestException
+
 
 MAX_MINUTES_TO_VERIFY_REAL_PHONE = 5
 LAST_CONTACT_TYPE_CHOICES = [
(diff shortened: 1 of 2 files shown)


phones/models.py
```diff
@@ -231,12 +233,20 @@ def save(self, *args, **kwargs):
...
+                client.messaging.v1.services(
+                    settings.TWILIO_MESSAGING_SERVICE_SID
+                ).phone_numbers.create(phone_number_sid=twilio_incoming_number.sid)
+            except TwilioRestException as err:
+                if err.status == 409 and err.code == 21710:
+                    # Ignore "Phone Number is already in the Messaging Service"
+                    # https://www.twilio.com/docs/api/errors/21710
+                    pass
+                else:
+                    raise
```
say-yawn, line 249: While this is happening because we are trying to allow multiple Relay Number for the same Real Number, but this could happen if we/Twilio accidentally serve a Relay Number that is already used (possible due to race condition etc?). Should we check here again that the Relay Number is correctly assigned to this user/Real Phone?

jwhitlock: Yes, this is the only code preventing two users from claiming the same relay number 🤦 
And, it won't work in Canada.

jwhitlock:
Thanks for the review @say-yawn! I'm closing this one to focus on the duplicate number issues, tracked in MPP-2434.

PR #2614, opened 2022-10-06
MPP-2434: Enforce one relay number per user

jwhitlock:
This PR fixes several things:

* Enforce that a user can have at most one `RelayNumber` at the Django level. Fix the test that checks this behavior.
* Enforce that a phone number can not be duplicated in two `RelayNumber`s, at the Django and database level
* Ignore duplicate phone registrations in Twilio (MPP-2332, from PR 2613, with a test)
* Refactor tests to use `pytest.raises` instead of a `try/except` pattern, check `ValidationError` messages


phones/models.py
```diff
@@ -231,12 +239,20 @@ def save(self, *args, **kwargs):
...
-            ).phone_numbers.create(phone_number_sid=twilio_incoming_number.sid)
+            try:
+                client.messaging.v1.services(
+                    settings.TWILIO_MESSAGING_SERVICE_SID
+                ).phone_numbers.create(phone_number_sid=twilio_incoming_number.sid)
+            except TwilioRestException as err:
+                if err.status == 409 and err.code == 21710:
+                    # Ignore "Phone Number is already in the Messaging Service"
+                    # https://www.twilio.com/docs/api/errors/21710
+                    pass
```
say-yawn, line 253: Nit: after the checks above, this means that we have a de-synced DB problem between our DB and Twilio's. We should log this. Can be a fast follow.

jwhitlock: I'm not sure I agree that we have a de-sync DB problem. Let's chat over Zoom and capture the next action in a ticket.

phones/tests/models_tests.py
```diff
@@ -122,13 +123,11 @@ def test_create_second_realphone_for_user_raises_exception(
     RealPhone.objects.create(user=phone_user, verified=True, number="+12223334444")
     mock_twilio_client = mocked_twilio_client
     mock_twilio_client.messages.create.assert_called_once()
-    try:
+    mock_twilio_client.reset_mock()
```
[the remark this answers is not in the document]
jwhitlock: I don't know of any case where you would want to create a copy of a mock, I don't think they work that way!

The reset is so that `mock_twilio_client.messages.assert_not_called()` will be true later. Without it, it will still have been called once from creating the original number.

jwhitlock:
Thanks @say-yawn! Let's sync up soon on video or in person and capture the future work.

phones/models.py
jwhitlock: Twillio allows a phone number to be reserved twice without an error. Twillio raises an error when registering a number with the Messaging Service. However, there's no additional information, so no DB sync issue if a number is already registered. No fix, but a log may be useful. In practice, this should only happen when the Relay Number is being manually added, such as via the Django Admin or a shell session, so it will be rare. A log may be useful to confirm that.

Validation in DRF / Django is not straightforward. Validation at the API level is good for error messages, but if you don't validate at save, there are code paths where invalid data can get in. There is room to try other approaches.

phones/tests/models_tests.py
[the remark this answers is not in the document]
jwhitlock: This was clearer in conversation. The issue is this line:

```
mock_twilio_client = mocked_twilio_client
```
This looks like a soft copy, but mocks don't work that way, it is more like making an alias for the mock, and is unneeded. The code can be refactored to remove this (and maybe change the fixture name to `mock_twilio_client` for less characters).

jwhitlock (commented):
@say-yawn and I talked in person. I'm adding comments for my new understanding of the issues.

Next PR:
* Add logging for duplicate registration event
* Refactor away `mocked_twilio_client` / `mock_twilio_client` renaming

Future work:
* Determine the most elegant and correct way to validate data at API call time and at Model save.

PR #2624, opened 2022-10-10
MPP-2449 Check remaining seconds not minutes

say-yawn:
This PR fixes [MPP-2449](https://mozilla-hub.atlassian.net/browse/MPP-2449).

How to test:
Check that the phone calls limits are enforced
1. From the second phone number, call the Relay Phone Mask
2. Keep up the call for at least 51 minutes, preferably even more
3. Close the call
4. Observe the Remaining call minutes counter

### DIFF
--- a/api/tests/phones_views_tests.py
+++ b/api/tests/phones_views_tests.py
@@ -1149,7 +1149,7 @@ def test_voice_status_completed_reduces_remaining_seconds_to_negative_value(
         "phone_limit_exceeded",
         extra={
             "fxa_uid": profile.fxa.uid,
-            "call_duration": 27,
+            "call_duration_in_seconds": 27,
             "relay_number_enabled": True,
             "remaining_seconds": -27,
             "remaining_minutes": 0,
--- a/api/tests/phones_views_tests.py
+++ b/api/tests/phones_views_tests.py
@@ -1149,6 +1149,7 @@ def test_voice_status_completed_reduces_remaining_seconds_to_negative_value(
         "phone_limit_exceeded",
         extra={
             "fxa_uid": profile.fxa.uid,
+            "call_duration": 27,
             "relay_number_enabled": True,
             "remaining_seconds": -27,
             "remaining_minutes": 0,
--- a/api/tests/phones_views_tests.py
+++ b/api/tests/phones_views_tests.py
@@ -1098,6 +1098,11 @@ def test_voice_status_completed_no_duration_error(phone_user):
 
 
 def test_voice_status_completed_reduces_remaining_seconds(phone_user):
+    # TODO: This test should fail since the Relay Number is disabled and
+    # the POST to our /api/v1/voice_status should ignore the the reduced remaining seconds.
+    # This is currently passing because the voice_status() is not checking
+    # if the user's Relay Number has hit the limit or is disabled (bug logged in MPP-2452).
+    # Keeping this test so we can correct it once MPP-2452 is completed.
     _make_real_phone(phone_user, verified=True)
     relay_number = _make_relay_number(phone_user, enabled=False)
     pre_request_remaining_seconds = relay_number.remaining_seconds
--- a/api/tests/phones_views_tests.py
+++ b/api/tests/phones_views_tests.py
@@ -1026,6 +1026,26 @@ def test_inbound_call_valid_twilio_signature_disabled_number(
     mocked_twilio_client.messages.create.assert_not_called()
 
 
+def test_inbound_call_valid_twilio_signature_no_remaining_seconds(
+    phone_user, mocked_twilio_client
+):
+    _make_real_phone(phone_user, verified=True)
+    relay_number = _make_relay_number(phone_user, remaining_seconds=0)
+    caller_number = "+15556660000"
+    mocked_twilio_client.reset_mock()
+
+    client = APIClient()
+    path = "/api/v1/inbound_call"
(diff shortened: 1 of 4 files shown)


api/tests/phones_views_tests.py
```diff
@@ -1026,6 +1026,26 @@ def test_inbound_call_valid_twilio_signature_disabled_number(
     mocked_twilio_client.messages.create.assert_not_called()
 
 
+def test_inbound_call_valid_twilio_signature_no_remaining_seconds(
```
jwhitlock, line 1029: 🎉 It is nice to have a test for this functionality!

jwhitlock (commented):
Thanks @say-yawn! 

This changes the message from "Number Is Out Of Minutes." to "Number Is Out Of Seconds". I'm not sure that is a positive change, since we use Minutes on the front-end. However, this is the 400 error sent to Twilio. The caller (calling the Relay number, not our user) gets the message "Sorry, an application error has occurered. Good-bye."

This also doesn't address the negative remaining minutes. That would require a change to [voice_status](https://github.com/mozilla/fx-private-relay/blob/603875c707596171dab7b633edd197cea010f72e/api/views/phones.py#L510) to prevent a negative number from being added to the database, or to [remaining_minutes](https://github.com/mozilla/fx-private-relay/blob/603875c707596171dab7b633edd197cea010f72e/phones/models.py#L195-L197) to prevent sending a negative number to the front-end. I still get a negative number in the UI.

api/tests/phones_views_tests.py
groovecoder, line 1029: It seems weird that the `test_voice_status_completed_reduces_remaining_seconds` test didn't catch this error? Should we update that test to make sure it's checking the right attribute is decremented?

groovecoder (approved):
This code looks good. We can make other fixes in other PRs.

api/tests/phones_views_tests.py
say-yawn: Added comments on a9314a8 why this test didn't catch the error

say-yawn: As per updated acceptance criteria on the MPP-2449, we are returning a non-negative integer on `remaining_minutes` completed in b9499ef

api/views/phones.py
```diff
@@ -509,6 +510,17 @@ def voice_status(request):
...
     relay_number.save()
+    if relay_number.remaining_seconds < 0:
+        profile = relay_number.user.profile_set.first()
+        info_logger.info(
+            "phone_limit_exceeded",
+            extra={
+                "fxa_uid": profile.fxa.uid,
+                "relay_number_enabled": relay_number.enabled,
+                "remaining_seconds": relay_number.remaining_seconds,
+                "remaining_minutes": relay_number.remaining_minutes,
```
jwhitlock, line 521: Please include the `call_duration` as well.

phones/models.py
```diff
@@ -194,7 +194,8 @@ class RelayNumber(models.Model):
 
     @property
     def remaining_minutes(self):
-        return floor(self.remaining_seconds / 60)
+        minutes_called = floor(self.remaining_seconds / 60)
+        return minutes_called if minutes_called > 0 else 0
```
jwhitlock, line 198: I'm used to seeing this using `max`:
```suggestion
        return floor(max(self.remaining_seconds, 0) / 60))
```

but I think explicit is better than implicit

jwhitlock (approved):
Thanks @say-yawn! This worked but I got a little weirdness.

I set my remaining time to 30 seconds in the admin, and called my relay number from Google Voice. I have AT&T Call Protect, which blocked the call and sent it to voicemail. I left myself a message and unblocked my Google Voice number with AT&T.

Now with less than 0 seconds remaining, I called myself again. This time I got the message  "Sorry, an application error has occurred. Good-bye.". Google Voice then proceeded to rapidly try to call again twice, creating 3 missed phone calls! Maybe it was Voice trying to be helpful, maybe Google doesn't like Firefox...

I think it would be useful to include the call duration in the `phone_limit_exceeded` log. That way, we can tell if the call itself was a short error call or the last call of the month. I'm happy to merge after that.

phones/models.py
say-yawn: completed 8e485f0

api/views/phones.py
say-yawn: Completed in b9e544d

api/tests/phones_views_tests.py
groovecoder: Cool, thanks for the `TODO`

jwhitlock (approved):

phones/tests/models_tests.py
```diff
@@ -380,6 +380,37 @@ def test_create_relaynumber_canada(phone_user, mocked_twilio_client):
...
+    mock_number_create.return_value = SimpleNamespace(sid=twilio_incoming_number_sid)
+
+    real_phone = "+14035551234"
+    RealPhone.objects.create(
+        user=phone_user, verified=True, number=real_phone, country_code="CA"
+    )
+
+    relay_number = "+17805551234"
+    relay_number_obj = RelayNumber.objects.create(user=phone_user, number=relay_number)
+    assert relay_number_obj.country_code == "CA"
```
groovecoder, line 398: 🇨🇦

groovecoder (approved):

PR #2641, opened 2022-10-11
Use new extension ".text" to avoid Dependabot

jwhitlock:
Dependabot assumes .txt files are Python requirements files, and is checking 650 terms in `blacklist.txt` against PyPI projects. There is not a way to ignore a file in the dependabot configuration if it is in a subdirectory (https://github.com/dependabot/dependabot-core/issues/1657)

Renaming this and `badwords.txt` to `.text` may speed up dependency scans.


Vinnl (approved):
:facepalm:

emails/apps.py
```diff
@@ -48,8 +48,8 @@ def __init__(self, app_name, app_module):
 
         # badwords file from:
         # https://www.cs.cmu.edu/~biglou/resources/bad-words.txt
```
Vinnl, line 50:
```suggestion
        # https://www.cs.cmu.edu/~biglou/resources/bad-words.txt
        # Using `.text` extension because of https://github.com/dependabot/dependabot-core/issues/1657
```

Vinnl:
Linking https://github.com/dependabot/dependabot-core/issues/2883, which the other issue was marked as a duplicate of, so I can find this ticket again if that ever gets resolved.

PR #2665, opened 2022-10-17
Remove references to unused partial

Vinnl:
I saw a Sentry error referencing these files. Generally people shouldn't come across a page that renders this template (it's only referenced in logout.html, which I think was only rendered in a rare case - see
https://github.com/mozilla/fx-private-relay/pull/1932), but when they do, might be good if it does render as expected.


jwhitlock (approved):
Thanks, works for me!

1. Change `privaterelay/settings.py`, `ACCOUNT_LOGOUT_ON_GET = False`
2. Log into dev server
3. Go to http://127.0.0.1:8000/accounts/logout/:
    * `main`: Get error "TemplateDoesNotExist at /accounts/logout/"
    * This branch: Get "Are you sure you want to sign out?" confirmation page

PR #2671, opened 2022-10-19
MPP-2451: Return JSON Validation error when a free user is at the email mask limit

jwhitlock:
For MPP-2451, add a call to `check_user_can_make_another_address` for validating `POST /api/v1/relayaddress`, so that it is a validation error returning JSON rather than a server error.

To test:

1. Create a free test user
2. Create 5 aliases
3. Attempt to create a new alias in the UI
    * [ ] Button changes from "+ Generate New Mask" to "Get Unlimited Email Masks"
4. Go to the browsable API at /api/v1/docs, attempt to create the new alias with this data (removing `block_list_emails` member):

    ```json
    {
      "enabled": true,
      "description": "description",
      "generated_for": "generated_for",
      "used_on": "used_on"
    }
    ```
    * [ ] Error is JSON:
    
        ```json
        {
          "non_field_errors": [
            "You must be a premium subscriber to make more than 5 aliases."
          ]
        }
        ```

### DIFF
--- a/api/serializers/__init__.py
+++ b/api/serializers/__init__.py
@@ -1,8 +1,16 @@
+from typing import Mapping
+
 from django.contrib.auth.models import User
 
 from rest_framework import serializers, exceptions
 
-from emails.models import Profile, DomainAddress, RelayAddress
+from emails.models import (
+    CannotMakeAddressException,
+    DomainAddress,
+    Profile,
+    RelayAddress,
+    check_user_can_make_another_address,
+)
 
 
 class PremiumValidatorsMixin:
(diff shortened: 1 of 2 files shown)


jwhitlock:
Force-push to remove debug code, and updated test instructions from local testing.

say-yawn (approved):
This looks good. I am guessing this means we are ok with the save on the model calling this check again? I think that is ok given that models during save does not call validate and as such this validation check happens twice, only during API calls.

jwhitlock:
> This looks good. I am guessing this means we are ok with the save on the model calling this check again? I think that is ok given that models during save does not call validate and as such this validation check happens twice, only during API calls.

I'm OK with the one in the save method for now. In general, Django doesn't like validation in the `save` method, but prefers it closer to the user input (like the DRF serializer or a view). We can revisit this when we have a good set of API-level tests in the far future.

PR #2675, opened 2022-10-20
MPP-2451: Add error_code, error_context to address creation errors

jwhitlock:
This addresses MPP-2451, and reverts a lot of PR #2671:

* Remove `RelayAddressSerializer.validate`. Validation remains in `RelayAddress.save()`
* Add `api.exceptions.RelayAPIException` as base class for Relay exceptions that will include an `error_code`
* Add a custom DRF error handler `api.views.relay_error_handler` that looks for `RelayAPIException` and adds more information:
    - `error_code` - An error code that identifies the error, and can be used as an index into client-side error handling or translation
    - `error_context` - Optional extra data needed for client-side error handling and translation. For example, the `"error_code": "free_tier_limit"` includes `"error_context": "{"free_tier_limit": 5}"
* Convert exceptions raised from creating `RelayAddress` and `DomainAddress` to raise one of a hierarchy of exceptions derived from `RelayAPIException`. These exceptions have the details of the status code, messages, `error_code`, etc., and are no longer passed at exception creation.
* Change `DomainAddressViewSet.perform_create` to no longer handle `CannotMakeAddressException`, since `relay_error_handler` now does the needed work.

The exception hierarchy:

* `APIException` - provided by Django REST Framework
    * `RelayAPIException` - includes `error_context()`, gets special error handling
        * `CannotMakeAddressException` - expected by lots of test code
            * `AccountIsPausedException` - `error_code` = `account_is_paused`, 403
            * `RelayAddrFreeTierLimitException` - `error_code` = `free_tier_limit`, 403
            * `DomainAddrFreeTierException` - `error_code` = `free_tier_no_subdomain_masks`, 403
            * `DomainAddrNeedSubdomainException` - `error_code` = `need_subdomain`, 400
            * `DomainAddrUnavailableException` - `error_code` = `address_unavailable`, 400

The exceptions for failing to set a subdomain and other runtime errors can be added into this hierarchy.

Some additional changes:
* Free users can pass `"block_list_emails": false` to `RelayAddressSerializer`. Previously, they had to omit the key entirely.
* `DomainAddress.make_domain_address` is now properly decorated as a `@staticmethod`
* Added type annotations to changed methods in `emails.models`
* Added type annotations to tests
* Refactored existing tests to use `with pytest.raises(CannotMakeAddressException)` instead of explicit `try`/`catch`/`return`/`fail`
* Tested models via the new `CannotMakeAddressException.get_codes()` rather than the sentence `.message`, which may change or get translated.

I'm going to tackle error translation (which we might not implement) in a future PR. However, I think this design would be compatible with translations, by generating a Fluent string ID from the error code (`account_is_paused` → `relay-api-error-account-is-paused`) and / or specifying a string ID on the exception.

## Test process

Suggested tests:

* Test website for flagged users
    * [x] Create or select a premium user  
    * [x] In the Django admin, add a `Last Account Flagged` of Today / Now for the test user. You may have to temporarily clear the Subdomain
    * [x] In the application, attempt to create a random mask. You get the message "This mask could not be created. Please try again"
    * [x] In the application, select a subdomain if needed, and attempt to create a subdomain mask. You get the message "This mask could not me created. Please try again."
    * [x] Go to http://127.0.0.1:8000/api/v1/docs/. Attempt to create a subdomain mask via `POST /domainaddresses`. You get the error JSON:
      ```json
      {
        "detail": "Your account is on pause.",
        "error_code": "account_is_paused"
      }
      ```
    * [x] Still in API docs, attempt to create a random mask via `POST /relayaddresses`. You get the error JSON:
      ```json
      {
        "detail": "Your account is on pause.",
        "error_code": "account_is_paused"
      }
      ```
    * [x] In the Django Admin, clear the Last Account Flagged and reset the subdomain.

Similar tests may be needed to check the add-on

### DIFF
--- a/api/tests/views_tests.py
+++ b/api/tests/views_tests.py
@@ -85,7 +85,7 @@ def test_post_domainaddress_no_subdomain_error(prem_api_client) -> None:
         format="json",
     )
 
-    assert response.status_code == 403
+    assert response.status_code == 400
     ret_data = response.json()
     assert ret_data == {
         "detail": (
@@ -123,7 +123,7 @@ def test_post_domainaddress_bad_address_error(prem_api_client) -> None:
         format="json",
     )
 
-    assert response.status_code == 403
+    assert response.status_code == 400
     ret_data = response.json()
     assert ret_data == {
         "detail": 'Domain address "myNewAlias" could not be created, try using a different value.',
@@ -167,7 +167,7 @@ def test_post_relayaddress_free_mask_email_limit_error(
 
     response = free_api_client.post(reverse("relayaddress-list"), {}, format="json")
 
-    assert response.status_code == 400
+    assert response.status_code == 403
     ret_data = response.json()
     assert ret_data == {
         "detail": (
@@ -187,7 +187,7 @@ def test_post_relayaddress_flagged_error(free_api_client) -> None:
 
     response = free_api_client.post(reverse("relayaddress-list"), {}, format="json")
 
-    assert response.status_code == 400
+    assert response.status_code == 403
     ret_data = response.json()
     assert ret_data == {
         "detail": "Your account is on pause.",
--- a/api/tests/views_tests.py
+++ b/api/tests/views_tests.py
@@ -1,6 +1,7 @@
 import pytest
 from model_bakery import baker
 
+from django.utils import timezone
 from django.urls import reverse
 from rest_framework.test import APIClient
 
@@ -35,7 +36,7 @@ def test_runtime_data(client):
 
 
 @pytest.mark.django_db
-def test_free_mask_email_limit_error(settings) -> None:
+def test_post_relayaddress_free_mask_email_limit_error(settings) -> None:
     """A JSON error is returned when a free user hits the mask limit"""
     free_user = make_free_test_user()
     free_profile = free_user.profile_set.get()
@@ -46,7 +47,6 @@ def test_free_mask_email_limit_error(settings) -> None:
     with pytest.raises(CannotMakeAddressException):
         check_user_can_make_another_address(free_profile)
 
-    url = reverse("relayaddress-list")
     client = APIClient()
     client.force_authenticate(user=free_user)
     origin = "https://login.example.com"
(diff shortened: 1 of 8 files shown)


jwhitlock:
After some sleep and re-reading the Django REST Framework (DRF) docs, I've re-worked the changes quite a bit.

The built-in DRF exceptions include the concept of error codes, used to identify errors. They also include the HTTP `status_code` that should be returned with the error. I've found it easier to work with DRF than against it, so I'm building on that. I'm also going to follow existing conventions, like `snake_case`.

After PR #2671, the error message for a free user failing to create a sixth random mask (`POST /api/v1/relayaddresses/`) is:

```json
{
  "non_field_errors": [
    "You must be a premium subscriber to make more than 5 aliases."
  ]
}
```

The new error message is:

```json
{
  "detail": "You must be a premium subscriber to make more than 5 aliases.",
  "error_code": "free_tier_limit",
  "error_context": {
    "free_tier_limit": 5
  }
}
```

A client could:

* Display `detail` to the end user
* Use `error_code` to handle the error
* Use `error_code` and `error_context` to construct their own error message

I'm also retaining the existing status codes, so `RelayAddress` errors return `400 Bad Request` and `DomainAddress` errors return `403 Permission Denied`. These details can be changed in the future.

I'm updating the PR message to reflect the current state of the code.

I think `error_code` could be turned into a Fluent string ID (such as `free_tier_limit` → `relay-api-error-free-tier-limit`) to facilitate optional translation of the `detail` message. I'm considering if that should be part of this PR or a new one.

jwhitlock:
Ready for review. I'm tagging @groovecoder, because he's touched API code the most, and @maxxcrawford, who said he can test the add-on locally. I believe the website is OK, since it just looks at status codes for errors, and those have not changed.

api/tests/views_tests.py
```diff
@@ -34,34 +60,136 @@ def test_runtime_data(client):
...
+
+def test_post_domainaddress_bad_address_error(prem_api_client) -> None:
+    """A domain address can be rejected due to the address format or content."""
+    response = prem_api_client.post(
+        reverse("domainaddress-list"),
+        data={"address": "myNewAlias"},
+        format="json",
+    )
+
+    assert response.status_code == 403
```
groovecoder, line 126: suggestion (non-blocking): this one seems like it should be `400` rather than `403`. I.e., no matter what `Authorization` or authentication is sent with this request, it will fail because of the content in the request - not because of the access level of the user.

api/tests/views_tests.py
```diff
@@ -34,34 +60,136 @@ def test_runtime_data(client):
...
+
+def test_post_relayaddress_flagged_error(free_api_client) -> None:
+    """A flagged user is unable to create a random mask."""
+    free_profile = get_user(free_api_client).profile_set.get()
+    free_profile.last_account_flagged = timezone.now()
+    free_profile.save()
+
+    response = free_api_client.post(reverse("relayaddress-list"), {}, format="json")
+
+    assert response.status_code == 400
```
groovecoder, line 190: suggestion (non-blocking): this one seems like it should be 403. i.e., if a different user made the same exact request, it could succeed.

groovecoder (approved):
Spot-checks and code look good.

(non-blocking) suggestion: Could we return a `403` status code when the request would succeed if it was sent by a different user, and return a `400` status code when the request would still fail no matter which user sent it?

By those rules, `account_is_paused`, `free_tier_limit`, `free_tier_no_subdomain_masks` would be `403` while `need_subdomain` and `address_unavailable` would be `400`. Is that intuitive enough to use as an overall guideline?

jwhitlock:
> (non-blocking) suggestion: Could we return a 403 status code when the request would succeed if it was sent by a different user, and return a 400 status code when the request would still fail no matter which user sent it?

I like it, and I think we'll have some time on stage to confirm that the website and the add-on can work with the new codes. It also simplifies the exception hierarchy - the status code does not distinguish between random and domain masks, so they two `account_is_paused` exceptions are unified and the intermediate exceptions go away.

groovecoder (approved):

PR #2677, opened 2022-10-20
Remove strings that are no longer pending

Vinnl:
Also, display a count of unmerged strings, to make it easier to check that I didn't accidentally remove a string that hadn't been submitted yet.


frontend/pendingTranslations.ftl
```diff
@@ -2,25 +2,14 @@
...
 nav-duo-email-mask-alt = Email masks
 nav-duo-phone-mask-alt = Phone masks
 
 setting-api-key-copied-alt = Click to copy
 
-waitlist-control-country-label-2 = What country or region do you live in?
-
-fx-desktop-2 = { -brand-name-firefox } for Desktop
-fx-mobile-2 = { -brand-name-firefox } for Mobile
 fx-containers = { -brand-name-firefox } Containers
```
Vinnl, line 13: @lloan Am I correct in thinking that this string does not get used anywhere in the codebase, and thus can be removed as well?

Vinnl:
I have submitted the remaining strings in https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/105.

PR #2679, opened 2022-10-21
Tell the browser to cache cache-busted resources

Vinnl:
All files in /_next/static/ apply cache busting, i.e. they are regenerated when they change with a different filename that includes a hash of the contents. This means that the browser should be able to cache them indefinitely.

(This change was suggested by [Webhint](https://webhint.io/).)

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -854,11 +854,14 @@ def set_index_cache_control_headers(
     headers: wsgiref.headers.Headers, path: str, url: str
 ) -> None:
     if DEBUG:
-        home_path = os.path.join(BASE_DIR, "frontend/out", "index.html")
+        base_path = os.path.join(BASE_DIR, "frontend/out")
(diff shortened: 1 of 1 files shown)


jwhitlock:
Thanks @Vinnl! This change looks good, but I'd like a day (Tuesday) to try the [ManifestStaticFilesStorage](https://docs.djangoproject.com/en/3.2/ref/contrib/staticfiles/#manifeststaticfilesstorage) and see if we can work with Django's solution to get similar results.

Vinnl:
@jwhitlock OK, let me know what the results are - it looks like Whitenoise does have explicit support for that. Keep in mind, though, that frontend files already get a hash appended to them during the build, and I'm not sure if that's something we can turn off. So I'm not sure if it's feasible to keep the old ones around...

Vinnl:
Superseded by #2704.

PR #2694, opened 2022-10-24
Tell virtual keyboards not to capitalise IDs

Vinnl:
Both email addresses and subdomains should be lowercase, so when the user is typing on a virtual keyboard, it should autocapitalise the first letter.

How to test: [try to](https://deploy-preview-2694--fx-relay-demo.netlify.app/?mockId=some) register a new subdomain and a new custom mask on mobile. The virtual keyboard should not uppercase the first letter.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, HTML attribute only


maxxcrawford (approved):
Confirmed behavior! I had one question about capitalization, but it's non-blocking. 

I love me some two-liner fixes. 🤠

frontend/src/components/dashboard/aliases/AddressPickerModal.tsx
```diff
@@ -118,6 +118,7 @@ export const AddressPickerModal = (props: Props) => {
                   placeholder={l10n.getString(
                     "modal-custom-alias-picker-form-prefix-placeholder"
                   )}
+                  autoCapitalize="none"
```
maxxcrawford, line 121: question: Looking at MDN, it's described as `autocapitalize`. 

Is having the camel casing (Capital "C") an issue? It worked as expected, but I wanted to double-check.

Vinnl: Yeah, that's one thing about React that it's good to be aware of: you're not writing HTML, but a description of what you want the DOM to look like - which gets built up programmatically by React. Hence, you're not actually passing attributes, but _properties_ (which is why their values can be actual JS objects, rather than just strings). This is why, for example, you write [`className`](https://developer.mozilla.org/en-US/docs/Web/API/Element/className) (the property) rather than [`class`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/class) (the attribute).

(Although it looks like `autocapitalize` [is actually all-lowercase](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement#instance_properties), but it's also listed as experimental, and [the type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d0244641c1191bf4c491cf545fc4a43a33255333/types/react/index.d.ts#L1886) to use CamelCase, and accompany it by a comment marking it as a "non-standard" attribute, so presumably React just picked a property name and went with it.)

Vinnl: Did some more digging to be sure, and that indeed seems to be the formatting they've chosen for this - and it's in a big list of other semi-standard attributes, it seems:

- https://github.com/facebook/react/blob/28a574ea8f5d0efa95f861abca26a67af47dd0c0/packages/react-dom-bindings/src/shared/possibleStandardNames.js#L22
- https://github.com/facebook/react/blob/28a574ea8f5d0efa95f861abca26a67af47dd0c0/fixtures/attribute-behavior/src/attributes.js#L143-L148
- https://github.com/facebook/react/blob/28a574ea8f5d0efa95f861abca26a67af47dd0c0/CHANGELOG.md?plain=1#L2189

PR #2697, opened 2022-10-24
MPP-2476 - Update bundle banner upsell UI

codemist:
This PR fixes #MPP-2476. It is a UI refresh of the bundle banner to call out the 40% savings for the VPN when purchasing the bundle, as well as clearly highlight the value props.

Preview/screenshot is in the comment section of this ticket: https://mozilla-hub.atlassian.net/jira/software/c/projects/MPP/boards/198?modal=detail&selectedIssue=MPP-2476

How to test:


frontend/src/components/landing/BundleBanner.tsx
```diff
@@ -114,15 +114,26 @@ export const BundleBanner = (props: Props) => {
...
-                <p />
-              </Localized>
+              <p>
+                {l10n.getString("bundle-banner-body-v2", { savings: "40%" })}
+              </p>
+              <p>
+                <strong>{l10n.getString("bundle-banner-plan-header")}</strong>
+              </p>
+              <div className={styles["bundle-banner-value-props"]}>
+                <div>
```
Vinnl, line 124: I feel like this would make more sense as a `<ul>` with `<li>`s in there, given that it's a list of included products? (With of course the same change applied to the CSS selectors.) Right now I think it reads a bit weird for screen reader users.

Vinnl (approved):
Looks good, would just like one small change to make things easier for people reliant on screen readers.

frontend/src/components/landing/BundleBanner.tsx
codemist: I was thinking about that using unordered list that but wasn't sure if it was the right use case. Good callout.

PR #2704, opened 2022-10-25
Add caching and compression for static files

jwhitlock:
This is an alternate for PR #2679 

Use Whitenoise's [CompressedManifestStaticFilesStorage](http://whitenoise.evans.io/en/stable/django.html#add-compression-and-caching-support) for hashed files and gzip-compressed versions, but extend it to skip hashing of Next.js files, which already include hashes in the filename.

Use Whitenoise's middleware, but consider all Next.js files to be immutable, so they get long cache times.

Next.js files are identified by being in the `_next/static` folder.

The behavior does not change with `DEBUG=True`. When logged out and requesting the homepage:

- The homepage has header `Cache-Control: no-cache, public`
- CSS like `_next/static/css/b10b975241d9d9b6.css` have header `Cache-Control: max-age=0, public`
- Images like `_next/static/media/hero-image-nopremium.ca4d2ad8.svg` have header `Cache-Control: max-age=0, public`
- Fonts like `fonts/Inter/Inter-Regular.woff2?v=3.18` have header `Cache-Control: max-age=0, public`

The change is with `DEBUG=False`:

* Stop the server if running
* Set `DEBUG=False` in `.env`
* Build frontend if needed (`cd frontend; npm install; npm run build; cd -`)
* Collect staticfiles with `./manage.py collectstatic -v 2 --clear --noinput`
    - Before this change, files are copied to `staticfiles/`, with an output like `1248 static files copied to 'my-projects/fx-private-relay/staticfiles'`
    - After this change, files are copied to `staticfiles/`, gzipped versions are created, and `staticfiles/staticfiles.json` is written, with output like `1248 static files copied to 'my-projects/fx-private-relay/staticfiles', 1664 post-processed.`
* Run the server with `./manage.py runserver`

Before this change, when logged out and requesting the `DEBUG=False` homepage, the cache times increase to 60 seconds:

- The homepage has header `Cache-Control: no-cache, public`
- CSS like `_next/static/css/b10b975241d9d9b6.css` have header `Cache-Control: max-age=60, public`
- Images like `_next/static/media/hero-image-nopremium.ca4d2ad8.svg` have header `Cache-Control: max-age=60, public`
- Fonts like `fonts/Inter/Inter-Regular.woff2?v=3.18` have header `Cache-Control: max-age=60, public`

After this change, with `DEBUG=False`, cache times increase to 10 years, and some are gzipped:

- The homepage still has header `Cache-Control: no-cache, public`, but also `Content-Encoding: gzip`
- CSS like `_next/static/css/b10b975241d9d9b6.css` have headers `Cache-Control: max-age=315360000, public, immutable`, `Content-Encoding: gzip`
- Images like `_next/static/media/hero-image-nopremium.ca4d2ad8.svg` have header `Cache-Control: max-age=315360000, public, immutable`, `Content-Encoding: gzip`
- Fonts like `fonts/Inter/Inter-Regular.4232a675a077.woff2?v=3.18` have header `Cache-Control: max-age=315360000, public, immutable` (URL has hash in name, but no gzip header)

### DIFF
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -127,6 +127,16 @@ class StoreFirstVisit:
 
 
 class RelayStaticFilesMiddleware(WhiteNoiseMiddleware):
+    """Customize WhiteNoiseMiddleware for Relay.
+
+    The WhiteNoiseMiddleware serves static files and sets headers. In
+    production, the files are read from staticfiles/staticfiles.json,
+    and files with hashes in the name are treated as immutable with
+    10-year cache timeouts.
+
+    This class also treats Next.js output files (already hashed) as immutable.
+    """
+
     def immutable_file_test(self, path, url):
         """
         Determine whether given URL represents an immutable file (i.e. a
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -10,6 +10,7 @@ from django.shortcuts import redirect
 
 from allauth.socialaccount.models import SocialToken
 from allauth.socialaccount.providers.fxa.views import FirefoxAccountsOAuth2Adapter
+from whitenoise.middleware import WhiteNoiseMiddleware
 
 from .views import _get_oauth2_session, update_social_token, NoSocialToken
 
@@ -123,3 +124,21 @@ class StoreFirstVisit:
         if first_visit is None and not request.user.is_anonymous:
             response.set_cookie("first_visit", datetime.now(timezone.utc))
         return response
+
+
+class RelayStaticFilesMiddleware(WhiteNoiseMiddleware):
+    def immutable_file_test(self, path, url):
+        """
+        Determine whether given URL represents an immutable file (i.e. a
+        file with a hash of its contents as part of its name) which can
+        therefore be cached forever.
+
+        All files outputed by next.js are hashed and immutable
+        """
+        if not url.startswith(self.static_prefix):
+            return False
+        name = url[len(self.static_prefix) :]
+        if name.startswith("_next/static/"):
+            return True
+        else:
+            return super().immutable_file_test(path, url)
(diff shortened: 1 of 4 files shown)


privaterelay/middleware.py
```diff
@@ -123,3 +124,21 @@ def __call__(self, request):
...
+        """
+        Determine whether given URL represents an immutable file (i.e. a
+        file with a hash of its contents as part of its name) which can
+        therefore be cached forever.
+
+        All files outputed by next.js are hashed and immutable
+        """
+        if not url.startswith(self.static_prefix):
+            return False
+        name = url[len(self.static_prefix) :]
```
jwhitlock, line 140: This code to remove the static prefix is from the Whitenoise code:

https://github.com/evansd/whitenoise/blob/ae01e9a63dccebf20de48df03f8f8b9cb457f7d8/src/whitenoise/middleware.py#L179-L181

Since our `STATIC_ROOT` is `/`, this trims off the initial slash from the URL. This will still work if we use a different `STATIC_ROOT` in the future.

https://github.com/evansd/whitenoise/blob/ae01e9a63dccebf20de48df03f8f8b9cb457f7d8/src/whitenoise/middleware.py#L173-L181

jwhitlock:
Our assets have gzip compression in production. It is possible that our nginx layer handles compression. In this case, we can probably use [ManifestStaticFileStorage](https://docs.djangoproject.com/en/stable/ref/contrib/staticfiles/#manifeststaticfilesstorage) instead.

Vinnl (approved):
This is great John, thanks for this. I verified that the heads are indeed updated, and that creating a new UI build with new hashes resulted in the new code getting loaded. Double gzipping is probably not too much of a problem, the `collectstatic` step is still pretty fast.

jwhitlock:
OK, merging this version. If relay-side gzip turns out to be Bad, the change in `privaterelay/storage`:

```python
from django.contrib.staticfiles.storage import ManifestStaticFilesStorage

class RelayStaticFilesStorage(ManifestStaticFilesStorage):
    ...
```

... and some docstring changes

PR #2707, opened 2022-10-26
MPP-2480 - Add bundle banner on dashboard

codemist:
# New feature description

This adds a promotional banner to the dashboard to both free and premium users who have not subscribed to the bundle plan yet.

# Screenshot (if applicable)

On this ticket:  https://mozilla-hub.atlassian.net/jira/software/c/projects/MPP/boards/198?modal=detail&selectedIssue=MPP-2480

# Checklist


frontend/src/components/Banner.module.scss
```diff
@@ -171,8 +170,8 @@
     display: block;
     padding: $spacing-md $spacing-lg;
     font-weight: 700;
-    background: $color-blue-50;
-    color: $color-white;
+    border: 2px solid $color-blue-50;
+    color: $color-blue-50;
```
Vinnl, line 174: You'll probably want to remove the `color` from the `:focus` styles - if you now tab to this button without hovering it with your mouse, the text is white-on-white.

frontend/src/components/dashboard/ProfileBanners.module.scss
```diff
@@ -5,3 +5,7 @@
   flex-direction: column;
   gap: $spacing-xl;
 }
+
+.bundle-logo {
+  min-width: 120px;
+}
```
Vinnl, line 11: I don't think this is needed; the width is constrained by this line: https://github.com/mozilla/fx-private-relay/blob/db6098a8ff27f007bf402d25901f6a84797cf8f3/frontend/src/components/Banner.module.scss#L100

`git blame` indicates that that line was added in 5f5c4c4cc20f2e25127e719979e9badd86f0ba42, for a banner (`<NoPremiumBanner>`) that is no longer being shown. So instead of adding more CSS that reverts some other CSS, it's probably better to just delete the line linked above :)

frontend/src/components/dashboard/ProfileBanners.tsx
```diff
@@ -67,6 +71,16 @@ export const ProfileBanners = (props: Props) => {
     );
   }
 
+  if (
+    isBundleAvailableInCountry(props.runtimeData) &&
+    isFlagActive(props.runtimeData, "bundle") &&
+    !props.profile.has_vpn
```
Vinnl, line 77: There's two things I'd double check with Tony here. First: do we also want to show this banner if the user already has phone masking? Or even regular Premium?

The second thing is that a free user can have a bit of a banner overload; this is a possible state:

image

We might want to choose between showing either the `<LoyalistPremiumBanner>` _or_ the `<BundlePromoBanner>` in some cases? Specifically, if the user has one or more masks, but does not have Premium nor VPN yet, which banner do we show?

Vinnl (approved):
Nice, you'd already submitted the strings earlier, good job!

I left a couple of things to check, but I think they're all easy to address without needing another review, so approving already.

frontend/src/components/dashboard/ProfileBanners.tsx
codemist: Confirmed with Tony this behaviour is fine for now.

frontend/src/components/dashboard/ProfileBanners.module.scss
codemist: @Vinnl Hmm, from my experience that line actually affects all of the banners with illustrations in them, not just `<NoPremiumBanner>`

Vinnl: Oh hmm, well, blame me then - would've been better to give the icons an explicit width I think. Ah well :P

PR #2711, opened 2022-10-27
Change User / Profile to a 1-to-1 relationship

### LINKED ISSUE #2708: In the database / Django models, a User should have exactly one Profile
We're using the user / profile pattern to store extended data for a user:

https://github.com/mozilla/fx-private-relay/blob/db6098a8ff27f007bf402d25901f6a84797cf8f3/emails/models.py#L84-L85

With a `ForeignKey`, it is possible for a `User` to have multiple `Profiles`, so we access the profile like this:

```python
profile1 = user1.profile_set.first()  # Could be None, mypy treats as Optional(Profile)
assert profile1                       # mypy treats as Profile, exception could be raised

profile2 = user2.profile_set.get()    # ObjectNotFound raised if no profile
```

The relationship is better modeled as a [One-to-one relationship](https://docs.djangoproject.com/en/3.2/topics/db/examples/one_to_one/):

```python
 class Profile(models.Model): 
     user = models.OneToOneField(User, on_delete=models.CASCADE)
```
We might be able to use `primary_key=true`, since we don't have any related models to `Profile`.

With this, the way to access a profile is:

```python
profile1 = user1.profile
```

With the shorter version, it is often easier to use `user.profile` rather than save it as a new variable.

We'll need to:
* Confirm that all users have one profile. A profile can be created for users without one (safe), and we can delete all but the earliest profile for others (risky). An exception is superusers, who do not need a Relay profile.
* Update the model and create a schema migration
* Update usage of `user.profile_set.first()` / `.get()` to `user.profile`

jwhitlock:
This PR fixes #2708.

Change `Profile.user` to a `OneToOneField`. This changes code like:

```python
profile1 = user1.profile_set.first()  # Could be None, mypy treats as Optional(Profile)
assert profile1                       # mypy treats as Profile, exception could be raised

profile2 = user2.profile_set.get()    # ObjectNotFound raised if no profile
```

to 

```python
profile1 = user1.profile
profile2 = user2.profile
```

Other changes:

* When a `profile` variable was used once, I removed it and instead used `user.profile`
* When `profile_set.prefetch_related` was used, I instead changed it to use [prefetch_related_objects](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#prefetch-related-objects) on the instance.
* I removed the `many_profile` detector of `./manage.py cleanup_data` also goes away, since it is now not possible to create Profiles for a User (database-enforced). We confirmed there are no instances in deployed instances in SVCSE-872.
* `bake.make(Profile)` tended to do the wrong thing, so I changed instances to `baker.make(User).profile`.


jwhitlock:
The migrations are failing, so I'll need to do this across two deployments 😢

PR #2712, opened 2022-10-28
MPP-2503: Prepare for profile.user as OneToOneField

jwhitlock:
This addresses issue #2708 / MPP-2503. It needs to be fixed over two deployments, first to prepare the code and tests, and second to make the migration and final code changes. PR #2711 attempted to do it all at once, and failed the migrations tests.

When `Profile.user` switches from a `ForeignKey` to a [OneToOneField](https://docs.djangoproject.com/en/3.2/topics/db/examples/one_to_one/), access will change from `user.profile_set.get()` to `user.profile`. Prepare the code for this change:
    
* Switch from `user.profile_set.first()` to `.get()`, which errors if there is not exactly one profile. We've confirmed that no users in deployed systems have multiple profiles, and ones with no profile appear to be unused, and could be fixed at runtime.
* Simplify some code if a `profile` variable was used once.
* Use [prefetch_related_objects](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#prefetch-related-objects) on a `User` object instead of `profile_set.prefetch_related`, which will stop working when the `profile_set` queryset is no longer available.
* In tests, use `baker.make(User).profile_set.get()` instead of `baker.make(Profile)`, because of weirdness using the later after the migration.
* Deprecate `./manage.py cleanup_data --many-profiles`, since soon the database will enforce that no user can have multiple profiles. Remove the test code that creates a use with multiple profiles, and the test that detects it.

**Testing**
I believe the integration tests are sufficient to show this code works.

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -82,6 +82,7 @@ def default_domain_numerical():
 
 
 class Profile(models.Model):
+    # TODO: #2708 Change to OneToOneField
     user = models.ForeignKey(User, on_delete=models.CASCADE)
     api_token = models.UUIDField(default=uuid.uuid4)
     num_address_deleted = models.PositiveIntegerField(default=0)
@@ -658,8 +659,7 @@ class RelayAddress(models.Model):
                         break
                     self.address = address_default()
                 locked_profile.update_abuse_metric(address_created=True)
-        profile = self.user.profile_set.first()
-        if not profile.server_storage:
+        if not self.user.profile_set.get().server_storage:
             self.description = ""
             self.generated_for = ""
             self.used_on = ""
@@ -851,7 +851,7 @@ class Reply(models.Model):
 
     @property
     def profile(self):
-        return self.address.user.profile_set.first()
+        return self.address.user.profile_set.get()
 
     @property
     def owner_has_premium(self):
--- a/emails/tests/cleaners_tests.py
+++ b/emails/tests/cleaners_tests.py
@@ -293,10 +293,7 @@ def setup_profile_mismatch_test_data(add_problems=False):
         return
 
     r2 = baker.make(User, email="regular2@example.com")
-    p2 = r2.profile_set.get()
-    r3 = baker.make(User, email="regular3@example.com")
-    # Assign user #2's profile to user #3, leaving r2 with none and r3 with 3
-    Profile.objects.filter(id=p2.id).update(user_id=r3.id)
+    r2.profile_set.get().delete()
 
 
 @pytest.mark.django_db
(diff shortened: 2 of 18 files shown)


jwhitlock:
Here's what the follow-up commit looks like:

https://github.com/mozilla/fx-private-relay/compare/prep-profile-user-one-to-one-2708...profile-user-one-to-one-2708

Once this PR is deployed, that commit should pass CircleCI checks, especially the migrations checks.

emails/tests/cleaners_tests.py
```diff
@@ -417,24 +409,3 @@ def test_many_profile_detector_with_users() -> None:
     One or no Profile: 4 (100.0%)
     Many Profiles    : 0 (  0.0%)"""
     assert report == expected
-
-
-@pytest.mark.django_db
-def test_many_profile_detector_with_problems() -> None:
```
groovecoder, line 423: question (non-blocking): should we remove this BEFORE we're enforcing single profiles with `OneToOneField`?

emails/views.py
```diff
@@ -368,9 +369,8 @@ def _sns_message(message_json):
         # RelayAddress or DomainAddress types makes the Rustacean in me throw
         # up a bit.
         address = _get_address(to_address, to_local_portion, to_domain_portion)
-        user_profile = address.user.profile_set.prefetch_related(
-            "user__socialaccount_set"
-        ).first()
+        prefetch_related_objects([address.user], "socialaccount_set", "profile_set")
```
groovecoder, line 372: praise: 😍

groovecoder (approved):
I ❤️ getting rid of `first()`! Thanks for doing this clean-up.

My only overall question is: do we also plan to get rid of `profile_set.get()` when we change to `OneToOneField` in the next PR? Seems like we better do that? So `profile_set.get()` is just a temporary change?

emails/models.py
```diff
@@ -658,8 +659,7 @@ def save(self, *args, **kwargs):
                         break
                     self.address = address_default()
                 locked_profile.update_abuse_metric(address_created=True)
-        profile = self.user.profile_set.first()
-        if not profile.server_storage:
+        if not self.user.profile_set.get().server_storage:
```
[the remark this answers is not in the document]
jwhitlock: No, `profile_set` will go away, so `self.user.profile_set.get()` will become `self.user.profile`.

emails/tests/cleaners_tests.py
```diff
@@ -297,10 +293,7 @@ def setup_profile_mismatch_test_data(add_problems=False):
         return
 
     r2 = baker.make(User, email="regular2@example.com")
-    p2 = r2.profile_set.first()
-    r3 = baker.make(User, email="regular3@example.com")
-    # Assign user #2's profile to user #3, leaving r2 with none and r3 with 3
-    Profile.objects.filter(id=p2.id).update(user_id=r3.id)
```
jwhitlock, line 303: These lines will fail when we switch to a `OneToOneField`. The `update(user_id=r3.id)` fails with an `IntegrityError` because it will violate the unique index on `user_id`.

It will work now, but will fail the sqlite / postgres migration tests in the future. Rather than try to detect the `IntegrityError`, I went ahead and removed the multiple profile code and just retained the "no profile" problem.

emails/tests/cleaners_tests.py
jwhitlock: Yes, we have to remove this test. The logic to test for many profiles remains for now, so `./manage.py cleanup_data --many-profiles` still works and will detect users with multiple profiles. However, the test setup will fail once `OneToOneField` is applied, so the migrations tests will fail. This (and `baker.make(Profile)`) blocked this from being a one-deployment change.

Issue #2726, opened 2022-10-31
On dev, styles are not applied on page changes


jwhitlock:
To reproduce:

1. Log on to dev server
2. Click "Phone Number" in the menu. The page content is unstyled.
3. Click "Home" in the menu. The page content is unstyled.

@Vinnl suspects the changes in PR #2704

**Update**: @Vinnl was correct. Django's `collectstatic` now changes some files so they refer to the cache-busting variants of icons, etc. Taking the hash of the collected files fixes the issue.

jwhitlock:
It was indeed the new static storage. In one file, the next.js version is:

```css
background-image:url(/icons/search-blue.svg);
```

Django updated the path to include the hash for the "permanent" version:

```css
background-image:url("/icons/search-blue.578593c4d4e3.svg")
```

Changing the code to calculate the SHA512 hash on the files in `staticfiles` fixes the issue.

jwhitlock:
PR #2729 fixed the issue in dev, and I believe it will work on stage as well. But, worth a check!

There are a few remaining issues that also appear on production. I'm going to write those up then close this ticket.

jwhitlock:
I've written the follow-up issues:

* #2744 
* #2745

PR #2728, opened 2022-10-31
Disable new storage/middleware

jwhitlock:
To address issue #2726, disable the new middleware and storage backends


jwhitlock:
This works, but I was also able to get it working by calculating the SHA512 from the copy in `staticfiles`. See PR #2729.

PR #2729, opened 2022-10-31
Calculate SHA512 for CSS from staticfiles

jwhitlock:
For issue #2726, calculate the SHA512 hash of CSS partials from the copy in `staticfiles/_next/static/css`, which appears to have a different hash than the source in `frontend/out/_next/static/css`.

Additionally, only add `.css` files, not `.css.map` files.


jwhitlock:
I found a difference, about 21,616 characters into a file. The next.js version is:

```css
background-image:url(/icons/search-blue.svg);
```

while the Django collected version is:

```css
background-image:url("/icons/search-blue.578593c4d4e3.svg")
```
This adds the permanent hash to the pathname, as well as quotes.

Calculating the SHA512 hashes on the `staticfiles` versions fixes the CSP issues, even with compression re-enabled.

This branch is currently running on the Heroku dev server.

groovecoder (approved):

Issue #2738, opened 2022-11-01
FxA profile updates rejected because created in future


jwhitlock:
FxA updates sent to `/fxa-rp-events` are being rejected because the creation time (`iat` field) is in the future. The rejection is `ImmatureSignatureError`. This is a new check introduced in `pyjwt` 2.6.0 (PR #2716).

* Team: https://sentry.io/organizations/mozilla/issues/3709316835
* Public: https://sentry.io/share/issue/a820378b604e49cbb1197f80b2c8b9ad/

Steps:
* Downgrade to pyjwt 2.5.0
* File upsteam issue
* Add logging of age (can be negative) of `iat`
* Determine mitigation (if a `leeway` parameter can cover `iat` variation, or FxA needs to fix)

jwhitlock:
The logging of `iat` claims has deployed to dev, stage, and prod. On stage and prod, the minimum age is 26 milliseconds, but no negative ages have been observed. On dev, we don't capture every log, but it looks like the minimum age is 200 milliseconds. This could be slower deliveries to Heroku, but may also be clock skew.

I believe the next step is to update to pyjwt 2.6.0, and use a leeway of 5 seconds or so. That should cover any minor clock skew. I hope to be able to log any future `iat` claims, so that we can see if our leeway was too low or if it looks like shenanigans.

jwhitlock:
I've checked again:

* Development: no negative `iat` in recent papertrail logs
* Stage: youngest `iat` was 26 milliseconds over 167,000 events, none negative
* Production: youngest `iat` was 28 milliseconds, 70,000 events, none negative

I'm dropping the logging of future `iat` claims, it looks like we just had a bad week the first time I deployed PyJWT 2.6.0.

Owenmiles17:
________________________________

PR #2739, opened 2022-11-01
Downgrade PyJWT from 2.6.0 to 2.5.0

jwhitlock:
PyJWT adds a checks that the RSA256 JWT iat claim (Issued At) is not in the future. This is failing for FxA updates (issue #2738).


Vinnl (approved):
Looks good with the comment, merging immediately because this is blocking the weekly stage deploy.

Issue #2736, opened 2022-11-01
django_ftl doesn't load brands.ftl


Vinnl:
Since we [split out](https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/103) our `app.ftl` into several files, I updated the UI to load all `.ftl` files (c6cf93d2c26c50bc97b308f1101f3186d132a6c6), but assumed that that would not be necessary for the back-end, since no phone and bundle strings are rendered there.

But of course, further rework also split out `brands.ftl`, which _is_ referenced in the email. So we'll at least have to load that as well.

PR #2737, opened 2022-11-01
Also load brand strings

Vinnl:
This PR fixes #2736.

How to test: I haven't been able to test this - when I load `wrapped_email_test`, it appears as expected. But, I deployed it to dev just now, and verified that the bug disappeared - the forwarded email contained no longer said `-brand-name-relay-premium⁩`, whereas the `main` version did.

- [x] l10n changes have been submitted to the l10n repository, if any.
- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, config change
- [x] I've added or updated relevant docs in the docs/ directory.
- [x] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #2740, opened 2022-11-01
Fail `test_wrapped_email_test` if Fluent errors for `en`

jwhitlock:
Detect if the email headers contain a Fluent ID that isn't in the English corpus. This detects issue #2736.


Vinnl (approved):
Verified that this no longer fails with #2737. Nice!

Vinnl:
Hmm, looks like GitHub doesn't detect that this has been merged in #2737 when using rebases rather than merges - closing manually.

PR #2743, opened 2022-11-01
Simplify base template, fix logout page

### LINKED ISSUE #2742: Logout page is broken
The logout page at `/accounts/logout/` is broken on `main`, returning a 500 Server Error. This is due to the changes in PR #2712, but also will fail due to references to static files that are no longer available:

https://github.com/mozilla/fx-private-relay/blob/bea767e096aca59374f2d3d067ba35c201edcd5a/privaterelay/templates/base.html#L20-L23

jwhitlock:
This PR fixes #2742. This page is viewed when a user goes directly to `/accounts/logout/`. Most users bypass this page when selecting "Sign Out", which POSTs to the page, but it is used by some django-allauth error paths.

* ~Fix~ Remove the reference to `request.user.profile_set.get`, which is unused.
* Drop references to static files that have been removed after the move to the frontend. The few views that are left can be mostly unstyled.
* Simplify HTML structure, since no CSS left to style the content
* Add a test that the logout page renders for a logged-in user.

### DIFF
--- a/privaterelay/templates/account/logout.html
+++ b/privaterelay/templates/account/logout.html
@@ -5,7 +5,6 @@
 
 {% block head_title %}{% ftlmsg 'nav-profile-sign-out' %}{% endblock %}
 
-
 {% block content %}
 <p>{% ftlmsg 'nav-profile-sign-out-confirm' %}</p>
 <form method="post" action="{% url 'account_logout' %}">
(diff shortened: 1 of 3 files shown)


Vinnl (approved):
I'm not sure how to get to the logout page myself (adding `ACCOUNT_LOGOUT_ON_GET = False` still logs me out immediately), but all the changes are sensible.

PR #2749, opened 2022-11-02
MPP-2503: Change Profile.user to OneToOneField

jwhitlock:
**Note**: This will fail CircleCI migration tests until `2022.11.02` is in production

Change Profile.user from ForeignKey to OneToOneField, to fix issue #2708 / MPP-2503. This requires some follow-on changes:

* Change `user.profile_set.get()` to `user.profile`
* Change `prefetch_related_objects('profile_set')` to `prefetch_related_objects('profile')`
* Remove `ManyProfileDetector` (no longer possible)
* Drop `./manage.py cleanup_data --many-profiles`

Also makes some changes suggested by `flake8`, such as removing unused imports.


emails/cleaners.py
```diff
@@ -215,66 +215,3 @@ def markdown_report(self) -> str:
             )
 
         return "\n".join(lines)
-
-
-class ManyProfileDetector(DetectorTask):
```
groovecoder, line 220: praise: 🪓

emails/tests/views_tests.py
```diff
@@ -217,21 +217,21 @@ def test_sns_message_with_hard_bounce(self):
 
         _sns_notification(BOUNCE_SNS_BODIES["hard"])
 
-        profile = self.user.profile_set.get()
-        assert profile.last_hard_bounce >= pre_request_datetime
+        self.user.refresh_from_db()
```
groovecoder, line 220: question (non-blocking): does `refresh_from_db()` also refresh related field data? Looks like most of these tests are written such that they might pass even if the data was stale?

groovecoder (approved):
Code and spot-checks look good! 1 non-blocking question.

PR #2751, opened 2022-11-04
MPP-2498: Localize API error messages

maxxcrawford:
This PR fixes [MPP-2498](https://mozilla-hub.atlassian.net/browse/MPP-2498)

~The following PR needs to land before this is merged in (Legacy L10N System Requirements): 
https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/111~

How to test:

- Be sure to check out the latest translations
- Set browser language to DE _(Note that as time of writing testing steps, DE worked)_ 
- Log in with free-tier account and create the maximum allowed email masks (5) 
- Go to API page: http://127.0.0.1:8000/api/v1/docs/
- Go to `relayaddresses` POST entry
- Try it out
   - Be sure to change `block_list_emails` value to FALSE
- **Expected:** The response should FAIL and the message should be in the German language. 

### Screenshot

image

TBD

- [X] I've added or updated relevant docs in the docs/ directory.

### DIFF
--- a/api/tests/views_tests.py
+++ b/api/tests/views_tests.py
@@ -88,9 +88,7 @@ def test_post_domainaddress_no_subdomain_error(prem_api_client) -> None:
     assert response.status_code == 400
     ret_data = response.json()
     assert ret_data == {
-        "detail": (
-            "You must select a subdomain before creating email address with subdomain."
-        ),
+        "detail": ("Please select a subdomain before creating a custom email address."),
         "error_code": "need_subdomain",
     }
 
@@ -125,8 +123,10 @@ def test_post_domainaddress_bad_address_error(prem_api_client) -> None:
 
     assert response.status_code == 400
     ret_data = response.json()
+    # Add unicode characters to get around Fluent.js using unicode isolation.
+    # See https://github.com/projectfluent/fluent.js/wiki/Unicode-Isolation for more info
     assert ret_data == {
-        "detail": 'Domain address "myNewAlias" could not be created, try using a different value.',
+        "detail": "“\u2068myNewAlias\u2069” could not be created. Please try again with a different mask name.",
         "error_code": "address_unavailable",
         "error_context": {"unavailable_address": "myNewAlias"},
     }
@@ -140,8 +140,10 @@ def test_post_domainaddress_free_user_error(free_api_client):
 
     assert response.status_code == 403
     ret_data = response.json()
+    # Add unicode characters to get around Fluent.js using unicode isolation.
+    # See https://github.com/projectfluent/fluent.js/wiki/Unicode-Isolation for more info
     assert ret_data == {
-        "detail": "You must be a premium subscriber to create subdomain aliases.",
+        "detail": "Your free account does not include custom subdomains for masks. To create custom masks, upgrade to \u2068Relay Premium\u2069.",
         "error_code": "free_tier_no_subdomain_masks",
     }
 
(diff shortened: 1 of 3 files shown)


jwhitlock:
Looks like Fluent is adding some Unicode magic for [bi-directional text](https://en.wikipedia.org/wiki/Bidirectional_text#Isolates):

* `\u2068` - [First Strong Isolate](https://unicodeplus.com/U+2068)
* `\u2069` - [Pop Directional Isolate](https://unicodeplus.com/U+2069)

I think those escaped values are needed in the expected results.

gregtatum:
https://github.com/projectfluent/python-fluent/blob/master/fluent.runtime/docs/internals.rst#using-fluentbundle

I see tests for it on the python, and it looks like there is a flag for turning it off `use_isolating=False`.

If you don't have any RTL text, it's fine to turn it off. If you expect on adding RTL languages (Hebrew, Arabic) in the future, then you'll want to use them.

api/tests/views_tests.py
```diff
@@ -125,8 +123,10 @@ def test_post_domainaddress_bad_address_error(prem_api_client) -> None:
 
     assert response.status_code == 400
     ret_data = response.json()
+    # Add unicode characters to get around Fluent.js using unicode isolation.
+    # See https://github.com/projectfluent/fluent.js/wiki/Unicode-Isolation for more info
     assert ret_data == {
-        "detail": 'Domain address "myNewAlias" could not be created, try using a different value.',
+        "detail": "“\u2068myNewAlias\u2069” could not be created. Please try again with a different mask name.",
```
groovecoder, line 129: question (non-blocking): I read the fluent doc, but I don't quite understand why `"myNewAlias"` needs unicode isolation markers? Is it because of the directional quote characaters `“` and `”`? But if it is, then don't those directional quote characters needs to be *inside* the unicode isolation markers?

api/views/__init__.py
```diff
@@ -212,18 +214,34 @@ def relay_exception_handler(exc: Exception, context: Mapping) -> Optional[Respon
     """
     Add error information to response data.
 
-    When the error is a RelayAPIException, these additional fields may be present:
+    When the error is a RelayAPIException, these additional fields may be present and
+    the information will be translated if a accept-language header is added to the request:
```
groovecoder, line 218: 🦅 👁️ (blocking):
```suggestion
    the information will be translated if an Accept-Language header is added to the request:
```

api/views/__init__.py
```diff
@@ -212,18 +214,34 @@ def relay_exception_handler(exc: Exception, context: Mapping) -> Optional[Respon
...
+
+            # Replace default message with Fluent string
+            response.data["detail"] = ftl_bundle.format(ftl_id, error_context)
+
         if error_context:
             response.data["error_context"] = error_context
+
+        response.data["error_code"] = error_codes
+
+    print(response)
```
groovecoder, line 245: change (blocking): remove debug(?) line.
```suggestion
```

groovecoder (changes requested):

api/views/__init__.py
maxxcrawford: 😅

api/tests/views_tests.py
maxxcrawford: I don't have a good answer for this. 

cc/ @jwhitlock

jwhitlock: I've got some Mozilla experience, but I don't claim to be an expert.

Other languages use [different characters for quotes](https://en.wikipedia.org/wiki/Quotation_mark#Summary_table), so I'd expect it as part of the translated pattern.

I'm not able to find guidance on quotation marks in the [styleguide](https://mozilla-l10n.github.io/styleguides/). There is an example in the [Urdu guide](https://mozilla-l10n.github.io/styleguides/ur/index.html?highlight=quotation#quotation-marks) (ur, right-to-left language):

```
Example: Instead, you might press “Enter”.
Correct: اس کے بجائے، آپ “اینٹر” دبا سکتے ہیں۔
```

I believe French would be inclined to use « » for quoting, as seen on https://github.com/mozfr/besogne/wiki/Guide-stylistique-pour-la-traduction

jwhitlock: @gregtatum confirmed that the isolation markers are at the right place, surrounding the value but not the quotes. This provides three benefits:

* The quote characters can be replaced, such as «»
* If the value is right-to-left and the translation is left-to-right, it is displayed correctly
* If the value is left-to-right and the translation is right-to-left, it is displayed correctly

PR #2769, opened 2022-11-07
Pin the versions of our GitHub Actions

Vinnl:
There are two reasons for doing this:
- To avoid our CI pipelines suddenly failing due to new (theoretically non-breaking) minor or patch versions of Actions we use.
- To be able to skip a single version of an Action, while still having Dependabot notify us of new later versions. This is something I had to do recently: https://github.com/mozilla/fx-private-relay/pull/2523#issuecomment-1250730713

Note that I did not pin the CodeQL jobs, since I can't find the
right version numbers and GitHub keeps complaining (even with .0
and .1) that it can't find them.


jwhitlock (approved):
Thanks @Vinnl! I'm a fan of pinning requirements, intentionally upgrading CI versions.

PR #2773, opened 2022-11-08
Fix menu overlap, adjust z-indexes

Vinnl:
This PR fixes MPPUX-754, aka the "Blocking" (or "Forwarding") label overlapping the mobile menu.

image

I fixed it by giving the content its own [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context), which should ensure that the mobile menu and header will now always overlap the content, even if there are z-indexes used internally in the content. And after having done that, I evaluated the rest of the z-indexes, lowered them where possible, did the same for the Help & Tips area to overlap the content, and removed a bunch of unneeded and undocumented z-indexes on underlays that were just copy-pasted (I'm using the passive voice, but I did that :stuck_out_tongue:) from the react-aria docs. I grouped each change in its own commit.

How to test: open the phone dashboard, and ensure your viewport is small. Open the mobile menu; the "Forwarding" and "Blocking" labels should disappear behind it when you scroll.

### DIFF
--- a/frontend/src/components/dashboard/tips/Tips.module.scss
+++ b/frontend/src/components/dashboard/tips/Tips.module.scss
@@ -52,8 +52,6 @@
   background-color: $color-white;
   width: $content-sm;
   max-width: calc(100% - $spacing-xl);
-  // Prevent the BlockLevelSlider thumb, which is also absolutely positioned
-  z-index: 1;
 
   .header {
     display: flex;
--- a/frontend/src/components/dashboard/aliases/AddressPickerModal.module.scss
+++ b/frontend/src/components/dashboard/aliases/AddressPickerModal.module.scss
@@ -9,7 +9,6 @@
   bottom: 0;
   left: 0;
   right: 0;
-  z-index: 10;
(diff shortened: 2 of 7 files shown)


frontend/src/components/dashboard/tips/Tips.module.scss
```diff
@@ -52,8 +52,6 @@
   background-color: $color-white;
   width: $content-sm;
   max-width: calc(100% - $spacing-xl);
-  // Prevent the BlockLevelSlider thumb, which is also absolutely positioned
```
Vinnl, line 55: I could remove this because I added `isolation: isolate` to `profile.module.scss`.

lloan (approved):
Very cool fix. I had not seen `isolation` used before. TIL. 

For anyone else curious as I was: https://developer.mozilla.org/en-US/docs/Web/CSS/isolation

Vinnl:
I know right! Same here :grin:

PR #2786, opened 2022-11-11
(draft) new API endpoints

groovecoder:
# API endpoit to create outbound calls
# TODOs
1. [ ] Put this behind waffle flag
2. [ ] Check the relay number has remaining call minutes
3. [ ] After the call, deduct the number of minutes of the call

## Screenshot (if applicable)
image

## How to test
1. Sign in to http://127.0.0.1:8000 with an account that has phone level of Relay
4. Go to http://127.0.0.1:8000/api/v1/docs/
5. Expand the `POST /call` endpoint at the top
6. Click "Try it out"
7. Replace `{"to": "string"}` with `{"to": "+12223334444"}` (use a real  number you have access to)
8. Click "Execute"
   * [ ] You should receive a call to your true phone number from your relay number
   * [ ] When you answer, it should say "Dialing +12223334444"
   * [ ] It should connect your call to +12223334444

# Checklist


api/views/phones.py
```diff
@@ -611,6 +613,87 @@ def sms_status(request):
...
+        )
+        outbound_messages = client.messages.list(
+            from_=relay_number.number, to=contact.inbound_number
+        )
+        return response.Response(
+            {
+                "inbound_messages": inbound_messages,
+                "outbound_messages": outbound_messages,
+            },
+            status=200,
```
codemist, line 683: Could you split the messages into two parts instead of having it contain both `inbound_messages` and `outbound_messages`?

groovecoder: How about a url parameter?

groovecoder:
image

api/views/phones.py
codemist: Yeah that works too

groovecoder:
Closing this PR. We can keep the branch and push it up to dev server as-needed for clients to work on it. We can also re-open this when we have work planned that needs new API endpoints.

PR #2804, opened 2022-11-14
MPP-2405: add new API_DOCS_ENABLED env var for api docs

groovecoder:
# API_DOCS_ENABLED
We want to show the API docs (especially on the dev server) even when `DEBUG=False`.

## How to test
1. Set env var `DEBUG=0` and remove any `API_DOCS_ENABLED` var value
2. Visit http://127.0.0.1:8000/api/v1/docs/
   * [ ] You should see "Not Found"
1. Set env var `DEBUG=0` and `API_DOCS_ENABLED=1`
2. Visit http://127.0.0.1:8000/api/v1/docs/
   * [ ] You should still see the API docs!
1. Set env var `DEBUG=0` and `API_DOCS_ENABLED=0`
2. Visit http://127.0.0.1:8000/api/v1/docs/
   * [ ] You should see "Not Found"

## Checklist

### DIFF
--- a/.env-dist
+++ b/.env-dist
@@ -1,3 +1,4 @@
+API_DOCS_ENABLED=False
 FXA_OAUTH_ENDPOINT=https://oauth.stage.mozaws.net/v1
 FXA_PROFILE_ENDPOINT=https://profile.stage.mozaws.net/v1
 FXA_BASE_ORIGIN=https://accounts.stage.mozaws.net
--- a/api/tests/views_tests.py
+++ b/api/tests/views_tests.py
@@ -2,6 +2,7 @@ import pytest
 from model_bakery import baker
 
 from django.contrib.auth.models import User
+from django.test import override_settings
 from django.utils import timezone
 from django.urls import reverse
 from rest_framework.test import APIClient
@@ -39,6 +40,7 @@ def get_user(client: APIClient) -> User:
 
 
 @pytest.mark.parametrize("format", ("yaml", "json"))
+@override_settings(API_DOCS_ENABLED=True)
 def test_swagger_format(client, format):
     path = f"/api/v1/swagger.{format}"
     response = client.get(path)
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -227,11 +227,17 @@ INSTALLED_APPS = [
     "api.apps.ApiConfig",
 ]
 
+API_DOCS_ENABLED = config("API_DOCS_ENABLED", False, cast=bool) or DEBUG
+if API_DOCS_ENABLED:
+    INSTALLED_APPS += [
+        "drf_yasg",
+    ]
+
 if DEBUG:
     INSTALLED_APPS += [
         "debug_toolbar",
-        "drf_yasg",
     ]
+
 if USE_SILK:
     INSTALLED_APPS.append("silk")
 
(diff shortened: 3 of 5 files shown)


say-yawn (changes requested):
Recommendation on `API_DOCS_ENABLED` to be falsey and that we use boolean rather than int.

.env-dist
```diff
@@ -1,3 +1,4 @@
+API_DOCS_ENABLED=1
```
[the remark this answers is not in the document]
groovecoder: Updated to `API_DOCS_ENABLED=False` by default.

privaterelay/settings.py
```diff
@@ -810,13 +816,11 @@ def set_index_cache_control_headers(
...
 
+DRF_RENDERERS = ["rest_framework.renderers.JSONRenderer"]
 if DEBUG and not IN_PYTEST:
-    DRF_RENDERERS = [
+    DRF_RENDERERS += [
         "rest_framework.renderers.BrowsableAPIRenderer",
-        "rest_framework.renderers.JSONRenderer",
     ]
-else:
-    DRF_RENDERERS = ["rest_framework.renderers.JSONRenderer"]
```
say-yawn, line 819: Nice clean up 👍

privaterelay/settings.py
```diff
@@ -227,11 +227,17 @@
...
+API_DOCS_ENABLED = config("API_DOCS_ENABLED", False, cast=bool)
+if API_DOCS_ENABLED:
+    INSTALLED_APPS += [
+        "drf_yasg",
+    ]
+
 if DEBUG:
     INSTALLED_APPS += [
         "debug_toolbar",
-        "drf_yasg",
```
say-yawn, line 233: Should keep this in here so that if in `DEBUG` mode we can always get Swagger. Or change the if condition in 231 to be `if API_DOCS_ENABLED or DEBUG:` so we don't accidentally add the app twice.

privaterelay/settings.py
```diff
@@ -227,11 +227,17 @@
     "api.apps.ApiConfig",
 ]
 
+API_DOCS_ENABLED = config("API_DOCS_ENABLED", False, cast=bool)
```
say-yawn, line 230:
```suggestion
API_DOCS_ENABLED = config("API_DOCS_ENABLED", False, cast=bool) or DEBUG
```
This should ensure that if we are in `DEBUG` mode we'll always get the swagger doc and the endpoints.

api/tests/views_tests.py
```diff
@@ -39,6 +40,7 @@ def get_user(client: APIClient) -> User:
 
 
 @pytest.mark.parametrize("format", ("yaml", "json"))
+@override_settings(API_DOCS_ENABLED=True)
```
say-yawn, line 43: Seems like the [test is failing](https://app.circleci.com/pipelines/github/mozilla/fx-private-relay/9848/workflows/cde4aef2-a439-4851-811b-f6051d9c5208/jobs/55033) because the url is not registered when the test started... This may be fixed if we update the logic for `API_DOCS_ENABLED` in privaterelay/settings.py (see comment below).

say-yawn (changes requested):
Requesting update on using either `API_DOCS_ENABLED` or `DEBUG` so that if either is true we get the swagger docs and comment on why I think the [Circle CI tests](https://app.circleci.com/pipelines/github/mozilla/fx-private-relay/9848/workflows/cde4aef2-a439-4851-811b-f6051d9c5208/jobs/55033) are failing.

api/tests/views_tests.py
groovecoder: Cool, thanks. Giving that a try.

say-yawn (approved):
LGTM.

PR #2811, opened 2022-11-16
MPP-2581 - Track last call and text times for inbound contacts

jwhitlock:
For MPP-2581, expand the `InboundContact` model to include new nullable datetime fields, `last_call_time` and `last_text_time`. When sending a text reply, `last_text_time` can be used to determine the last contact that texted, even when the contact called after it texted.

## How to test

Manual testing requires an environment setup for relaying voice calls and text messages. You'll need:

* An account that has phones enabled with a true phone (number 1) and a relay number (number 2)
* A third number (contact 1) that can send and receive texts
* A fourth number (contact 2) that can text and call

The testing starts on the released tag:

* Checkout `2022.11.09`, and run the webserver and any other services needed for phone relay.
  * [x] Send a message from contact 1 to the relay number. The true phone gets the relayed message.
  * [x] Reply from the true phone to the relay number. The reply goes to contact 1.
* Optionally confirm the issue:
   * [x] Send a message from contact 2 to the relay number. The true phone gets the relayed message.
   * [x] Call the relay number from contact 2. The true phone gets a voice call.
   * [x] Reply from the true phone to the relay number. The reply goes to contact 1, not contact 2
* Update to this branch. Run migrations `./manage.py migrate phones`, and restart the server as needed
* Test the changes:
   * [x] Reply from the true phone to the relay number. The reply goes to contact 1.
   * [x] Send a message from contact 2 to the relay number. The true phone gets the relayed message.
   * [x] Call the relay number from contact 2. The true phone gets a voice call.
   * [x] Reply from the true phone to the relay number. The reply goes to contact 2.

## Checklist


api/tests/phones_views_tests.py
```diff
@@ -899,6 +902,111 @@ def test_inbound_sms_reply(phone_user, mocked_twilio_client):
...
+    client = APIClient()
+    path = "/api/v1/inbound_sms"
+    data = {"From": real_phone.number, "To": relay_number.number, "Body": "test reply"}
+    response = client.post(path, data, HTTP_X_TWILIO_SIGNATURE="valid")
+
+    assert response.status_code == 200
+    mocked_twilio_client.messages.create.assert_called_once()
+    call_kwargs = mocked_twilio_client.messages.create.call_args.kwargs
+    assert call_kwargs["to"] != inbound_contact1.inbound_number
+    assert call_kwargs["to"] == inbound_contact2.inbound_number
```
groovecoder, line 992: question (blocking) Is this test right? This test seems like it's saying that a text reply should go to the number that sent the last *call* ? Is that really what we want?

api/views/phones.py
```diff
@@ -704,6 +704,8 @@ def _check_and_update_contact(inbound_contact, contact_type, relay_number):
     inbound_contact.last_inbound_type = contact_type[:-1]
     attr = f"num_{contact_type}"
     setattr(inbound_contact, attr, getattr(inbound_contact, attr) + 1)
+    last_date_attr = f"last_{contact_type[:-1]}_date"
+    setattr(inbound_contact, last_date_attr, inbound_contact.last_inbound_date)
```
groovecoder, line 708: micro-optimization (non-blocking): pull `contact_type[:-1]` into a var.
```suggestion
    contact_type_value = contact_type[:-1]
    inbound_contact.last_inbound_type = contact_type_value
    attr = f"num_{contact_type}"
    setattr(inbound_contact, attr, getattr(inbound_contact, attr) + 1)
    last_date_attr = f"last_{contact_type_value}_date"
    setattr(inbound_contact, last_date_attr, inbound_contact.last_inbound_date)
```

groovecoder (changes requested):

api/tests/phones_views_tests.py
jwhitlock: This is the sequence from the user's perspective:

1. Contact 1 texts
2. Contact 2 texts
3. Contact 2 calls

I think the text reply should go to the last texter, Contact 2, even though they then called. Before this PR, it goes to Contact 1.

groovecoder: Makes sense, but I don't see those steps 1 & 2 in this test?

groovecoder (changes requested):

jwhitlock:
> As per Slack - let's try to update this test to make it a little more clear that it's testing the specific sequence in [MPP-2581](https://mozilla-hub.atlassian.net/browse/MPP-2581).

I've refactored the tests to use the API, which makes the sequences clearer and hides the details of `InboundContact` (except when needed to emulate the existing code).

groovecoder (approved):
Spot-check works too!

PR #2814, opened 2022-11-16
MPP-2581 - Track last call and text times for inbound contacts

jwhitlock:
For MPP-2581, expand the `InboundContact` model to include new nullable datetime fields, `last_call_time` and `last_text_time`. When sending a text reply, `last_text_time` can be used to determine the last contact that texted, even when the contact called after it texted.

This is the same as PR #2811, but against the `main` branch.

## How to test

See reproduction steps on PR #2811.

## Checklist


groovecoder (approved):
Already reviewed https://github.com/mozilla/fx-private-relay/pull/2811.

PR #2835, opened 2022-11-21
Explicitly enumerate known Waffle flags

Vinnl:
This ensures that we're at least using the same spelling everywhere (e.g. consistently use underscores rather than regular hyphens). The downside is that you'll have to know to update the list in `runtimeData.ts` if you want to make use of a new flag.


lloan (approved):
This will be very useful. LGTM!

Wonder if we should add a step to our checklist at some point, regarding flags. Making sure we add/remove. Or I wonder what a good approach to making sure we keep our list clean and up to date is.

PR #2839, opened 2022-11-21
mypy 0.991 update and related changes

jwhitlock:
Update from mypy 0.982 to 0.991 (opened in PR #2822), along with the changes needed to check types without errors or warnings:

* Update djangorestframework-stubs from 1.7.0 to 1.8.0, so that `api/permissions.py` will not error with `Metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases [misc]`
* Switch to downloading FxA verification keys on first use, to speed up application startup and fully type `privaterelay.apps.PrivateRelayConfig`, avoiding warning `By default the bodies of untyped functions are not checked, consider using --check-untyped-defs`
* Add types to `emails.tests.cleaners_tests.setup_server_storage_test_data`, to avoid the same warning
* Convert the helper function `api.tests.views_tests.get_user` into two fixtures `free_user` and `premium_user`, to avoid error `"ClientHandler" has no attribute "_force_user"`.

### DIFF
--- a/api/tests/views_tests.py
+++ b/api/tests/views_tests.py
@@ -12,31 +12,33 @@ from emails.tests.models_tests import make_free_test_user, make_premium_test_use
 
 
 @pytest.fixture
-def free_api_client(db) -> APIClient:
+def free_user(db) -> User:
+    return make_free_test_user()
+
+
+@pytest.fixture
+def free_api_client(free_user: User) -> APIClient:
     """Return an APIClient for a newly created free user."""
-    free_user = make_free_test_user()
     client = APIClient()
     client.force_authenticate(user=free_user)
     return client
 
 
 @pytest.fixture
-def prem_api_client(db) -> APIClient:
-    """Return an APIClient for a newly created premium user."""
+def premium_user(db) -> User:
     premium_user = make_premium_test_user()
     premium_profile = premium_user.profile
     premium_profile.subdomain = "premium"
     premium_profile.save()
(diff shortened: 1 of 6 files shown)


api/tests/views_tests.py
```diff
@@ -12,33 +12,35 @@
 
 
 @pytest.fixture
-def free_api_client(db) -> APIClient:
+def free_user(db) -> User:
+    return make_free_test_user()
+
+
+@pytest.fixture
+def free_api_client(free_user: User) -> APIClient:
```
groovecoder, line 20: question (non-blocking): so when multiple pytest fixtures are passed as parameters in test functions, the later fixtures receive the preceding fixture's `return` value as an argument?

pyproject.toml
```diff
@@ -4,9 +4,18 @@ mypy_path = "$MYPY_CONFIG_FILE_DIR/mypy_stubs"
 exclude = "env"
 python_version = "3.9"
 show_error_codes = true
-warn_return_any = true
-warn_unused_configs = true
-warn_unused_ignores = true
+strict = true
```
groovecoder, line 7: so strict

groovecoder (approved):

api/tests/views_tests.py
jwhitlock: yes, or `yeild whatever` if you use that to allow post-test teardown

PR #2842, opened 2022-11-23
Update Credential Manager integration date

Vinnl:
The Credential Manager integration will miss a train, so the announcement should go live the release after. This fixes MPP-2596.


lloan (approved):
Straightforward change - LGTM. Thanks!

PR #2844, opened 2022-11-23
Update dismissal status when the key changes

Vinnl:
This is useful when e.g. the profile API is still loading, and we're using a profile ID as part of the dismissal key. (That was specifically what @codemist was running into.)

How to test: merge this into `mpp-2417-phone-splash`, visit the phone onboarding as described in #2806, then select "Go to dashboard". Without this PR, a refresh would land you back at the onboarding. _With_ this PR, it should remember that you've seen it before.

- [ ] I've added a unit test to test for potential regressions of this bug. - Simulating async loading is a bit of a pain, but I could certainly add a test for this. Maybe I'll look into that tomorrow.


codemist (approved):

codemist:
Tested and works.

Vinnl:
Well, I decided to add a test after all, and then to cover the rest of the hook as well, and I'm happy I did, because I found a bug: f2c4984a

(The bug was that the `useEffect` would run every time if no options were provided, because then `options` would get the default value — a new empty object. And since every new empty object is a different instance, the effect would re-run. I fixed it by making it look at the `duration` directly, which is a `number` and therefore has referential equality.)

lloan (approved):
New tests! Thanks! LGTM.

PR #2866, opened 2022-11-29
MPP-2609 Ignore updating phone stats if not yet billing cycle

say-yawn:
Expecting that this job will run once everyday, there was a bug where when the `update_phone_remaining_stats` command was ran a user whose phone subscription has not renewed yet will also get their phone stats reset and will use the date the stats renewed to checked for future renewals erroneously. To fix that bug, this PR checks that if the user is being updated for the first time (`date_phone_subscription_checked==None`) we skip the user from updating phone stats if their billing cycle **has not** past.

How to test:
## Before this change, check that phone stats were updated even though it has not reached billing cycle
1. Create a user with phone subscription that started today and use some minutes and/or texts
2. Run `./manage.py update_phone_remaining_stats`
3. Check that the minutes and texts are updated. **Expected Results**: Phone stats are not renewed because the subscription has not reached billing cycle

## After this change, check that phone stats **WERE NOT** updated since the billing cycle was not reached
1. Create a user with phone subscription that started today and use some minutes and/or texts
2. Run `./manage.py update_phone_remaining_stats`
3. Phone stats did not change

### DIFF
--- a/privaterelay/management/commands/update_phone_remaining_stats.py
+++ b/privaterelay/management/commands/update_phone_remaining_stats.py
@@ -23,6 +23,13 @@ def update_phone_remaining_stats():
     updated_profiles = []
     for social_account in accounts_with_phones:
         profile = social_account.user.profile
+        if not profile.date_phone_subscription_checked:
+            first_billing_date = profile.date_subscribed_phone + timedelta(
+                settings.DAYS_PER_BILLING_CYCLE
+            )
+            if datetime.now(timezone.utc) < first_billing_date:
+                # Not yet billing cycle, ignore updating phone stats
+                continue
         # Has it been more than 30 days since we last checked the user's phone
         # subscription?
         if (
--- a/privaterelay/tests/mgmt_update_phone_remaining_stats_tests.py
+++ b/privaterelay/tests/mgmt_update_phone_remaining_stats_tests.py
@@ -35,7 +35,20 @@ def test_no_accounts_with_phones(db):
     update_phone_remaining_stats()
 
 
-def test_one_account_with_phones_checked_1_day_ago(phone_user):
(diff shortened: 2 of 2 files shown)


privaterelay/tests/mgmt_update_phone_remaining_stats_tests.py
```diff
@@ -35,7 +35,20 @@ def test_no_accounts_with_phones(db):
     update_phone_remaining_stats()
 
 
-def test_one_account_with_phones_checked_1_day_ago(phone_user):
+def test_one_account_with_phones_subscribed_3_day_ago_does_not_update_stats(phone_user):
```
jwhitlock, line 38: Thanks for the test!

privaterelay/management/commands/update_phone_remaining_stats.py
```diff
@@ -23,6 +23,13 @@ def update_phone_remaining_stats():
     updated_profiles = []
     for social_account in accounts_with_phones:
         profile = social_account.user.profile
+        if not profile.date_phone_subscription_checked:
+            first_billing_date = profile.date_subscribed_phone + timedelta(
```
jwhitlock, line 27: `profile.date_subscribed_phone` is nullable, and may be unset for some users since we added it in the middle of the phones feature. I think we skip this path if it is null, something like:

```python
if (not profile.date_phone_subscription_checked) and (profile.date_subscribed_phone is not null):
```

jwhitlock (changes requested):
Thanks @say-yawn! I think one change is needed if we want to merge this. There's very little chance that `profile.date_subscribed_phone` will be null going forward, but I also don't want to break the update script for one weird row.

groovecoder:
This PR is long in the tooth. @say-yawn has agreed to re-open this when it's ready for 👀 .

PR #2884, opened 2022-12-07
Api auth refactoring

groovecoder:
This PR supports #MPP-2405. The Firefox credential manager is receiving `403` errors from the API endpoints, but without reliable steps to reproduce. This change refactors the API `FxaTokenAuthentication` to add more detailed logging of the various possible error scenarios. This log data will help us determine if the `403` errors have a common underlying cause and how to fix it.

### First, you need to get a "Relay-scoped" stage FXA access token
Firefox [will](https://phabricator.services.mozilla.com/D158155) requests "Relay-[scoped](https://github.com/mozilla/fxa/blob/main/packages/fxa-auth-server/docs/oauth/scopes.md#url-scopes)" access tokens from FXA. (Our Relay scope URL is `https://identity.mozilla.com/apps/relay`.) It's not easy, but this is honestly maybe the easiest way to get such a token:
1. Clone [the add-on repo](https://github.com/mozilla/fx-private-relay-add-on/) locally
2. Change it to the `new-auth-mpp-1531` branch
3. `npm instsall` dependencies
4. `npm run dev` to run Firefox with the Relay add-on installed
5. Go to about:debugging#/runtime/this-firefox
6. Click "Inspect" next to the Firefox Relay add-on to open Developer Tools for it
7. Open the "Network" panel of the Developer Tools
8. Back in the `moz-extension://...` tab, click "Sign Up / In" to go thru the FXA stage auth flow
9. In the "Network" panel, look for the request to http://127.0.0.1:8000/api/v1/relayaddresses/
10. The request headers should include `authorization: Bearer ...`

### Use the API endpoints with the access token 
```
curl -H "Content-Type: application/json" -H "Authorization: Bearer abc123" http://127.0.0.1:8000/api/v1/relayaddresses/
curl -H "Content-Type: application/json" -H "Authorization: Bearer abc123" http://127.0.0.1:8000/api/v1/profiles/
etc.
```

### DIFF
--- a/api/authentication.py
+++ b/api/authentication.py
@@ -10,70 +10,93 @@ from django.core.cache import cache
 
 from allauth.socialaccount.models import SocialAccount
 from rest_framework.authentication import BaseAuthentication, get_authorization_header
+from rest_framework.exceptions import AuthenticationFailed
 
 
 logger = logging.getLogger("events")
 
 
 def get_cache_key(token):
+    # TODO: token can be a JWT which makes for a really big cache key
     return f"fxa_token_{token}"
+
+
+def introspect_token(introspect_token_url, token, cache_key, cache_timeout):
+    try:
+        fxa_resp = requests.post(introspect_token_url, json={"token": token})
+    except:
+        logger.error(
+            "Could not introspect token with FXA.",
+            extra={"fxa_response": shlex.quote(fxa_resp.text)},
+        )
+        # cache empty response data to prevent FXA failures from causing runaway
+        # HTTP retries
+        fxa_resp_data = {"status_code": None, "json": {}}
+        cache.set(cache_key, fxa_resp_data, cache_timeout)
+        raise AuthenticationFailed("Could not introspect token with FXA.")
+
+    fxa_resp_data = {"status_code": fxa_resp.status_code, "json": {}}
+    try:
+        fxa_resp_data["json"] = fxa_resp.json()
+    except requests.exceptions.JSONDecodeError:
+        logger.error(
+            "JSONDecodeError from FXA introspect response.",
+            extra={"fxa_response": shlex.quote(fxa_resp.text)},
+        )
+        cache.set(cache_key, fxa_resp_data, cache_timeout)
+        raise AuthenticationFailed("JSONDecodeError from FXA introspect response")
+    return fxa_resp_data
 
 
 class FxaTokenAuthentication(BaseAuthentication):
     def authenticate(self, request):
         authorization = get_authorization_header(request).decode()
         if not authorization or not authorization.startswith("Bearer "):
+            # If the request has no Bearer token, return None to attempt the next
+            # auth scheme in the REST_FRAMEWORK AUTHENTICATION_CLASSES list
             return None
 
         token = authorization.split(" ")[1]
         cache_key = get_cache_key(token)
         cached_fxa_resp_data = fxa_resp_data = cache.get(cache_key)
         # set a default cache_timeout, but this will be overriden to match
-        # the 'exp' time returned by FXA
+        # the 'exp' time in the JWT returned by FXA
         cache_timeout = 60
         if not fxa_resp_data:
             introspect_token_url = (
                 "%s/introspect"
                 % settings.SOCIALACCOUNT_PROVIDERS["fxa"]["OAUTH_ENDPOINT"]
             )
-            fxa_resp = requests.post(introspect_token_url, json={"token": token})
-            fxa_resp_data = {"status_code": fxa_resp.status_code, "json": None}
-            try:
-                fxa_resp_data["json"] = fxa_resp.json()
-            except requests.exceptions.JSONDecodeError:
-                logger.error(
-                    "JSONDecodeError from FXA introspect response.",
-                    extra={"fxa_response": shlex.quote(fxa_resp.text)},
-                )
+            fxa_resp_data = introspect_token(
+                introspect_token_url, token, cache_key, cache_timeout
+            )
 
         user = None
-        if (
-            fxa_resp_data
-            and fxa_resp_data["status_code"] == 200
-            and fxa_resp_data["json"]
-            and fxa_resp_data["json"].get("active")
-        ):
+        if not fxa_resp_data["status_code"] == 200:
+            raise AuthenticationFailed("Did not receive a 200 response from FXA.")
 
-            # FxA user is active, check for the associated Relay account
-            fxa_uid = fxa_resp_data.get("json").get("sub")
-            if fxa_uid:
-                try:
-                    sa = SocialAccount.objects.get(uid=fxa_uid, provider="fxa")
-                except SocialAccount.DoesNotExist:
-                    # No Relay account associated with the FxA ID. It might be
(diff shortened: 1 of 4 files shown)


api/authentication.py
```diff
@@ -10,77 +10,101 @@
...
-                    fxa_token_exp_time = int(
-                        fxa_resp_data.get("json").get("exp") / 1000
-                    )
-                    now_time = int(datetime.now(timezone.utc).timestamp())
-                    cache_timeout = fxa_token_exp_time - now_time
+        if not fxa_resp_data["status_code"] == 200:
+            raise AuthenticationFailed("Did not receive a 200 response from FXA.")
+
+        if not fxa_resp_data["json"].get("active"):
+            raise AuthenticationFailed("FXA returned active: False for token.")
```
jwhitlock, line 80: I was able to get a good response using the test directions, and to get this error by removing a letter from the bearer token. The response content is:

```json
{"detail":"FXA returned active: False for token."}
```

api/authentication.py
```diff
@@ -10,77 +10,101 @@
...
+        fxa_uid = fxa_resp_data.get("json").get("sub")
+        if not fxa_uid:
+            raise AuthenticationFailed(
+                "Authenticated user does not have a Relay account."
+            )
+        try:
+            sa = SocialAccount.objects.get(uid=fxa_uid, provider="fxa")
+        except SocialAccount.DoesNotExist:
+            raise AuthenticationFailed(
+                "Authenticated user does not have a Relay account."
```
jwhitlock, line 92: Nit: should this message be different than the one if there is no `fxa_uid`?

api/authentication.py
```diff
@@ -10,77 +10,101 @@
...
+    cache.set(cache_key, fxa_resp_data, cache_timeout)
+    return fxa_resp_data
+
+
 class FxaTokenAuthentication(BaseAuthentication):
     def authenticate(self, request):
         authorization = get_authorization_header(request).decode()
         if not authorization or not authorization.startswith("Bearer "):
+            # If the request has no Bearer token, return None to attempt the next
+            # auth scheme in the REST_FRAMEWORK AUTHENTICATION_CLASSES list
```
jwhitlock, line 57: Nice comment, this is not obvious!

api/authentication.py
```diff
@@ -10,77 +10,101 @@
...
+        fxa_resp = requests.post(introspect_token_url, json={"token": token})
+    except:
+        logger.error(
+            "Could not introspect token with FXA.",
+            extra={"fxa_response": shlex.quote(fxa_resp.text)},
+        )
+        # cache empty response data to prevent FXA failures from causing runaway
+        # HTTP retries
+        fxa_resp_data = {"status_code": None, "json": {}}
+        cache.set(cache_key, fxa_resp_data, cache_timeout)
```
jwhitlock, line 35: Nit: This code is repeated three times, since it is trying to ensure the cache is always, which is really a concern of the calling function. Here's what it might look like in ``FxaTokenAuthenication.authenticate``:

```python
            fxa_resp_data = None
            try:
                fxa_resp_data = introspect_token(introspect_token_url, token)
            except AuthenticationFailed:
                # cache empty response data to prevent FXA failures from causing runaway
                # HTTP retries
                fxa_resp_data = {"status_code": None, "json": {}}
                raise
            finally:
                cache.set(cache_key, fxa_resp_data, cache_timeout)

File: api/tests/authentication_tests.py [hunk 5]
```diff
@@ -31,10 +33,11 @@ def test_no_bearer_in_authorization_returns_none(self):
     def test_no_token_returns_none(self):
         headers = {"HTTP_AUTHORIZATION": "Bearer "}
         get_addresses_req = self.factory.get(self.path, **headers)
-        assert self.auth.authenticate(self.auth, get_addresses_req) == None
+        with pytest.raises(AuthenticationFailed):
```
jwhitlock, line 36: 👍 using `with pytest.raises`

jwhitlock (approved):
Thanks @groovecoder, this looks good, and the test steps worked. I'm a little worried about giving the user information about auth failures in general, but the messages are fairly generic and we can easily convert this to log data later.

docs/api_auth.md
```diff
@@ -1,12 +1,15 @@
 # Relay API Authentication
+
 The Relay API is built on [Django REST Framework][drf] and authenticates
 requests with any of 3 methods:
-* [FXA OAuth Token Authentication](fxa-oauth-token-authentication)
-* [`SessionAuthentication`][SessionAuthentication]
-* [`TokenAuthentication`][TokenAuthentication]
+
+- [FXA OAuth Token Authentication](fxa-oauth-token-authentication)
```
jwhitlock, line 6: Maybe this link should be:

```suggestion
- [FXA OAuth Token Authentication](#fxa-oauth-token-authentication)
```

api/authentication.py
groovecoder: Updated the previous one to "FXA did not return an FXA UID."

api/authentication.py
groovecoder: Changed the calling function to include `finally: cache.set` thanks!

api/tests/authentication_tests.py
```diff
@@ -64,49 +68,55 @@ def test_non_200_non_json_resp_from_fxa_returns_none_and_caches(self):
...
         headers = {"HTTP_AUTHORIZATION": f"Bearer {not_found_token}"}
         get_addresses_req = self.factory.get(self.path, **headers)
-        auth_return = self.auth.authenticate(self.auth, get_addresses_req)
+        with pytest.raises(AuthenticationFailed):
+            self.auth.authenticate(self.auth, get_addresses_req)
 
-        assert auth_return == None
         assert responses.assert_call_count(self.fxa_verify_path, 1) is True
-        expected = {"status_code": 503, "json": None}
+        expected = {"status_code": None, "json": {}}
```
jwhitlock, line 75: I haven't thought about this aspect of the change. I _think_ it is OK, we get the same result of avoiding another call to FxA for 60 seconds.

api/authentication.py
```diff
@@ -10,77 +9,99 @@
...
-                    user = sa.user
-
-                    # cache fxa_resp_data for as long as access_token is valid
-                    # Note: FXA iat and exp are timestamps in *milliseconds*
-                    fxa_token_exp_time = int(
-                        fxa_resp_data.get("json").get("exp") / 1000
-                    )
-                    now_time = int(datetime.now(timezone.utc).timestamp())
-                    cache_timeout = fxa_token_exp_time - now_time
+        if not fxa_resp_data["status_code"] == 200:
```
jwhitlock, line 75: What if we add this before the 200 check?

```python
if fxa_resp_data["status_code"] is None:
   raise AuthenticationFailed("Previous FXA call failed, wait to retry")
```

That way we'll get a slightly different response for a cached FXA failure.

Also, I always type it `FxA` not `FXA`. What will Mozilla Accounts be?

jwhitlock (approved):
Thanks @groovecoder, I appreciate the changes! The test steps still work for me.

One more suggestion, but I think FxA server failures _should_ be rare enough that it doesn't matter

jwhitlock (approved):

PR #2898, opened 2022-12-12
Postpone the credential manager integration date

Vinnl:
It is now planned to go out with Firefox 110.


maxxcrawford (approved):
Confirmed for [110 date](https://whattrainisitnow.com/release/?version=110).

PR #2902, opened 2022-12-13
Lint backend pending translations

jwhitlock:
Expand the GitHub action to also lint the backend pending translations file.

### DIFF
--- a/.github/workflows/lint-pending-strings.yml
+++ b/.github/workflows/lint-pending-strings.yml
@@ -1,7 +1,10 @@
-name: Lint Reference Files
+name: Lint Pending Translations
 on:
   push:
-    paths: ['frontend/pendingTranslations.ftl', '.github/workflows/lint-pending-strings.yml'  ]
+    paths:
+      - 'frontend/pendingTranslations.ftl'
+      - '.github/workflows/lint-pending-strings.yml'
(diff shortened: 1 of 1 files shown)


jwhitlock:
The linter ran on both the frontend and backend translation file.

.github/workflows/lint-pending-strings.yml
```diff
@@ -20,6 +23,9 @@ jobs:
       - name: Install Python dependencies
         run: |
           pip install -r privaterelay/locales/.github/requirements.txt
-      - name: Lint reference
+      - name: Lint frontend pending translations
         run: |
           moz-fluent-lint frontend --config privaterelay/locales/.github/linter_config.yml
+      - name: Lint backend pending translations
+        run: |
+          moz-fluent-lint privaterelay/pending_locales
```
Vinnl, line 31: Shouldn't we add this to make sure we're adhering to the same rules as applied in the l10n repository? Or was there a reason you left it out?

```suggestion
          moz-fluent-lint privaterelay/pending_locales --config privaterelay/locales/.github/linter_config.yml
```

Vinnl (approved):

.github/workflows/lint-pending-strings.yml
jwhitlock: Good catch!

PR #2905, opened 2022-12-15
Preserve query params on dashboard redirect

Vinnl:
This PR fixes MPP-2636. When the user is linked to relay.firefox.com from somewhere else with query parameters are appended, and the user is logged in, they will be redirected to `/account/profile/`. However, that means the query parameters will not be preserved. This is problematic for e.g. `utm_` parameters, which are processed by client-side JavaScript and hence only read after the redirect has completed.

With this PR, the redirect simply appends the query parameters for the original URL to the new URL.

How to test: be logged in, navigate to the root of the website with query parameters (e.g. `?test`) appended. They should be appended to `/account/profile/` (e.g. `/account/profile/?test`).

I didn't see any other middleware tests, so I didn't add a test for this. With some pointers I could give that a shot.

I also don't think there's a security risk here, but it's probably good if someone else gave that a bit of thought as well.

### DIFF
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -56,7 +56,12 @@ class RedirectRootIfLoggedIn:
         # in, use a server-side redirect to send them to the dashboard,
         # rather than handling that on the client-side:
         if request.path == "/" and settings.SESSION_COOKIE_NAME in request.COOKIES:
-            return redirect("accounts/profile/")
+            query_string = (
+                "?" + request.META["QUERY_STRING"]
+                if request.META["QUERY_STRING"]
+                else ""
+            )
+            return redirect("accounts/profile/" + query_string)
 
         response = self.get_response(request)
         return response


privaterelay/middleware.py
```diff
@@ -56,7 +56,12 @@ def __call__(self, request):
         # in, use a server-side redirect to send them to the dashboard,
         # rather than handling that on the client-side:
         if request.path == "/" and settings.SESSION_COOKIE_NAME in request.COOKIES:
-            return redirect("accounts/profile/")
+            query_string = (
+                "?" + request.META["QUERY_STRING"]
+                if request.META["QUERY_STRING"]
+                else ""
+            )
```
jwhitlock, line 63: I was considering using [GET.urlencode](using [https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.QueryDict.urlencode]) instead:

```suggestion
            query_string = "?" + request.GET.urlencode() if request.GET else ""
```

However, I think there is no difference in performance or security. If we were also manipulating the querystring, I'd want to use the Python functions that work with the querystring, but we're not so `META["QUERY_STRING"]` is perfect.

jwhitlock (approved):
Thanks @Vinnl! I think this is a great feature.

PR #2907, opened 2022-12-16
Prevent content from overflowing on small screens

Vinnl:
flex-shrink was set to 0, so it wouldn't shrink when there wasn't enough space.

And since it can now shrink, I had to set flex-shrink to 0 for the icon to prevent *that* from shrinking when there wasn't enough space - the title text next to it is better suited for shrinking.

This PR fixes MPP-2644.

How to test: ensure your viewport is thing, then open a page with a banner (e.g. the settings page with data collection disabled). The contents of the banner should stay in the banner.

Before:

image

After:

image


lloan (approved):
Looks good, approved - only question is what is the smallest device width we support?

Vinnl:
That is a good question, and I don't think we have an answer to that :sweat_smile: But apparently this is a device QA tests on, so it's a good baseline, and other than that just best effort, I suppose.

PR #2925, opened 2022-12-19
Use correct l10n checkout when testing migrations

Vinnl:
When switching to the version of the code that's currently in production, make sure to also switch to the version of the l10n repository that's currently in production. Otherwise, if strings have been removed from the l10n repository that are still being referenced in production, the tests would start failing despite nothing being broken.


jwhitlock:
✅ Thanks @Vinnl!

PR #2927, opened 2022-12-19
Prevent subdomain explainer overflowing on mobile

Vinnl:
This changes the popover into a dialog on small screens, so it can overlap the indicator.

This PR fixes MPP-2648.

How to test: log in with a user with a custom subdomain set. On a small screen, position the viewport such that the button labelled with the chosen subdomain is in the middle of the screen, then click it. Before this change, there would not be enough room for the popover above the button, so its content would break out of it:

image

After this change, it is instead a dialog, and overlaps the button:

image

On wider screens, the behaviour should be unchanged.


lloan (approved):
Looks great! Tested and works as intended: 

image

PR #2928, opened 2022-12-19
Prevent mask cards from horizontall overflowing

Vinnl:
Flexbox items by default can't shrink beyond its content sizes, but since we're ellipsizing (is that a word?) the mask address anyway, that is fine here. Hence, we override the default min-width: 0. Also see https://stackoverflow.com/a/38383437

This PR fixes MPP-2645. Or at least, I think so, since it seems to be related to long mask names. It fixes a long-standing annoying issue anyway.

How to test: create a mask with a long name, then resize the viewport. The mask card should no longer overflow. In other words, this:

image

instead of this:

image

----

I also noticed there was a vertical overflow if the contents of a card was larger than the viewport:

image

So I also fixed that in 8994889e:

image


lloan (approved):
LGTM!

PR #2931, opened 2022-12-20
Mobile fixes for landing page

Vinnl:
This PR fixes two issues mentioned in MPP-2660 for very thin screens (iPhone 12 pro):

- The landing page has horizontal scrolling because the pricing cards don't fit (f2998f567eed7ddee6e9947b81e334e0c4e6767d)
- The frame of the phone with the website demo wasn't visible (d003d421df3814179711dbddbc1aede8578097b6)

Note that the latter was because both the frame and the page background were white. I took some (bad) design liberty and just added a grey backdrop there for now:

image

How to test: be logged out, resize your viewport to the iPhone 12 Pro's size, then check for horizontal scrolling and that the phone frame is visible.

### DIFF
--- a/frontend/src/pages/premium.module.scss
+++ b/frontend/src/pages/premium.module.scss
@@ -8,7 +8,7 @@
   width: $content-max;
   max-width: 100%;
   margin: 0 auto;
-  padding: $spacing-2xl;
+  padding: $spacing-lg;
 
   @media screen and #{$mq-md} {
     flex-direction: row;
--- a/frontend/src/components/landing/DemoPhone.module.scss
+++ b/frontend/src/components/landing/DemoPhone.module.scss
@@ -2,8 +2,12 @@
 
 .container {
   position: relative;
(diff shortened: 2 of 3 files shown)


frontend/src/pages/premium.module.scss
```diff
@@ -8,7 +8,7 @@
   width: $content-max;
   max-width: 100%;
   margin: 0 auto;
-  padding: $spacing-2xl;
+  padding: $spacing-lg;
```
Vinnl, line 11: Note: this change just brings it in line with `index.page.ts`, where a similar change was applied. Without it, the margin around the phone image on small screens is really big, which now means that there's a big layer of white padding surrounding a relatively thin grey padding around the phone.

maxxcrawford (approved):
Works great! Reviewed on Xcode Simulator for iPhone SE (Smallest phone). 

Screen Shot 2022-12-20 at 9 56 39 AM

PR #2932, opened 2022-12-20
Mark localtunnel URLs as being local

Vinnl:
This makes the website accept cross-origin requests from http://localhost:3000, aka `npm run dev` for the frontend.


jwhitlock:
I've had to do similar gymastics to get `SENTRY_ENVIRONMENT` to have a good default on local:

https://github.com/mozilla/fx-private-relay/blob/b560cc30b095f9592a5062279aaf400c625fc5d1/privaterelay/settings.py#L899-L902

I'm using `ngrok.io`, not `.loca.lt`.

A couple of other options:

Use `local` for the default:

```python
RELAY_CHANNEL = ORIGIN_CHANNEL_MAP.get(SITE_ORIGIN, "local")
```

Allow an `.env` setting for `RELAY_CHANNEL`:

```python
_DEFAULT_RELAY_CHANNEL = ORIGIN_CHANNEL_MAP.get(SITE_ORIGIN, "prod")
RELAY_CHANNEL = config("RELAY_CHANNEL", default=_DEFAULT_RELAY_CHANNEL)
```

Vinnl:
@jwhitlock Ah, so with a change like this that could be removed, I suppose?

Which option would you prefer? I think I'm partial towards using `local` for the default, but don't have a strong preference.

jwhitlock:
I like `local` as default as well, but no strong preference. Let @groovecoder decide?

jwhitlock (changes requested):
I've slept on it, and I'm more sure. Let's combine the two:

```python
 _DEFAULT_RELAY_CHANNEL = ORIGIN_CHANNEL_MAP.get(SITE_ORIGIN, "local")
RELAY_CHANNEL = config("RELAY_CHANNEL", default="_DEFAULT_RELAY_CHANNEL")
```
Production, stage, and dev will work the same way. Local development with `.loca.it` and `ngrok.io` will default to `local`. If someone wants to force a different config, they can set `RELAY_CHANNEL` in the environment or `.env`.

Vinnl:
@jwhitlock Done. Only thing I wasn't sure about is why you needed `"` around `_DEFAULT_RELAY_CHANNEL` the second time around — it looks like a mistake, but maybe there's just some Python I'm not aware of?

jwhitlock:
> Only thing I wasn't sure about is why you needed `"` around `_DEFAULT_RELAY_CHANNEL` the second time around — it looks like a mistake, but maybe there's just some Python I'm not aware of?

You are correct, it was a mistake.

jwhitlock (approved):
Thanks @Vinnl, looks good to me!

PR #2968, opened 2023-01-03
Update custom dimension data object names to match Google documentation

maxxcrawford:


privaterelay/views.py
```diff
@@ -143,8 +143,8 @@ def metrics_event(request):
         request_data.get("action", None),
         request_data.get("label", None),
         request_data.get("value", None),
-        dimension5=request_data.get("dimension5", None),
-        dimension7=request_data.get("dimension7", "website"),
+        cd5=request_data.get("dimension5", None),
+        cd7=request_data.get("dimension7", "website"),
```
maxxcrawford, line 147: Note: Based on this [documentation](https://developers.google.com/analytics/devguides/collection/gtagjs/custom-dims-mets), the previous setting is the correct way? 
Screen Shot 2023-01-03 at 2 48 12 PM


We may need to wrap both items in a nested object: 
Screen Shot 2023-01-03 at 2 48 44 PM

jwhitlock: The library [google-measurement-protocol](https://github.com/mirumee/google-measurement-protocol) is using the [measurement protocol (universal analytics)](https://developers.google.com/analytics/devguides/collection/protocol/v1), so the documentation to use is [Measurement Protocol Parameters](https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#customs). Yes, `cd5` is custom dimension 5, I believe the change in this PR is correct.

The library is using `https://ssl.google-analytics.com/collect` as their base URL (see [the code](https://github.com/mirumee/google-measurement-protocol/blob/f7a0749ab76fc397904583f4768eba96f983073e/google_measurement_protocol/report.py#L5)), but the Google docs say to use `https://www.google-analytics.com/collect`. I wonder if the library is out of date? It looks like a fairly thin wrapper around `requests`, a Python HTTP request library.

This user reports a change of user agent is needed to get to the GA server: https://github.com/mirumee/google-measurement-protocol/issues/22. However, the docs imply that is only needed to set the browser, platform, and mobile capabilities.

maxxcrawford:
Closing as this is not necessary. Having a custom field may still be helpful in the future but updating the label name is not needed.

maxxcrawford:
https://mozilla-hub.atlassian.net/browse/MPP-2674

PR #2972, opened 2023-01-04
Allow stats area to shrink if not all are visible

Vinnl:
I'm not quite sure why I thought the flex-basis of $content-xs was necessarily initially, although I do remember struggling with a number of different values, so I might just have gotten them mixed up in 39602f04526083c12c7da4995c6377892d4dfd64.

However, with this change, both Chinese and English still have the expected (lack of) line breaks, but when not all four mask-related statistics are shown, the stats area doesn't take up more space than necessary.

Specifically, instead of

image

it now looks like

image

How to test: open the dashboard as a free user (i.e. without promo-blocking available). There should be no superfluous whitespace around the mask stats on desktop.

### DIFF
--- a/frontend/src/components/dashboard/aliases/Alias.module.scss
+++ b/frontend/src/components/dashboard/aliases/Alias.module.scss
@@ -214,7 +214,7 @@ $trackerRemovalIndicatorWidth: 20px;
     // stylelint-disable-next-line no-duplicate-selectors
     .main-data & {
       display: flex;
-      flex: 1 0 $content-xs;
+      flex: 0 0 0;
       text-align: center;
     }
     // I've grouped these together under the media query to emphasise that they


maxxcrawford (approved):
🤝  LGTM

Screen Shot 2023-01-04 at 16 06 43

Vinnl:
Thanks for sharing that screenshot @maxxcrawford, cause that still didn't look right to me - too much room around the "Trackers removed" stat. The reason is that labels were disabled for you, with that element therefore not pushing the stats area aside enough to shrink. I've now set it to not grow and not shrink either, so it will simply be flush with the right-hand side: 8b36c81e

maxxcrawford (approved):
Hey! Glad I included the screenshot. 😅 

Looks good both ways now: 
Screen Shot 2023-01-05 at 9 07 09 AM

image

PR #2975, opened 2023-01-05
Preserve locally-stored mask-website associations

Vinnl:
When the user enables server-side label storage, we used to only upload the locally-stored labels, but website associations would be dropped.

Users won't notice immediately since the old associations are still stored in their local cache, but as soon as that gets cleared, the associations will disappear.

How to test: disable server-side storage, then use the add-on to create a mask on a website. If you then open the in-page menu on that website, the mask should be listed as "From this website". Now enable server-side storage, and delete the `relayAddresses` value in the extension storage. The association should be preserved with this fix, whereas it wasn't before.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, I don't think, as it'd just verify parameters to a mock call without any logic beforehand.


lloan (approved):
Awesome fix!

PR #2997, opened 2023-01-11
(WIP) MPP-2479 - Prevent modal from scrolling up on iOS Safari

codemist:
This PR fixes #MPP-2479. I have a couple questions for this one, which I added in a self-review.

How to test:

### DIFF
--- a/frontend/src/components/dashboard/aliases/AddressPickerModal.tsx
+++ b/frontend/src/components/dashboard/aliases/AddressPickerModal.tsx
@@ -15,7 +15,6 @@ import {
   useDialog,
   useModal,
   useOverlay,
-  usePreventScroll,
   useButton,
   AriaOverlayProps,
 } from "react-aria";
@@ -189,7 +188,6 @@ type PickerDialogProps = {
 const PickerDialog = (props: PickerDialogProps & AriaOverlayProps) => {
   const wrapperRef = useRef<HTMLDivElement>(null);
   const { overlayProps, underlayProps } = useOverlay(props, wrapperRef);
-  usePreventScroll();
   const { modalProps } = useModal();
   const { dialogProps, titleProps } = useDialog({}, wrapperRef);
 
--- a/frontend/src/styles/globals.scss
+++ b/frontend/src/styles/globals.scss
@@ -23,7 +23,6 @@ a {
   margin: 0;
 }
 
-html,
 body,
 #__next,
 #overlayProvider {
@@ -35,12 +34,23 @@ body,
   // empty space at the bottom of the page, below the footer.
   // Since the user can scroll in `.non-header-wrapper`, everything outside that
   // that still overflows these elements can safely be hidden:
-  overflow: hidden;
+  // overflow: hidden;
 }
 
 html {
   scroll-behavior: smooth;
   scroll-padding-top: $layout-xl;
+  font-family: $font-stack-base;
+  height: auto;
+
+  @media #{$mq-sm} {
+    // In <Layout> all content is wrapped in a `.non-header-wrapper` that has its
+    // own `overflow` applied, to allow the header and mobile menu to remain fixed
+    // in place. However, in Blink- and WebKit-based browsers, this resulted in
+    // empty space at the bottom of the page, below the footer.
+    // Hence, a height of 100% is applied to the html document
+    height: 100%;
+  }
 
   @media (prefers-reduced-motion) {
     & {


frontend/src/components/dashboard/aliases/AddressPickerModal.tsx
```diff
@@ -189,7 +188,6 @@ type PickerDialogProps = {
 const PickerDialog = (props: PickerDialogProps & AriaOverlayProps) => {
   const wrapperRef = useRef<HTMLDivElement>(null);
   const { overlayProps, underlayProps } = useOverlay(props, wrapperRef);
-  usePreventScroll();
```
codemist, line 192: @Vinnl So I discovered that the reason it was scrolling up in a strange fashion on Safari was due to this `usePreventScroll()` hook. I removed it and I didn't see a noticeable difference in UI when checked across the desktop and the simulators. Could you explain in detail what this is doing for the component or is it imperative to leave it in?

frontend/src/styles/globals.scss
```diff
@@ -23,7 +23,6 @@ a {
   margin: 0;
 }
 
-html,
```
codemist, line 26: I found that the `100%` height on the `html` element was causing a weird keyboard bug as seen here, where after I close the soft keyboard on iOS it wipes out the back of it due to the changed height that isn't rerendered.

https://user-images.githubusercontent.com/13066134/211921434-4d9b7d97-06b2-49ba-90f2-807a9f63c0a6.mov

Based on the documentation within this section, I think that was only applied to prevent whitespace from appearing in the body size is smaller than the document size.

frontend/src/styles/globals.scss
```diff
@@ -35,12 +34,23 @@ body,
...
+  height: auto;
+
+  @media #{$mq-sm} {
+    // In <Layout> all content is wrapped in a `.non-header-wrapper` that has its
+    // own `overflow` applied, to allow the header and mobile menu to remain fixed
+    // in place. However, in Blink- and WebKit-based browsers, this resulted in
+    // empty space at the bottom of the page, below the footer.
+    // Hence, a height of 100% is applied to the html document
+    height: 100%;
+  }
```
codemist, line 53: So I simply added a media query here to only apply the `100%` height to devices sizes `>480px`, since on mobile I don't think we need to worry about content being too short.


https://user-images.githubusercontent.com/13066134/211921665-52d1da39-b1f8-440a-8a65-e58e8d9fdad6.mov

codemist (commented):
@Vinnl I have some questions about the solution to this bug, I'm still a bit confused about the `usePreventScroll` bug in this context.

frontend/src/components/dashboard/aliases/AddressPickerModal.tsx
Vinnl: Hmm, interesting. Looking at [its docs](https://react-spectrum.adobe.com/react-aria/usePreventScroll.html), it seems to have been created specifically for `useDialog`, [whose documentation](https://react-spectrum.adobe.com/react-aria/useDialog.html#example) no longer mentions it. It's not even mentioned in the sidebar of that page, so I've [logged a bug upstream](https://github.com/adobe/react-spectrum/issues/3909), but I'm guessing that indeed it is no longer necessary?

The scrolling behaviour in your preview deployment when opening the custom mask dialog seems normal to me, so let's indeed remove it and keep an eye on QA reports related to scrolling/the custom mask dialog.

frontend/src/styles/globals.scss
Vinnl: Nice detective work! Yes, with the clear comment next to it, this seems like a fine workaround to me. Some of the [new CSS units like `dvh`](https://developer.mozilla.org/en-US/docs/Web/CSS/length#relative_length_units_based_on_viewport) might be able to provide some relief here as well, but given that I don't expect this to negatively impact maintainability, and that the result is that it works just fine for users, I don't think that's worth the time to investigate right now. Nice job!

frontend/src/styles/globals.scss
```diff
@@ -35,12 +34,23 @@ body,
...
   // Since the user can scroll in `.non-header-wrapper`, everything outside that
   // that still overflows these elements can safely be hidden:
-  overflow: hidden;
+  // overflow: hidden;
 }
 
 html {
   scroll-behavior: smooth;
   scroll-padding-top: $layout-xl;
+  font-family: $font-stack-base;
```
Vinnl, line 43: Was the font change related to this ticket? Applying this to _everything_ within `<html>` seems like potentially a fairly big change in the sense that I can't quite predict what that will affect?

frontend/src/styles/globals.scss
```diff
@@ -35,12 +34,23 @@ body,
   // empty space at the bottom of the page, below the footer.
   // Since the user can scroll in `.non-header-wrapper`, everything outside that
   // that still overflows these elements can safely be hidden:
-  overflow: hidden;
+  // overflow: hidden;
```
Vinnl, line 37: I'm not sure if it was necessary to remove this, but if it was, might as well delete the full line and the comment above it. (And ideally also an explanation of why it needed to be removed :) )

frontend/src/styles/globals.scss
```diff
@@ -35,12 +34,23 @@ body,
...
 }
 
 html {
   scroll-behavior: smooth;
   scroll-padding-top: $layout-xl;
+  font-family: $font-stack-base;
+  height: auto;
```
Vinnl, line 44: Hmm crap. Unfortunately it's been a while since I touched this, so I don't have top of mind why this is happening, but it appears that without `height: 100%`, the mobile menu is no longer sticky:

[Screencast from 2023-01-12 11-14-05.webm](https://user-images.githubusercontent.com/4251/212040074-fd3100d5-2255-4712-b7d1-27cf38fb638f.webm)

Vinnl (commented):
Crap, the workaround looked good to me, but then I did find a bug :/ I think we could at least extract out the removal of `usePreventScroll` into a separate PR to fix the weird scrolling in Safari, but the keyboard bug will need some extra work :cry: Do you know if it's only happening with the dialog, or anywhere where the keyboard pops up?

PR #2998, opened 2023-01-11
CSP: Allow fonts on stage, block objects

jwhitlock:
Fonts are unable to load on stage, because it used to have the default of `'self'`, but changed to `'https://relay.firefox.com/'` in https://github.com/mozilla/fx-private-relay/pull/2882, to support emails. This adds `'self'` as well.

I used https://csp-evaluator.withgoogle.com/, and it recommended a `object-src: 'none'` directive, so I'm adding that as well.

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -109,6 +109,7 @@ CSP_SCRIPT_SRC = [
     "https://www.google-analytics.com/",
 ]
 CSP_FONT_SRC = ["'self'", "https://relay.firefox.com/"]
+CSP_OBJECT_SRC = ("'none'",)
 if USE_SILK:
     CSP_SCRIPT_SRC.append("'unsafe-inline'")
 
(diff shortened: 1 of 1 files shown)


Vinnl (approved):
Oh nice, I had just ran into the same issue and [applied a similar quick fix](https://github.com/mozilla/fx-private-relay/commit/d72ca822dfb51fddcc92ae30d7d663c72ead5be7), still intending to figure out the proper fix and submitting a PR for that - nice job.

(I wonder if something changed recently though, since we both noticed around the same time?)

jwhitlock:
> (I wonder if something changed recently though, since we both noticed around the same time?)

https://github.com/mozilla/fx-private-relay/pull/2882 was included in https://github.com/mozilla/fx-private-relay/releases/tag/2023.01.11, which went out yesterday. I _think_ production would be OK, since relay.firefox.com is explicitly mentioned. We could tag again to fix stage, or wait until next week's tag

Vinnl:
Ah gotcha, it was just newly added. Nah let's wait, I didn't jump into action either because it wouldn't be a production issue.

PR #3013, opened 2023-01-16
Consistently allow missing trailing slashes API routes

Vinnl:
Like API routes added by api_router, it is still allowed to omit the trailing slash, but with this change, it's _also_ allowed to add a trailing slash, just like for all the api_router routes.

This works because by default, Django's APPEND_SLASH option is True (see [1]), so if a route without a trailing slash it found, it will append it, but not vice versa (i.e. a trailing slash will not be automatically stripped).

[1] https://docs.djangoproject.com/en/dev/ref/middleware/#django.middleware.common.CommonMiddleware

- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege. - N/A
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process. - N/A
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage. - N/A

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -38,7 +38,17 @@ api_router.register(r"flags", FlagViewSet, "flag")
 
 
 urlpatterns = [
+    path(
+        "v1/runtime_data",
+        runtime_data,
+        name="runtime_data_deprecate_after_updating_clients",
+    ),
     path("v1/runtime_data/", runtime_data, name="runtime_data"),
+    path(
+        "v1/report_webcompat_issue",
+        report_webcompat_issue,
+        name="report_webcompat_issue_deprecate_after_updating_clients",
+    ),
     path(
         "v1/report_webcompat_issue/",
         report_webcompat_issue,
@@ -75,11 +85,44 @@ if settings.PHONES_ENABLED:
     api_router.register(r"relaynumber", RelayNumberViewSet, "relay_number")
     api_router.register(r"inboundcontact", InboundContactViewSet, "inbound_contact")
     urlpatterns += [
+        # TODO: Update Twilio webhooks to versions with trailing slashes,
+        #       then remove versions without trailing slashes (Django's
+        #       APPEND_SLASH option will then make those redirect).
+        path(
+            "v1/inbound_sms",
+            inbound_sms,
+            name="inbound_sms_deprecate_after_updating_clients",
+        ),
         path("v1/inbound_sms/", inbound_sms, name="inbound_sms"),
+        path(
+            "v1/inbound_call",
+            inbound_call,
+            name="inbound_call_deprecate_after_updating_clients",
+        ),
         path("v1/inbound_call/", inbound_call, name="inbound_call"),
+        path(
+            "v1/voice_status",
+            voice_status,
+            name="voice_status_deprecate_after_updating_clients",
+        ),
         path("v1/voice_status/", voice_status, name="voice_status"),
+        path(
+            "v1/sms_status",
+            sms_status,
+            name="sms_status_deprecate_after_updating_clients",
+        ),
         path("v1/sms_status/", sms_status, name="sms_status"),
+        path(
+            "v1/vCard/<lookup_key>",
+            vCard,
+            name="vCard_deprecate_after_updating_clients",
+        ),
         path("v1/vCard/<lookup_key>/", vCard, name="vCard"),
+        path(
+            "v1/realphone/resend_welcome_sms",
+            resend_welcome_sms,
+            name="resend_welcome_sms_deprecate_after_updating_clients",
+        ),
         path(
             "v1/realphone/resend_welcome_sms/",
             resend_welcome_sms,
--- a/api/urls.py
+++ b/api/urls.py
@@ -38,14 +38,14 @@ api_router.register(r"flags", FlagViewSet, "flag")
 
 
 urlpatterns = [
-    path("v1/runtime_data", runtime_data, name="runtime_data"),
+    path("v1/runtime_data/", runtime_data, name="runtime_data"),
     path(
-        "v1/report_webcompat_issue",
+        "v1/report_webcompat_issue/",
         report_webcompat_issue,
         name="report_webcompat_issue",
     ),
     path(
-        "v1/swagger<swagger_format:format>",
+        "v1/swagger<swagger_format:format>/",
         enable_if_setting("API_DOCS_ENABLED")(schema_view.without_ui(cache_timeout=0)),
         name="schema-json",
     ),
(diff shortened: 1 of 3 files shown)


groovecoder (changes requested):
issue (blocking): If we change the URLs we need to change the Twilio webhook URLs to match, and we need to do it at the EXACT time as we deploy this change.

Let's keep the webhook urls as they are now, so we don't disrupt any webhook operations. Or, we need a nicer way to roll this out with 0 downtime.

Vinnl:
@groovecoder As mentioned in the description,

> it is still allowed to omit the trailing slash, but with this change, it's also allowed to add a trailing slash, just like for all the api_router routes.

So this doesn't _change_ route, it only _additionally_ allows routes with trailing slashes, just like Django's routes. (The tests had to be updated because they check the status codes, and the routes without trailing slash will now return a redirect to the ones with status codes.) So there should indeed be 0 downtime.

groovecoder:
Update: When I changed the tests to remove the `/` in urls, I got:
```
>       assert response.status_code == 400
E       assert 301 == 400
E        +  where 301 = <HttpResponsePermanentRedirect status_code=301, "text/html; charset=utf-8", url="/api/v1/inbound_call/">.status_code
```
I'm not sure if Twilio webhook client will follow those redirects. So we're going to test this on the dev server before we merge it.

jwhitlock:
When testing this locally, Twilio got a 500 Internal Server Error with a `POST /api/v1/inbound_sms`, with this error message on the debug HTML:

> You called this URL via POST, but the URL doesn&#x27;t end in a slash and you have APPEND_SLASH set. Django can&#x27;t redirect to the slash URL while maintaining POST data. Change your form to point to <my_subdomain>.ngrok.io/api/v1/inbound_sms/ (note the trailing slash), or set APPEND_SLASH=False in your Django settings.

I think this will fail on the dev server as well, but I'd be willing to try it.

api/urls.py
```diff
@@ -75,13 +75,13 @@ def to_url(self, value):
...
-        path("v1/inbound_sms", inbound_sms, name="inbound_sms"),
-        path("v1/inbound_call", inbound_call, name="inbound_call"),
-        path("v1/voice_status", voice_status, name="voice_status"),
-        path("v1/sms_status", sms_status, name="sms_status"),
-        path("v1/vCard/<lookup_key>", vCard, name="vCard"),
+        path("v1/inbound_sms/", inbound_sms, name="inbound_sms"),
+        path("v1/inbound_call/", inbound_call, name="inbound_call"),
+        path("v1/voice_status/", voice_status, name="voice_status"),
+        path("v1/sms_status/", sms_status, name="sms_status"),
+        path("v1/vCard/<lookup_key>/", vCard, name="vCard"),
```
jwhitlock, line 82: For POST URLs that we need to switch remotely as well, we could allow both URLs for a time:

```suggestion
        path("v1/inbound_sms", inbound_sms, name="legacy_inbound_sms"),
        path("v1/inbound_call", inbound_call, name="legacy_inbound_call"),
        path("v1/voice_status", voice_status, name="legacy_voice_status"),
        path("v1/sms_status", sms_status, name="legacy_sms_status"),
        path("v1/inbound_sms/", inbound_sms, name="inbound_sms"),
        path("v1/inbound_call/", inbound_call, name="inbound_call"),
        path("v1/voice_status/", voice_status, name="voice_status"),
        path("v1/sms_status/", sms_status, name="sms_status"),
        path("v1/vCard/<lookup_key>/", vCard, name="vCard"),
```

Vinnl: Hmm, I like the idea, implemented in [b9da2cf](https://github.com/mozilla/fx-private-relay/pull/3013/commits/b9da2cfc1d821bf33489b92b1a1dc2376da0648f).

Unfortunately, my texts still don't seem to go through on the dev server, and calls receive a robot voice saying something like "An application error has occurred".

Since this was just intended to be a quick fix to solve a minor annoyance, I might close this PR, since it's starting to get into not-worth-the-hassle territory.

Although I just tried it out with `main`, and got the same error, so maybe something else is wrong at the moment.

Vinnl: Never mind, there was a general error on dev, and calling and text did actually work with these changes!

api/urls.py
```diff
@@ -38,14 +38,14 @@ def to_url(self, value):
 
 
 urlpatterns = [
-    path("v1/runtime_data", runtime_data, name="runtime_data"),
+    path("v1/runtime_data/", runtime_data, name="runtime_data"),
     path(
-        "v1/report_webcompat_issue",
+        "v1/report_webcompat_issue/",
```
jwhitlock, line 43: I believe this is a `POST`, needs the legacy version as well

api/urls.py
```diff
@@ -75,13 +75,21 @@ def to_url(self, value):
     api_router.register(r"relaynumber", RelayNumberViewSet, "relay_number")
     api_router.register(r"inboundcontact", InboundContactViewSet, "inbound_contact")
     urlpatterns += [
+        # TODO: Update Twilio webhooks to versions with trailing slashes,
+        #       then remove versions without trailing slashes (Django's
+        #       APPEND_SLASH option will then make those redirect).
         path("v1/inbound_sms", inbound_sms, name="inbound_sms"),
+        path("v1/inbound_sms/", inbound_sms, name="inbound_sms"),
```
jwhitlock, line 82: Nit: Remove duplicate names

I was surprised that you can re-use the name `"inbound_sms"`, but I believe this is how you can override a URL handler from a third party library, so it makes sense that names can be reused, in which case the last one wins.

api/urls.py
```diff
@@ -75,13 +75,21 @@ def to_url(self, value):
...
+        path("v1/inbound_call/", inbound_call, name="inbound_call"),
         path("v1/voice_status", voice_status, name="voice_status"),
+        path("v1/voice_status/", voice_status, name="voice_status"),
         path("v1/sms_status", sms_status, name="sms_status"),
+        path("v1/sms_status/", sms_status, name="sms_status"),
         path("v1/vCard/<lookup_key>", vCard, name="vCard"),
+        path("v1/vCard/<lookup_key>/", vCard, name="vCard"),
         path(
-            "v1/realphone/resend_welcome_sms",
+            "v1/realphone/resend_welcome_sms/",
```
jwhitlock, line 92: `v1/realphone/resend_welcome_sms/` is a `POST`, so we may need the duplicate as well during the transition.

api/urls.py
```diff
@@ -75,13 +75,21 @@ def to_url(self, value):
...
         path("v1/inbound_sms", inbound_sms, name="inbound_sms"),
+        path("v1/inbound_sms/", inbound_sms, name="inbound_sms"),
         path("v1/inbound_call", inbound_call, name="inbound_call"),
+        path("v1/inbound_call/", inbound_call, name="inbound_call"),
         path("v1/voice_status", voice_status, name="voice_status"),
+        path("v1/voice_status/", voice_status, name="voice_status"),
         path("v1/sms_status", sms_status, name="sms_status"),
+        path("v1/sms_status/", sms_status, name="sms_status"),
         path("v1/vCard/<lookup_key>", vCard, name="vCard"),
+        path("v1/vCard/<lookup_key>/", vCard, name="vCard"),
```
jwhitlock, line 90: This one is a `GET`, so we might be able to remove the version without a trailing slash. Or, we could wait until after the next deploy.

jwhitlock (changes requested):
I think for safety we should have duplicates for any URL involved in a `POST`, until we can fully update the callers (Twilio, the add-on, etc.)

Vinnl:
Oops crap, I did intend to preserve all the existing endpoints, but got derailed by the dev issue :/ Good point about the names @jwhitlock, updated those too: 5bcfe392

The only one I did not preserve was the Swagger one, because I'm not sure where (if anywhere) that gets called? Any of our CI tools perhaps?

jwhitlock (approved):
Thanks @Vinnl! I think this is ready to merge.

I don't think the swagger endpoint is used. I thought _maybe_ `/api/v1/docs/` used it, but no.

groovecoder (approved):

Issue #2709, opened 2022-10-27
Handle 412 error, "Number Pool size limit reached", when registering with Twilio messaging service


jwhitlock:
Twilio responds with an error when attempting to register a new number when at the pool limit (default 400):

```
HTTP 412 error: Unable to create record: Number Pool size limit reached.
```

On the website, it looks like this:
1. A US-based premium user browses the available numbers
2. The user selects a number
2. They get a confirmation dialog "Please confirm that this is the phone number mask you want. This can’t be changed later."
3. They click the "Confirm" button
4. The website sends `POST /api/v1/relaynumber/` with the selected number is sent. After a delay it returns a `500 Server Error` with no further data
5. The website sends `GET /api/v1/relaynumber/`, which returns `200 OK` and `[]` (an empty JSON list)
6. The user sees the available numbers again, gets no message that the request fails, and is able to restart the process.

Canadian numbers are not registered with the messaging service, so Canadian-based premium users do not see this issue.

Sentry error: https://sentry.io/organizations/mozilla/issues/3699489374

jwhitlock:
Fixed by PR 3014

PR #3014, opened 2023-01-17
MPP-2728: Allow multiple Twilio messaging services

jwhitlock:
This PR fixes issue #2709:

* Add a test to replicate the issue
* Change `TWILIO_MESSAGING_SERVICE_SID` to a comma-separated value
* Log a 412 response / [21714 errors](https://www.twilio.com/docs/api/errors/21714) when assigning a number to a pool, and try the next SID. This will allow us to pre-allocate the next messaging pool and avoid user signup issues.
* Log 409 response / [21710 errors](https://www.twilio.com/docs/api/errors/21710) when assigning an already assigned number to a pool. This was suggested by @say-yawn in the review for PR #2614.
* Skip registration and log a warning if `TWILIO_MESSAGING_SERVICE_SID` is undefined. This is new behavior - previously, it would be a runtime exception. It minimizes changes to existing tests, which expect to pass with an undefined `TWILIO_MESSAGING_SERVICE_SID` and a mocked twilio client.

Additional changes:

* Address flake8 linting issues
* Rename `mocked_twilio_client` fixture to `mock_twilio_client`, and remove unneeded copying of the mock object. This was also suggested in PR #2614.
* Update `mock_twilio_client` fixture to implement the subset of the client needed for tests
* Add `real_phone_us` fixture to consolidate setup for phone mask creation
* Drop some duplicate test assertions

## How to test

Unit tests should be sufficient. Manual testing can be done by:

* Update `TWILIO_MESSAGING_SERVICE_SID` to two comma-separated values - a full pool SID, and a SID with room
* Create or update a user to have the phone feature and a registered true phone
* On the website, select a relay phone mask:
   - A log message is sent that a message pool is full
   - The relay phone mask was allocated to the second SID
   - The user proceeds to the next step in setting up their phone mask


phones/tests/models_tests.py
```diff
@@ -251,52 +272,61 @@ def test_create_duplicate_relaynumber_raises_error(phone_user, mocked_twilio_cli
     mock_messages_create.assert_not_called()
 
 
+@pytest.fixture
+def real_phone_us(phone_user, mock_twilio_client):
```
groovecoder, line 276: praise: nice helper!

phones/tests/models_tests.py
```diff
@@ -251,52 +272,61 @@ def test_create_duplicate_relaynumber_raises_error(phone_user, mocked_twilio_cli
...
-    mock_services.assert_called_once()
-    call_args = mock_services.call_args
-    assert call_args[0][0] == settings.TWILIO_MESSAGING_SERVICE_SID
-    mock_messaging_number_create = (
-        mock_twilio_client.messaging.v1.services().phone_numbers.create
+    mock_number_create.assert_called_once_with(
+        phone_number=relay_number,
+        sms_application_sid=settings.TWILIO_SMS_APPLICATION_SID,
+        voice_application_sid=settings.TWILIO_SMS_APPLICATION_SID,
+    )
```
groovecoder, line 306: praise: nice clean-up! When I first wrote a bunch of these assertions, for some reason the `assert_called_once_with` method didn't work for me, so I ended up pulling out the `call_args.kwargs` property and asserting on that. I was probably just doing something wrong.

phones/tests/models_tests.py
```diff
@@ -315,37 +345,191 @@ def test_create_relaynumber_already_registered_with_service(
...
 
-    relay_number = "+19998887777"
-    relay_number_obj = RelayNumber.objects.create(user=phone_user, number=relay_number)
 
-    mock_number_create.assert_called_once()
-    call_kwargs = mock_number_create.call_args.kwargs
-    assert call_kwargs["phone_number"] == relay_number
-    assert call_kwargs["sms_application_sid"] == settings.TWILIO_SMS_APPLICATION_SID
-    assert call_kwargs["voice_application_sid"] == settings.TWILIO_SMS_APPLICATION_SID
+def test_create_relaynumber_full_service(
```
groovecoder, line 366: quibble (non-blocking): change this to `test_create_relaynumber_all_services_full`?

When I first read this, I was expecting this to be the "fallback" test.

phones/tests/models_tests.py
```diff
@@ -315,37 +345,191 @@ def test_create_relaynumber_already_registered_with_service(
...
+    )
+    mock_messages_create.assert_not_called()
+    assert caplog.messages == ["twilio_messaging_service"]
+    assert caplog.records[0].code == 21714
+
+
+def test_create_relaynumber_no_service(
+    phone_user, real_phone_us, mock_twilio_client, test_settings, caplog
+):
+    """If no Twilio Messaging Service IDs are defined, registration is skipped."""
```
groovecoder, line 405: comment (non-blocking): Interesting. This means the numbers will not have a service ID, so they won't be in a 10DLC campaign. Which is probaby fine for dev instances

phones/tests/models_tests.py
```diff
@@ -315,37 +345,191 @@ def test_create_relaynumber_already_registered_with_service(
...
+    phone_user, real_phone_us, mock_twilio_client, test_settings, caplog, django_cache
+):
+    """If adding to a pool raises a different error, it is skipped."""
+    twilio_incoming_number_sid = "PNXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
+    twilio_service_sid = test_settings.TWILIO_MESSAGING_SERVICE_SID[0]
+    mock_messages_create = mock_twilio_client.messages.create
+    mock_number_create = mock_twilio_client.incoming_phone_numbers.create
+    mock_number_create.return_value = SimpleNamespace(sid=twilio_incoming_number_sid)
+    mock_services = mock_twilio_client.messaging.v1.services
+    mock_messaging_number_create = mock_services.return_value.phone_numbers.create
```
groovecoder, line 504: suggestion (non-blocking): a lot of this mock setup code seems repeated in each of these tests? refactor into a function?

groovecoder (approved):

phones/models.py
```diff
@@ -280,21 +287,67 @@ def save(self, *args, **kwargs):
...
+def register_with_messaging_service(client: Client, number_sid: str) -> None:
+    """Register a Twilio US phone number with a Messaging Service."""
+
+    assert settings.TWILIO_MESSAGING_SERVICE_SID
+
+    cache_value = cache.get("twilio_messaging_service_closed", "")
+    if cache_value:
+        closed_sids = cache_value.split(",")
+    else:
+        closed_sids = []
```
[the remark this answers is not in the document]
jwhitlock: I think a second function would be needed for updating it later. I instead wrapped it in a class.

phones/tests/models_tests.py
jwhitlock: That was what I thought would be the most common case. However, it does add something we need to be careful about in the deployments (which we're about to update with multiple SIDs anyway)

I see a few options:

1. Skip registration when no service ID is given (this code)
2. Fail if no service ID is given (the phone number is already reserved, however) - requires updating other tests to pass
3. Add an invalid service ID as the default (tests pass, but deployments / local dev fail until a real number is entered)
4. Add a valid service ID as the default (tests pass, real use may fail if service is full, but after allocating number)

I think "no service ID = no registration" might be the best quick option.

phones/tests/models_tests.py
jwhitlock: I was able to move the setup of `twilio_incoming_number_sid` to `mock_twilio_client` setup. I was able to remove a lot of setup code, but only saved one line 😆.

phones/tests/models_tests.py
groovecoder: Yeah, that makes sense.

PR #3024, opened 2023-01-19
Update PyJWT from 2.50 to 2.6.0, with 5 seconds leeway

jwhitlock:
This PR fixes issue #2738.

PyJWT 2.6.0 checks that the "Issued At" claim (iat) is in the past. Add a leeway of 5 seconds to verification to account for clock skew. If it is further in the future, log the "negative age" and return a server error.

Future `iat` claims were an issue on Heroku the week I tried to update to PyJWT 2.6.0. I reverted to PyJWT 2.5.0, added logging of the age, and found that `iat` claims were always in the past on stage and production. This makes some sense, since the clocks across AWS servers should be in sync. This PR also removes logging of all `iat` ages, replacing it with logging when the `iat` claim is rejected as too far in the future.

### DIFF
--- a/privaterelay/tests/views_tests.py
+++ b/privaterelay/tests/views_tests.py
@@ -288,6 +288,7 @@ def get_fxa_event_jwt(
     client_id: str,
     signing_key: RSAPrivateKey,
     event_data: dict[str, Any],
+    iat_skew: int = 0,
 ) -> str:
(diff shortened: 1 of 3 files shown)


groovecoder (approved):
praise: nice sleuthing and good fix!

PR #3062, opened 2023-02-01
Upgrade to Node 18, the current LTS release

Vinnl:
Second attempt, after https://github.com/mozilla/fx-private-relay/pull/2904. Last time, installing dependencies ran into conflicts, but locally, they appear to work for me now, so hopefully all our (transitive) dependencies are compatible with Node 18 now.

The reason I'm revisiting it is because #3010 is blocked by https://github.com/vercel/next.js/issues/44856, which appears to be resolved by using a newer version of Node.


maxxcrawford (approved):
Good by me! Volta setup made this a piece of 🍰.

PR #3074, opened 2023-02-07
ci: configure codeql to ignore python dependencies

groovecoder:
How to test:
Check  if the codeql python job on this PR took less than the 7 minutes it usually takes.


groovecoder:
It looks like CodeQL walks "up" the code imports and checks all code imported by our code even if we ignore the dependency directory. Closing this unless or until we find another way to optimize.

PR #3104, opened 2023-02-14
Update to Django 3.2.18

jwhitlock:
3.2.18 is a [security release](https://www.djangoproject.com/weblog/2023/feb/14/security-releases/) that addresses a denial-of-service attack via multi-part upload forms. Since we do not use Django's upload feature, this is not a risk to Relay, but we'd like to update for security releases as soon as we can.


jwhitlock:
Dependabot opened PR #3108

PR #3125, opened 2023-02-21
Py dep updates 2023 02 21

groovecoder:


jwhitlock:
In the future, please use `pmac prmsg` to generate the PR message, so that we have a summary of what changed and links to the dependabot PRs in the PR message.

Also, please do not close the related dependabot PRs. It signals to dependabot that we are not interested in that version, and may be used to signal to other dependabot users that there is something wrong with that version. Instead, let dependabot notice that the PR is unneeded after the combo PR is merged. 

You can also review the dependabot PRs, and self-merge the combo PR, rather than ask for a review. I'm assuming you didn't review the individual PRs, I'll look at the changelogs.

jwhitlock (approved):

PR #3145, opened 2023-02-28
Documentation: running npm scripts on frontend as a windows user

lloan:
# New feature description

Update readme to provide debug help for windows users running npm scripts. We were running into issues where environment variables were breaking our usage of scripts, such as `npm run watch`. 

Example: `"watch_build": "NEXT_PUBLIC_DEBUG=true npm run build"`

This would break OR we would have to rewrite is as ` "watch_build": "set NEXT_PUBLIC_DEBUG=true && npm run build"` which is Windows specific.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ Please refer to our [coding standards](docs/coding-standards.md) for code styles
    cd ../
    ```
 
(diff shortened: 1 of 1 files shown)


lloan:
> Thanks @lloan. As a Windows-based dev, what do you think is the best way to include these Windows-specific notes?
> 
> * Like this, where we put a `Note` next to each step that may have Windows-specific notes
> * In a separate "Windows" section of the README
> 
> ?

I apologize for the late reply on this. I'm not the best at looking at the little notification icon, top right. However, funny enough, here is an instance where I updated the documentation to reflect something specific to Windows users: 

image

So based on that, I would say the first suggestion/option would be my choice. :)

PR #3175, opened 2023-03-09
Allow togglings flags other than manage_flags

Vinnl:
The previous validation pattern blocked toggling the `manage_flags` flags _and all other flags_... The updated regex should allow everything except for `manage_flags`.


jwhitlock (approved):
Thanks for the quick fix @Vinnl! Reading https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern, this appears to validate the new pattern:

```js
$ node
Welcome to Node.js v18.12.1.
Type ".help" for more information.
> const re_old = RegExp("^(?:(?!manage_flags))$")
undefined
> const re_new = RegExp("^(?:(?!manage_flags)(.*))$")
undefined
> const good_flag = "mask_redesign"
undefined
> re_old.test(good_flag)
false
> re_new.test(good_flag)
true
> const bad_flag = "manage_flags"
undefined
> re_old.test(bad_flag)
false
> re_new.test(bad_flag)
false
> 
```

This looks good to me, ship it!

Vinnl:
Heh, I guess this should succeed as well, but it's probably a good idea not to create such flags anyway :P

```js
re_new.test("manage_flags_suffix")
```

PR #3218, opened 2023-03-22
Revert "Bump @fluent/bundle from 0.17.1 to 0.18.0"

codemist:
Reverts mozilla/fx-private-relay#3206 because `@fluent/react` [wants a lower version](https://github.com/projectfluent/fluent.js/blob/master/fluent-react/package.json#L65) than 0.18 for `@fluent/bundle`


Vinnl (approved):
Relevant issue: https://github.com/projectfluent/fluent.js/issues/610

(I also see upstream has a fix coming up: https://github.com/projectfluent/fluent.js/pull/612)

PR #3245, opened 2023-03-29
Ignore changes for Black 23

jwhitlock:
Ignore linting changes needed when updating from Black 22 to 23. This is follow-on work for PR #3068.


groovecoder (approved):
Whoa cool. TIL (or probably re-learned, because I bet I had to look up this git feature when we first migrated code style to black)

Issue #3223, opened 2023-03-23
Ensure that there is no white space underneath the footer for shorter pages


codemist:
![Screenshot from 2023-03-24 11-25-32](https://user-images.githubusercontent.com/4251/227496893-66c96249-943a-4880-ac95-86732b09edbf.png)

The fix for this was removed in https://github.com/mozilla/fx-private-relay/pull/3222, to fix [MPP-2920](https://mozilla-hub.atlassian.net/browse/MPP-2920), so make sure that doesn't regress when fixing this bug.

[MPP-2920]: https://mozilla-hub.atlassian.net/browse/MPP-2920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jwhitlock:
Leaving open until someone with an Android 12 / 13 browser checks that MPP-2920 did not regress:

With Android 12 or 13, on the Relay dashboard, change a Relay mask status (such as block promotional or block all). Changing this setting breaks the entire page (most of the page is now whitespace).

Uploading 20230310_104241.mp4…

Vinnl:
Verified on Mulch browser (Chromium-based) on Android 11 that MPP-2920 did not regress, which seems safe enough to me to close this issue.

I did ran into https://github.com/mozilla/fx-private-relay/pull/3714, which seems pretty serious as well, and also affects Firefox.

PR #3268, opened 2023-04-05
MPP-2995 Push footer to bottom of desktop screens

codemist:
Quick fix for issue #3223 and MPP-2995. This sets a media query to set the body `width` to `100%` for large device widths (targeting desktop users), without triggering the bug found on mobile browsers related to MPP-2920.


frontend/src/styles/globals.scss
```diff
@@ -28,6 +28,11 @@ body,
 #__next,
 #overlayProvider {
   font-family: $font-stack-base;
+
+  @media screen and #{$mq-lg} {
+    // Ensures that the footer stays at the bottom
+    height: 100vh;
```
maxxcrawford, line 34: Question: What does this do when there's more content than `100vh`?  Should this be `min-height` instead? Does that have the same effect? 

Also, would doing a [screen _height_ media query](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/height) be helpful?

maxxcrawford (commented):
Added question/comment

frontend/src/styles/globals.scss
Vinnl: Yeah, IIRC the original was `100%`, which should allow it to grow to the full height of its parent. (Specifically, the footer is in an element with `display: flex; flex-direction: vertical`, and the element above the footer has a `flex-grow` higher than 0. That allows it to push the footer to the bottom, as long as all its parents elements have `height: 100%`.)

Which is not to say that `100vh` might not work or that you may not have good reasons for using that - just sharing the context behind the original value :)

frontend/src/styles/globals.scss
```diff
@@ -28,6 +28,11 @@ body,
 #__next,
 #overlayProvider {
   font-family: $font-stack-base;
+
+  @media screen and #{$mq-lg} {
+    // Ensures that the footer stays at the bottom
+    height: 100%;
```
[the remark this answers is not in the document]
Vinnl: Looks good! Maybe add a comment with a link to the bug/previous PR to tell future readers why this is in a media query?

Vinnl (approved):
Works for me!

PR #3250, opened 2023-03-31
send RealPhone verification message with iQ

groovecoder:
Add IQ_FOR_VERIFICATION env var and setting.

Refactor api.phones.views._send_iq_sms into PhonesConfig.send_iq_sms

Use PhonesConfig.send_iq_sms to send real phone verification when IQ_FOR_VERIFICATION is True.

## How to test:
1. [Set up local phone dev with iQ](https://github.com/mozilla/fx-private-relay/blob/ec1a69e1432248c98e9105f71ed8c1277695ddee/docs/end-to-end-local-phone-dev-iq.md).
   * Note: this PR updates that doc
   * [x] When you get to the step "Go to http://127.0.0.1:8000/phone/ and verify your real phone: it will send an SMS via inteliquent from the IQ_MAIN_NUMBER above." - verify it sends the verification message from the `IQ_MAIN_NUMBER`!

# Screenshot (if applicable)
image

# Checklist (Definition of Done)
- [x] Product Owner [accepted the User Story](https://mozilla-hub.atlassian.net/browse/MPP-2850?focusedCommentId=667856) (demo of functionality completed) or waived the privilege.
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.
- [x] I've added or updated relevant docs in the docs/ directory
- [x] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to ~stage~ dev.
- [x] ~All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).~
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] ~l10n changes have been submitted to the l10n repository, if any.~

PR #3269, opened 2023-04-05
Cherry picked iq verification mpp 2850

groovecoder:
This is 3 commits from https://github.com/mozilla/fx-private-relay/pull/3250.

I cherry-picked them and made a new branch to get rid of the stray commits that somehow ended up in https://github.com/mozilla/fx-private-relay/pull/3250.

I also added ec4d161b4d92b52f6dc7269c9b7085b8d20731ba to address review comments.


api/tests/iq_views_tests.py
```diff
@@ -54,6 +52,13 @@ def mocked_twilio_client():
         yield mock_twilio_client
 
 
+def _make_real_phone_with_mock_iq(phone_user, **kwargs):
```
jwhitlock, line 55: 👍 nice helper function, cleans up the test setup

phones/iq_utils.py
```diff
@@ -0,0 +1,24 @@
+import json
+
+import requests
+
+from django.conf import settings
+
+from rest_framework import exceptions
+
+
+def send_iq_sms(to_num: str, from_num: str, text: str) -> None:
```
jwhitlock, line 10: 👍 Thanks for moving this. I agree, "utils" is often a code smell for me, but I think it is better to have these in the same place. I know some object-oriented ways to solve this, like a class `InteliquentService` that inherits from `BasePhoneService` and has a method `send_sms()`. We may want something like that to support both services. However, that feels like a future PR, since it would touch the Twilio stuff too, and this is a good focused PR.

phones/models.py
```diff
@@ -39,6 +41,14 @@ def twilio_client() -> Client:
     return phones_config.twilio_client
 
 
+def phones_config():
```
jwhitlock, line 44: No longer used, now that `send_iq_sms` lives in `iq_utils`

privaterelay/settings.py
```diff
@@ -201,9 +201,16 @@
...
 MAX_DAYS_IN_MONTH = 31
 IQ_ENABLED = config("IQ_ENABLED", False, cast=bool)
+IQ_FOR_VERIFICATION = config("IQ_FOR_VERIFICATION", False, cast=bool)
 IQ_FOR_NEW_NUMBERS = config("IQ_FOR_NEW_NUMBERS", False, cast=bool)
+IQ_MAIN_NUMBER = config("IQ_MAIN_NUMBER", "")
 IQ_OUTBOUND_API_KEY = config("IQ_OUTBOUND_API_KEY", None)
 IQ_INBOUND_API_KEY = config("IQ_INBOUND_API_KEY", "")
+IQ_MESSAGE_API_ORIGIN = config(
+    "IQ_MESSAGE_API_ROOT", "https://messagebroker.inteliquent.com"
+)
```
jwhitlock, line 211: This is cleaner than the last PR. But,

1. When will the API Origin change? Is this premature flexibility?
2. Should the setting `IQ_MESSAGE_API_ORIGIN` match the env variable `IQ_MESSAGE_API_ROOT`

jwhitlock (approved):
Thanks @groovecoder! Sending the verification code through iq still works.

Approving with nits, merge when you are ready

jwhitlock:
the `mypy` errors can be fixed with some annotations in settings:

```python
IQ_OUTBOUND_API_KEY: str | None = config("IQ_OUTBOUND_API_KEY", None)
```

and

```python
IQ_PUBLISH_MESSAGE_URL: str = f"{IQ_MESSAGE_API_ORIGIN}{IQ_MESSAGE_PATH}"
```

I'm not sure why this is a problem now that the code has been moved to a new file. There's some issues that smell like this in the `django-stubs` project.

privaterelay/settings.py
groovecoder: 1. The API origin could change if we use the iQ "sandbox" environment in local, dev, or stage.
2. I updated the env var to match setting name `IQ_MESSAGE_API_ORIGIN`

PR #3270, opened 2023-04-06
2023.04.05 stage hotfix

say-yawn:
This PR fixes Sentry errors:
- [accounts_subscription_subscription_date_invalid](https://mozilla.sentry.io/share/issue/946a50b1591a45b583510f8bdd884d02/)
  - which is a detailed error of [no_subscription_data_in_fxa_for_user_with_phone_subscription](https://mozilla.sentry.io/share/issue/9ad57bdd26304379a44d70bbff358261/)
- `TokenExpiredError`
  - [CustomOAuth2Error](https://mozilla.sentry.io/share/issue/ed8b34eb701b49d783eb2aa9ed0cf848/)

How to test:
- `accounts_subscription_subscription_date_invalid`: set the `extra_data` of the social account to include `relay-phones` and have a non-phone subscription
- `TokenExpiredError`: change the social token's `expires_at` attribute to a datetime value in the past

 #


privaterelay/fxa_utils.py
```diff
@@ -67,6 +67,9 @@ def _token_updater(new_token):
         "expires_in": expires_in,
     }
 
+    # TODO: find out why the auto_refresh and token_updater is not working
+    # and instead we are manually refreshing the token at
+    # FxAToRequest and get_subscription_data_from_fxa
```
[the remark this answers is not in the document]
groovecoder: Oh, hmm ... there's no "auto refresh" feature in allauth yet. It uses `requests_oauthlib` which has this feature, but `allauth` doesn't seem to call that feature of it. That's the best I can come up with here.

privaterelay/tests/fxa_utils_tests.py
```diff
@@ -0,0 +1,188 @@
...
+
+@pytest.fixture()
+def phone_user(db):
+    yield make_phone_test_user()
+
+
+@patch(f"{MOCK_BASE}.get_subscription_data_from_fxa")
+def test_get_phone_subscription_dates_refreshed_twice(mocked_phone_dates, phone_user):
+    social_account = SocialAccount.objects.get(user=phone_user)
+    mocked_phone_dates.return_value = {"social_token": "new token", "refreshed": True}
```
groovecoder, line 34: suggestion (blocking): it seems like this `mocked_phone_dates` variable name should be `mocked_data_from_fxa`?

Or maybe this is an indication that the `get_subscription_data_from_fxa` is returning 2 very different things in different situations? I.e., if the token is expired or the requested scopes are not allowed, it returns a dictionary with a new token and a "refreshed" indicator. But if the token is valid, then it returns the data from FXA.

That smells like this function is doing 2 very different things and returning 2 very different kinds of data.

groovecoder (changes requested):
Nice thorough tests. I left a blocking suggestion to at least change the variable name of the mock object when it's returning mocked data of a refreshed token vs. mocked data from FXA.

privaterelay/tests/fxa_utils_tests.py
say-yawn: No you are correct and I updated the tests to use the new variable name `mocked_data_from_fxa` on 73858e9. Agreed on potentially confusing `get_subscription_data_from_fxa` but with the refresh also using the client it's a bit complicated to refactor now and hopefully when we revisit the refresh logic we can handle it there.

PR #3289, opened 2023-04-11
Add docs on how to verify npm dependency updates

Vinnl:
I realised when I had a holiday that I'm not comfortable yet merging dependency updates on Monitor, and that the same probably holds for other people for Relay's frontend. So here's an attempt at docs for that: if you see an npm Dependabot PR come in and are not sure if it's safe to merge, you should be able to look up that dependency in this doc.

It's probably a good idea to add something similar for our pip dependencies, if anyone has a moment.

In the process I also found that we don't actually use some dependencies, so I removed those.

[**Rendered docs here**](https://github.com/mozilla/fx-private-relay/blob/docs/dependencies/docs/dependency-updates.md).


docs/dependency-updates.md
```diff
@@ -0,0 +1,164 @@
...
+You can line up multiple Dependabot PRs for merging by adding the comment
+`@dependabot merge`. Dependabot will then automatically keep the PR up-to-date
+with changes in `main`, and merge it if the Checks continue to run successfully.
+Alternatively, you can use the CLI tool [`pmac`](https://github.com/willkg/paul-mclendahand).
+
+## npm
+
+We have two npm projects in this repository: the website frontend (identified by
+`/frontend/package.json`), and the root project from which our end-to-end tests
+are run (identified by `/package.json`). The former is defined as a
```
groovecoder, line 43: question (non-blocking): should we move the E2E test `package.json` to `e2e-tests/package.json` ?

groovecoder (changes requested):
When I do dependency updates, I do them all together to make sure I'm checking for any possible integration bugs between the dependency updates. And I want to do it as efficiently as possible. So, I like to use `pmac` to get all the dependency updates together into a single branch, and then review it all at once.

In that workflow, it makes more sense to organize this doc as a list of the actions that humans should perform to check any and all dependnecy updates, rather than listing out each depenency on its own and the steps to check it in isolation.

E.g., when I combine any number of js dependency updates together, I can cover nearly all of them by: "open the preview deployments and check that you can still ..."

* interact with UI via both keyboard and mouse
* see website content in different languages
* navigate between different pages
* interact with the API
* see the right styling
* generally interact with everything
* see toast notifications
* see that help & tips stick to the bottom of the viewport
* see GA reuests in dev tools

... which will cover react (aria, stately, toastify, intersection observer, ga), fluent, next, swr, msw, sass, and protocol.

So ...

Suggestion (blocking): Change the outline of the doc to more like:

1. Check CI
when __ is updated, to ensure ___.
2. Check preview deployment
when __ is updated, to ensure ___.
3. (Optional) Local updates + checks
when __ is updated, to ensure ___.
4. (Optional) Local E2E tests
when __ is updated, to ensure ___.

I realize this means doing E2E human testing, which our automated test suite should really do for us on every PR (including a PR that updates all js dependencies). But if I really want to feel confident in clicking the "merge" button on a bunch of js dependency updates, that's how rigorous I want to be with it.

docs/dependency-updates.md
```diff
@@ -0,0 +1,164 @@
...
+if a dependency's [major version](https://semver.org/) is bumped, or if a
+dependency doesn't use Semantic Versioning.
+
+If you're unsure whether anything looks as expected, try to replicate what
+you're doing on the `main` branch. If the results differ, there's probably an
+issue with the update.
+**Make sure to do a fresh install of the dependencies when switching branches!**
+
+You can line up multiple Dependabot PRs for merging by adding the comment
+`@dependabot merge`. Dependabot will then automatically keep the PR up-to-date
```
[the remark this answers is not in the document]
maxxcrawford: Yes! 

> `@dependabot rebase` rebases the pull request.


See the [full documentation](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates#managing-dependabot-pull-requests-with-comment-commands) for all the commands.

frontend/package.json
```diff
@@ -46,14 +45,12 @@
...
     "@typescript-eslint/parser": "^5.55.0",
     "babel-jest": "^29.4.2",
-    "chokidar-cli": "^3.0.0",
     "eslint": "^8.37.0",
     "eslint-config-next": "^13.3.0",
     "eslint-plugin-jest-dom": "^4.0.3",
     "eslint-plugin-testing-library": "^5.10.2",
     "fast-check": "^3.7.0",
     "husky": "^8.0.3",
-    "identity-obj-proxy": "^3.0.0",
```
maxxcrawford, line 56: question (non-blocking): How did you know these were not in use? Did you use a library to detect this?

docs/dependency-updates.md
```diff
@@ -0,0 +1,164 @@
...
+deployment and returns mock data instead. Both can be verified by opening the
+preview deployment and making sure that you can still interact with the API
+(e.g. by creating a new mask).
+
+### `sass` and `@mozilla-protocol/core`
+
+[Sass](https://sass-lang.com/) gives us some handy tooling to ease working with
+CSS, and [Protocol](https://protocol.mozilla.org/) is Mozilla's design system
+from which we use some Sass variables. Both can be verified by checking the
+preview deployment and making sure the styling doesn't look weird.
```
maxxcrawford, line 120: suggestion: You may add an extra line about how Protocol adds "migration tips" to their releases if anything is broken with new updates.

https://github.com/mozilla/protocol/releases/tag/v16.0.0

docs/dependency-updates.md
Vinnl: Maybe! :shrug: It's not necessarily part of the frontend (one end of the end-to-end is the backend :) ), but it is pretty confusing to have two `package.json`s, I'm not always 100% sure where to run which command, and it sometimes confuses our tooling (e.g. Dependabot) as well, so I'd be in favour of that.

One downside is that you'd no longer be able to run `npm install` in the root, but I think npm gives a clear error message there, so that's probably not a biggie.

docs/dependency-updates.md
```diff
@@ -0,0 +1,164 @@
...
+That said, it can't harm to occassionally scan the results of the
+`build_frontend` or `test_frontend` CI runs to see if any new warnings were
+added. If `jest-junit` is updated, you'll want to check that CircleCI can still
+see code coverage for the frontend.
+
+### `@playwright/test` and `dotenv`
+
+Our Playwright tests are currently only triggered in CI once a day from the
+`main` branch. Thus, there's unfortunately no easy way to verify that it still
+runs successfully until after merging. To do so, after merging the PR, open the
```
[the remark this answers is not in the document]
Vinnl: I was going back-and-forth on this since I believe it's still pretty hard to get a consistent result with the current tests between the local environment and the CI, but I guess it's unlikely to succeed locally even though it would fail in CI, so added in d9acd6a5.

docs/dependency-updates.md
```diff
@@ -0,0 +1,164 @@
...
+verify that the website was put together successfully, and that you can navigate
+between different pages (e.g. from the home to the FAQ) without issues.
+
+### `@next/bundle-analyzer`
+
+We use this package to investigate if any of our dependencies take up an
+inordinate amount of bytes in our payload. To verify that this still runs
+successfully, run
+
+    ANALYZE=true npm run build
```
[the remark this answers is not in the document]
Vinnl: The thing is that there's not really a pass or fail result; it rather gives you a report of which modules take up which percentage of the final bundle. So it's a tool to help investigate in case issues with our bundle size occur. (I think there's also something to be said for removing this dependency, since I don't think we've used it more than once or twice - so we can just temporarily install it if we need it in the future.)

docs/dependency-updates.md
Vinnl: TIL, nice! But I think a number of our dependencies do the same (see e.g. the changelog in https://github.com/mozilla/fx-private-relay/pull/3164 referencing a migration guide), so I think that's covered by this note at the top:

> and a cursory inspection of the changelog (if any) for anything that might be relevant to Relay

So it might be a bit noisy to add that to every dependency that does that?

Vinnl:
@groovecoder I'm 100% with you that I'd like automated E2E tests to cover most of this. That aside, I think it would be good to allow for people to choose their own workflows. For example, I prefer doing them one by one so it's easier to locate the source of problems. And perhaps more importantly: many of the tests aren't relevant for many of the dependencies, which can make the job a lot quicker. For example, `react-toastify` almost never updates, so will almost never need to be checked. (And this aligns with my experience: I don't think I've missed regressions due to being selective in my testing. The few regressions we had so far were due to Dependabot doing something different from npm, IIRC.)

docs/dependency-updates.md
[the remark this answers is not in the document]
groovecoder: I see. Looks like we could run the underlying `webpack-bundle-analyzer` tool with `-m json` and then make sure the total size is under a certain value. But again - outside the scope of this PR.

groovecoder (approved):
That makes sense, and I don't want to let the perfect be the enemy of the good, so I'm good to merge this as-is now.

PR #3316, opened 2023-04-18
MPP-2847 - Document the switch to optional TLS

jwhitlock:
# New feature description

We currently require TLS for the incoming email server, but will change it to optional TLS. This is not obvious from a security perspective. Relay users will have a different experience as well. Relay masks will be accepted more places, and some additional emails (mostly good, some unwanted) will be delivered.

This PR adds ADR 0001 to document the change and the pros / cons

# How to test
SREs will make the change in AWS. There is nothing to review here but the documentation.


docs/0001-optional-tls-for-incoming-smtp.md
```diff
@@ -0,0 +1,155 @@
...
+
+**Good** - Relay masks will be accepted on more websites. This will protect
+our user's true email.
+
+**Good** - Valid emails from servers and clients that do not implement TLS /
+`STARTTLS` will now be forwarded to Relay users.
+
+**Bad** - Relay users will not know if an email was sent with TLS or not.
+Amazon SES does not provide this data. This prevents Relay from allowing users
+to decide about requiring TLS, or creating filter rules based on TLS.
```
groovecoder, line 128: quibble (non-blocking):
```suggestion
**Bad** - Relay users will not know if an email was sent with TLS or not.
Amazon SES does not provide this data. So Relay users can't require TLS
or create filter rules based on TLS.
```

groovecoder (changes requested):
This is great!

suggestions (blocking):
* [x] Re-name this to an ADR.
* [ ] Include a link to https://docs.google.com/document/d/10mZUt8TiVflmx-d4pIyZ2YyFZh4NOPydo1l_i4Prlhk/edit

jwhitlock:
Thanks @groovecoder! I move the document to `docs/adr`, where I _meant_ to put it in the first place, and added to the references. Back to you!

PR #3356, opened 2023-04-25
Add X-Forwarded-For meta field to email

maxxcrawford:


jwhitlock (changes requested):
Quick feedback, I haven't looked closely.

This seems to work well for messages sent to a relay mask. Gmail displays the mail as coming from my original email, and when I reply, it goes to the replies@ address.

However, the reply email is shown as coming from my true email, not my email mask. So, a reply will expose the true email.

We'll need to ensure messages to a mask are treated differently than messages from a mask.

PR #3387, opened 2023-05-07
Remove double magnifying glass icon in search aliases input

### LINKED ISSUE #1779: Double magnifying glass icon in Search aliases bar on iOS
**Build:**
Stage

**Affected platforms:**
iOS 15.2

**Browsers tested on:**
Firefox, latest release

**Prerequisites:**
The user has at least one alias already created.

**Steps to reproduce:**

2. Go to "Search aliases" bar just above the aliases list and observe it.

**Actual result:**
Two magnifying glass icons are displayed in the Search bar.

**Expected result:**
Only one magnifying glass icon is displayed in the Search bar.

**Notes:**
N/A

double magnifying glass icon

joaonunomota:
This PR fixes #1779.

The search aliases input shows two magnifying glass icon when viewed in iOS browsers, this is due to additional styling provided to inputs with the attribute `type="search"`. The simplest solution to avoid this was changing the attribute to `type="text`, which according to https://caniuse.com/input-search will only affect the styling added by browsers. In terms of accessibility the affected inputs still have associated labels or title attributes that describe their purpose for searching.

How to test:

1. Go to Private Relay website (on iOS browser) and search inputs for:
   a. Email Masks.
   b. Phone Numbers.
   c. Subdomains.
3. Review the number of magnifying glasses on search filter.

Before (Windows 11)

image

Before (iOS)

before

After (Windows 11)

image

After (iOS)

after


Vinnl:
Thanks @joaonunomota! I'm a bit wary about approaching it like this; the "Known issues" tab also mentions the keyboard not adapting on Safari (implying that Chrome does adapt it), and more importantly, I'm not sure whether screen readers adjust their behaviour to this attribute, or will do so in the future.

Did you happen to try adding `display: hidden` to the [`::-webkit-search-results-button` pseudo-element](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-search-results-button) as well? If that doesn't work, or if there's good reason not to use it, then I suppose this is fine, but otherwise I think I'd lean towards that approach.

joaonunomota:
Thanks for the feedback @Vinnl! I had no luck with `::-webkit-search-results-button`, but `::-webkit-search-decoration` removed the icon from the input. It's one of a few pseudo-elements used in these stylesheets:
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/css/html.css
https://chromium.googlesource.com/chromium/blink/+/refs/heads/main/Source/core/css/html.css

I placed the rule in `globals.scss` to avoid repetition, but let me know if there is somewhere more appropriate for it.

Vinnl (approved):
That's great @joaonunomota; globals.scss seems fine to me, I think we're OK with this icon not appearing anywhere in all browsers. And the pseudo element seems to work well. Thanks!

PR #3408, opened 2023-05-11
Add demo data to take screenshots from

Vinnl:
Based on [a Slack discussion](https://mozilla.slack.com/archives/C03DQBV9X5L/p1683792638115959?thread_ts=1683743368.418369&cid=C03DQBV9X5L), this might help product to create screenshots for illustrations.

Still in draft while waiting for confirmation that it's demo-worthy enough, or if we need more data.


lloan (approved):
Looks good! Awesome job!

Are you going to add more to this demo data?

Vinnl:
We could add phone data to demo if that's desired but I'd like Product to determine what kind of demo data they'd like to show. I'm not even sure how often they'd use it anyway - it looks like Tony has already used the preview environment. Ah well, it's probably not a big maintenance burden anyway, and if it is, we can remove it again.

PR #3428, opened 2023-05-15
pyjwt 2.6.0 to 2.7.0, with type fixes

jwhitlock:
* Update to pyjwt 2.7.0, from PR #3413, which includes expanded and more accurate type hints. 
* Update our code for the new hints
* Arrange imports and remove some unused imports

### DIFF
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -1,20 +1,10 @@
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timezone
 from functools import lru_cache
 from hashlib import sha256
 from typing import Any, Iterable, Optional, TypedDict
 import json
 import logging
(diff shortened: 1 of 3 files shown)


privaterelay/views.py
```diff
@@ -1,19 +1,10 @@
...
 import os
-from rest_framework.decorators import api_view, schema
-
-# from silk.profiling.profiler import silk_profile
-
-from google_measurement_protocol import event, report
-import jwt
-from oauthlib.oauth2.rfc6749.errors import CustomOAuth2Error
-from requests_oauthlib import OAuth2Session
-import sentry_sdk
```
groovecoder, line 16: praise: thanks for the clean-up!!

PR #3432, opened 2023-05-15
Translation fixes

jwhitlock:
Several changes:

* Use `Bundle.reload()` to force Fluent files to reload before tests that use Fluent. This allows failing a test for Fluent errors, used to find many of the issues fixed in this PR.
* Load `faq.ftl` on the backend, which includes the `email-size-limit` used in "misc.ftl"
* Remove `other-reply-not-forwarded-2`, already merged to the l10n repo
* Remove strings unused in the current wrapper email template
* Add translator notes for variables like `{ $url }` and `{ $attrs }`
* Add `attrs=""` for strings that use `{ $attrs }`
* In the wrapper email template, convert partial strings, such as `"by" <a> "Firefox Relay"</a>` to HTML fragments, such as `"by <a> Firefox Relay </a>"`, to allow translators to match their language's sentence construction.

After this change is merged, we can more easily submit upstream to the l10n repo.

# How to test

Locally, view:

* http://127.0.0.1:8000/emails/wrapped_email_test
* http://127.0.0.1:8000/emails/reply_requires_premium_test
* http://127.0.0.1:8000/emails/first_time_user_test

They should be the same before and after this PR. No Fluent errors are printed to the console.

### DIFF
--- a/emails/templates/emails/first_time_user.html
+++ b/emails/templates/emails/first_time_user.html
@@ -294,7 +294,7 @@
                           </p>
 
                           <p style="line-height: 140%; text-align: left; word-wrap: break-word;">
-                            {% ftlmsg 'first-time-user-email-how-item-1-subhead-html' url='https://addons.mozilla.org/firefox/addon/private-relay/?utm_campaign=relay-onboarding&utm_source=relay-onboarding&utm_medium=email&utm_content=addon-cta' %}
+                            {% ftlmsg 'first-time-user-email-how-item-1-subhead-html' url='https://addons.mozilla.org/firefox/addon/private-relay/?utm_campaign=relay-onboarding&utm_source=relay-onboarding&utm_medium=email&utm_content=addon-cta' attrs='' %}
                           </p>
 
                       </td>
@@ -334,7 +334,7 @@
                           </p>
 
                           <p style="line-height: 140%; text-align: left; word-wrap: break-word;">
-                            {% ftlmsg 'first-time-user-email-how-item-3-subhead-html' url=SITE_ORIGIN|add:'/accounts/profile/' %}
+                            {% ftlmsg 'first-time-user-email-how-item-3-subhead-html' url=SITE_ORIGIN|add:'/accounts/profile/' attrs='' %}
                           </p>
 
                       </td>
--- a/emails/templates/emails/wrapped_email.html
+++ b/emails/templates/emails/wrapped_email.html
@@ -123,20 +123,18 @@
           <img width="30" src="{{ SITE_ORIGIN }}/static/images/email-images/relay-icon.png" style="margin-right: 5px; display: inline-block; vertical-align: top;" alt="relay icon"/>   
            
           <p style="margin-top: 0; margin-bottom: 5px; display: inline-block;"> 
-            {% with display_email|striptags|urlencode as mask_url %}
             <span class="forwarded-from-email" style="display: block; color: #FFFFFF;"> 
-            {% ftlmsg 'relay-email-forwarded-from' %} 
-            <a class="container-link" href="{{ SITE_ORIGIN }}/accounts/profile/#{{ mask_url }}" style="margin-right: 30px;color: #FFFFFF;"> {{ display_email|striptags }} </a>
+            {% with display_email|striptags|urlencode as mask_url %}
+            {% ftlmsg 'relay-email-forwarded-from-html' url=SITE_ORIGIN|add:'/accounts/profile/#'|add:mask_url attrs='class="container-link" style="margin-right: 30px;color: #FFFFFF;"' email_address=display_email|striptags %}
             {% endwith %}
-            </span> 
+            </span>
 
-          <span style="margin-top: 0; margin-bottom: 5px; display: block;color: #FFFFFF;">{% ftlmsg 'relay-email-by-line' %} <a class="container-link" href="{{ SITE_ORIGIN }}/accounts/profile" style="margin-right: 30px;color: #FFFFFF;">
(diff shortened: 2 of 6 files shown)


emails/templates/emails/first_time_user.html
```diff
@@ -294,7 +294,7 @@ <h1 style="margin-bottom: 30px; line-height: 140%; text-align: center; word-wrap
                           </p>
 
                           <p style="line-height: 140%; text-align: left; word-wrap: break-word;">
-                            {% ftlmsg 'first-time-user-email-how-item-1-subhead-html' url='https://addons.mozilla.org/firefox/addon/private-relay/?utm_campaign=relay-onboarding&utm_source=relay-onboarding&utm_medium=email&utm_content=addon-cta' %}
+                            {% ftlmsg 'first-time-user-email-how-item-1-subhead-html' url='https://addons.mozilla.org/firefox/addon/private-relay/?utm_campaign=relay-onboarding&utm_source=relay-onboarding&utm_medium=email&utm_content=addon-cta' attrs='' %}
```
groovecoder, line 297: question (non-blocking): do we have to, or is it best practice, to include arguments even when they are meant to be blank?

emails/templates/emails/wrapped_email.html
```diff
@@ -123,20 +123,18 @@
           <img width="30" src="{{ SITE_ORIGIN }}/static/images/email-images/relay-icon.png" style="margin-right: 5px; display: inline-block; vertical-align: top;" alt="relay icon"/>   
            
           <p style="margin-top: 0; margin-bottom: 5px; display: inline-block;"> 
-            {% with display_email|striptags|urlencode as mask_url %}
             <span class="forwarded-from-email" style="display: block; color: #FFFFFF;"> 
-            {% ftlmsg 'relay-email-forwarded-from' %} 
-            <a class="container-link" href="{{ SITE_ORIGIN }}/accounts/profile/#{{ mask_url }}" style="margin-right: 30px;color: #FFFFFF;"> {{ display_email|striptags }} </a>
+            {% with display_email|striptags|urlencode as mask_url %}
+            {% ftlmsg 'relay-email-forwarded-from-html' url=SITE_ORIGIN|add:'/accounts/profile/#'|add:mask_url attrs='class="container-link" style="margin-right: 30px;color: #FFFFFF;"' email_address=display_email|striptags %}
```
groovecoder, line 128: change (blocking): Don't change the template yet until the message has been translated into our "top supported languages". I'm not sure what those are for email-receiving users, but our top 10 languages for web traffic are:

image

So maybe German, French, Russian, Spanish, Portuguese (Brazil), and Italian? Those cover 82% of our traffic.

If we change the template from an existing ftl msg ID that is translated to one that is only in the `pending.ftl` file, this will effectively "un-translate" this message in the email wrapper until translators have translated this new message, and we release the new translations.

I think the best practice here is to add this new ftl msg to the file for translation, and then after it's been translated into enough langauges, change the template.

emails/templates/emails/wrapped_email.html
```diff
@@ -123,20 +123,18 @@
...
             {% endwith %}
-            </span> 
+            </span>
 
-          <span style="margin-top: 0; margin-bottom: 5px; display: block;color: #FFFFFF;">{% ftlmsg 'relay-email-by-line' %} <a class="container-link" href="{{ SITE_ORIGIN }}/accounts/profile" style="margin-right: 30px;color: #FFFFFF;">
-            {% if has_premium %} 
-              {% ftlmsg 'relay-email-premium-by-line-link' %} 
+          <span style="margin-top: 0; margin-bottom: 5px; display: block;color: #FFFFFF;">
+            {% if has_premium %}
+              {% ftlmsg 'relay-email-premium-byline-html' url=SITE_ORIGIN|add:'/accounts/profile/' attrs='class="container-link" style="margin-right: 30px;color: #FFFFFF;"' %}
```
groovecoder, line 134: change (blocking): same change here - don't change the template until the message is translated.

emails/templates/emails/wrapped_email.html
```diff
@@ -123,20 +123,18 @@
...
             {% else %}
-              {% ftlmsg 'relay-email-by-line-link' %}
+              {% ftlmsg 'relay-email-byline-html' url=SITE_ORIGIN|add:'/accounts/profile/' attrs='class="container-link" style="margin-right: 30px;color: #FFFFFF;"' %}
```
groovecoder, line 136: change (blocking): same change here - don't change the template until the message is translated.

emails/tests/views_tests.py
```diff
@@ -1143,6 +1144,10 @@ def test_wrapped_email_test(
     has_tracker_report_link,
     num_level_one_email_trackers_removed,
 ):
+    # Reload Fluent files to regenerate errors
+    if language == "en":
+        main.reload()
```
groovecoder, line 1149: praise: nice fix-up.

groovecoder (changes requested):

emails/templates/emails/wrapped_email.html
jwhitlock: None of the `relay-email-` strings have been submitted to the translation repository, so there are no existing translations that will break

jwhitlock (commented):
@groovecoder, thanks for the quick review!

None of the strings in `wrapped_email.html` have been submitted upstream, so I felt it was OK to remove unused ones and rewrite others. Once this PR is merged, I can submit these and the `first-time-user-email` strings to https://github.com/mozilla-l10n/fx-private-relay-l10n .

groovecoder (approved):

PR #3435, opened 2023-05-17
MPP-3131 - Updates to the email wrapper

jwhitlock:
Make a few changes to the email wrapper (Jan 2023 version) for forwarded emails. User-facing changes:

* If a user is in a country that doesn't allow premium, then don't link to the premium upsell page
* In the plain-text wrapper, stop linking to the FAQ with the old attachment limit of 150KB. The FAQ link was removed in the HTML version back in Jan 2023.

For developers:
* Remove trailing whitespace (use the "hide whitespace" option to ignore for review)
* Convert `/emails/wrapped_email_test` to include links to try other options, like changing language and removing trackers. The tracker link now varies based on the number of removed trackers.
* Remove unused "has_attachment" from `wrap_html_email`
* Remove unused "survey_text" and "survey_link" from email context, removed in the Jan 2023 version

For translators:
* Vary `relay-email-trackers-removed` on the number of trackers, so it is '0 email trackers removed' or '1 email tracker removed'. This was submitted in https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/149

To test:
* Load locally, go to http://127.0.0.1:8000/emails/wrapped_email_test, click around.


emails/templates/emails/wrapped_email.html
```diff
@@ -136,41 +136,41 @@
...
             {% endif %}
-          </p> 
+          </p>
 
           {% with display_email|striptags|urlencode as mask_url %}
-          <p class="relay-mask" style="margin: 0; display: inline-block;"> 
-              {% if has_premium %}
-              <a class="container-link" href="{{ SITE_ORIGIN }}/accounts/profile/#{{ mask_url }}" style="color: #FFFFFF;"> 
+          <p class="relay-mask" style="margin: 0; display: inline-block;">
+              {% if has_premium or not in_premium_country %}
```
groovecoder, line 159: suggestion (blocking): Remove this change. We redirect these users from `/premium` to `/premium/waitlist` and we use that wait-list to help inform us which countries we should expand into.

jwhitlock: This was the only usage of `in_premium_country`, so I was able to remove that as well. I left a comment in the template about why it is OK to send all users to `/premium`.

jwhitlock:
I removed `in_premium_country` from the template logic as well as the context, which required more code changes.

I updated the Fluent warning at the top of the templates to refer to `/privaterelay/ftl_bundles.py` to get the latest list of loaded files, rather than the comment continually being out of date.

Finally, I changed the test tracker link to make it a little less weird.

groovecoder (approved):

PR #3454, opened 2023-05-23
fix MPP-3123: include per-language hero images

groovecoder:
This PR fixes #MPP-3123.

How to test:

1. Set up end-to-end email, or use the dev server
2. Set you browser to one of the 16 languages of the images in this PR
3. Sign up for a new Relay account
4. Check the email inbox of the FXA
   * [ ] The email header "hero image" should be in the language matching the language of the browser when you signed up


emails/templates/emails/first_time_user.html
```diff
@@ -231,7 +231,7 @@
             <tbody>
               <tr>
                 <td align="center"> 
-                  <img align="center" border="0" src="{{ SITE_ORIGIN }}/static/images/email-images/first-time-user/hero-image-en.png" alt="" title="" /> 
+                  <img align="center" border="0" src="{{ SITE_ORIGIN }}/static/images/email-images/first-time-user/hero-image-{{request.LANGUAGE_CODE}}.png" alt="" title="" style="width:425px;max-width:425px;"/>
```
jwhitlock, line 234: This fails for at least `zh-cn`, probably for other supported languages. Attempting to load https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/static/images/email-images/first-time-user/hero-image-zh-cn.png results in a blank spot in the email:

PR 3454 - no zh-cn

I believe `hero-image-cn.png` should be renamed `hero-image-zh-cn.png`, or even `hero-image-zh.png`.

jwhitlock (changes requested):
Thank @groovecoder. After attempting to review this PR, I'm negative on translating the images. I don't think much value is added versus the English version. The UI in the image is very small, and it is hard to tell that it is translated. Additionally, we did not translate any of the labels, which are English phrases rather than website URLs, so it looks partially translated.

If we keep this feature, we need to ensure that we only link when there is a provided image that matches the language code, and fallback to the English version rather than a 404 image.

I'd also like a test page that lets me flip between languages more easily than creating a FxA user, but that can be another day.

emails/templates/emails/first_time_user.html
groovecoder: Updated to default to `hero-image-en.png` and only change to an l10n image if one is available for the `request.LANGUAGE_CODE`.

jwhitlock (approved):
Thanks @groovecoder, works for me!

groovecoder:
Okay, I tested on the dev server with:
* `pt-br`: correctly picked `hero-image-pt.png`
* `de`: correctly picked `hero-image-de.png`
* `af`: correctly fell back to `hero-image-en.png`

PR #3456, opened 2023-05-23
Replace SMS term with text in FAQ page

codemist:
This swaps out the old string containing the 'SMS' term for the updated one which contains the 'text' term instead.


jwhitlock (approved):
Thanks @codemist!

PR #3487, opened 2023-06-01
MPP-2117: Add type annotations for sending email, allow ASCII encoding

jwhitlock:
This PR adds type annotations to `ses_relay_email` and all the `utils` functions it calls.

## How to Test
* Get the code in an email-relaying environment
* Send an HTML message to a mask, the wrapped message is delivered
* Send a plain text, ASCII only message to a mask. The wrapped message is delivered. The "original" email has the text version in ASCII ("us-ascii")
* Send a plain text message with non-ASCII unicode to a mask. The wrapped message is delivered. The "original" email has the text version in base64-encoded UTF-8.

## Details

When adding annotations, `mypy` flagged this code:

```python
charset = "UTF-8"
...
textpart = MIMEText(body_text.encode(charset), "plain", charset)
```

with the error:

```
Argument 1 to "MIMEText" has incompatible type "bytes"; expected "str"  [arg-type] 
```

Looking in the Python code, `MIMEText` does expect a `str` and passes it to [email.Message.set_payload](https://github.com/python/cpython/blob/ede89af605b1c0442353435ad22195c16274f65d/Lib/email/message.py#L330C1-L348), which detects bytes and decodes it back to a string with `payload.decode('ascii', 'surrogateescape')` (see [PEP 383](https://peps.python.org/pep-0383/)), and then re-encodes it as UTF-8. This is extra work at best and will introduce encoding errors at worst.

Instead, pass the content as a `str`. For text content, let [MIMEText](https://github.com/python/cpython/blob/3698fda06eefb3c01e78c4c07f46fcdd0559e0f6/Lib/email/mime/text.py#L12-L40) detect the encoding (`us-ascii` with a `utf-8` fallback), which will be smaller and look better for ASCII messages. For HTML content, our translations always output unicode ([U+2068](https://www.compart.com/en/unicode/U+2068) and [U+2069](https://www.compart.com/en/unicode/U+2069)), so pass `utf-8` and skip encoding detection.

### DIFF
--- a/emails/utils.py
+++ b/emails/utils.py
@@ -7,6 +7,8 @@ from email.mime.text import MIMEText
 from email.mime.application import MIMEApplication
 from email.utils import parseaddr
 from functools import cache
+from io import IOBase
+from typing import cast, Any, Callable, Literal, TypeVar
 import json
 import pathlib
 import re
@@ -32,6 +34,7 @@ from django.template.defaultfilters import linebreaksbr, urlize
 
 from privaterelay.utils import get_countries_info_from_request_and_mapping
 
+from .apps import EmailsConfig
 from .models import (
     DomainAddress,
     RelayAddress,
@@ -101,8 +104,11 @@ def strict_trackers():
     return get_trackers(level=2)
 
 
-def time_if_enabled(name):
-    def timing_decorator(func):
+_TimedFunction = TypeVar("_TimedFunction", bound=Callable[..., Any])
+
+
+def time_if_enabled(name: str) -> Callable[[_TimedFunction], _TimedFunction]:
+    def timing_decorator(func: _TimedFunction) -> _TimedFunction:
         def func_wrapper(*args, **kwargs):
             ctx_manager = (
                 metrics.timer(name)
@@ -112,7 +118,7 @@ def time_if_enabled(name):
             with ctx_manager:
                 return func(*args, **kwargs)
 
-        return func_wrapper
+        return cast(_TimedFunction, func_wrapper)
 
     return timing_decorator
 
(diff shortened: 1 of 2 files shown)


jwhitlock:
I'm getting `mypy` failures in CircleCI on code I didn't change. I've seen this before - using a setting in `emails/apps.py` appears to fail because it is treated as `Any` on the first run. I'd like to fix this so others stop tripping on it.

emails/utils.py
```diff
@@ -101,8 +104,11 @@ def strict_trackers():
     return get_trackers(level=2)
 
 
-def time_if_enabled(name):
-    def timing_decorator(func):
+_TimedFunction = TypeVar("_TimedFunction", bound=Callable[..., Any])
```
groovecoder, line 107: praise: TIL `TypeVar(bound=)` argument creates an "upper bound" on the type.

emails/utils.py
```diff
@@ -188,27 +194,36 @@ def get_welcome_email(request: HttpRequest, user: User, format: str) -> str:
     )
 
 
+MessageBodyContent = dict[Literal["Charset", "Data"], str]
+MessageBody = dict[Literal["Text", "Html"], MessageBodyContent]
+AttachmentPair = tuple[str, IOBase]
+MailJSON = dict[str, Any]
```
groovecoder, line 200: suggestion (non-blocking): maybe someday we can start making modules like `emails/types.py` to hold our types.

emails/utils.py
```diff
@@ -237,21 +254,23 @@ def _start_message_with_headers(subject, from_address, to_address, reply_address
...
     msg_body = MIMEMultipart("alternative")
 
-    # Encode the text and HTML content and set the character encoding.
-    # This step is necessary if you're sending a message with characters
-    # outside the ASCII range.
     if "Text" in message_body:
         body_text = message_body["Text"]["Data"]
-        textpart = MIMEText(body_text.encode(charset), "plain", charset)
+        # Let MIMEText determine if us-ascii encoding will work
+        textpart = MIMEText(body_text, "plain")
```
groovecoder, line 266: praise: good catch on this.

emails/utils.py
```diff
@@ -237,21 +254,23 @@ def _start_message_with_headers(subject, from_address, to_address, reply_address
...
         msg_body.attach(textpart)
     if "Html" in message_body:
         body_html = message_body["Html"]["Data"]
-        htmlpart = MIMEText(body_html.encode(charset), "html", charset)
+        # Our translated strings contain U+2068 (First Strong Isolate) and
+        # U+2069 (Pop Directional Isolate), us-ascii will not work
+        # so save time by suggesting utf-8 encoding
+        htmlpart = MIMEText(body_html, "html", "utf-8")
```
groovecoder, line 273: praise: good catch on this.

groovecoder (approved):

PR #3490, opened 2023-06-03
fix MPP-3155: add total_masks and at_mask_limit to profile endpoint

groovecoder:
This PR fixes #MPP-3155.

How to test:
1. Go to `/api/v1/profiles/`
   * [ ] It should show new `total_masks` and `at_mask_limit` values


emails/models.py
```diff
@@ -195,20 +195,28 @@ def fxa_locale_in_premium_country(self):
     def avatar(self):
         return self.fxa.extra_data.get("avatar")
 
-    @cached_property
+    @property
```
jwhitlock, line 198: 👍 cleanup!

emails/models.py
```diff
@@ -195,20 +195,28 @@ def fxa_locale_in_premium_country(self):
...
 
+    @property
+    def total_masks(self) -> int:
+        return len(self.relay_addresses) + len(self.domain_addresses)
+
+    @property
+    def at_mask_limit(self) -> bool:
+        if self.has_premium:
+            return False
+        return len(self.relay_addresses) == settings.MAX_NUM_FREE_ALIASES
```
jwhitlock, line 218: Please use ``.count()`` instead of ``len(queryset)`` so that the DB return is just a number, not all the `RelayAddress` rows

```suggestion
        return self.relay_addresses.count() == settings.MAX_NUM_FREE_ALIASES
```

jwhitlock (changes requested):
Thanks @groovecoder! This looks like a good small change w/ tests. Some nits, but I think the changes from `len(qs)` to `qs.count()` are required.

I'm slightly bothered that each API call will count the relay addresses twice, but I don't think there is a clean way to avoid it, so I'm willing to wait until profiling to see if it is an issue.

emails/models.py
```diff
@@ -195,20 +195,28 @@ def fxa_locale_in_premium_country(self):
...
     def domain_addresses(self):
-        # TODO: Remove cached_property, to avoid cache invalidation
         return DomainAddress.objects.filter(user=self.user)
 
     @property
     def num_active_address(self):
         return len(self.relay_addresses) + len(self.domain_addresses)
 
+    @property
+    def total_masks(self) -> int:
```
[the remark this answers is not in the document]
groovecoder: heh, good catch. there were no references to `num_active_address` in either the website or add-on code. removing.

emails/models.py
```diff
@@ -195,20 +195,28 @@ def fxa_locale_in_premium_country(self):
...
+        return len(self.relay_addresses) + len(self.domain_addresses)
```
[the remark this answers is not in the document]
groovecoder: Oh yeah I always forget that. There were actually a couple other uses of `len` on a QuerySet, so I changed those to `count()` also.

emails/models.py
```diff
@@ -195,19 +194,26 @@ def fxa_locale_in_premium_country(self):
...
+        ra_count: int = self.relay_addresses.count()
+        da_count: int = self.domain_addresses.count()
+        return ra_count + da_count
+
+    @property
+    def at_mask_limit(self) -> bool:
+        if self.has_premium:
+            return False
+        ra_count: int = self.relay_addresses.count()
+        return ra_count == settings.MAX_NUM_FREE_ALIASES
```
jwhitlock, line 216: Let's make this work for downgraded users too - they had premium, and now they are on the free plan, but have over the limit:

```suggestion
        return ra_count >= settings.MAX_NUM_FREE_ALIASES
```

emails/models.py
```diff
@@ -195,19 +194,26 @@ def fxa_locale_in_premium_country(self):
...
+    @property
     def domain_addresses(self):
-        # TODO: Remove cached_property, to avoid cache invalidation
         return DomainAddress.objects.filter(user=self.user)
 
     @property
-    def num_active_address(self):
-        return len(self.relay_addresses) + len(self.domain_addresses)
+    def total_masks(self) -> int:
+        ra_count: int = self.relay_addresses.count()
```
jwhitlock, line 207: I'm curious, were the annotations needed?

jwhitlock (changes requested):
Looks good, one more small change

emails/models.py
jwhitlock: hmm might be because `self.relay_addresses` is not annotated. Someday...

jwhitlock (approved):

PR #3514, opened 2023-06-06
for MPP-2867: remove FxAToRequest middleware

groovecoder:
This PR supports #MPP-2867.

When our API automatically created a Relay user for Firefox, it created a `SocialAccount` object without any `SocialToken` objects, because it did not go thru the regular OAuth2 flow. This breaks the (old) `FxAToRequest` middleware that tried to use the `SocialToken` to retrieve FXA data and attach it to the request object as an `fxa_account` attribute.

But, we don't have any code that uses the `fxa_account` attribute. (I searched both this code-base and the add-on codebase for `fxa_account` and didn't find a single reference to it.) So, we can remove the `FxAToRequest` middleware completely.

We also checked that after Firefox creates a user without `SocialToken` objects, the django-allauth code automatically gets and stores proper `SocialToken` objects on the correct `SocialAccount` as soon as the user signs in tru the web-based OAuth flow.

How to test:

1. Run `pytest`
   * [ ] Everything still passes
2. Run this branch of code and run thru regular functionality: signing up/in still works, creating/deleting masks still works, changing settings still works, etc.

### DIFF
--- a/privaterelay/fxa_utils.py
+++ b/privaterelay/fxa_utils.py
@@ -68,8 +68,7 @@ def _get_oauth2_session(social_account: SocialAccount) -> OAuth2Session:
     }
 
     # TODO: find out why the auto_refresh and token_updater is not working
-    # and instead we are manually refreshing the token at
-    # FxAToRequest and get_subscription_data_from_fxa
+    # and instead we are manually refreshing the token at get_subscription_data_from_fxa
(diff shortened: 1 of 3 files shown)


say-yawn (approved):
Glad that we are removing unused code. Thanks for the PR!

PR #3513, opened 2023-06-05
For MPP-2867 Refactor authentication helper functions

say-yawn:
# About this PR
Separate out the introspect call to FxA and getting the FxA uid from the response of introspect.

How to test:

## Relay integration to Firefox for existing Relay users (Phase 1) still works
1. Run Firefox with Relay pointed to local/dev (see instructions to do this below)
2. Login into Firefox with FxA user who already has Relay
3. Set `signon.firefoxRelay.feature` to `available`
4. Go to [testing sign-in page](https://mozilla.github.io/form-fill-examples/password_manager/login_and_pw_change_forms.html)
5. Click on username or email field and go through the flow to generate a new alias
6. Check that new alias was created on the user dashboard on Relay as well

# Checklist (Definition of Done)
- [ ] l10n changes have been submitted to the l10n repository, if any.
- [ ] I've added a unit test to test for potential regressions of this bug.
- [ ] I've added or updated relevant docs in the docs/ directory.
- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

# How to run Firefox to point to local Relay
Full document: https://firefox-source-docs.mozilla.org/setup/macos_build.html#building-firefox-on-macos
To run Local Firefox:
```
cd mozilla-unified
hg up -C central
./mach build
./mach run
```

Build and run
```
./mach build
./mach run
```
For faster build
```
./mach build faster
```

To use a specific profile:
```
./mach run --profile {profile path}
```

If you want to change the the Firefox Relay env change the signon.firefoxRelay.base_url value to the url of the environment you want Relay to be in
To get the profile already pointing to local Relay and stage FxA, message me or @groovecoder on Relay team.

PR #3515, opened 2023-06-06
MPP-2867 Register new relay user from firefox

say-yawn:
# New feature description

- Add a new endpoint, `/api/v1/terms-accepted-user/`, so that Firefox can create new users on Relay with a scoped token from Firefox Accounts
   -  Returns appropriate response for happy and unhappy cases
- New field, `created_by`, added on `Profile` object to record where the profile was first created by
   -  `Profile.created_by` is `"firefox_resource"` if the profile is created during resource flow in Firefox integration
   -  `Profile.created_by` is `""` if the profile is created during normal relying party flow on Relay website

# Screen recording

- Happy case new account created:

- Happy case account exists

- Unhappy case account access token is expired

- 

# How to test

- [ ] I've added or updated relevant docs in the docs/ directory

# How to run Firefox to point to local Relay with latest changes for Firefox integration
Full document: https://firefox-source-docs.mozilla.org/setup/macos_build.html#building-firefox-on-macos
1. To run Local Firefox:
```
cd mozilla-unified
hg up -C central
./mach build
./mach run
```

2. Check that local Firefox pop-ups and works.
3. Apply the patch for Firefox integration:
```
moz-phab patch D175565
```

4. Build and run
```
./mach build faster
./mach run
```

5. Apply the following diff on Firefox
```
--- a/toolkit/components/passwordmgr/FirefoxRelay.sys.mjs
+++ b/toolkit/components/passwordmgr/FirefoxRelay.sys.mjs
@@ -23,7 +23,7 @@ const gConfig = (function() {
     undefined
   );
   return {
-    scope: ["https://identity.mozilla.com/apps/relay"],
+    scope: ["profile", "https://identity.mozilla.com/apps/relay"],
     addressesUrl: baseUrl + `relayaddresses/`,
     acceptTermsUrl: baseUrl + `terms-accepted-user/`,
     profilesUrl: baseUrl + `profiles/`,
@@ -68,13 +68,14 @@ async function getRelayTokenAsync() {
 }

 async function hasFirefoxAccountAsync() {
-  if (!lazy.fxAccounts.constructor.config.isProductionConfig()) {
+/*  if (!lazy.fxAccounts.constructor.config.isProductionConfig()) {
     return false;
   }
-
+*/
```
The first change adds the scope `profile` on the scoped token Firefox generates so Relay can use the access token to get profile data from Firefox Accounts. The second change ignores that Firefox Accounts (FxA) is production so we can use staging FxA. 

6. Use a specific profile that @say-yawn or @groovecoder has that sets the locally built browser to use Staging FxA and Stage Relay:
```
./mach run --profile {profile path}
```


api/tests/views_tests.py
```diff
@@ -1,12 +1,25 @@
...
+)
 from django.utils import timezone
 from django.urls import reverse
 from rest_framework.test import APIClient
 
+from api import authentication
+from api.authentication import get_cache_key
+from api.tests.authentication_tests import (
+    _setup_fxa_response,
+    _setup_fxa_response_no_json,
```
groovecoder, line 21: question (non-blocking): do we (i.e., also @jwhitlock ) care about the number of places we're importing functions that start with the `_` character, which usually signals that the function is only used in the same module that defines it?

api/tests/views_tests.py
```diff
@@ -201,3 +214,166 @@ def test_post_relayaddress_flagged_error(free_user, free_api_client) -> None:
         "detail": "Your account is on pause.",
         "error_code": "account_is_paused",
     }
+
+
+class TermsAcceptedUserViewTest(TestCase):
+    def setUp(self):
+        self.factory = RequestFactory()
+        self.path = "/api/v1/terms-accepted-user/"
+        self.fxa_verify_path = authentication.INTROSPECT_TOKEN_URL
```
groovecoder, line 223: note: if you accept the above suggestion, this would change too:
```suggestion
        self.fxa_verify_path = INTROSPECT_TOKEN_URL
```

api/tests/views_tests.py
```diff
@@ -201,3 +214,166 @@ def test_post_relayaddress_flagged_error(free_user, free_api_client) -> None:
...
+        fxa_response = _setup_fxa_response(401, {"error": "401"})
+        not_found_token = "not-found-123"
+        self._setup_client(not_found_token)
+
+        assert cache.get(get_cache_key(not_found_token)) is None
+
+        response = self.client.post(self.path)
+        assert response.status_code == 500
+        assert response.json()["detail"] == "Did not receive a 200 response from FXA."
+        assert responses.assert_call_count(self.fxa_verify_path, 1) is True
```
groovecoder, line 261: question (blocking): In some phone tests, I used `assert rsp.call_count == 1` because it's less verbose. Is there a reason we need to use `assert_call_count` like this?

api/tests/views_tests.py
```diff
@@ -201,3 +214,166 @@ def test_post_relayaddress_flagged_error(free_user, free_api_client) -> None:
...
+        assert cache.get(get_cache_key(not_found_token)) == fxa_response
+
+        # now check that the code does NOT make another fxa request
+        response = self.client.post(self.path)
+        assert response.status_code == 500
+        assert response.json()["detail"] == "Did not receive a 200 response from FXA."
+        assert responses.assert_call_count(self.fxa_verify_path, 1) is True
```
groovecoder, line 268: suggestion (non-blocking): Looks like this block is repeated 5 times. Refactor into a method?

```python
def _assert_cached_response(self, status_code: int, detail: str):
    response = self.client.post(self.path)
    assert response.status_code == status_code
    assert response.json()["detail"] == detail
    assert responses.assert_call_count(self.fxa_verify_path, 1) is True
```

Though this function depends on the `responses` scope in addition to `self` ... so maybe it's not as clean?

api/tests/views_tests.py
```diff
@@ -201,3 +214,166 @@ def test_post_relayaddress_flagged_error(free_user, free_api_client) -> None:
...
+        assert (
+            response.json()["detail"] == "Jsondecodeerror From Fxa Introspect Response"
+        )
+        assert responses.assert_call_count(self.fxa_verify_path, 1) is True
+        assert cache.get(cache_key) == err_fxa_response
+
+        # now check that the 2nd call did NOT make another fxa request
+        response = self.client.post(self.path)
+        assert response.status_code == 500
+        assert response.json()["detail"] == "Previous FXA call failed, wait to retry."
```
groovecoder, line 294: question: Does this pass? The previous post got a `401` and `Jsondecodeerror` while this one gets `500` and `Previous FXA call failed` ?

api/tests/views_tests.py
```diff
@@ -1,12 +1,25 @@
...
+    override_settings,
+    RequestFactory,
+    TestCase,
+)
 from django.utils import timezone
 from django.urls import reverse
 from rest_framework.test import APIClient
 
+from api import authentication
+from api.authentication import get_cache_key
```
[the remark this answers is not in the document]
say-yawn: I'm copying the import style that was recommended in @jwhitlock's comment [here](https://github.com/mozilla/fx-private-relay/pull/3513#discussion_r1218389746). I am open to both styles the only style enforced [according to PEP](https://peps.python.org/pep-0008/#constants) is:
> Constants are usually defined on a module level and written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL.

api/tests/views_tests.py
say-yawn: [PEP recommends](https://peps.python.org/pep-0008/#method-names-and-instance-variables) single underscore in front of methods like:
> Use one leading underscore only for non-public methods and instance variables.

The helper functions are for non-public, for tests only, so using them like this makes sense. However, if we continue to use the helper function outside of where it is defined I think it could be better suited in a separate module that lists the helper functions for a given app. Whichever we decide on, I recommend documenting decision or further conversation in [MPP-3157 Jira ticket](https://mozilla-hub.atlassian.net/browse/MPP-3157).

api/tests/views_tests.py
say-yawn: 😆 this is a following previous tests/copy-paste of code that was introduced in PR #1514 for `authentications_tests.py`. But looking at the `responses` [documentation](https://github.com/getsentry/responses/blob/master/README.rst#assert-based-on-the-exact-url), it also uses the `assert_call_count()` and `... is True`. I think using the library's recommendation is better.

api/tests/views_tests.py
say-yawn: Adding this as a part of clean-up ticket created in [MPP-3162](https://mozilla-hub.atlassian.net/browse/MPP-3162).

api/views/__init__.py
```diff
@@ -163,6 +175,76 @@ def get_queryset(self):
...
+        # Complete social login is called by callback
+        # (see https://github.com/pennersr/django-allauth/blob/77368a84903d32283f07a260819893ec15df78fb/allauth/socialaccount/providers/oauth/views.py#L118)
+        # which is what we are mimicking to
+        # create new SocialAccount, User, and Profile for the new Relay user from Firefox
+        # Since this is a Resource Provider/Server flow and are NOT a Relying Party (RP) of FXA
+        # No social token information is stored (no Social Token object created).
+        complete_social_login(request, social_login)
+        # complete_social_login writes ['account_verified_email', 'user_created', '_auth_user_id', '_auth_user_backend', '_auth_user_hash']
+        # on request.session which sets the cookie because complete_social_login does the "login"
+        # Clear the session since the user did not actually login by logging out
```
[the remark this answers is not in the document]
say-yawn: Much cleaner, thanks!

api/tests/views_tests.py
say-yawn: Yeah the tests passes. Here's a branch I created with before the refactor but this same error. https://github.com/mozilla/fx-private-relay/tree/pre-refactor-fxa-authentication-test

api/tests/views_tests.py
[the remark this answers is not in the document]
groovecoder: Looks like @jwhitlock's meant to import the constant from the module rather than hard-code another string. Let's do this so we're only importing the symbols we need:
```python
from api.authentication import get_cache_key, INTROSPECT_TOKEN_URL
```

api/tests/views_tests.py
groovecoder: Cool, sounds good. Do you mind updating those phones tests too so we're consistent?

api/tests/views_tests.py
say-yawn: Resolved 0c7673f

api/tests/views_tests.py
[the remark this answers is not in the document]
say-yawn: @jwhitlock also confirmed that @groovecoder recommendation is preferred. Resolved 0c7673f

api/tests/views_tests.py
```diff
@@ -229,6 +233,58 @@ def _setup_client(self, token):
...
+        assert response.status_code == 201
+        assert response.data is None
+        # ensure no session cookie was set
+        assert len(response.cookies.keys()) == 1
+        assert "csrftoken" in response.cookies
+        assert responses.assert_call_count(self.fxa_verify_path, 1) is True
+        assert responses.assert_call_count(FXA_PROFILE_URL, 1) is True
+        assert cache.get(cache_key) == fxa_response
+        assert SocialAccount.objects.filter(user__email=email).count() == 1
+        assert Profile.objects.filter(user__email=email).count() == 1
```
groovecoder, line 278: suggestion (non-blocking): these could use `.exists()`:
```python
        assert SocialAccount.objects.filter(user__email=email).exists()
        assert Profile.objects.filter(user__email=email).exists()
```

groovecoder (approved):

PR #3521, opened 2023-06-09
add necessary mypy stubs for allauth

groovecoder:


say-yawn:
I still think we should merge this, did you close this for a reason?

say-yawn:
@jwhitlock can we go over this change to use stub files to make typing better on our project?

jwhitlock:
I'd be OK with adding these stubs this as a stand-alone change instead of a patch to your PR.

PR #3522, opened 2023-06-09
MPP-2117: Change relayed email headers with flag "resender_headers"

jwhitlock:
When the new flag "resender_headers" is active for a user, change the headers of emails relayed via a Relay mask. The biggest change in the "From:" header, which now uses the original sender's email instead of ours. This is similar to the [ReSender](https://datatracker.ietf.org/doc/html/rfc5598#section-5.2) example from RFC-5598, although I'm not fully following that format to keep replies going back to Relay.

The header changes:

* `From:` - Changed to the original sender's email, with changes to the display name
* `Reply-To` - Change to the free / premium reply address, previously the `From:` address
* `Resent-From` - The Relay mask used by the original sender
* `Resent-Sender` - The replay address again

Some examples:

* Email sent from `john@example.com`:
    - Old: `From: "\"john@example.com [via Relay]\"" <replies@relay.firefox.com>`
    - New: `From: "john@example.com [via Relay]" <john@example.com>`
* Email sent from `"jwhitlock" <jwhitlock@example.com>`:
    - Old: `From: "\"jwhitlock@example.com [via Relay]\"' <replies@relay.firefox.com>`
    - New: `From: "jwhitlock [via Relay]" <jwhitlock@example.com>`

The double-quoting is gone, and the original sender's display name is used if available, falling back to the email. A long display name is truncated to about 80 characters.

Additional changes (some which could be moved to a new PR)

* The email sender at the SMTP level is the plain reply address, not an address with a display name
* Remove Flag `new_from_address` and related code
* Add django-waffle type stubs

# How to test

* Deploy code to dev, or enable email locally
* **Before adding the flag** remind yourself how it works:
    - Send an email to a relay random mask
        - [x] The email arrives
        - [x] The email "from" has the reply address
        - [x] *Optional* View original message, note headers - `From:`, `To:`, and `Reply-To:`
        - [x] "Reply" or "Reply All" goes to the reply address, **not** the original sender
        - [x] A reply is delivered to the original sender
    - Send an email to a relay premium mask
        - [ ] The email arrives
        - [ ] The email "from" has the reply address
        - [ ] *Optional* View original message, note headers - `From:`, `To:`, and `Reply-To:`
        - [ ] "Reply" or "Reply All" goes to the reply address, **not** the original sender
        - [ ] A reply is delivered to the original sender     
* Add the `resender_headers` flag for the test user (or everyone)
    - View `/api/v1/runtime_data`
        - [x] `WAFFLE_FLAGS` has `["resender_headers", true]`
    - Send an email to a relay random mask with the flag
        - [x] The email arrives
        - [x] The email "from" has the **original sender's address** (a change)
        - [ ] *Optional* View original message, note headers - `From:` (*changed*), `To:` (*same*), `Reply-To:` (*probably same*), `Resent-From` (*new, mask*), `Resent-Sender` (*new, same as Reply-To*)
        - [ ] "Reply" or "Reply All" goes to the reply address, **not** the original sender
        - [ ] A reply is delivered to the original sender
    - Send an email to a relay premium mask
        - [ ] The email arrives
        - [ ] The email "from" has the reply address
        - [ ] *Optional* View original message, note headers - `From:` (*changed*), `To:` (*same*), `Reply-To:` (*probably same*), `Resent-From` (*new, mask*), `Resent-Sender` (*new, same as Reply-To*)
        - [ ] "Reply" or "Reply All" goes to the reply address, **not** the original sender
        - [ ] A reply is delivered to the original sender


.env-dist
```diff
@@ -29,7 +29,7 @@ SOCKETLABS_SECRET_KEY="dummy-value"
 SOCKETLABS_API_KEY="dummy-value"
 SOCKETLABS_VALIDATION_KEY="dummy-value"
 RELAY_FROM_ADDRESS="relay@127.0.0.1:8000"
-NEW_RELAY_FROM_ADDRESS="new_from@127.0.0.1:8000"
+NEW_RELAY_FROM_ADDRESS="new_from@127.0.0.1:8000"  # TODO: remove
```
jwhitlock, line 32: Removing this line breaks the migrations tests. We can remove it when this code goes to production.

emails/tests/views_tests.py
```diff
@@ -130,38 +132,42 @@ def setUp(self):
...
-                return headers
+                return source, destinations[0], headers
             assert ": " in line
             key, val = line.split(": ", 1)
             assert key not in headers
             headers[key] = val
         raise Exception("Never found message body!")
 
-    def test_single_recipient_sns_notification(self) -> None:
+    def test_single_recipient_sns_notification_legacy_headers(self) -> None:
```
jwhitlock, line 156: These changes are harder to read than I'd like (I'm tempted to open yet another cleanup PR). The pattern is an existing test, with `_legacy_headers` added to the name, that tests the headers without the new flag. The new test, with the old name, tests the headers with the new flag. When we remove the flag, we remove the `_legacy_headers` tests and the `@override_flag` decorators.

emails/utils.py
```diff
@@ -202,17 +203,29 @@ def get_welcome_email(request: HttpRequest, user: User, format: str) -> str:
 
 @time_if_enabled("ses_send_raw_email")
 def ses_send_raw_email(
+    *,
```
jwhitlock, line 206: This syntax means that all the following parameters are keyword-only, allowing me to mix parameters with and without defaults

jwhitlock (commented):
Tests pass!

This PR is uglier than desired. I'm very willing to split into smaller PRs to make review easier. I've added some review notes.

emails/views.py
```diff
@@ -617,17 +620,46 @@ def _sns_message(message_json):
...
-        attachments,
-        mail,
-        address,
+    resender_headers_flag, _ = Flag.objects.get_or_create(
+        name=RESENDER_HEADERS_FLAG_NAME,
+        defaults={"note": "MPP-2117: Filter-friendly mail headers"},
     )
+    resender_headers = (
+        resender_headers_flag.is_active_for_user(user_profile.user)
+        or resender_headers_flag.everyone
```
groovecoder, line 629: this still boggles my mind that `is_active_for_user` doesn't return `True` when the flag is set to everyone.

emails/views.py
```diff
@@ -617,17 +620,46 @@ def _sns_message(message_json):
...
+            resent_from_header=to_address,
+            resent_sender_header=reply_address,
+            from_address=reply_address,
+            to_address=true_address,
+            message_body=message_body,
+            attachments=attachments,
+            mail=mail,
+            address=address,
+        )
+    else:
```
groovecoder, line 649: suggestion (non-blocking): `else` code-smell to refactor this logic out of `_sns_message` and into a separate function with a single `if` block with an early `return` statement.

emails/views.py
```diff
@@ -874,14 +906,16 @@ def _handle_reply(from_address, message_json, to_address):
...
+            from_header=outbound_from_address,
+            to_header=to_address,
+            subject_header=subject,
+            reply_to_header=outbound_from_address,
+            from_address=outbound_from_address,
+            to_address=to_address,
+            message_body=message_body,
+            attachments=attachments,
+            mail=mail,
+            address=address,
```
groovecoder, line 918: praise: oh nice, the change from positional args to keyword args makes this read much easier now!!

mypy_stubs/waffle/defaults.pyi
```diff
@@ -0,0 +1,28 @@
...
+SWITCH_DEFAULT: bool
+READ_FROM_WRITE_DB: bool
+CREATE_MISSING_FLAGS: bool
+CREATE_MISSING_SAMPLES: bool
+CREATE_MISSING_SWITCHES: bool
+LOG_MISSING_FLAGS: Incomplete
+LOG_MISSING_SAMPLES: Incomplete
+LOG_MISSING_SWITCHES: Incomplete
+OVERRIDE: bool
+ENABLE_ADMIN_PAGES: bool
```
groovecoder, line 28: question (non-blocking): when you create stubs, do you keep everything that `mypy` creates for the module, even if it's not used in our code?

emails/views.py
```diff
@@ -617,17 +620,46 @@ def _sns_message(message_json):
...
-        formatted_from_address,
-        to_address,
-        subject,
-        message_body,
-        attachments,
-        mail,
-        address,
+    resender_headers_flag, _ = Flag.objects.get_or_create(
+        name=RESENDER_HEADERS_FLAG_NAME,
+        defaults={"note": "MPP-2117: Filter-friendly mail headers"},
```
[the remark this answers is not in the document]
jwhitlock: What do you have in mind? The rest of the defaults were fine with me.

groovecoder:
Uh-oh. Sending an email from ProtonMail to Gmail with the flag puts the email into the Spam folder and marks it as suspicious.

image

`_dmarc.protonmail.com` policy is strict alignment with the `From:` domain: https://mxtoolbox.com/SuperTool.aspx?action=dmarc%3aprotonmail.com&run=toolpage

From what domain did you test? It might be possible to check a sending domain's dmarc policy for `aspf=s` or `adkim=s`, and DON'T change the `From:` value for those domains? But then end users will receive some Relay emails with the original `From:` and some Relay emails with the Relay `From:` ...

groovecoder (changes requested):
Uh-oh. Sending an email from ProtonMail to Gmail with the flag puts the email into the Spam folder and marks it as suspicious.

image

`_dmarc.protonmail.com` policy is strict alignment with the `From:` domain: https://mxtoolbox.com/SuperTool.aspx?action=dmarc%3aprotonmail.com&run=toolpage

From what domain did you test? It might be possible to check a sending domain's dmarc policy for `aspf=s` or `adkim=s`, and DON'T change the `From:` value for those domains? But then end users will receive some Relay emails with the original `From:` and some Relay emails with the Relay `From:` ...

mypy_stubs/waffle/defaults.pyi
jwhitlock: In this case, I generated but didn't alter `defaults.pyi`. Others like `testutils.pyi` I did clean up. I'm not sure what the policy should be! I think I justified my laziness in that these would be removed next release, but who knows when that will be, `django-waffle` might be abandoned...

jwhitlock:
OK, back to the design phase - the `From:` address will need to be from our domain so it will pass DMARC.

One option is to set the email mask as the sender. An email from `"Your special offer!" <offers@sender.example.com>" to "a1b2c3d4@mozmail.com" becomes `From: "offers@sender.example.com [via Relay] <a1b2c3d4@mozmail.com>`. It at least allows filtering on the alias.

I'm going to break out the non-functional parts of this PR while I work on that change.

PR #3540, opened 2023-06-13
Change email tests for MPP-2117

jwhitlock:
This PR makes some changes to the email tests that will make it easier to test the changes with MPP-2117, which will include changing the interface of the helper function `emails.views.ses_relay_email`. The changes are:

* Update all successful `SNSNotificationTest` tests to also check the source and destination email addresses
* Add global `SEND_RAW_EMAIL_FAILED`, use when `ses_client.send_raw_email` should fail
* For `SNSNotificationValidUserEmailsInS3Test`, switch mock from `emails.views.ses_relay_email` to `ses_client.send_raw_email`, with lower-level errors and successes. This required making attachments an empty list in some mocks.
* For `SnsMessageTest`, also switch mock from `emails.views.ses_relay_email` to `ses_client.send_raw_email`, mocking in the `setUp` instead of with a decorator. Refactor `setUp` to read the message JSON once, and to not store unused objects.

How to test:

* Try to read changes, verify they should be similar.
* Run `pytest`

### DIFF
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -844,45 +844,39 @@ class SNSNotificationValidUserEmailsInS3Test(TestCase):
 
 class SnsMessageTest(TestCase):
     def setUp(self) -> None:
-        self.user = baker.make(User)
-        self.profile = self.user.profile
-        assert self.profile is not None
-        self.sa: SocialAccount = baker.make(
-            SocialAccount, user=self.user, provider="fxa"
-        )
+        self.message_json = json.loads(EMAIL_SNS_BODIES["s3_stored"]["Message"])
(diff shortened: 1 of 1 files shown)


emails/tests/views_tests.py
```diff
@@ -755,7 +769,7 @@ def test_blocked_list_email_in_s3_deleted(
 
     @patch("emails.views._get_text_html_attachments")
     @patch("emails.views.remove_message_from_s3")
-    def test_get_text_hteml_s3_client_error_email_in_s3_not_deleted(
+    def test_get_text_html_s3_client_error_email_in_s3_not_deleted(
```
groovecoder, line 772: praise: 🦅 👁️ 🖊️

groovecoder (approved):

PR #3542, opened 2023-06-14
MPP-2117: Add `flag_is_active_for_task` function, `waffle` type stubs

jwhitlock:
Add a helper function `flag_is_active_for_task("flag_name", user)`, so we don't forget to check `Flag.everyone`. This does similar checks to `flag_is_active(request, "flag_name")`, or the decorator `@waffle_flag("flag_name")`, or the function `Flag.is_active(request)`, except it doesn't take a `request`.

As the docstring says, it does these things that [AbstractBaseFlag.is_active()](https://github.com/django-waffle/django-waffle/blob/v3.0.0/waffle/models.py#L250-L322) does:

* Logs missing flags, if configured
* Creates missing flags, if configured
* Returns default for missing flags
* Checks flag.everyone
* Checks flag.users and flag.groups, if a user is passed
* Returns random results for flag.percent

Since it does not take a request, it does not include:

* Overriding a flag with a query parameter
* Persisting a flag in a cookie (includes percent flags)
* Language-specific overrides (could be added later from user's language)
* Read-only mode for percent flags

In support of this code, I've added type stubs for some of `django-waffle`. I omitted types for code we do not use, an that weren't needed for the code we do use. The next release of `django-waffle` should include types, and we can remove our stubs then. However, the project has the dreaded "Is this still active?" issue, as well as some unmerged PRs for Django 4.2 and Python 3.11, so it may be a while for the next release.

I plan to use `flag_is_active_for_task` in MPP-2117. There are a few existing places where it is useful, and I've made those changes in this PR.

**To Test**:

* `pytest` runs without errors
* Check that code switched to `flag_is_active_for_task` looks correct (some may not be covered by tests)


say-yawn (approved):
Thanks for the PR and nudging me to learn about stubs, generating stubs via stubgen, and Waffle behaviors. @jwhitlock and I also reviewed the flags on Stage and Prod to see if this may disrupt the environments with the clarified definition of active flag since `everyone` set to `False` will short circuit the flag to inactive instead of continuing to check that the flag is active for user.

privaterelay/tests/utils_tests.py
```diff
@@ -20,3 +37,190 @@ def test_get_premium_country_lang(self):
...
+def test_flag_is_active_for_task_with_everyone_on_flag(flag_user: User | None) -> None:
+    Flag.objects.create(name=TEST_FLAG_NAME, everyone=True)
+    assert flag_is_active_in_task(TEST_FLAG_NAME, flag_user)
+
+
+def test_flag_is_active_for_task_with_everyone_off_flag(flag_user: User | None) -> None:
+    Flag.objects.create(name=TEST_FLAG_NAME, everyone=False)
+    assert not flag_is_active_in_task(TEST_FLAG_NAME, flag_user)
+
+
```
say-yawn, line 151: Suggestion (blocking) add a test for `everyone=False`, `user=True`, and confirm `flag_is_active_in_task` returns `False`

say-yawn (changes requested):
Confirmed that Stage does not have `everyone` field set to `False`. However, the `/flags/` page accepts "enable" or "disable" which respectively sets the flag to `everyone` `True` or `False` instead of `None`. Which means using `flag_is_active_in_task` instead of the previous code (below) returns a different `has_free_phones` value:
https://github.com/mozilla/fx-private-relay/blob/5021a583465789f1db5b3251d790eebef2167496/privaterelay/fxa_utils.py#L133-L136
Truth table comparison, old `has_free_phones` logic:
| everyone | user | has_free_phones |
| --- | --- | --- |
| True | False | True |
| Null | True | True |
| False | True | True |

New `has_free_phones` logic using `flag_is_active_in_task`:
| everyone | user | has_free_phones |
| --- | --- | --- |
| True | False | True |
| Null | True | True |
| False | True | False |

Suggestion (blocking): Add logic in `FlagSerializer.validate()` to convert `disable` to `None` https://github.com/mozilla/fx-private-relay/blob/5021a583465789f1db5b3251d790eebef2167496/api/serializers/__init__.py#L172

privaterelay/tests/utils_tests.py
```diff
@@ -20,3 +37,190 @@ def test_get_premium_country_lang(self):
...
+def test_flag_is_active_for_task_everyone_overrides_user(
+    flag_user: User | None,
+) -> None:
+    if flag_user is None:
+        return  # Nothing further to test without a user
+    flag = Flag.objects.create(name=TEST_FLAG_NAME, everyone=False)
+    flag.users.add(flag_user)
+    assert not flag_is_active_in_task(TEST_FLAG_NAME, flag_user)
```
jwhitlock, line 180: @say-yawn this is the test that check everyone=False, user=True returns that the flag is inactive

say-yawn (approved):

PR #3543, opened 2023-06-15
MPP-3192 - Refactor email sending functions

jwhitlock:
For MPP-2117, refactor the email sending functions to take a dictionary of email headers (`OutgoingHeaders`), and to move email assembly to its own function. The code has the same behavior and there are no test changes (except for import order).

Function changes:

* `ses_relay_email` - the function now creates the email message, sends it, creates a `Reply` record, and creates the `HttpResponse`. This was previously done by `ses_send_raw_email`. This more flexible function can be used by `_send_reply_requires_premium_email`, rather than `ses_send_raw_email`. Parameter changes:
   - `from_address` becomes `source_address`, to match `ses_client.send_raw_email`'s `Source` parameter
   - `to_address` becomes `destination_address`, to match `ses_client.send_raw_email`'s `Destinations` parameter
   - new `headers` dict has the email headers, at least `Subject`, `From`, and `To`, but can include `Reply-To`, `In-Reply-To`, and `References` (static checks w/ mypy). These headers can now vary from the params sent to `ses_client.send_raw_email`
   - drop `subject` (in `headers`)
* `ses_send_raw_email` - the function is now focused on calling `ses_client.send_raw_email`. Parameter changes:
   - `from_address` becomes `source_address`, to match `ses_client.send_raw_email`'s `Source` parameter
   - `to_address` becomes `destination_address`, to match `ses_client.send_raw_email`'s `Destinations` parameter
   - new `message` is a `MIMEMultipart` email message
   - drop `subject`, `message_body`, `attachements`, and `reply_address`, part of the `message` param
   - drop `mail` and `address`, functionality moved to `ses_relay_email`
* `_store_reply_record` now takes a `message_id`, rather than the entire SES response to get the `MessageId`.
* new `create_message` function takes `headers`, a `message_body` dict, and optional `attachments` and returns a `MIMEMultipart` email message. This more flexible function can be used by `_build_reply_requires_premium_email`.  
* new `get_reply_to_address` can return the premium (default) or free reply address. This _might_ be the same address these days...

How to test:

* `pytest` passes

### DIFF
--- a/emails/types.py
+++ b/emails/types.py
@@ -1,0 +1,14 @@
+"""Types for email functions"""
+from typing import Any, Literal
+from io import IOBase
+
+# Relay container for email content
+MessageBodyContent = dict[Literal["Charset", "Data"], str]
+MessageBody = dict[Literal["Text", "Html"], MessageBodyContent]
+
+# Attachment path and data stream
+AttachmentPair = tuple[str, IOBase]
+
+# AWS "mail" element in Received notification
+# See https://docs.aws.amazon.com/ses/latest/dg/notification-contents.html
+AWS_MailJSON = dict[str, Any]
--- a/emails/types.py
+++ b/emails/types.py
@@ -18,6 +18,9 @@ OutgoingHeaderName = Literal[
 ]
 OutgoingHeaders = dict[OutgoingHeaderName, str]
 
+# Generic AWS message over SNS - Notification, Bounce, Complaint, ...
+AWS_SNSMessageJSON = dict[str, Any]
+
 # AWS "mail" element in Received notification
 # See https://docs.aws.amazon.com/ses/latest/dg/notification-contents.html
 AWS_MailJSON = dict[str, Any]
--- a/emails/types.py
+++ b/emails/types.py
@@ -9,6 +9,15 @@ MessageBody = dict[Literal["Text", "Html"], MessageBodyContent]
 # Attachment path and data stream
 AttachmentPair = tuple[str, IOBase]
 
+# Headers for outgoing emails
+OutgoingHeaderName = Literal[
+    "From",
+    "Reply-To",
+    "Subject",
+    "To",
+]
+OutgoingHeaders = dict[OutgoingHeaderName, str]
+
 # AWS "mail" element in Received notification
 # See https://docs.aws.amazon.com/ses/latest/dg/notification-contents.html
 AWS_MailJSON = dict[str, Any]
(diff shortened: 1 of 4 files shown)


emails/views.py
```diff
@@ -861,21 +860,26 @@ def _handle_reply(from_address, message_json, to_address):
         # we are returning a 500 so that SNS can retry the email processing
```
say-yawn, line 860: Nit:
```suggestion
        # we are returning a 5XX so that SNS can retry the email processing
```

emails/types.py
```diff
@@ -0,0 +1,28 @@
+"""Types for email functions"""
```
say-yawn, line 1: Very cool that we are adding a `types.py`!

say-yawn (commented):
Changes look good to me. I am wondering if we should use this opportunity to relocate functions in `views.py`. It might be better to move most functions to `utils.py` since most, except `sns_inbound` view, are being only or also used on the `process_emails_from_sqs` job and are not handling requests.

emails/views.py
```diff
@@ -705,49 +712,42 @@ def _strip_localpart_tag(address):
...
+    from_address: str,
+    reply_record: Reply,
+    message_id: str | None,
+    decrypted_metadata: dict[str, Any] | None,
+) -> MIMEMultipart:
+    # If we haven't forwarded a first reply for this user yet, _reply_allowed
     # will forward.  So, tell the user we forwarded it.
     forwarded = not reply_record.address.user.profile.forwarded_first_reply
-    sender = ""
+    sender: str | None = None
```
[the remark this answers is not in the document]
jwhitlock: They aren't quite equivalent... if you pass `ctx = {"sender": None}`, then you'll get an FTL error:

```
FTL exception for locale [], message 'first-reply-forwarded', args {'sender': None}: TypeError("Unsupported external type: sender, <class 'NoneType'>")
```

but it can be set to `""` here, since we protect against passing `None` a few lines down, and that will make it clearer that we're adding types.

jwhitlock (commented):
Thanks @say-yawn!

> I am wondering if we should use this opportunity to relocate functions in `views.py`. It might be better to move most functions to `utils.py` since most, except `sns_inbound view`, are being only or also used on the `process_emails_from_sqs` job and are not handling requests.

I'm open to moving files, but this PR is pretty big already 😅 

There's a few other changes I want to make before moving functions around, such as building a logging context instead of returning HTTP responses. In general, I think stuff in `utils.py` shouldn't return HTTP responses. Also, we may have a more useful name than `emails/utils.py` as a home for the email processing code.

say-yawn (approved):

PR #3602, opened 2023-06-26
MPP-3199 - Split `ProfileTest` into 19 smaller test classes

jwhitlock:
While making changes to the Profile tests, I found that `ProfileTest` had grown to many tests, often prefixed with the `Profile` method or property they were testing.  This PR breaks `ProfileTest` into individual test classes that focus on one `Profile` method or property. I've also added type hints so `mypy` will check the test code.

For example, `ProfileNextEmailTryDateTest` has the tests for the`Profile.next_email_try` property. Because that is implied by the class name, the test names can be shortened, from `test_next_email_try_no_bounces_returns_today` to `test_no_bounces_returns_today`. The other tests use the helper functions `add_hard_bounce` and `add_soft_bounce`, from a base class `ProfileBounceTestCase`, removing some code duplication.

Some tests in `ProfileTest` were for functions outside of `Profile`. For example, `ValidAvailableSubdomainTest` tests `valid_available_subdomain()`, which can involve creating a `Profile`. I've moved tests for `measure_feature_usage`, a signal handler fired before `Profile.save()`, to a new file `emails/tests/signal_tests.py`, so that the file names match.

There are no code changes except tests, and CircleCI will ensure those pass. The review challenge will be to follow the moved tests, which might be easier if the code is reviewed commit by commit.

*Update* - this work is tracked in MPP-3199


88rogikl88 (approved):
####rii#

emails/tests/models_tests.py
```diff
@@ -719,327 +564,292 @@ def test_save_server_storage_false_deletes_data(self):
...
+        return baker.make(
             RelayAddress,
-            user=self.profile.user,
-            description=test_desc,
-            generated_for=test_generated_for,
+            user=user,
+            description=self.TEST_DESCRIPTION,
+            generated_for=self.TEST_GENERATED_FOR,
+            used_on=self.TEST_USED_ON,
             _quantity=4,
```
groovecoder, line 576: praise: TIL `baker.make(_quantity=)` (I know it was in there before but I didn't notice it!)

emails/tests/models_tests.py
```diff
@@ -475,242 +469,93 @@ def test_total_masks(self):
...
+
+    def test_premium_user_returns_False(self) -> None:
+        self.upgrade_to_premium()
         assert self.profile.at_mask_limit is False
-        for _ in list(range(settings.MAX_NUM_FREE_ALIASES)):
+
+    def test_free_user(self) -> None:
+        assert self.profile.at_mask_limit is False
+        for _ in range(settings.MAX_NUM_FREE_ALIASES):
             baker.make(RelayAddress, user=self.profile.user)
```
groovecoder, line 483: suggestion (non-blocking): since we're in here and I just learned about `baker.make(_quantity=)`, maybe we can change this from a for loop?
```suggestion
        baker.make(
            RelayAddress,
            user=self.profile.user,
            _quantity=settings.MAX_NUM_FREE_ALIASES
        )
```

emails/tests/models_tests.py
```diff
@@ -719,327 +564,292 @@ def test_save_server_storage_false_deletes_data(self):
...
-    def test_locale_in_premium_country_returns_False_if_premium_unavailable(self):
-        baker.make(
-            SocialAccount,
-            user=self.profile.user,
-            provider="fxa",
-            extra_data={"locale": "en;q=0.8"},
-        )
+    def test_when_premium_unavailable_returns_False(self) -> None:
+        self.set_fxa_locale("en;q=0.8")
         assert self.profile.fxa_locale_in_premium_country is False
```
groovecoder, line 713: question (blocking): why is premium unavailable in `en` locale in this test?

emails/tests/signals_tests.py
```diff
@@ -0,0 +1,72 @@
...
+class MeasureFeatureUsageSignalTest(TestCase):
+    """Test measure_feature_usage signal handler"""
+
+    def setUp(self) -> None:
+        user = make_free_test_user()
+        self.profile = user.profile
+
+        patcher_incr = patch("emails.signals.incr_if_enabled")
+        self.mocked_incr = patcher_incr.start()
+        self.addCleanup(patcher_incr.stop)
```
groovecoder, line 19: praise: TIL `addCleanup` !

groovecoder (changes requested):
Just 1 blocking question. If we have a good answer for it, I'll approve and add to merge queue. Great work!

jwhitlock:
I've rebased and added a few more commits to use `baker.make(..., _quantity=#)` in more places where it made sense. I also expanded the premium test to actually test that the user can create more than `MAX_NUM_FREE_ALIASES`.

emails/tests/models_tests.py
```diff
@@ -185,27 +186,30 @@ def test_create_assigns_to_user(self):
 
     @override_settings(MAX_NUM_FREE_ALIASES=5, MAX_ADDRESS_CREATION_PER_DAY=10)
     def test_create_has_limit(self) -> None:
-        for _ in range(10):
-            RelayAddress.objects.create(user=self.premium_user_profile.user)
+        baker.make(RelayAddress, user=self.premium_user, _quantity=10)
```
groovecoder, line 189: suggestion (blocking):
```suggestion
        baker.make(RelayAddress, user=self.premium_user, _quantity=MAX_ADDRESS_CREATION_PER_DAY)
```

emails/tests/models_tests.py
groovecoder: That's fine. This PR isn't about cleaning up EVERY sloppy test, so yeah - let's land this nice re-organizing work.

groovecoder (approved):

PR #3606, opened 2023-06-28
Profile and optimize fx endpoints mpp 3021

groovecoder:
This PR is for MPP-3021. It includes a few optimizations for the API endpoints that are used during the new Firefox integration.

https://github.com/mozilla/fx-private-relay/pull/3606/commits/2eaf759acb4f66fd8115fe6e19b17c4e9044a125 is the most significant optimization so if there's an issue with any of the others, we can remove them so long as we keep this one.

How to test:

### DIFF
--- a/api/authentication.py
+++ b/api/authentication.py
@@ -134,8 +134,11 @@ class FxaTokenAuthentication(BaseAuthentication):
                 use_cache = True
         fxa_uid = get_fxa_uid_from_oauth_token(token, use_cache)
         try:
-            sa = SocialAccount.objects.get(uid=fxa_uid, provider="fxa")
-        except SocialAccount.DoesNotExist:
+            # MPP-3021: select_related user object to save DB query
+            sa = SocialAccount.objects.filter(
+                uid=fxa_uid, provider="fxa"
+            ).select_related("user")[0]
+        except IndexError:
             raise PermissionDenied("Authenticated user does not have a Relay account.")
         user = sa.user
 
--- a/emails/models.py
+++ b/emails/models.py
@@ -682,7 +682,10 @@ class RelayAddress(models.Model):
 def check_user_can_make_another_address(profile: Profile) -> None:
     if profile.is_flagged:
         raise AccountIsPausedException()
-    if profile.at_max_free_aliases and not profile.has_premium:
+    # MPP-3021: return early for premim users to avoid at_max_free_aliases DB query
+    if profile.has_premium:
+        return
+    if profile.at_max_free_aliases:
         raise RelayAddrFreeTierLimitException()
 
 
--- a/emails/models.py
+++ b/emails/models.py
@@ -682,7 +682,7 @@ class RelayAddress(models.Model):
 def check_user_can_make_another_address(profile: Profile) -> None:
     if profile.is_flagged:
         raise AccountIsPausedException()
-    # MPP-3021: return early for premim users to avoid at_max_free_aliases DB query
+    # MPP-3021: return early for premium users to avoid at_max_free_aliases DB query
     if profile.has_premium:
         return
     if profile.at_max_free_aliases:
--- a/emails/templates/emails/first_time_user.html
+++ b/emails/templates/emails/first_time_user.html
@@ -2,8 +2,10 @@
   Note that Django only loads strings from some Fluent files.
   See privaterelay/ftl_bundles.py.
 {% endcomment %}
+{% load cache %}
 {% load ftl %}
 {% load email_extras %}
+{% cache 86400 welcome_email_html language %}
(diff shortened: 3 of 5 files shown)


privaterelay/settings.py
```diff
@@ -867,6 +867,12 @@ def _get_initial_middleware() -> list[str]:
     }
     SESSION_ENGINE = "django.contrib.sessions.backends.cache"
     SESSION_CACHE_ALIAS = "default"
+if RELAY_CHANNEL == "local":
+    CACHES = {
+        "default": {
+            "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
+        }
+    }
```
say-yawn, line 875: (question) Are we using this instead of Redis on local? If so this is better set as `elif`

privaterelay/settings.py
```diff
@@ -1171,6 +1177,7 @@ def set_index_cache_control_headers(
 if USE_SILK:
     SILKY_PYTHON_PROFILER = True
     SILKY_PYTHON_PROFILER_BINARY = True
+    SILKY_PYTHON_PROFILER_RESULT_PATH = ".silk-profiler"
```
say-yawn, line 1180: (question) Can you elaborate on why this config value was added? Also, if there is an appropriate documentation to go with these configs that would be nice.

(suggestion, non-blocking) Looking through the code to enable `silk` on my local I have to have `DEBUG` and `HAS_SILK` be true. To get `HAS_SILK` I need to install `silk`. Perhaps we should consider creating a `requirements-dev.txt` file so that we are not manually reading code to see what packages we need to install?

emails/templates/emails/first_time_user.html
```diff
@@ -2,8 +2,10 @@
   Note that Django only loads strings from some Fluent files.
   See privaterelay/ftl_bundles.py.
 {% endcomment %}
+{% load cache %}
```
say-yawn, line 5: TIL [Template Fragment Caching](https://docs.djangoproject.com/en/4.2/topics/cache/#template-fragment-caching) 😎

emails/templates/emails/first_time_user.html
```diff
@@ -2,8 +2,10 @@
...
 {% load ftl %}
 {% load email_extras %}
+{% cache 86400 welcome_email_html language %}
```
say-yawn, line 8: (suggestion, non-blocking) Consider setting a cache timeout variable in the context to be used in the templates so we can use a single variable and expect same cache behavior for caching email templates.

emails/models.py
```diff
@@ -682,7 +682,10 @@ def full_address(self):
 def check_user_can_make_another_address(profile: Profile) -> None:
     if profile.is_flagged:
         raise AccountIsPausedException()
-    if profile.at_max_free_aliases and not profile.has_premium:
+    # MPP-3021: return early for premim users to avoid at_max_free_aliases DB query
```
say-yawn, line 685:
```suggestion
    # MPP-3021: return early for premium users to avoid at_max_free_aliases DB query
```
(compliment) I like this practice of adding the Jira issue number in the comments. While I think it makes it easier to understand code, I also think it makes the code messy. Let's revisit how and when we use the Jira issue on comments versus commits on [MPP-3216](https://mozilla-hub.atlassian.net/browse/MPP-3216).

api/authentication.py
```diff
@@ -134,8 +134,11 @@ def authenticate(self, request):
                 use_cache = True
         fxa_uid = get_fxa_uid_from_oauth_token(token, use_cache)
         try:
-            sa = SocialAccount.objects.get(uid=fxa_uid, provider="fxa")
-        except SocialAccount.DoesNotExist:
+            # MPP-3021: select_related user object to save DB query
+            sa = SocialAccount.objects.filter(
+                uid=fxa_uid, provider="fxa"
+            ).select_related("user")[0]
+        except IndexError:
```
say-yawn, line 141: (compliment) NICE!

say-yawn (approved):
Lots of non-blocking comments. Would like to review them before we merge this. Logic looks good though.

emails/templates/emails/first_time_user.html
groovecoder: This is how I learned it too!

privaterelay/settings.py
groovecoder: This change broke CI and I'd rather not debug it in this PR.

emails/templates/emails/first_time_user.html
groovecoder: I created https://mozilla-hub.atlassian.net/browse/MPP-3220 for a follow-up.

emails/models.py
groovecoder: fixed typo.

PR #3609, opened 2023-06-29
MPP-3200 - Refactor Relay plan data

jwhitlock:
This PR moves the plan data from `privaterelay/settings.py` to a new file `privaterelay/plans.py`, adds some functions to access the data, and refactors the raw data structures to reduce duplication and support static type checking.

The public interface is:

* `get_premium_country_language_mapping()` - Get mapping of countries → languages → plan period → Stripe data for the countries that can purchase a premium plan (unlimited masks, custom subdomain).
* `get_premium_countries()` - Get the countries that can purchase a premium plan
* `get_phone_country_language_mapping()` -> Get the mapping for the phone plan (premium plus a phone mask)
* `get_bundle_country_language_mapping()` -> Get the mapping for the bundle plan (premium plus phone mask plus Mozilla VPN)

The tests in `privaterelay/tests/plan_tests.py` verify that none of the data changed during the refactor, and using the public functions returns the same data that accessing the settings did (for example, `get_phone_country_language_mapping()` returns the same data that was in `settings.PHONE_PLAN_COUNTRY_LANG_MAPPING`)

The data is moved to:
* `_STRIPE_PLAN_DATA` - the same data that was in `settings.PERIODICAL_PREMIUM_PLAN_ID_MATRIX`, `settings.PHONE_PLAN_ID_MATRIX`, and `settings.BUNDLE_PLAN_ID_MATRIX`, but re-organized for less duplication, to use ISO 3166 and RFC 5646 capitalization, and with type hints
* `_RELAY_PLANS_BY_COUNTRY_AND_LANGUAGE` - the same data that was in `settings.PERIODICAL_PREMIUM_PLAN_COUNTRY_LANG_MAPPING`, `settings.PHONE_PLAN_COUNTRY_LANG_MAPPING`, and `settings.BUNDLE_PLAN_COUNTRY_LANG_MAPPING`, except not statically joined to the Stripe data.

The remaining code provided static typing hints for these data structures, as well as helper functions that create the mappings and cache them on first use. The US price IDs are part of the cache key, so it is possible to test that the settings like `settings.PREMIUM_PLAN_ID_US_MONTHLY` change the price IDs.

# How to Test

- [x] I've added unit tests to test for potential regressions.


privaterelay/plans.py
```diff
@@ -0,0 +1,610 @@
...
+
+
+#
+# Data
+#
+
+# Selected Stripe data
+# The "source of truth" is the Stripe data, this copy is used for upsell views
+# and directing users to the correct Stripe purchase page.
+_STRIPE_PLAN_DATA: "_StripePlanData" = {
```
groovecoder, line 100: question (blocking): why are we using strings for some type annotations instead of symbols? I don't see anything in these type definitions that need run-time evaluation. can we move the type declarations up above the code that references them so we can use symbols instead of strings?

privaterelay/plans.py
```diff
@@ -0,0 +1,610 @@
...
+
+from django.conf import settings
+
+#
+# Public functions
+#
+
+
+def get_premium_country_language_mapping(
+    eu_country_expansion: bool | None,
```
groovecoder, line 69: note (non-blocking): might have to change this to an `int` if we do the eu expansion in 2 steps instead of a single on/off step.

privaterelay/tests/plans_tests.py
```diff
@@ -0,0 +1,302 @@
...
+    "gr",
+    "lv",
+    "lt",
+    "lu",
+    "mt",
+    "pt",
+    "sk",
+    "si",
+]
+_NON_PREMIUM_COUNTRY = "mx"
```
groovecoder, line 60: note (non-blocking): 😢 🇲🇽

groovecoder (changes requested):

privaterelay/tests/plans_tests.py
```diff
@@ -0,0 +1,302 @@
...
+        ("lv", "lv", "lv"),
+        ("lt", "lt", "lt"),
+        ("lu", "en", "en-LU"),
+        ("mt", "en", "en-MT"),
+        ("pt", "pt", "pt"),
+        ("sk", "sk", "sk"),
+        ("si", "sl", "sl"),
+    ),
+)
+@pytest.mark.parametrize("eu_expansion", (True, None))
```
groovecoder, line 179: praise: wow impressive use of `parametrize`!
```
73 passed in 0.27s
```

groovecoder (changes requested):
Whoops I left some other comments as individual comments. Only 1 blocking suggestion.

jwhitlock:
Updates for code review:

* Copy-edit on the docstrings, made them more consistent
* Sorted the plans by the country code, rather than the original order in settings
* Moved types around so less annotations are strings. Continued to define data structure types in top-down format with string annotations.

groovecoder (approved):

PR #3634, opened 2023-07-06
Windows instruction additions in Step 2 of the readme

rafeerahman:
Added a change to the local installation instructions for Windows in Step 2 because 'source' is not a command in Windows. Also added a note for an issue I ran into.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,17 +37,22 @@ Please refer to our [coding standards](docs/coding-standards.md) for code styles
 
 2. Create and activate a virtual environment:
 
-   Unix based systems: 
+   Unix based systems:
+
    ```sh
    virtualenv env
    source env/bin/activate
    ```
-   Windows: 
+
+   Windows:
+
    ```sh
    python -m venv env
-   source env/Scripts/activate
+   .\env\Scripts\activate
    ```
-   
+
+   Note: If you're running on Windows and get an error message stating that executing scripts are disabled on your computer, go into the Windows powershell and type `Set-ExecutionPolicy -Scope CurrentUser -ExecutionPolicy Unrestricted`, then try again.
+
 3. Install Python and Node requirements:
 
    ```sh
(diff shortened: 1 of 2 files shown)


groovecoder (changes requested):
@rafeerahman - do you feel comfortable changing this branch so it's rebased on `main` instead of using a merge commit from `main`?

README.md
```diff
@@ -37,17 +37,24 @@ Please refer to our [coding standards](docs/coding-standards.md) for code styles
...
    source env/bin/activate
    ```
-   Windows: 
+
+   Windows:
+
    ```sh
    python -m venv env
-   source env/Scripts/activate
+   source env/bin/activate
```
lloan, line 51:
```suggestion
   source env/Scripts/activate
```

This would stay the same as the original - its the correct path. Only difference is Git Bash users do use `source` to set the source.

lloan (changes requested):
Minor change and we should be good to go! Great job!

rafeerahman:
> @rafeerahman - do you feel comfortable changing this branch so it's rebased on `main` instead of using a merge commit from `main`?

Rebased to main and pushed, was this correct?

README.md
rafeerahman: Changed, thanks for checking

say-yawn:
It looks like something went wrong on your rebase. Consider creating a new branch out of the latest from `main`, cherry-pick your changes to the new branch (effectively simulating the rebase), and then git push force.

rafeerahman:
> It looks like something went wrong on your rebase. Consider creating a new branch out of the latest from `main`, cherry-pick your changes to the new branch (effectively simulating the rebase), and then git push force.

Resolved now.

lloan (approved):
Thanks for making the change. 

@groovecoder are we good with adding locales to .gitignore? 

Thanks @rafeerahman !

groovecoder (approved):

PR #3648, opened 2023-07-11
Update @testing-library/dom and @testing-library/react

jwhitlock:
Combine these dependabot PRs, to get the tests passing.

* PR #3119 
* PR #3562

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -39,7 +39,7 @@
   "devDependencies": {
     "@next/bundle-analyzer": "^13.4.9",
     "@next/eslint-plugin-next": "^13.4.9",
-    "@testing-library/dom": "^9.2.0",
+    "@testing-library/dom": "^9.3.1",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^14.0.0",
     "@testing-library/user-event": "^14.4.3",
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -39,9 +39,9 @@
   "devDependencies": {
     "@next/bundle-analyzer": "^13.4.9",
     "@next/eslint-plugin-next": "^13.4.9",
-    "@testing-library/dom": "^9.2.0",
+    "@testing-library/dom": "^9.3.1",
```
jwhitlock, line 42: I wonder if we should remove `@testing-library/dom` from `devDependencies`, since `@testing-library/react` pulls it in as a dependency.

lloan: From [Next discussion boards](https://github.com/vercel/next.js/discussions/14000): 


The short answer is: it doesn't matter, unless you are only installing dependencies (with npm/yarn install --production). For consistency, I add all build/dev/test tools in devDependencies, and the rest of the stuff in dependencies.

The situation where you might face problems is when you are only installing the dependencies and not devDependencies. Keep in mind Next.js will run your code both in the client and the server, so if you have a server-rendered page, and the app does not have less listed in the dependencies list, then your server will crash when trying to render the page.

PR #3677, opened 2023-07-17
MPP-3249: Add `./manage.py delete-phone-data`

jwhitlock:
For MPP-3249, add a command `./manage.py delete_phone_data <fxa-id>`, to replace the manual SQL commands on [How To: Reset a User's Real Phone Number and Phone Mask](https://mozilla-hub.atlassian.net/wiki/spaces/SECPRV/pages/193332682/How+To+Reset+a+User+s+Real+Phone+Number+and+Phone+Mask).

## How to test

In an environment with phones enabled, try the command on a known and unknown FxA ID.

## Checklist (Definition of Done)
- ~[ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.~
- ~[ ] I've added or updated relevant docs in the docs/ directory~
- ~[ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.~
- ~[ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).~
- ~[ ] l10n changes have been submitted to the l10n repository, if any.~


phones/management/commands/delete_phone_data.py
```diff
@@ -0,0 +1,118 @@
...
+
+        return cls(
+            fxa=fxa,
+            real_phone=real_phone,
+            relay_phone=relay_phone,
+            inbound_contact_count=inbound_contact_count,
+        )
+
+    @property
+    def has_phone_data(self) -> bool:
```
groovecoder, line 47: suggestion (non-blocking): `has_phone_data` seems a bit ambiguous on a `PhoneData` class. Maybe this should be `has_real_phone`?

groovecoder: Or if it's meant to return `True` so long as the user has *ANY* phone data, maybe `has_data`?

phones/management/commands/delete_phone_data.py
```diff
@@ -0,0 +1,118 @@
...
+    help = "Deletes phone data, so a user can re-enroll."
+
+    def add_arguments(self, parser: ArgumentParser) -> None:
+        parser.add_argument("fxa_id", help="The user's FxA ID")
+        parser.add_argument(
+            "-n",
+            "--dry-run",
+            action="store_true",
+            help="Report but do not delete phone data",
+        )
```
groovecoder, line 96: suggestion (non-blocking): Consider the UX here, would it be safer to make "dry run" the default behavior and then make the user have to pass a scary-named flag to delete the data? Like `python manage.py delete_phone_data --DELETE <fxa_id>`

Or, since the `delete_phone_data` command is already scary sounding, can we add a confirmation prompt to the command?
```
python manage.py delete_phone_data <fxa_uid>
> Found a matching user: ...
...
> Please type the real phone number to confirm you want to delete:
```

phones/tests/mgmt_delete_phone_data_tests.py
```diff
@@ -0,0 +1,272 @@
...
+    """Override settings for tests"""
+    settings.SUBSCRIPTIONS_WITH_PHONE = ["PHONE_SUBSCRIPTION"]
+    return settings
+
+
+@pytest.fixture(autouse=True)
+def mock_twilio_client() -> Iterator[Mock]:
+    """Mock PhonesConfig with a mock twilio client"""
+    with patch("phones.apps.PhonesConfig.twilio_client") as mock_twilio_client:
+        yield mock_twilio_client
```
groovecoder, line 35: question (non-blocking): interesting. does this management command trigger a Twilio API call?

phones/tests/mgmt_delete_phone_data_tests.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+@pytest.fixture
+def phone_user(db: None, test_settings: SettingsWrapper) -> User:
```
groovecoder, line 39: suggestion (blocking): let's reconcile this with [the other utility functions in `model_tests.py`](https://github.com/mozilla/fx-private-relay/blob/eae992837a2c0a007591ec2803d2548919b635ec/phones/tests/models_tests.py#L77-L99).

phones/tests/mgmt_delete_phone_data_tests.py
```diff
@@ -0,0 +1,272 @@
...
+    assert InboundContact.objects.filter(relay_number=relay_number).count() == 2
+
+    stdout = StringIO()
+    call_command(THE_COMMAND, phone_user.profile.fxa.uid, stdout=stdout)
+
+    expected_stdout = f"""\
+Found a matching user:
+
+* FxA ID: decd14f50fce4865b0fd06afcd9d2c4d
+* User ID: {phone_user.id}
```
groovecoder, line 107: question (non-blocking): Why reference the `phone_user` object for this value but not the others? E.g., why not this?
```python
* FxA ID: {phone_user.profile.fxa.uid}
* User ID: {phone_user.id}
* Email: {phone_useremail}
* etc.
```

groovecoder (changes requested):

phones/tests/mgmt_delete_phone_data_tests.py
jwhitlock: No, it's creating the test data. Creating a `RealPhone` calls [realphone_post_save](https://github.com/mozilla/fx-private-relay/blob/62db400cdd4ad1a57627d8d06fe2536ba0bfcc74/phones/models.py#L192), which calls Twilio to send a text message, and creating a `RelayNumber` will call [.save](https://github.com/mozilla/fx-private-relay/blob/62db400cdd4ad1a57627d8d06fe2536ba0bfcc74/phones/models.py#L263) which registers the number with Twilio and adds US numbers to a message pool.

I think mocking will only work in the Twilio case, not the Intelliquint case, so I think I need to fix that stuff first.

phones/tests/mgmt_delete_phone_data_tests.py
jwhitlock: because in the test cases, those values do not change, but the user ID changes depending on what tests ran before this one.

jwhitlock (commented):
Sorry for the delay in reading your review. I'll make some changes, but the feedback makes me think a management command is the wrong strategy, and we should skip ahead to a customer support portal, while we improve the testing situation in phones

phones/management/commands/delete_phone_data.py
groovecoder: Yeah that sounds good especially if we're still running this command on a user-by-user basis.

Can the command-line flag to skip confirmation be `-f --force` to match up with other CLI ux?

phones/tests/mgmt_delete_phone_data_tests.py
groovecoder: Oh no worries. Just curios.

phones/tests/mgmt_delete_phone_data_tests.py
groovecoder: Got it, thanks for the clarification.

groovecoder:
Hmm ... can we avoid falling into [the "rethinking" valley](https://imgs.xkcd.com/comics/automation.png) here? [The sacred text](image) says if you're spending 5-10 minutes 2-3 times per week, we should only spend 1-5 days on this level of automation. This PR is 3 days old and improves the situation a bit already, right?

phones/tests/mgmt_delete_phone_data_tests.py
jwhitlock: I'm adding `settings.PHONES_NO_CLIENT_CALLS_IN_TEST`, and using it to skip calling Twilio or IQ clients for code like this that is not testing the client interaction. I've got bigger ideas for later.

phones/management/commands/delete_phone_data.py
jwhitlock: oops went with `--yes`, quick change to `--force` coming up

jwhitlock:
I agree, some smaller changes can get this ready to merge, bigger changes can wait. Ready for re-review.

phones/management/commands/delete_phone_data.py
```diff
@@ -0,0 +1,129 @@
...
+        self.stdout.write(report)
+
+        if not data.has_data:
+            return "User has NO PHONE DATA to delete."
+
+        confirmed = skip_confirmation or self.confirm()
+        if confirmed:
+            data.reset()
+            return "Deleted user's phone data."
+        return "User still has their phone data... FOR NOW!"
```
groovecoder, line 43: 😄

groovecoder (approved):

groovecoder:
Since we're touching more `phones` code than just the management command, let's be sure to spot-check phone functionality on dev server before this goes to stage & prod.

PR #3678, opened 2023-07-17
MPP-3236: Adjust languages for countries

jwhitlock:
For MPP-3236, adjust the languages for countries:

* Cyprus (`CY`): Greek (`el`) instead of English (`en`)
* Croatia (`HR`): Croatian (`hr`) instead of English (`en`)
* Luxembourg (`LU`): French (`fr`)~, then German (`de`) and English (`en`),~ instead of ~just~ English (`en`)
* Malta (`MT`): English (`en`) ~then Maltese (`mt`), instead of just English~

My investigation of the code suggests these language maps don't actually matter, but removing them will take longer than this simple change, and can be done in a future PR.

# How To Test
* Languages are close to those on MPP-3213


privaterelay/plans.py
say-yawn, line 1: (suggestion, non-blocking) If you are manually adding the data to the variables can you please create a document on which vars and how you add the additional entries to the vars so other engineers can also do this? This can be a separate fast follow work.

privaterelay/plans.py
```diff
@@ -543,17 +545,24 @@ def get_bundle_country_language_mapping() -> PlanCountryLangMapping:
...
+        "lu": {  # Luxembourg
+            "fr": "LU",
+            "de": "LU",
+            "en": "LU",
+        },
         "lv": {"lv": "LV"},  # Latvia
-        "mt": {"en": "MT"},  # Malta (language may change from English)
+        "mt": {  # Malta
+            "en": "MT",
+            "mt": "MT",
```
say-yawn, line 564: (comment) Interesting that while these are both official language English is higher up than Maltese. Should we do that for other languages too that have multiple official languages? But defaulting to English to ensure a more consistent experience--we write everything in English first--makes sense to me.

say-yawn (approved):
Thanks for creating the PR to provide better translations for people in Europe.
I verified that the languages being added/updated in the value of `"premium_eu_expansion"` key for `_RELAY_PLANS_BY_COUNTRY_AND_LANGUAGE` is also in the `LanguageStr` variable in plans.py

privaterelay/plans.py
jwhitlock: Looking at the front-end code, it is only used to distinguish when there are multiple plans for a country, so it doesn't make sense to have several languages map to the same plan, other than expanding the data transfer. I'm going to reduce these to one item.

PR #3680, opened 2023-07-18
Update to Prettier 3.0.0, apply new formatting

jwhitlock:
This includes the upgrade to prettier 3.0.0 on PR #3663, and applies the new rules with `npx prettier --write ./src`. The changes are:

* [trailing-commas](https://prettier.io/docs/en/options.html#trailing-commas) changed from `es5` to `all`, affecting `.ts` and `.tsx` files
* Multi-line `.scss` rules had the first line inline, and now it is indented with the rest of the lines.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -3,3 +3,5 @@
 5fe13eb71813d00ba32eebb6adc4021f6e9c4d66
 # Update to Black 23
 870af2ac817aca795d1bb20184a9ad90c82cef4c
+# 2023-07-18 - Update to prettier 3.0.0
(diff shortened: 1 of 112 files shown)


stephaniehobson (approved):

stephaniehobson:
Gave it an R+ so I'm not blocking but I did not do any actual reviewing. But maybe it's not blocked if I don't R+ since the button is green now. 🤔

lloan (approved):
I went ahead and checked out https://prettier.io/blog/2023/07/05/3.0.0.html#breaking-changes to make I had an idea of all the breaking changes. It looks good from what I can see. Thanks a ton for this!

PR #3682, opened 2023-07-19
MPP-3264: Re-implement guessing the country from the Accept-Langugage header

jwhitlock:
As part of MPP-3264, this PR improves how an `Accept-Language` header is turned into a country, mostly to see if premium plans are allowed for the user.

In some cases, the country is guessed from the `Accept-Language` header:

* When the infrastructure does not set the `X-Client-Region` header
* When a background task is using the `Accept-Language` header from Firefox Accounts (FxA) signup.

This was handled in two ways:
* Web request code used `privaterelay.utils.get_premium_country_lang`, which returned a country and language, and discarded the language.
* Background code used `emails.models.Profile.fxa_locale_in_premium_country`, which duplicated some of the logic in `privaterelay.utils.get_premium_country_lang`

This code replaces `get_premium_country_lang` with `guess_country_from_accept_lang`, and uses the same logic in both places. `guess_country_from_accept_lang` only returns a lower-cased country code.

The new function fixes several issues with language to country mapping for existing premium and EU expansion countries, when the language code is not identical to the lower-cased country code, such as:

* Czech (`cs`) returns the Czech Republic (`cz`), instead of invalid country code `cs`
* Danish (`da`) returns Denmark (`dk`), instead of invalid country code `da`
* Estonian (`et`) returns Estonia (`ee`), instead of Ethiopia (`et`)
* Greek (`el`) returns Greece (`gr`), instead of invalid country code `el`
* Luxembourgish (`lb`) return Luxembourg (`lu`), instead of Lebanon (`lb`)
* Slovenian (`sl`) returns Slovenia (`si`), instead of Sierra Leone (`sl`)
* Welsh (`cy`) returns the United Kingdom (`gb`) rather than Cyprus (`cy`)

It also handles cases not handled by the previous code, such as:

* A language tag with a quality specification, such as `de; q=1.0` (should be Germany (`de`), was invalid `de; q=1.0`)
* A Language-Script tag, such as `zh-Hant` (treated as China (`cn`), was invalid code `hant`)
* A Language-Variant tag, such as `sl-rozaj` (treated as Slovenia (`si`), was invalid code `rozaj`)
* A Language-Script-Region tag, such as `zh-Hant-TW` (treated as Taiwan (`tw`), was invalid code `hant`)
* An "any language" tag, such as `*` (was invalid code `*`, now treated as an error)
* An invalid language tag, such as `**` (was invalid code `**`, now treated as an error)

The test cases, which drove the development of the function, are pulled from several sources:

* Previous tests for `get_premium_country_lang`
* Django's tests for [test_parse_spec_http_header](https://github.com/django/django/blob/848fe70f3ee6dc151831251076dc0a4a9db5a0ec/tests/i18n/tests.py#L1332), which include good and bad `Accept-Language` headers and good headers with quality strings (`en; q=0.8`)
* Firefox's default `Accept-Language` headers. These are set in Pontoon as the translation for "en-US, en" in Firefox, `intl.properties`, such as the [German translation](https://pontoon.mozilla.org/de/firefox/toolkit/chrome/global/intl.properties/?string=81017) to `de, en-US, en`. The full list is in the `Locales` tab, and can be downloaded (mostly) as:
    ```sh
    curl 'https://pontoon.mozilla.org/other-locales/?entity=81017&locale=de' \
    -H 'x-requested-with: XMLHttpRequest'
    ```
   This JSON includes all but the German (`de`) translation, which can be manually added. I transformed it into test cases, then selected the expected country from CLDR data (more on that later)
* Examples of language tags from [RFC 5646, "Tags for Identifying Languages"]((https://www.rfc-editor.org/rfc/rfc5646.html), [Appendix A, "Examples of Language Tags (Informative)"](https://www.rfc-editor.org/rfc/rfc5646.html#appendix-A), which include many different formats and corner cases not previously handled.

The mapping from languages to countries `_PRIMARY_LANGUAGE_TO_COUNTRY` comes from the Unicode consortium's [Common Locale Data Repository (CLDR) v43.0 Supplemental Data](https://www.unicode.org/cldr/charts/43/supplemental/index.html). The official data is in XML, but the same data is exported as JSON at https://github.com/unicode-org/cldr-json, and as [human-readable HTML](https://www.unicode.org/cldr/charts/42/supplemental/language_territory_information.html). I processed the file [territoryInfo.json](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-core/supplemental/territoryInfo.json), limited to the 135 [Firefox languages](https://pontoon.mozilla.org/es-ES/firefox/toolkit/chrome/global/intl.properties/?string=81017) for `intl.properties`, to find the country with the highest population of speakers of that language. This worked as expected for most countries, and helped find the cases where the language code does not match the lower-cased country code (for example, Japanese `ja` to Japan `jp`). It also included official languages (Luxembourgish `lb` to Luxembourg `lu`), which may improve country detection for the EU expansion. 

I tweaked one entry. Spanish (`es`) is mapped to Spain (`es`), even though there are more Spanish speakers in Mexico, and I expect Mexicans to use `es-MX` as their primary language.

There were some additional special cases which might be rare in the wild, but required special handling.

* `ca-valencia`, from Firefox's [Catalan (Valencian)](https://pontoon.mozilla.org/ca-valencia/firefox/toolkit/chrome/global/intl.properties/?string=81017). Catalan (`ca`) has the most speakers in France (`fr`), but this Language-Script combination for the [Valencian language](https://en.wikipedia.org/wiki/Valencian_language) implies Spain (`es`).
* `es-419`, a special test case of a [UN M49](https://en.wikipedia.org/wiki/UN_M49) code for [Latin America and the Caribbean](https://en.wikipedia.org/wiki/Latin_America_and_the_Caribbean). I picked Mexico `mx`, which has the most Spanish speakers. Other UN M49 codes are not handled, and get the default country for that language.
* `gl-gl`, from Firefox's [Galician](https://pontoon.mozilla.org/gl/firefox/toolkit/chrome/global/intl.properties/?string=81017). This means "Galician as spoken in Greenland", which appears to be a mistake by the translators, since [Galician](https://en.wikipedia.org/wiki/Galician_language) is mostly spoken in [Galucia, Spain](https://en.wikipedia.org/wiki/Galicia_(Spain)) (`ES-GA`), with some speakers in Portugal, and (almost?) none in Greenland. I picked Spain (`es`) for this combo.

## How to test
I used the unit tests to develop and test this code, rather than integration testing. I don't think any value will be added by integration testing in a browser, but you can try it.

You may want to cross-check the languages with the premium countries in [privaterelay/plans.py](https://github.com/mozilla/fx-private-relay/blob/main/privaterelay/plans.py#L523), as well as CLDR's [Territory-Language Information](https://www.unicode.org/cldr/charts/42/supplemental/territory_language_information.html), to see if you agree with the languages that would lead to a user being placed in those countries.

See if the code and comments are sufficient to understand what is going on, or if more documentation is needed.

## Follow-on Work
I was tempted to add some logging to see how often `X-Client-Region` agrees with `Accept-Language`, and what errors are raised. That data could be useful for validating `guess_country_from_accept_lang` and potentially replacing it with a machine learning algorithm. However, this PR is already big enough, and that task can be started when we plan to use the data.

We should also capitalize country codes (`US` instead of `us`), to make it clearer when a code is a language code (`es` for Spanish) versus a country code (`ES` for Spain), and to conform to [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2).

### DIFF
--- a/emails/tests/models_tests.py
+++ b/emails/tests/models_tests.py
@@ -724,8 +724,12 @@ class ProfileFxaLocaleInPremiumCountryTest(ProfileTestCase):
         self.set_fxa_locale("de-DE,en-xx;q=0.9,en;q=0.8")
         assert self.profile.fxa_locale_in_premium_country is True
 
-    def test_when_premium_unavailable_returns_False(self) -> None:
+    def test_en_implies_premium_available(self) -> None:
         self.set_fxa_locale("en;q=0.8")
+        assert self.profile.fxa_locale_in_premium_country is True
+
+    def test_when_premium_unavailable_returns_False(self) -> None:
+        self.set_fxa_locale("en-IN, en;q=0.8")
         assert self.profile.fxa_locale_in_premium_country is False
 
     def test_when_premium_available_by_language_code_returns_True(self) -> None:
--- a/emails/tests/models_tests.py
+++ b/emails/tests/models_tests.py
@@ -736,9 +736,9 @@ class ProfileFxaLocaleInPremiumCountryTest(ProfileTestCase):
         self.set_fxa_locale("de;q=0.8")
         assert self.profile.fxa_locale_in_premium_country is True
 
-    def test_when_invalid_language_code_returns_True(self) -> None:
-        self.set_fxa_locale("xx;q=0.8")  # Invalid code, treated as 'us'
-        assert self.profile.fxa_locale_in_premium_country is True
+    def test_invalid_language_code_returns_False(self) -> None:
+        self.set_fxa_locale("xx;q=0.8")
+        assert self.profile.fxa_locale_in_premium_country is False
 
     def test_when_premium_unavailable_by_language_code_returns_False(self) -> None:
         self.set_fxa_locale("zh;q=0.8")
--- a/emails/tests/models_tests.py
+++ b/emails/tests/models_tests.py
@@ -736,8 +736,12 @@ class ProfileFxaLocaleInPremiumCountryTest(ProfileTestCase):
         self.set_fxa_locale("de;q=0.8")
         assert self.profile.fxa_locale_in_premium_country is True
 
+    def test_when_invalid_language_code_returns_True(self) -> None:
+        self.set_fxa_locale("xx;q=0.8")  # Invalid code, treated as 'us'
+        assert self.profile.fxa_locale_in_premium_country is True
+
     def test_when_premium_unavailable_by_language_code_returns_False(self) -> None:
-        self.set_fxa_locale("xx;q=0.8")
+        self.set_fxa_locale("zh;q=0.8")
         assert self.profile.fxa_locale_in_premium_country is False
 
     def test_no_fxa_account_returns_False(self) -> None:
--- a/privaterelay/tests/utils_tests.py
+++ b/privaterelay/tests/utils_tests.py
@@ -14,7 +14,6 @@ from waffle.testutils import override_flag
 from waffle.utils import get_cache as get_waffle_cache
 import pytest
 
-from ..plans import get_premium_country_language_mapping
 from ..utils import (
     flag_is_active_in_task,
     get_premium_country_lang,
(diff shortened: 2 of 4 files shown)


emails/tests/models_tests.py
```diff
@@ -724,8 +724,12 @@ def test_when_premium_available_returns_True(self) -> None:
         self.set_fxa_locale("de-DE,en-xx;q=0.9,en;q=0.8")
         assert self.profile.fxa_locale_in_premium_country is True
 
-    def test_when_premium_unavailable_returns_False(self) -> None:
+    def test_en_implies_premium_available(self) -> None:
```
jwhitlock, line 727: This was a "wat" from https://github.com/mozilla/fx-private-relay/pull/3602#discussion_r1261725670 - @groovecoder expected `en` to imply `us`, and now it does, both in `Profile.fxa_locale_in_premium_country` and in a web request.

jwhitlock:
I made a significant change since I opened this earlier today. In the first version, the code defaulted to United States (`us`) if there were issues with the `Accept-Language` header. Now, it raises a new exception `AcceptLanguageError`, and the callers catch this and do something appropriate (in a web request, use an empty string for the country, and in the backend, return that the user is _not_ in a premium country).

privaterelay/utils.py
```diff
@@ -27,34 +40,242 @@ def get_countries_info_from_request_and_mapping(request, mapping):
...
+        )
+
+    for maybe_region_raw in subtags[1:]:
+        maybe_region = maybe_region_raw.lower()
+
+        # Look for a special case
+        try:
+            return _LANGUAGE_TAG_TO_COUNTRY_OVERRIDE[(lang, maybe_region)]
+        except KeyError:
+            pass
```
groovecoder, line 253: comment (non-blocking): hmm ... I'm not sure why but this `try ... except KeyError: pass` logic feels looser than:
```python
if (lang, maybe_region) in _LANGUAGE_TAG_TO_COUNTRY_OVERRIDE:
    return _LANGUAGE_TAG_TO_COUNTRY_OVERRIDE[(lang, maybe_region)]
```

I suppose doing it this way reduces all that code duplication, and this code is well tested. I just don't think I've ever seen this approach before.

privaterelay/tests/utils_tests.py
```diff
@@ -16,60 +15,305 @@
...
-        cc, lang = get_premium_country_lang("et-ee", mapping)
-        assert cc == "ee"
-        assert lang == "et"
-
-    def test_eu_country_expansion_inactive(self) -> None:
-        mapping = get_premium_country_language_mapping(eu_country_expansion=False)
-        cc, lang = get_premium_country_lang("et-ee", mapping)
-        assert cc == "ee"
-        assert lang == "en"
+@pytest.mark.parametrize(
```
groovecoder, line 25: As soon as I read the PR description, I knew there was about to be an *EPIC* `parametrize` decorator in here. 🤯

groovecoder (approved):

privaterelay/utils.py
jwhitlock: It's a long debate, here's an argument from 2016:

https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

After re-reading it, it is much more common for the override to NOT be present, so Brett would agree with you that I picked the wrong one. I'm looking for ways to use the walrus operator, is this too ugly?

```suggestion
        if (override := _LANGUAGE_TAG_TO_COUNTRY_OVERRIDE.get((lang, maybe_region))):
            return override
```

groovecoder: TIL (Today I loved): the walrus operator.

PR #3707, opened 2023-07-26
MPP-3262: save utms in cookies to restore after redirects

groovecoder:
When a browser goes directly to a page that requires login, we redirect them to FxA. The redirect drops any utm_* url params, so we can't see where these browsers came from.

So, this change:
1. Adds a backend middleware which stores any utm_* params into cookies.
2. Adds frontend code to set GA fields for the utm cookie values.

### How to test:
1. Open a browser session that is NOT signed into 127.0.0.1:8000
2. Open the console and choose "Persist Logs"
3. Visit http://127.0.0.1:8000/accounts/profile/?utm_medium=firefox-desktop&utm_source=modal&utm_campaign=limit&utm_content=manage-masks-global
   * [ ] The GA output should show that it set `campaign*` fields to the values in the `utm_*` url params
   * [ ] It should redirect to FxA sign-in
4. Complete sign in via FxA
   * [ ] You should land back on the dashboard again

### Checklist

### DIFF
--- a/frontend/src/functions/cookies.ts
+++ b/frontend/src/functions/cookies.ts
@@ -1,3 +1,5 @@
+import ReactGa from "react-ga";
+
 export function getCookie(cookieId: string): string | undefined {
   if (typeof document === "undefined") {
     // When server-side rendering:
@@ -40,4 +42,32 @@ export function clearCookie(cookieId: string) {
 
 export function getCsrfToken(): string | undefined {
   return getCookie("csrftoken");
+}
+
+export function convertUtmCookieToGaField(
+  utmCookie: string,
+): ReactGa.FieldsObject {
+  /*
+   * To preserve utm url param values thru FxA redirects, the server-side
+   * middleware:StoreUtmsInCookie stores utm params in cookies with this format:
+   * utm_medium=firefox-desktop;
+   * utm_source=modal;
+   * utm_content=manage-masks-global;
+   *
+   * analytics.js traffic source field names are in this format:
+   * campaignMedium
+   * campaignSource
+   * campaignContent
+   *
+   * So, convert the cookie names and values to GA field names & values
+   */
+  const campaignField = utmCookie.split("=");
+  const campaignFieldName = campaignField[0].replace("utm_", "");
+  const capitalizedCampaignFieldName =
+    campaignFieldName.charAt(0).toUpperCase() + campaignFieldName.slice(1);
+  const campaignFieldValue = campaignField[1];
+  const gaField: ReactGa.FieldsObject = {};
+  const gaFieldName = `campaign${capitalizedCampaignFieldName}`;
+  gaField[gaFieldName] = campaignFieldValue;
+  return gaField;
 }
--- a/frontend/src/pages/_app.page.tsx
+++ b/frontend/src/pages/_app.page.tsx
@@ -14,6 +14,7 @@ import { initialiseApiMocks } from "../apiMocks/initialise";
 import { mockIds } from "../apiMocks/mockData";
 import { useIsLoggedIn } from "../hooks/session";
 import "@stripe/stripe-js";
+import { clearCookie, convertUtmCookieToGaField } from "../functions/cookies";
 
 if (process.env.NEXT_PUBLIC_MOCK_API === "true") {
   initialiseApiMocks();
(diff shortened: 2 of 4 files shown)


frontend/src/functions/cookies.ts
```diff
@@ -41,3 +43,17 @@ export function clearCookie(cookieId: string) {
...
+  const campaignField = utmCookie.split("=");
+  const campaignFieldName = campaignField[0].replace("utm_", "");
+  const capitalizedCampaignFieldName =
+    campaignFieldName.charAt(0).toUpperCase() + campaignFieldName.slice(1);
+  const campaignFieldValue = campaignField[1];
+  const gaField: ReactGa.FieldsObject = {};
+  const gaFieldName = `campaign${capitalizedCampaignFieldName}`;
+  gaField[gaFieldName] = campaignFieldValue;
+  return gaField;
+}
```
lloan, line 59: suggestion (non-blocking): This seems pretty efficient to me and readable. As Vincent mentioned, perhaps adding a comment for the input/output expected would be useful. We could add a test later.

lloan (approved):
non-blocking comments left - thanks!

frontend/src/functions/cookies.ts
groovecoder: Added comment.

Vinnl (approved):
I haven't been able to test this locally, but I think that's because I couldn't correctly disable tracking protections where needed. The code looks good. The only non-blocking suggestion I'd have is, given the hairiness of the cookie API's and general code readability, that it might be preferable to just enumerate the couple of possible UTM params and re-use the `getCookie` method to read them.

frontend/src/pages/_app.page.tsx
```diff
@@ -65,11 +66,19 @@ function MyApp({ Component, pageProps }: AppProps) {
...
       anonymizeIp: true,
       transport: "beacon",
-    });
+    };
     const cookies = document.cookie.split("; ");
+    cookies.forEach((item) => {
+      if (item.trim().startsWith("utm_")) {
+        const cookieId = item.split("=")[0];
+        Object.assign(gaFields, convertUtmCookieToGaField(item));
+        clearCookie(cookieId);
```
say-yawn, line 78: (question, blocking) I think this code might be clearing the cookie early when it first finds a `utm_` value even though there are other keys that start with `utm_`. 
(comment) Glad we are clearing the data after we use them for analytics!

privaterelay/middleware.py
```diff
@@ -12,6 +13,23 @@
...
+    def __call__(self, request):
+        # Store utm_ values in cookie so we can send them in GA pings after redirecting
+        # thru FxA flow.
+        response = self.get_response(request)
+        if "utm_" in request.get_full_path():
+            parsed_url = urlparse(request.build_absolute_uri())
+            parsed_qs = parse_qs(parsed_url.query)
+            for param in parsed_qs:
+                if param.startswith("utm_"):
+                    response.set_cookie(param, parsed_qs[param][0])
```
say-yawn, line 29: (suggestion, blocking) Maximum size of the cookie differs from browser to browser. Do we want to set a limit on how big the cookies can get here?

say-yawn (changes requested):
Most of the changes look good but I have a question and suggestion that is blocking.

frontend/src/pages/_app.page.tsx
groovecoder: It shouldn't be. I tested and saw all the GA fields ARE assigned. It only clears the single cookie matching the `cookieId` value of the `item` that it just assigned to the `gaFields`.

privaterelay/middleware.py
groovecoder: Truncating the value to 4095 bytes.

say-yawn (changes requested):

privaterelay/middleware.py
```diff
@@ -12,6 +13,24 @@
...
+        # Store utm_ values in cookie so we can send them in GA pings after redirecting
+        # thru FxA flow.
+        response = self.get_response(request)
+        if "utm_" in request.get_full_path():
+            parsed_url = urlparse(request.build_absolute_uri())
+            parsed_qs = parse_qs(parsed_url.query)
+            for param in parsed_qs:
+                if param.startswith("utm_"):
+                    # Prevent setting cookie too large by truncating to 4096 bytes
+                    response.set_cookie(param, parsed_qs[param][0][:4095])
```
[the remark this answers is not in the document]
groovecoder: Hmm ... now I'm worrying that we're over-engineering this. It seems like the worst-case scenario of setting a cookie past the limit is that the cookie just isn't set. And the only consequence here is that we'll lose the traffic campaign/source values of this particular browser session in Google Analytics.

So, I think I want to just go back to setting the cookie and ignoring the limits completely.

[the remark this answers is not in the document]
say-yawn: Sure that sounds good to me.

say-yawn (changes requested):
Changes LGTM. @groovecoder and I agreed that it's too early to be worried about the cookies being over the limit for browsers given that the worst case scenarios is that we lose analytics.

say-yawn (approved):
Meant to approve this 😓

privaterelay/middleware.py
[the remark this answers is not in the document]
groovecoder: Cool, reverted back to plain `set_cookie` call without worrying about limits.

PR #3652, opened 2023-07-13
Improve scrolling experience across browsers

jwhitlock:
**Note**: _This is a copy of PR #3426, on mozilla/fx-private-relay so tests will run_

This PR fixes #3407, fixes #3406, fixes #3223, fixes #3079.

Updates layout structure and styling to avoid the issues above. It also moves the menu for small/mobile screens inside the header so that it is able to scroll smoothly with the header.

How to test:

1. Go to Private Relay website (on iOS / Windows / Android browser).
2. Scroll through each page (including tracker report)
3. Observe the following:
   1. Short pages have the footer reach the bottom of the screen
   2. Mobile menu remains directly under the header during scrolling
   3. There are not multiple scroll bars on the page
   4. No blank areas while scrolling on mobile

- [ ] ~l10n changes have been submitted to the l10n repository, if any.~ (N/A)
- [ ] ~I've added a unit test to test for potential regressions of this bug.~ (N/A)
- [ ] ~I've added or updated relevant docs in the docs/ directory.~ (N/A)
- [x] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #3710, opened 2023-07-27
MPP-3298: Fix for interactions being blocked in mobile view

rafeerahman:
This PR fixes MPP-3298.

- [x] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

# New feature description

Disabled the the parent mobile-menu element when the toggle menu is not expanded. This prevents the invisible mobile menu from appearing and blocking interactions with elements under it.

# Screenshot (if applicable)
New (we can see here that the info icon is clickable): 

image

Old (mobile-menu blocking elements, see in prod):

image

# How to test

* Check whether interactions with the website are valid and whether buttons are clickable in a mobile view (inspect element, and click the mobile symbol).

# Checklist (Definition of Done)
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.
- [x] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] l10n changes have been submitted to the l10n repository, if any.

PR #3714, opened 2023-07-28
Fix not being able to tap elements on mobile

Vinnl:
The mobile menu intercepted all taps underneath it, even when it wasn't visible, leading to the user not being able to interact with most of the page. This fixes that.

This is probably fallout from https://github.com/mozilla/fx-private-relay/pull/3652, and I found out about it by trying to verify https://github.com/mozilla/fx-private-relay/issues/3223#issuecomment-1644319245.

How to test: in responsive design mode with tap simulation enabled, or on an actual phone (in either Firefox and Chrome, at least), don't expand the mobile menu, but try to tap elements present in the area where the mobile menu would be visible when expanded. Before this fix, that shouldn't work, but after this fix, it should.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, CSS fix

### DIFF
--- a/frontend/src/components/layout/navigation/MobileNavigation.module.scss
+++ b/frontend/src/components/layout/navigation/MobileNavigation.module.scss
@@ -9,6 +9,11 @@
   z-index: 1;
   max-height: calc(100vh - 100%);
   overflow-y: auto;
+  pointer-events: none;
+
+  &.is-active {
+    pointer-events: auto;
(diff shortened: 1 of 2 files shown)


Vinnl:
Ah, looks like https://github.com/mozilla/fx-private-relay/pull/3710 also solves this - closing this in favour of that.

PR #3715, opened 2023-07-28
Variable name updated

rafeerahman:
Updated the variable name to match the date, MPP-3235 for reference


jwhitlock (approved):
Thanks @rafeerahman, nice attention to detail!

PR #3716, opened 2023-07-31
MPP-3321: Clear block_list_emails for non-premium users

jwhitlock:
Users who were formerly premium customers may have `block_list_emails` set on their random or domain masks. This would block list emails when forwarding, forward other emails, and then fail when updating the forward / block counts for the address, due to the changes in PR #3651. This may cause our code to re-process the email multiple times.

This PR fixes this bug (MPP-3321) with three changes:

* Remove the pre-save signal handler that raises a `BadRequest` when a non-premium user attempts to set `block_list_emails`. This does not change the API, which checks this in the serializer and returns an error, but does fix the bug.
* Check if the user has premium in the email-handling code before blocking list emails. This prevents previously premium users from using this feature.
* Clear `block_list_emails` when saving a random or domain mask. This will fix the data going forward, as masks are used.

This is sufficient to fix the bug. Users may still see blocking set to "Promotional" on the website. A cleaner task could be added to periodically clear `block_list_emails` from formerly premium accounts, but that would be a larger change than this bug fix

## How to test

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -662,6 +662,8 @@ class RelayAddress(models.Model):
             self.description = ""
             self.generated_for = ""
             self.used_on = ""
+        if not self.user.profile.has_premium:
+            self.block_list_emails = False
         return super().save(*args, **kwargs)
 
     @property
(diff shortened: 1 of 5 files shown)


groovecoder (approved):
no notes. thanks for the quick turn-around when catching this!

PR #3718, opened 2023-08-02
MPP-3323: Info icon tooltip in generate mask modal should only be active on hover

rafeerahman:
This PR fixes MPP-3323.

# Bug description

After clicking the "Generate new mask" button then clicking the info icon button next to "Block promotional emails", clicking the "Learn more" link in the tool tip cause it to disappear, making the "Learn more" unclickable. _See MPP-3323 for full steps to reproduce._

# Screenshot 
(Old)
blob:https://mozilla-hub.atlassian.net/1f407440-8fd8-4b84-878f-b0413447f54c#media-blob-url=true&id=ffdcc805-f428-4a55-b79a-42579841f75d&collection=&contextId=342591&height=972&width=1905&alt=

(New)
[New version](https://i.gyazo.com/20bf1d3ef1a7731e28fc0c16b2b1cc6e.gif)

# How to test

1. Navigate to the Relay dashboard;

2. Click on the “Generate new mask” button;

3. Select “@<custom domain> mask”;

4. Click on the “i” tooltip button;

5. Hover over the tooltip to open it again;

6. Hover over the tooltip content and try to click on the “Learn more” link;

Newly expected behaviour is that the "Learn more" button should bring you to /faq/#faq-promotional-email-blocking.

### DIFF
--- a/frontend/src/components/InfoTooltip.tsx
+++ b/frontend/src/components/InfoTooltip.tsx
@@ -12,7 +12,7 @@ export type Props = {
 
 export const InfoTooltip = (props: Props) => {
   const tooltipTriggerState = useTooltipTriggerState({ delay: 0 });
-  const triggerRef = useRef<HTMLButtonElement>(null);
+  const triggerRef = useRef<HTMLSpanElement>(null);
 
   const tooltipTrigger = useTooltipTrigger({}, tooltipTriggerState, triggerRef);
   const { tooltipProps } = useTooltip(
@@ -22,14 +22,10 @@ export const InfoTooltip = (props: Props) => {
 
   return (
     <span className={styles.wrapper}>
-      <button
+      <span
         ref={triggerRef}
-        // Set to type="button" to prevent wrapping forms from being submitted
-        // when the info icon is clicked:
-        type="button"
         {...tooltipTrigger.triggerProps}
         className={styles.trigger}
-        disabled
       >
         <InfoIcon
           alt={props.alt}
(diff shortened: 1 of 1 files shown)


lloan:
@rafeerahman I asked the other UI team members to pitch in to review this one. Thanks!

Vinnl (commented):
Ah hmm, I think this was my mistake initially - I think the reason [the docs](https://react-spectrum.adobe.com/react-aria/useTooltipTrigger.html) use a `button` is just so that the tooltip can describe what the button does. But since in this case, nothing happens when you click it, it should just not have been a button - a plain `<span>` for example should do.

Adding `disabled` only adds to the confusion that I introduced - a screen reader would now announce it as a disabled button, so the user might think they'd have to perform some action to enable it.

Sorry about the initial mistake and not approving this fix yet. Luckily the real fix (removing the `type="button"` and changing the `<button>` into a `<span>`) should be pretty quick too, I think.

rafeerahman:
> Ah hmm, I think this was my mistake initially - I think the reason [the docs](https://react-spectrum.adobe.com/react-aria/useTooltipTrigger.html) use a `button` is just so that the tooltip can describe what the button does. But since in this case, nothing happens when you click it, it should just not have been a button - a plain `<span>` for example should do.
> 
> Adding `disabled` only adds to the confusion that I introduced - a screen reader would now announce it as a disabled button, so the user might think they'd have to perform some action to enable it.
> 
> Sorry about the initial mistake and not approving this fix yet. Luckily the real fix (removing the `type="button"` and changing the `<button>` into a `<span>`) should be pretty quick too, I think.

Ah I see, didn't think about screen readers. Made the change now. Thanks!

Vinnl (approved):
Looking good, thank you!

jwhitlock:
@rafeerahman, you can merge when ready. Do you have the permissions needed to merge this yourself?

PR #3719, opened 2023-08-04
Group related Dependabot JS updates

Vinnl:
This should hopefully cut down on some of the noise of ESLint updates, and ensure that packages that break unless they are updated together (like https://github.com/mozilla/fx-private-relay/pull/3648) are actually updated together.

See https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/


jwhitlock (approved):
Thanks @Vinnl!

PR #3747, opened 2023-08-08
MPP-3282: Refactor plans to make it clear when language is not a factor in plan choice

jwhitlock:
This PR updates the code to use standards for formatting countries and languages:

* For languages, use the [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) code of two lower-case letters.code. For example, `"en"` for English. This is the current standard in Relay code.
* For countries, use [ISO 3116](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes
) code, specifically [ISO 3166 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) code of two upper-case letters when possible. For example `"DE"` for Germany. Before this PR, a lower-case code was used, which made it easy to confuse a language with a country (`"de"` could mean the language German or the country Germany).

This change is big enough that it could be its own PR, as well as some other cleanup (alpha-sorting test cases, re-implementing type definitions).

For MPP-3282, this PR rewrites `_RELAY_PLANS_BY_COUNTRY_AND_LANGUAGE` to `_RELAY_PLANS`, and further splits the structure to distinguish between:

* Plans determined by country and language (German-speaking Swiss have a different Stripe plan than French-speaking Swiss)
* Plans determined by country, but use a different country's plan (Canada uses the US plan, priced in USD)
* Plans where the country matches the Stripe country (United States gets the US plan)

This makes it clearer that:

1. In the majority of the cases, the language is _not_ used to distinguish between Stripe plans
2. It is unnecessary to determine the majority language in a country when only the country is used to determine the plan

In these cases, instead of a majority language (or Pontoon language with active translators), a language code of `"*"` is used as a placeholder in the plan data available at `/api/v1/runtime_data`. The front-end code already defaults to the first entry when the user's language isn't in the country's language map, so it will now default to this `"*"` entry in cases where only the country matters.

## How to test:

- View `/api/v1/runtime_data` response in Firefox, or with a JSON formatter like `jq`:
    - [ ] Countries are now consistently two upper-case letters
    - [ ] The plan maps are sorted by country code
    - [ ] The plan maps use `"*"` when language does not matter
    - [ ] The premium plan map (`PERIODICAL_PREMIUM_PLANS.plan_country_lang_mapping`) for Belgium (`BE`) has entries for French (`fr`), German (`de`) and Dutch (`nl`), in that order.
    - [ ] The premium plan map for Switzerland (`CH`) has entries for French (`fr`), German (`de`) and Italian (`it`), in that order.
    - [ ] The phone plan map (`PHONE_PLANS.plan_country_lang_mapping`) has only `CA` (Canada) and `US` (United States), the language is `"*"`, and the plan details are the same.
    - [ ] The bundle plan map (`BUNDLE_PLANS.plan_country_lang_mapping`) has only `CA` (Canada) and `US` (United States), the language is `"*"`, and the plan details are the same.
- View the front page when logged out
    - [ ] The available plans are listed the prices and currency for your region
- [x] I've updated unit tests to test for potential regressions of this bug.
- ~[ ] I've added or updated relevant docs in the docs/ directory.~ - covered by MPP-3283


privaterelay/plans.py
```diff
@@ -18,8 +18,8 @@
 They all return a PlanCountryLangMapping dict, which has this structure:
```
groovecoder, line 18: question (non-blocking): is there a reason (other than just historical?) we call this `PlanCountryLangMapping` when this structure seems to be more accurately called `CountryLangPlanMapping`?

groovecoder (approved):
1 non-blocking question.

privaterelay/plans.py
jwhitlock: Yep, historical. These structures were previously in settings:

https://github.com/mozilla/fx-private-relay/blob/085142c4148e607a009931f454e1a854c499a68f/privaterelay/settings.py#L662
https://github.com/mozilla/fx-private-relay/blob/085142c4148e607a009931f454e1a854c499a68f/privaterelay/settings.py#L796
https://github.com/mozilla/fx-private-relay/blob/085142c4148e607a009931f454e1a854c499a68f/privaterelay/settings.py#L818

So the structure type was named after the common element from the settings names. The "Plan" is premium, phones, or bundle.

The structure is:

* country code
    * language code
        * term (monthly, yearly)
            * price details
            
so maybe `CountryLanguageTermPriceMapping`? `CountryLangTermPriceMappingForPlan`? `PlanMap`? I give up.

groovecoder: No worries - we don't need to change it.

groovecoder:
I'm merging my PR, so this one may need to resolve conflicts.

PR #3754, opened 2023-08-11
MPP-2117: ADR 0002 - Use mask address as `From:` address

jwhitlock:
Better late than never - here is the ADR documenting the change to the `From:` address for forwarded emails, started under MPP-2117 and continued under MPP-3113. It included the aborted attempt in MPP-3189 as "Option 2: Use the Service's Sender Address as the `From:` Address"

### DIFF
--- a/docs/adr/0002-use-mask-as-from-address.md
+++ b/docs/adr/0002-use-mask-as-from-address.md
@@ -146,9 +146,9 @@ Resent-From: offers@servicex.example.com
 - Bad, because of a small increased probability of delivery to spam, due to the
   reputation of `mozmail.com` versus `firefox.com`.
 
-In QA testing in August 2023, emails from one service were sent to spam with
-the new `From:` address, while were delivered to the inbox with the old `From:`
-address from Option 1, while over 100 saw no difference in Inbox delivery.
+In QA testing in August 2023, over 100 services had no delivery
+change with the new `From:` address. For one service, the email went
+to spam instead of the inbox.
 
 ### Option 4: Use the Relay Mask Plus Derived Prefix as the `From:` Address
 
--- a/docs/adr/0002-use-mask-as-from-address.md
+++ b/docs/adr/0002-use-mask-as-from-address.md
@@ -1,0 +1,200 @@
+# Use the Relay mask address as From: address for forwarded emails
+
+- Status: Proposed
+- Deciders: Luke Crouch, Doris Deng
+- Date: 2023-08-11
+
+Technical Story: [MPP-2117](https://mozilla-hub.atlassian.net/browse/MPP-2117)
+
+## Context and Problem Statement
+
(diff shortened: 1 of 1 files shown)


docs/adr/0002-use-mask-as-from-address.md
```diff
@@ -0,0 +1,200 @@
...
+Resent-From: offers@servicex.example.com
+```
+
+- Good, because the original sender is visible in the display name.
+- Good, because the email mask is available for filtering.
+- Good, because Proton Mail organizes by mask in conversation mode.
+- Good, because it rewards usage of a mask on a single service, which is better
+  for user privacy and is compatible with the Relay business model.
+- Bad, because of a small increased probability of delivery to spam, due to the
+  reputation of `mozmail.com` versus `mozilla.com`.
```
groovecoder, line 147:
```suggestion
- Good, because it rewards usage of a mask on a single service, which is better
  for user privacy.
- Bad, because of a small increased probability of delivery to spam, due to the
  reputation of `mozmail.com` versus `firefox.com`.
```

docs/adr/0002-use-mask-as-from-address.md
```diff
@@ -0,0 +1,200 @@
...
+
+In QA testing in August 2023, emails from one service were sent to spam with
+the new `From:` address, while were delivered to the inbox with the old `From:`
```
groovecoder, line 150: suggestion (blocking): seems like there's a number missing here? "... while where delievered to the inbox ..."

docs/adr/0002-use-mask-as-from-address.md
```diff
@@ -0,0 +1,200 @@
...
+Resent-From: offers@servicex.example.com
+```
+
+- Good, because the original sender is visible in the display name.
+- Good, because the email mask plus original email is available for filtering.
+- Good, because Proton Mail organizes by mask in conversation mode.
+- Bad, because of a small increased probability of delivery to spam, due to the
+  reputation of `mozmail.com` versus `mozilla.com`.
+- Bad, because it encourages reuse of masks, which is worse for user privacy
+  and for the Relay business model.
```
groovecoder, line 179:
```suggestion
- Bad, because of a small increased probability of delivery to spam, due to the
  reputation of `mozmail.com` versus `firefox.com`.
- Bad, because it encourages reuse of masks, which is worse for user privacy.
```

groovecoder (changes requested):

docs/adr/0002-use-mask-as-from-address.md
jwhitlock: good catch! Changing the paragraph to:

> In QA testing in August 2023, over 100 services had no delivery change with the new `From:` address. For one service, the email went to spam instead of the inbox.

PR #3773, opened 2023-08-14
MPP-3008 iOS not displaying close buttons

rafeerahman:
This PR fixes MPP-3008.

# Bug description
On iOS devices (tested on iOS 16), the "X" icon disappears in areas where the SVG element is a child of the button element. The issue describes this for the info modal, but John also noticed this on a survey banner (https://mozilla-hub.atlassian.net/browse/MPP-3008?focusedCommentId=715256). From the inspector, I noticed that this happens because some extra padding gets added from the user-agent (browser) style sheets, although there might be some other reasoning related to svg viewports and iOS. 

To fix this, I added a global padding of 0 to buttons (instead of going into every component that has a close button and adding 0 padding to it). The buttons that use padding for design reasons are defined in their respective scss modules, and this overrides the global css. 

# Screenshot (if applicable)
IMG-20230814-WA0000

# How to test
There is no clear way run the local server on your phone, because there will be an error when trying to log in to your FxA (redirect_uri error). Because of this, testing is a bit difficult. I tested by using my Mac and connecting an iPhone that has the dev site loaded, then used the inspector.

A way you can test this specific branch is by doing the following: 
* Follow the instructions to release this branch to dev, https://github.com/mozilla/fx-private-relay/blob/main/docs/release_process.md#release-to-dev 
* Then go to https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/accounts/profile/ on an iOS device
* Tap on the “i” button next to the “Set your unique ⁨Relay⁩ email domain“;
* Observe the opened modal;

### DIFF
--- a/frontend/src/styles/globals.scss
+++ b/frontend/src/styles/globals.scss
@@ -53,6 +53,10 @@ body {
   -moz-osx-font-smoothing: grayscale;
 }
 
+button {
+  padding: 0;
+}
+
 img,
 picture,
 video,


frontend/src/styles/globals.scss
```diff
@@ -53,6 +53,10 @@ body {
   -moz-osx-font-smoothing: grayscale;
 }
 
+button {
+  padding: 0;
+}
+
```
groovecoder, line 59: suggestion (non-blocking): This seems like a pretty broad change. When this is merged on the dev server, can we ask QA to pay special attention to `button` elements across the site to make sure this change doesn't break any other buttons?

groovecoder (approved):

frontend/src/styles/globals.scss
rafeerahman: Will make sure to let QA know.

Vinnl:
Oh yeah, I'd generally be weary of such global changes, because it's really hard to see all repercussions. One example is that e.g. the `x` in a popover is no longer center-aligned when a border is visible because it is focused with the keyboard:



You can see that it affects a whole lot of elements if you enable/disable the style:

[](https://github.com/mozilla/fx-private-relay/assets/4251/aef9ab4d-18ce-42d1-8b08-860035d4fee3)

It's very much possible that all those changes are very minor, but it's also possible that one of them somewhere breaks in unexpected ways. So I'll leave it up to your judgement, but it might be safer to just only add the change to specific buttons that you know break.

rafeerahman:
> Oh yeah, I'd generally be weary of such global changes, because it's really hard to see all repercussions. One example is that e.g. the `x` in a popover is no longer center-aligned when a border is visible because it is focused with the keyboard:

Thanks for checking this. I agree and think that adding the change to the specific buttons that are breaking is more practical. Will make a new PR later today

PR #3795, opened 2023-08-17
MPP-2959: Subdomain can be edited through PATCH request

rafeerahman:
This PR fixes MPP-2959.

How to test:

* Verify that the test is successful by running `pytest api/tests/views_tests.py"

Testing manually:

1. Go to the Relay dashboard
2. Send a request to /api/v1/profiles/<uid\>/  with the body {"subdomain": "<a new subdomain\>"}. Screenshot 2023-08-17 091222
3. Refresh the page, and observe the subdomain to ensure that it was not changed.
4. Observe that you get a response code of "400"

- [x] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] ~l10n changes have been submitted to the l10n repository, if any.~

### DIFF
--- a/api/serializers/__init__.py
+++ b/api/serializers/__init__.py
@@ -132,6 +132,7 @@ class ProfileSerializer(serializers.ModelSerializer):
         ]
         read_only_fields = [
             "id",
+            "subdomain",
             "has_premium",
             "has_phone",
             "has_vpn",
--- a/api/serializers/__init__.py
+++ b/api/serializers/__init__.py
@@ -106,7 +106,8 @@ class DomainAddressSerializer(PremiumValidatorsMixin, serializers.ModelSerialize
 
 
 class StrictReadOnlyFieldsMixin:
-    """Raises a validation error (400) if read only fields are in the body of PUT/PATCH requests.
+    """
+    Raises a validation error (400) if read only fields are in the body of PUT/PATCH requests.
 
     This class comes from https://github.com/encode/django-rest-framework/issues/1655#issuecomment-1197033853,
     where different solutions to mitigating 200 response codes in read-only fields are discussed.
@@ -122,9 +123,17 @@ class StrictReadOnlyFieldsMixin:
         read_only_fields = set(
             field_name for field_name, field in self.fields.items() if field.read_only
         ).union(set(getattr(self.Meta, "read_only_fields", set())))
-        received_read_only_fields = set(self.initial_data).intersection(
-            read_only_fields
+
+        # Getting implicit read only fields that are in the Profile model, but were not defined in the serializer.
+        # By default, they won't update if put in the body of a request, but they still give a 200 response (which we don't want).
+        implicit_read_only_fields = set(
+            field for field in vars(self.Meta.model) if field not in self.fields
         )
+
+        received_read_only_fields = set(self.initial_data).intersection(
+            read_only_fields.union(implicit_read_only_fields)
+        )
+
         if received_read_only_fields:
             errors = {}
             for field_name in received_read_only_fields:
--- a/api/serializers/__init__.py
+++ b/api/serializers/__init__.py
@@ -6,10 +6,6 @@ from waffle import get_waffle_flag_model
 
 from emails.models import DomainAddress, Profile, RelayAddress
 
-import logging
-
-log = logging.getLogger(__name__)
-
 
 class PremiumValidatorsMixin:
     # the user must be premium to set block_list_emails=True
@@ -110,10 +106,7 @@ class DomainAddressSerializer(PremiumValidatorsMixin, serializers.ModelSerialize
 
 
 class StrictReadOnlyFieldsMixin:
-    """Raises a validation error (400) if read only fields are in the body of requests."""
-
-    # Gets added to "error_messages" in the Serializer class using this mixin
-    default_error_messages = {"read_only": ("This field is read only")}
+    """Raises a validation error (400) if read only fields are in the body of PUT/PATCH requests."""
 
     def validate(self, attrs):
         attrs = super().validate(attrs)
@@ -128,12 +121,11 @@ class StrictReadOnlyFieldsMixin:
         received_read_only_fields = set(self.initial_data).intersection(
             read_only_fields
         )
-
         if received_read_only_fields:
             errors = {}
             for field_name in received_read_only_fields:
                 errors[field_name] = serializers.ErrorDetail(
-                    self.error_messages["read_only"], code="read_only"
+                    "This field is read only", code="read_only"
                 )
 
             raise serializers.ValidationError(errors)
--- a/api/serializers/__init__.py
+++ b/api/serializers/__init__.py
@@ -106,7 +106,11 @@ class DomainAddressSerializer(PremiumValidatorsMixin, serializers.ModelSerialize
 
 
 class StrictReadOnlyFieldsMixin:
-    """Raises a validation error (400) if read only fields are in the body of PUT/PATCH requests."""
+    """Raises a validation error (400) if read only fields are in the body of PUT/PATCH requests.
+
+    This class comes from https://github.com/encode/django-rest-framework/issues/1655#issuecomment-1197033853,
+    where different solutions to mitigating 200 response codes in read-only fields are discussed.
+    """
 
     def validate(self, attrs):
         attrs = super().validate(attrs)
--- a/api/serializers/__init__.py
+++ b/api/serializers/__init__.py
@@ -6,6 +6,10 @@ from waffle import get_waffle_flag_model
 
 from emails.models import DomainAddress, Profile, RelayAddress
 
+import logging
+
+log = logging.getLogger(__name__)
+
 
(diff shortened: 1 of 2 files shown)


groovecoder (changes requested):
suggestion (blocking): This seems to return a "200" status code even though the request has failed. Can we change this to somehow return a "4xx" status code so it's clear tothe client that the request failed.

say-yawn (changes requested):
👍 on adding the `subdomain` field to read-only. In addition to recommended change from @groovecoder to return more accurate response on requests that were not reflected in the DB, I recommend adding another test for making a PATCH request on a field that is not listed on `ProfileSerializer` AND verifying that no changes did occur.

rafeerahman:
> suggestion (blocking): This seems to return a "200" status code even though the request has failed. Can we change this to somehow return a "4xx" status code so it's clear tothe client that the request failed.

By default, read-only fields have a 200 code on PATCH/PUT requests. https://github.com/encode/django-rest-framework/issues/1655 goes more in depth, but its by Django's design. I added a mixin that overrides this behavior and gives a validation error (400 response) if a user tries to update read-only fields.

api/serializers/__init__.py
```diff
@@ -105,7 +109,39 @@ class Meta:
         ]
 
 
-class ProfileSerializer(serializers.ModelSerializer):
+class StrictReadOnlyFieldsMixin:
```
groovecoder, line 108: doc (blocker): add a link from this code to https://github.com/encode/django-rest-framework/issues/1655#issuecomment-1197033853 where it came from. So future us's know where this code came from.

groovecoder: suggestion (non-blocking): Also, the work-around in [this comment](https://github.com/encode/django-rest-framework/issues/1655#issuecomment-1478741260) is less code. I'm not sure if it's better code, but did you consider it too?

groovecoder (changes requested):

api/tests/views_tests.py
```diff
@@ -111,6 +111,7 @@ def test_patch_premium_user_subdomain_cannot_be_changed(
     premium_profile = premium_user.profile
     original_subdomain = "helloworld"
     premium_profile.subdomain = original_subdomain
+    READ_ONLY_RES = "This field is read only"
```
groovecoder, line 114: suggestion (blocking): until we re-use this string value another team, let's hard-code it into the line where it's used. I think `CAPITAL_VARS` are typically reserved for module-level values.

groovecoder (changes requested):

api/serializers/__init__.py
rafeerahman: > suggestion (non-blocking): Also, the work-around in [this comment](https://github.com/encode/django-rest-framework/issues/1655#issuecomment-1478741260) is less code. I'm not sure if it's better code, but did you consider it too?

With the new mixin class, all of the read-only fields give a 400 response on PUT/PATCH requests. I think this makes sense because we don't want any of them giving a 200 response. And this class would also be reusable for the other serializers with read-only fields. 

However, if we want just the subdomain field to give a 400 code, I think the latter code ([this comment](https://github.com/encode/django-rest-framework/issues/1655#issuecomment-1478741260)) is better. 

@groovecoder let me know if we should make all of the read-only fields respond with 400 or just the subdomain.  Will also make sure to add this into the docs. Thanks!

groovecoder: > With the new mixin class, all of the read-only fields give a 400 response on PUT/PATCH requests ... And this class would also be reusable for the other serializers with read-only fields.

Sounds good! Let's keep the mixin.

groovecoder (approved):

api/serializers/__init__.py
```diff
@@ -105,7 +105,39 @@ class Meta:
...
+    """Raises a validation error (400) if read only fields are in the body of PUT/PATCH requests.
```
say-yawn, line 109:
```suggestion
    """
    Raises a validation error (400) if read only fields are in the body of PUT/PATCH requests.
```

say-yawn: (suggestion - blocking) Let's utilize similar code style where the multi-line comment `"""` starts on a new line when there is multiple lines of comment while it is on the same line as `"""` when there is only one line of comment.

api/tests/views_tests.py
```diff
@@ -104,6 +104,55 @@ def test_post_domainaddress_no_subdomain_error(premium_user, prem_api_client) ->
...
+    ret_data = response.json()
+    premium_profile.refresh_from_db()
+
+    assert ret_data["subdomain"][0] == "This field is read only"
+    assert premium_profile.subdomain == original_subdomain
+    assert response.status_code == 400
+
+
+def test_patch_profile_fields_are_read_only_by_default(premium_user, prem_api_client):
+    """A field in the Profile model should be read only by default if it is not mentioned in the ProfileSerializer class fields.
```
say-yawn, line 132:
```suggestion
    """
    A field in the Profile model should be read only by default if it is not mentioned in the ProfileSerializer class fields.
```

api/serializers/__init__.py
```diff
@@ -105,7 +105,39 @@ class Meta:
...
+
+    This class comes from https://github.com/encode/django-rest-framework/issues/1655#issuecomment-1197033853,
+    where different solutions to mitigating 200 response codes in read-only fields are discussed.
+    """
+
+    def validate(self, attrs):
```
say-yawn, line 115: (suggestion - blocking) Can you add tests for this code?
"happy" case - where the attribute that is read only on the model and the attribute that is read only on the serializer is passed to the `validate()` and returns the error
happy case - where the attribute that is editable is passed to the `validate()` and DOES NOT returns the error

say-yawn (changes requested):
Almost there, changes requested for code style and more test coverage. The functionality looks good though!

api/tests/views_tests.py
```diff
@@ -104,6 +104,149 @@ def test_post_domainaddress_no_subdomain_error(premium_user, prem_api_client) ->
...
+    )
+    assert premium_profile.remove_level_one_email_trackers is True
+    assert ret_data.get("remove_level_one_email_trackers", None) is True
+    assert ret_data.get("onboarding_state", None) != old_onboarding_state
+    assert premium_profile.onboarding_state == 1
+    assert ret_data.get("onboarding_state", None) == 1
+    assert response.status_code == 200
+
+
+def test_profile_patch_with_model_and_serializer_fields(premium_user, prem_api_client):
```
rafeerahman, line 199: Case where one field is from the model, and another is from the serializer. Both are read-only and should return an error.

api/tests/views_tests.py
```diff
@@ -104,6 +104,149 @@ def test_post_domainaddress_no_subdomain_error(premium_user, prem_api_client) ->
...
+    premium_profile.refresh_from_db()
+
+    assert premium_profile.num_address_deleted == expected_num_address_deleted
+    assert premium_profile.sent_welcome_email == expected_sent_welcome_email
+    assert ret_data.get("sent_welcome_email", [""])[0] == "This field is read only"
+    assert ret_data.get("num_address_deleted", [""])[0] == "This field is read only"
+    assert response.status_code == 400
+
+
+def test_profile_non_read_only_fields_update_correctly(premium_user, prem_api_client):
```
rafeerahman, line 161: Case where fields are non-read-only, and should update correctly.

api/tests/views_tests.py
```diff
@@ -104,6 +104,149 @@ def test_post_domainaddress_no_subdomain_error(premium_user, prem_api_client) ->
...
+    )
+
+    premium_profile.refresh_from_db()
+
+    assert response.status_code == 400
+    assert premium_profile.subdomain != "vanilla"
+    assert premium_profile.num_address_deleted == 0
+
+
+def test_profile_patch_with_non_read_only_and_read_only_fields(
```
rafeerahman, line 215: Case where the body of the request has both a read only, and a non-read-only field. Expected to return an error.

api/tests/views_tests.py
```diff
@@ -104,6 +104,149 @@ def test_post_domainaddress_no_subdomain_error(premium_user, prem_api_client) ->
...
+
+    ret_data = response.json()
+    premium_profile.refresh_from_db()
+
+    assert premium_profile.onboarding_state == old_onboarding_state
+    assert ret_data.get("subdomain", [""])[0] == "This field is read only"
+    assert response.status_code == 400
+
+
+def test_profile_patch_fields_that_dont_exist(premium_user, prem_api_client):
```
rafeerahman, line 236: When the body of the request is comprised of fields that don't exist, the default behavior is to return a 200 response. This test covers this.

api/serializers/__init__.py
rafeerahman: Added these cases (https://github.com/mozilla/fx-private-relay/pull/3795#discussion_r1317838049, https://github.com/mozilla/fx-private-relay/pull/3795#discussion_r1317839215) and some additional ones.

say-yawn (approved):

PR #3840, opened 2023-08-30
fix MPP-3262: send pageview before redirecting to FXA login

groovecoder:
This PR fixes #MPP-3262.

How to test:
1. Open the browser devtools console
   * [ ] Make sure it's set to show "info" level debug messages
2. Go to http://127.0.0.1:8000/accounts/profile/?utm_medium=firefox-desktop&utm_source=modal&utm_campaign=limit&utm_content=manage-masks-global
   * [ ] In the devtools console, you should see output like:
```
[react-ga] called ga('set', fieldsObject);
[react-ga] with fieldsObject: {"anonymizeIp":true,"transport":"beacon"}
[react-ga] called ga('send', 'pageview', path);
[react-ga] with path: /accounts/profile/?utm_medium=firefox-desktop&utm_source=modal&utm_campaign=limit&utm_content=manage-masks-global
```
3. After you finish FXA login, you should get back to your profile without any problems

- [x] ~l10n changes have been submitted to the l10n repository, if any.~
- [ ] I've added a unit test to test for potential regressions of this bug.
- [x] ~I've added or updated relevant docs in the docs/ directory.~
- [x] ~All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).~
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #3844, opened 2023-09-01
fix MPP-3262: preserve utms thru FXA redirect

groovecoder:
This PR fixes #MPP-3262 ... again.

The [previous fix](https://github.com/mozilla/fx-private-relay/pull/3840) DOES send the utm param values to GA before redirecting to FXA, but when the browser is redirected BACK FROM FXA, the subsequent `pageview` ping changes the `source/medium` value of the entire browser session from the `modal/firefox-desktop` sent in the 1st ping, to the `accounts.stage.mozaws.net/referral` value of the 2nd ping.

So, we need to preserve the utm params thru the entire FXA redirect flow. There's 4 steps to this:
1. The `accounts/profile.page.tsx` code performs this redirect, so it needs to send the utm params to `django-allauth` login url. But, `django-allauth` doesn't automatically send all url params to FXA. 😢 
2. Luckily, `django-allauth` has an [un-documented `dynamic_auth_params` functionality](https://github.com/pennersr/django-allauth/blob/fccd33f68b4c19acfa4277c97fbbf0a9226d2367/allauth/socialaccount/providers/oauth2/provider.py#L34). So rather than put the utm params on the `django-allauth` login url, we encode them and put them in an `auth_params` param instead. Then, `django-allauth` sends these to the auth provider, so FXA receives them.
3. [For certain query parameters, including utms, FXA sends them back to the Relying Party in the redirect url](https://mozilla.github.io/ecosystem-platform/relying-parties/reference/query-parameters#generic-parameters).
4. When FXA redirects back to Relay, `django-allauth` doesn't automatically preserve all url params to the post-login redirect url. So, we use a new custom `AccountAdapter.get_login_redirect_url` method to do that.

How to test:

1. Be signed out of Relay
2. Go to http://127.0.0.1:8000/accounts/profile/?utm_medium=firefox-desktop&utm_source=modal&utm_campaign=limit&utm_content=manage-masks-global
3. Go thru the FXA login proces
   * [x] You should return back to http://127.0.0.1:8000/accounts/profile/?utm_medium=firefox-desktop&utm_source=modal&utm_campaign=limit&utm_content=manage-masks-global: note that the utm params are STILL THERE!

## Checklist


jwhitlock (approved):
👍 works locally

PR #3858, opened 2023-09-04
MPP-79: Lint Python code with ruff

jwhitlock:
Check and fix Python code style with [ruff](https://github.com/astral-sh/ruff). Ruff is an extremely fast linter with autofix capabilities, is compatible with `black`, and implements most of the `flake8` rules by default. There are many [rulesets](https://beta.ruff.rs/docs/rules/) that can be added by updating the configuration. I've enabled the pyflakes (F) and pycodestyle (E and W) rules, as well as the isort rules (I). We can add more if they are useful.


groovecoder (changes requested):
suggestion (docs): I don't have strong opinions on code styles, but should we make an ADR doc for adopting ruff?

jwhitlock:
OK, closing until we have an ADR for adopting ruff.

PR #3870, opened 2023-09-07
use more active voice in ADR template

groovecoder:


say-yawn (commented):
I looked into more on the use of active and passive voice and after learning more I would recommend we continue to use passive voice when it comes to describing consequences. In particular [this article](https://www.grammarly.com/blog/active-vs-passive-voice/) on active and passive voice describes the reasoning behind why passive voice is helpful in describing results:
> when you want your reader to focus on the action being described or the action’s target rather than on who or what is performing the action. This is why[ the passive voice is used in lab reports](https://guides.lib.purdue.edu/c.php?g=352816&p=2377936)—it conveys scientific objectivity by minimizing the focus on the doer of the action.

groovecoder:
Hmm ... interesting perspective. I still don't like passive voice, but it seems fine for results & consequences.

PR #3874, opened 2023-09-08
Do Not Merge - Spike for MPP-3375: Handle inline images

jwhitlock:
This is a spike for MPP-3375, changing email processing to handle inline images, opened for feedback purposes. **DO NOT MERGE** I think this is too much for a single PR, so I'm going to break it up, but this is where I'm headed.

Previously, the code would:

* Load an email as a Python `email.EmailMessage`
* Walk the email to extract the HTML content, plaintext content, and any attachments
* Create a wrapped version of the HTML and plaintext content
* Create a new email (`multipart/mixed`) with the content and attachments
* Send that email

Inline images were not attachments, but instead top-level objects in a `multipart/related` section, so they were dropped by this method.

The new code does this:

* Load an email as a Python `email.EmailMessage`
* Replace the headers, dropping most, keeping some, and changing others
* Find the HTML and plaintext content, and replace it in-place
* Send that email

This preserves any inline attachments, as well as other MIME sections that we don't quite understand. The new email has the same structure as the previous email.

Other changes:

* Add a way to load serialized email files as fixtures, and generate a fake SNS JSON envelope
* Change mocks to mock the S3 loading function, rather than the html/text/attachments level
* Add types to more functions
* Refactors, renaming, and deleting code

### DIFF
--- a/emails/tests/fixtures/inline-image.email
+++ b/emails/tests/fixtures/inline-image.email
@@ -1,0 +1,55 @@
+Mime-Version: 1.0 (MailClient 1.1.1)
+Date: Tue, 5 Sep 2023 14:42:46 -0700
+Subject: Test Email
+From: Your Friend <friend@mail.example.com>
+To: Your Relay Address <ebsbdsan7@test.com>
+Content-Type: multipart/related;
+	type="multipart/alternative";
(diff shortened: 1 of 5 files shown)


jwhitlock:
Thanks for looking @groovecoder! I consider this spike code, so there's no need for deeper review.

PR #3891, opened 2023-09-13
MPP-3404 - Retire resender flag

jwhitlock:
The `resender_headers` flag was turned on for everyone in production on September 5, 2023, and has been mostly error free. This PR removes it from the code, and removes code paths where it is off.

### DIFF
--- a/emails/utils.py
+++ b/emails/utils.py
@@ -374,9 +374,9 @@ def generate_from_header(original_from_address: str, relay_mask: str) -> str:
     display_name, original_address = parseaddr(oneline_from_address)
     parsed_address = Address(addr_spec=original_address)
 
-    # Truncate the to 71 characters, so the sender portion fits on the first
-    # line of a multi-line "From:" header, if it is ASCII. A utf-8 encoded
-    # header will be 226 chars, still below the 998 limit of RFC 5322 2.1.1.
+    # Truncate the display name to 71 characters, so the sender portion fits on the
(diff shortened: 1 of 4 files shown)


emails/utils.py
```diff
@@ -346,31 +344,6 @@ def get_reply_to_address(premium: bool = True) -> str:
...
-
-    display_name = Header('"%s [via Relay]"' % (original_from_address), "UTF-8")
-    user_has_premium = bool(user_profile and user_profile.has_premium)
-    relay_from_address = get_reply_to_address(user_has_premium)
-    formatted_from_address = str(
-        Address(display_name.encode(maxlinelen=998), addr_spec=relay_from_address)
-    )
-    return formatted_from_address
-
-
```
groovecoder, line 373: change (blocking, though not part of this particular code change): When I looked for the corresponding code in the new `generate_from_header` function, I saw this:
```
      # Truncate the to 71 characters, so the sender portion fits on the first
```

Looks like it should say "Truncate the display_name to 71 characters ..."?

groovecoder (changes requested):
Small but important code doc change. Everything else looks good!

jwhitlock:
Good catch!

As long as we're touching that code, I added an exception handler to capture data for MPP-3407 and MPP-3417, where some `From:` addresses are causing exceptions. The handler logs the address and re-raises the exception, so no logic change (yet).

groovecoder (approved):
Nice. thanks for adding more tests for more possible scenarios.

PR #3911, opened 2023-09-21
MPP-55: Email types and tests

jwhitlock:
This makes some changes to prepare for handling emails differently for inline images.

The big, consequential changes:

* Add `create_email_from_notification` to create a test email from the headers in an SNS notification fixture and plain text
* Add `assert_email_equals` to test email outputs against fixtures. The emails are adjusted to remove the randomness of MIME boundary strings. If the expected email doesn't match the actual email, the actual (adjusted) email is written to disk. This makes it easier to keep email fixtures up-to-date with content changes - the tests fail, and you copy the actual fixtures over the expected fixtures.
* Add four email fixtures to test email output. I'll add more for inline images and other cases in the future.
* Change tests to mock `emails.views.get_message_content_from_s3` instead of `_get_text_html_attachments`, often using `create_email_from_notification`. Loading for S3 will remain in the new email handler, but extracting content with `_get_text_html_attachments` is going away.
* Switch email creation to use `EmailMessage` instead of the lower-level `MIMEMultipart` type. This had a relatively small impact on the output email, switching around some headers.

The less consequential changes:

* Add a README to the fixtures folder.
* Refactor fixture loading into generic function `load_fixtures` and related functions.
* Refactor mock of `emails.views.remove_message_from_s3` to the test setup. This saves several lines of code in tests.
* Fix a few errors in tests, some caught by mypy once tests were annotated.

How to test:

- [ ] Test that email forwarding still works. I trust the unit tests, but a test run in Heroku would confirm.


emails/tests/fixtures/README.md
```diff
@@ -0,0 +1,59 @@
...
+- `single_recipient_expected.email`
+- `single_recipient_list_expected.email`
+
+When the expected mail does not match the actual forwarded mail, the test creates a file
+with the actual output (after standardization). These files end in `_actual.email`
+instead of `_expected.email`. These can be used in a different tool, such as XCode's
+[Opendiff][], to view the differences. They can also be used when changing the email
+content. Once the developer is happy with the new content, they can replace
+`_expected.email` with the contents of `_actual.email`. These files are ignored by git,
+via the `.gitignore` file in this directory.
```
groovecoder, line 43: suggestion (non-blocking): Can we add 1 more sentence here with an example of **_why_** the expected mail wouldn't match the actual forwarded mail?

emails/tests/fixtures/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+Some fixtures represent SNS Bounce notifications:
+
+- `hard_bounce_sns_body.json` - A hard bounce. Re-trying will not fix the issue.
+- `soft_bounce_sns_body.json` - A soft bounce. Re-trying may fix the issue.
+- `spam_bounce_sns_body.json` - A soft bounce. Sending with different content may work.
+
+### Other notifications
+
+This SNS notification is not handled by our system.
```
groovecoder, line 57: question (non-blocking): can we remove this fixture then?

emails/tests/fixtures/s3_stored_replies_expected.email
```diff
@@ -0,0 +1,29 @@
+Subject: Re: Test Mozilla User New Domain Address
+From: a1b2c3d4@test.com
+To: sender@external.example.com
+Reply-To: a1b2c3d4@test.com
```
groovecoder, line 4: question (non-blocking): When/why do we have `test.com` in the `Reply-To` header sometimes, and other times we have `default.com` in the `Reply-To` header?

emails/tests/views_tests.py
```diff
@@ -109,12 +112,76 @@
 )
 
 
+def create_email_from_notification(
```
groovecoder, line 115: praise: nice and useful helper function!

emails/tests/views_tests.py
```diff
@@ -109,12 +112,76 @@
...
+    email.set_content(text)
+    return email.as_bytes()
+
+
+def assert_email_equals(output_email: str, name: str) -> None:
+    """
+    Assert the output equals the expected email, after replacements.
+
+    The MIME boundary strings have random content, are replaced with
+    "==[BOUNDARY#]==", with a number based on order found in the email.
```
groovecoder, line 143: question (non-blocking): I don't understand why the MIME boundary strings have random content? Is that just how the `SESClient.send_raw_email` function works? And since we mock it, we "inherit" that behavior?

emails/tests/views_tests.py
```diff
@@ -138,9 +205,11 @@ def setUp(self):
         send_raw_email_patcher.start()
         self.addCleanup(send_raw_email_patcher.stop)
 
-    def get_details_from_mock_send_raw_email(self) -> tuple[str, str, dict[str, str]]:
+    def get_details_from_mock_send_raw_email(
+        self,
+    ) -> tuple[str, str, dict[str, str], str]:
```
groovecoder, line 210: joke: It's `str`s all the way down.

groovecoder (approved):
Whew, that's a lot of good refactoring. No blockers - just a bunch of non-blocking questions or suggestions.

emails/tests/fixtures/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Incoming Emails
+
+Some fixtures capture the [SNS][] Delivery [notification JSON][] representing an
+incoming email. Their names all end in `_email_sns_body.json`.
+
+- `dmarc_failed_email_sns_body.json` - Email that failed the [DMARC][] check,
+  and the relevant domain's policy is to reject the email.
+- `domain_recipient_email_sns_body.json` - Email to a premium domain email mask.
+- `replies_email_sns_body.json` - Email from a Relay user
```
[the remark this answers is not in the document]
jwhitlock: I'm going to be more assertive, since we know this is a reply to a previous message

emails/tests/fixtures/README.md
jwhitlock: Sounds good, added "They can also be used when changing the email content. *For example, if we change the
format of the HTML wrapper, then the email content will change*."

emails/tests/fixtures/README.md
jwhitlock: Maybe, but that would be a separate effort. It is a valid SNS message, just one we don't currently handle. It is used for test coverage at:

https://github.com/mozilla/fx-private-relay/blob/29512b38ad47e1103a9cbdc683b3e28888824b6d/emails/tests/views_tests.py#L705-L709

The SNS notification is received as part of the email setup:

https://github.com/mozilla/fx-private-relay/blob/main/docs/end-to-end-local-email-dev.md#create-sns-topic-subscription-that-sends-https-posts-to-your-local-server

In the future, we may want to handle this or other SNS messages differently, or notify an admin for unexpected ones.

emails/tests/fixtures/s3_stored_replies_expected.email
jwhitlock: Good question, looking into it...

emails/tests/views_tests.py
jwhitlock: [RFC 1521, section 7.2.1, The Content-Type Header Field](https://www.rfc-editor.org/rfc/rfc1521#section-7.2.1) is lengthy, but the key bit is:

> Encapsulation boundaries must not appear within the encapsulations,
>  and must be no longer than 70 characters, not counting the two
>  leading hyphens.

The Python library handles this by generating a large random value, then searching the email content for that value, and expanding it if found:

https://github.com/python/cpython/blob/22a8efa7408e00552c9845420caf925685dde92f/Lib/email/generator.py#L367-L388

This means the MIME boundaries change every time you run the test. Some Relay-generated boundaries looks like:

```
--===============3715842840874147262==
--===============5503079870118974321==
```

If you "view source" on your own emails, you'll see different methods used by different providers. Here's a few I found:

```
--91066fc9851403c96763f5d61c0a3917177cf1a64591a02b6e6bb5efb00b
--6ff9379f9f7fe9a87181d055163c0f4c5b4646e16fa0ad005c3144927afb
--=-fT/v9bIhfLoYa9WR+nQldA==
--9Ev1BlIFPznK=_?:
------=_Part_269670996_371849946.1695657797625
```

emails/tests/fixtures/s3_stored_replies_expected.email
jwhitlock: Here's the `Reply-To` header for each:

- `domain_recipient_expected.email` - `Reply-To: replies@default.com`
- `s3_stored_replies_expected.email` - `Reply-To: a1b2c3d4@test.com`
- `single_recipient_expected.email` - `Reply-To: replies@default.com`
- `single_recipient_list_expected.email` - `Reply-To": "replies@default.com`

The `s3_stored_replies_expected.email` is a reply from the Relay user back to the external email. So, if the external sender wants to reply, they should reply to the email mask. For tests, that domain is `test.com`.

https://github.com/mozilla/fx-private-relay/blob/29512b38ad47e1103a9cbdc683b3e28888824b6d/emails/models.py#L40-L47

emails/tests/views_tests.py
groovecoder: Ah, that makes sense. With that in mind, I wrote this doc-block out to myself to try to make it more clear. Take it or leave it as another way to write this doc-block:

```
    Per RFC 1521, 7.2.1, MIME boundary strings must not appear in the bounded content.
    So python's email library generates a large random value for email boundary strings.
    So python makes different MIME boundary strings every time a test is run.
    So replace python's random boundary strings with a generic "==[BOUNDARY#]==",
    where the number is not random, but rather based on order found in the email.
```

jwhitlock: OK, expanding the doc string with this content.

PR #3912, opened 2023-09-22
MPP-3446 and MPP-3450 Labels Misalignment Fix

rafeerahman:
This PR fixes MPP-3446 and MPP-3450.

# Bug description
The promotions label next to the relay mask can go off the page when the mask is long. This issue was caused by a fix I made for MPP-3380. 

Also, on mobile screens such as an iPhone 12, the copied label causes the page width to be smaller then the screen size.

Bug 1             |  Bug 2
:-------------------------:|:-------------------------:
Untitled  |  Untitled

# Screenshot of fix

Copied label will be placed above the correctly aligned promo label, and dissapears after a few seconds.
image

Better positioning for copied label in mobile screens to avoid going off page.
image

# How to test
1. Go to [Relay](https://stage.fxprivaterelay.nonprod.cloudops.mozgcp.net/accounts/profile/) dashboard and sign into a premium account
2. Click “Generate mask” and enter a long custom domain (60 chars)
3. Click on the block promotions checkbox, and generate the mask.
4. [Acceptance Criteria] Observe that the “Block promotions” label is aligned correctly.
5. Enter a mobile viewport via inspector (i.e iPhone 12 or others)
6. [Acceptance Criteria] Click copy on the relay mask, observe that the label does not go off the page.

### DIFF
--- a/frontend/src/components/dashboard/aliases/MaskCard.module.scss
+++ b/frontend/src/components/dashboard/aliases/MaskCard.module.scss
@@ -90,10 +90,9 @@
         @media screen and (min-width: $content-lg) {
           max-width: 400px;
         }
-        @media screen and (min-width: calc($screen-lg + 120px)) {
-          max-width: calc(
-            $content-md - 100px
-          ); // Handles long custom domains, prevents going off page.
+        // Stops long mask domain from cutting into the "Block promotions label"
+        @media screen and (min-width: calc($content-xl)) {
+          max-width: calc($content-md - 100px);
         }
       }
 
@@ -151,7 +150,7 @@
     color: $color-grey-40;
     border-radius: $border-radius-sm;
     width: $layout-sm;
-    z-index: 999;
+    z-index: 3; // Ensures that the button gets placed above the copied label (needed for long custom domains at viewport width ~ 780px)
 
     @media (any-pointer: coarse) {
       width: $layout-md;
--- a/frontend/src/components/dashboard/aliases/MaskCard.module.scss
+++ b/frontend/src/components/dashboard/aliases/MaskCard.module.scss
@@ -48,10 +48,12 @@
     .copy-button-wrapper {
       display: flex;
       align-items: center;
-      position: relative;
 
       @media screen and #{$mq-md} {
         gap: $spacing-sm;
+      }
+      @media screen and (max-width: $screen-xs) {
+        overflow: hidden;
       }
     }
 
@@ -112,7 +114,7 @@
       @media screen and (max-width: $screen-md) {
         @include text-body-sm;
       }
-      @media screen and (max-width: 380px) {
+      @media screen and (max-width: $content-sm) {
         position: absolute;
         right: 0;
       }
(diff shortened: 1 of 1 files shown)


frontend/src/components/dashboard/aliases/MaskCard.module.scss
```diff
@@ -142,6 +151,7 @@
     color: $color-grey-40;
     border-radius: $border-radius-sm;
     width: $layout-sm;
+    z-index: 999;
```
lloan, line 154: suggestion: We definitely want to add a comment here and want to stay away from using such a big number for z-index. 

Reason why we want to be mindful with z-index values: 

- Maintainability: Using arbitrary large numbers makes the code harder to understand and maintain. Future US may struggle to understand why a particular z-index value was set so high (most likely me :)).

- Conflicts: If you use a very high z-index value, you might inadvertently cause stacking issues with other elements, especially third-party components that also use z-index - in our case, we use very little third-party CSS (protocol?), but its a good practice.
 
- Debugging: It becomes difficult to debug issues related to stacking context when z-index values are set arbitrarily high. You may end up spending more time figuring out the correct z-index than you would if you had a more organized approach.
 
- Semantics: z-index values are meant to be relative to each other, so it's easier to understand the stacking context when the numbers are closer together and logically organized.
 
- Performance: While not a significant issue, browsers do have to keep track of the stacking context. Using unnecessarily large numbers could, in theory, consume more memory, although this is not going to break our project 🤞🏻 .

I think overall, its important because it makes debugging easier - easier to identify issues related to stacking context specifically. Also, makes it easier for team members to understand the stacking context, which is particularly useful in larger projects like this one.

We should be able to get by with z-index values like 1, 2, 3, etc., and reserve higher values for elements that genuinely need to be above everything else, just not 999+ (hopefully).

lloan (changes requested):
@rafeerahman two minor CSS issues, I was there when you demoed this and it works great.

frontend/src/components/dashboard/aliases/MaskCard.module.scss
```diff
@@ -83,7 +90,7 @@
         @media screen and (min-width: $content-lg) {
           max-width: 400px;
         }
-        @media screen and #{$mq-lg} {
+        @media screen and (min-width: calc($screen-lg + 120px)) {
```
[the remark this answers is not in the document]
rafeerahman: Thanks! I added a token instead

frontend/src/components/dashboard/aliases/MaskCard.module.scss
rafeerahman: Thanks for the detailed review! I made a change now.

frontend/src/components/dashboard/aliases/MaskCard.module.scss
```diff
@@ -83,10 +90,9 @@
         @media screen and (min-width: $content-lg) {
           max-width: 400px;
         }
-        @media screen and #{$mq-lg} {
-          max-width: calc(
-            $content-md - 100px
-          ); // Handles long custom domains, prevents going off page.
+        // Stops long mask domain from cutting into the "Block promotions label"
+        @media screen and (min-width: calc($content-xl)) {
```
lloan, line 94:
```suggestion
        @media screen and (min-width: $content-xl) {
```

lloan (approved):
Looks good to me. Great job with the comments! Will definitely save us time in the future! 

I left one minor change, I don't think we need the `calc()` anymore, but approving pending change. 

Thanks!

PR #3913, opened 2023-09-22
MPP-3407: Log SNS notification details

jwhitlock:
After PR #3891 shipped, I had more data for MPP-3407. Two types of addresses caused `InvalidHeaderDefect`: 

* UTF-8 encoded addresses, like `=?utf-8?b?dGVzdEBleGFtcGxlLmNvbQ==?=` for `test@example.com`
* Unicode strings that do not appear to be email addresses, like `EÍlÍoÍnÍ`

and MPP-3417, `IndexError`, the `From` address is always an empty string.

I was unable to replicate these issues locally, because I couldn't get my mail providers to send emails with UTF-8 encoded headers, weird Unicode, or blank `From` addresses.

I am curious if this is a decoding error, or a double-encoding error, or something else. I'm logging the multiple ways the `From:` header can be encoded in the AWS SNS notification. I'm also "handling" the error by return a 503, so we should stop getting tracebacks in Sentry, just logs.

How to test:

### DIFF
--- a/emails/tests/utils_tests.py
+++ b/emails/tests/utils_tests.py
@@ -1,5 +1,4 @@
 from base64 import b64encode
-from email.errors import InvalidHeaderDefect
 from email.utils import parseaddr
 from typing import Literal
 from urllib.parse import quote_plus
@@ -16,6 +15,7 @@ from emails.utils import (
     generate_from_header,
     get_email_domain_from_settings,
     remove_trackers,
+    InvalidFromHeader,
 )
 
 from .models_tests import make_free_test_user, make_premium_test_user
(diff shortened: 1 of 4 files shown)


emails/tests/utils_tests.py
```diff
@@ -133,24 +133,21 @@ def test_generate_from_header(params: _GENERATE_FROM_TEST_CASE_DEF) -> None:
...
+    "no_domain": '"I am groot" <groot>',
+    "no_address": '"I am groot" <>',
+    "utf_8_encoded": _encode_as_base64_utf8_str('"I am groot" <groot@groot.gr>'),
+}
+
+
+@pytest.mark.parametrize(
+    "address",
+    GENERATE_FROM_HEADER_RAISES_CASES.values(),
+    ids=GENERATE_FROM_HEADER_RAISES_CASES.keys(),
```
groovecoder, line 146: praise: TIL `parametrize`'s `ids` parameter.

groovecoder (approved):

PR #3931, opened 2023-09-26
MPP-55: Use ASCII in wrapped HTML email when possible

jwhitlock:
Add a new Django temple filter `convert_fsi_to_span` that converts the Unicode bidi control characters to HTML equivalents. 

The one string with these characters is:

```
&#x2068;0&#x2069; email trackers removed
```

However, instead of the HTML escaped characters, the string contains the Unicode characters. `&#x2068` is the [First Strong Isolate (FSI)](https://www.compart.com/en/unicode/U+2068), which means "start a span of text where the text direction is determined by the first content character with a preferred text direction". The character `&#x2069;` is [Pop Directional Isolate (PDI)](https://www.compart.com/en/unicode/U+2069), which ends that span of text.

The filter changes the output to:
```
<span dir="auto">0</span> email trackers removed
```

This is the HTML equivalent of these characters, according to [How to use Unicode controls for bidi text](https://www.w3.org/International/questions/qa-bidi-unicode-controls) from w3c.org.

This allows our test emails to be encoded in 7-bit ASCII rather than base64-encoded UTF-8, so that you can see the actual content in the raw email. If the email is translated, it will still require UTF-8, although we could do work to replace these with the HTML escape codes if we want these in 7-bit ASCII as well./


emails/tests/fixtures/domain_recipient_expected.email
```diff
@@ -22,127 +22,157 @@ should work
 
 Sent with [ProtonMail](https://protonmail.com/) Secure Email.
 --==[BOUNDARY1]==
-Content-Type: text/html; charset="utf-8"
+Content-Type: text/html; charset="us-ascii"
 MIME-Version: 1.0
-Content-Transfer-Encoding: base64
+Content-Transfer-Encoding: 7bit
```
groovecoder, line 27: question (blocking): Doesn't AWS SES (or even the sending MTA?) control the `Content-Type` and `Content-Transfer-Encoding` ? So some of our fixtures should still use `utf-8` `base64` right? To make sure we're still handling those ?

emails/views.py
```diff
@@ -134,13 +134,13 @@ def reply_requires_premium_test(request):
...
-    num_level_one_email_trackers_removed=None,
-    tracker_report_link=0,
-):
+    original_html: str,
+    language: str,
+    has_premium: bool,
+    display_email: str,
+    num_level_one_email_trackers_removed: int | None = None,
+    tracker_report_link: str | None = None,
+) -> str:
```
groovecoder, line 143: praise: 😍 types!

emails/views.py
```diff
@@ -431,6 +434,37 @@ def _sns_message(message_json: AWS_SNSMessageJSON) -> HttpResponse:
...
+    * 200 if the email was sent, the Relay address is disabled, the Relay user is
+      flagged for abuse, the email is under a bounce pause, the email was suppressed
+      for spam, the list email was blocked, or the noreply address was the recipient.
+    * 400 if commonHeaders entry is missing, the Relay recipient address is malformed,
+      the email failed DMARC with reject policy, or the email is a reply chain to a
+      non-premium user.
+    * 404 if an S3-stored email was not found, no Relay address was found in the "To",
+      "CC", or "BCC" fields, or the Relay address is not in the database.
+    * 503 if the "From" address is malformed, the S3 client returned an error different
+      from "not found", or the SES client fails
```
groovecoder, line 459: praise: 😍 code comments and docs!

emails/views.py
```diff
@@ -738,6 +746,67 @@ def _strip_localpart_tag(address):
...
+    language: str,
+    has_premium: bool,
+    sample_trackers: bool,
+    remove_level_one_trackers: bool,
+    now: datetime | None = None,
+) -> tuple[str, int]:
+    # frontend expects a timestamp in milliseconds
+    now = now or datetime.now(timezone.utc)
+    datetime_now_ms = int(now.timestamp() * 1000)
+    # user_profile = address.user.profile
```
groovecoder, line 762: change (blocking): We try to remove commented lines, right?

groovecoder (changes requested):
Some specific changes to make, but also an overall question: this looks like it's changing the charset and encoding that we're using for *ALL* of our emails? And I don't see any tests that make sure that non-ascii characters in the body of emails still work? Is there something that automatically fixes that I'm not seeing?

emails/views.py
jwhitlock: yep, supposed to be removed...

emails/tests/fixtures/domain_recipient_expected.email
jwhitlock: [RFC 5598](https://datatracker.ietf.org/doc/html/rfc5598#section-4.3) calls the first server the Mail Submission Agent (MSA). It verifies the message is formatted correctly, and then sends it on it's way. It can change the form of the data, like the `Content-Transfer-Encoding`, but it does not have to.

I agree, I'll add a non-English test so that we'll have a UTF-8 example.

jwhitlock:
I've removed the commented code (nice catch!) and added a test case where the content requires UTF-8 encoding, because the HTML wrapper is in French.

emails/tests/fixtures/single_recipient_fr_expected.email
```diff
@@ -0,0 +1,146 @@
...
+Content-Type: multipart/alternative;
+ boundary="==[BOUNDARY1]=="
+MIME-Version: 1.0
+
+--==[BOUNDARY1]==
+Content-Type: text/plain; charset="us-ascii"
+MIME-Version: 1.0
+Content-Transfer-Encoding: 7bit
+
+This email was sent to your alias ebsbdsan7@test.com. To stop receiving emails sent to this alias, update the forwarding settings in your dashboard.
```
groovecoder, line 19: question (blocking): Should we test a non-ascii character in the plain text too?

groovecoder (changes requested):

jwhitlock:
I found an error with `create_email_from_notification` from PR #3911, which is fixed here. The headers in the constructed email were being set with blank values. I've added a commit that fixes this, which required a slightly different way to set the text content.

I've added a test with non-ASCII text content, by expanding `test_reply`, which has a ton of setup and calls `create_email_from_notification`.

groovecoder (approved):
Cool, thanks for the extra test!

groovecoder:
Can you make a sub-task of MPP-55 for QA to test email forwards with/without non-ASCII characters and check that the `text/plain` and `text/html` content all works as expected? Just to make sure nothing with the SES config causes unexpected behavior with these changes.

PR #3932, opened 2023-09-26
Add Rafee in the release engineer rotation

say-yawn:


docs/release-engineer-playbook.md
```diff
@@ -15,6 +15,8 @@
 
    b. Co-ordinate, re-tag for [stage fixes][stage-fixes] as needed
 
+   - Latest tag to auto-deploy to Stage should not happen after Thursday to ensure stage is stable and had plenty of time to be test by QA before the tag is pushed to production on Tuesday.
+
```
lloan, line 19:
```suggestion
   - Latest tag to auto-deploy to Stage should not happen after Thursday to ensure stage is stable and had plenty of time to be tested by QA before the tag is pushed to production on Tuesday.

```


Suggestion (non-blocking): Perhaps something like this would also work? 

_To ensure stability and allow ample time for QA testing, the latest tag should be auto-deployed to the staging environment no later than Thursday. This will facilitate a smooth transition when pushing the tag to production on the following Tuesday._

lloan (approved):
Left a non-blocking comment (typo) and a content suggestion.

docs/release-engineer-playbook.md
groovecoder: 👍  @lloan 

The rest of the sentences in this doc all use active 2nd-person voice: "Ping all the engineers ...", "Confirm any hotfixes ...", "Coordinate, re-tag for stage fixes ..."

To keep the same voice and tone: "Don't tag after Thursday, so QA has a stable stage environment to test before the Tuesday deployment to production."

PR #3937, opened 2023-09-29
MPP-55: Test more emails outside of SNS notifications

jwhitlock:
This PR adds new tests for the existing email processing pipeline, before it is replaced to pass inline images in forwarded emails. There is no change to the production code, just tests and docs.

This PR adds a new test helper `create_notification_from_email` that creates a fake SNS notification from an email message. This makes it easier to test header and multipart structures, and compare incoming messages to the Relay forwarded messages.

This PR add new tests and fixtures for incoming emails using `create_notification_from_email`:

* `inline_image_incoming.email` / `_expected.email` - Show that inline images are dropped by the current email forwarding (to be fixed by MPP-55)
* `russian_spam_incoming.email` / `_expected.email` - Show that UTF-8 encoded headers and email bodies are correctly handled
* `plain_text_incoming.email` / `_expected.email` - Show that Relay adds an HTML version to a plain text email when forwarding to the Relay user. This adds the HTML header and footer to the message.

This PR also adds expected email fixtures for the emails sent to free Relay users who attempt to reply to a message:

* `reply_requires_premium_first_expected.email` - The first email tells them the reply was forwarded
* `reply_requires_premium_second_expected.email` - The second and future emails tell them that the reply was _not_ forwarded.

The tests directly call `_build_reply_requires_premium_email` in a different way from the running code. They used to pass the local part of the Relay email mask (`w41fwbt4q`) rather than the full address (`w41fwbt4q@test.com`). This was more visible in the full email, and has been fixed. The local part is no longer randomly generated from test to test, so that the fixture will match between test runs.


emails/tests/fixtures/README.md
```diff
@@ -1,10 +1,21 @@
...
 
-### Incoming Emails
+## Incoming Emails
 
-Some fixtures capture the [SNS][] Delivery [notification JSON][] representing an
+Incoming Email fixtures come in two varieties. The original fixtures capture the [SNS][]
+Delivery [notification JSON][] representing an incoming email. The newer fixtures are
+the raw incoming emails.
+
+Both fixtures have related fixtures for the expected output email.
```
groovecoder, line 11: suggestion (non-blocking): I didn't understand this at first glance. Maybe this?
```suggestion
Every incoming email fixture has a corresponding fixture for the expected output email.
```

emails/tests/fixtures/README.md
```diff
@@ -21,24 +32,47 @@ incoming email. Their names all end in `_email_sns_body.json`.
...
+- `inline_image_incoming.email` - Contains an inline image, referenced in the HTML by
+  content ID
+- `plain_text_incoming.email` - A simple email with only plain text content
+- `russian_spam_incoming.email` - Contains UTF-8 encoded headers, Base64-encoded content
+
+[Internet Message Format]: https://datatracker.ietf.org/doc/html/rfc5322
+
+### Expected output email
+
+There are related fixtures for the expected output email. They are stored in the
```
groovecoder, line 51:
```suggestion
These are related fixtures for the expected output email. They are stored in the
```

emails/tests/fixtures/README.md
```diff
@@ -48,16 +82,17 @@ directory.
...
 - `hard_bounce_sns_body.json` - A hard bounce. Re-trying will not fix the issue.
 - `soft_bounce_sns_body.json` - A soft bounce. Re-trying may fix the issue.
 - `spam_bounce_sns_body.json` - A soft bounce. Sending with different content may work.
 
-### Other notifications
+## Other notifications
 
-This SNS notification is not handled by our system.
+This SNS notification is not handled by our system. It is included to test how our
+system handles unknown notifications.
```
groovecoder, line 96: praise: nice additional info!

emails/tests/fixtures/plain_text_expected.email
```diff
@@ -0,0 +1,179 @@
...
+MIME-Version: 1.0
+Content-Transfer-Encoding: 7bit
+
+This email was sent to your alias ebsbdsan7@test.com. To stop receiving emails sent to this alias, update the forwarding settings in your dashboard.
+---Begin Email---
+This is a text-only email.
+See https://text.example.com
+
+--==[BOUNDARY1]==
+Content-Type: text/html; charset="us-ascii"
```
groovecoder, line 25: comment (non-blocking): This seems strange, but IIRC we intentionally "enhance" text-only emails with an additional `text/html` part so we can put our rich header and footer into them.

emails/tests/views_tests.py
```diff
@@ -137,31 +139,110 @@ def create_email_from_notification(
     return email.as_bytes()
 
 
-def assert_email_equals(output_email: str, name: str) -> None:
+def create_notification_from_email(email_text: str) -> AWS_SNSMessageJSON:
```
groovecoder, line 142: praise: great utility function!

emails/tests/views_tests.py
```diff
@@ -137,31 +139,110 @@ def create_email_from_notification(
...
+                "to": [str(addr) for addr in email["To"].addresses],
+                "messageId": email["Message-ID"],
+                "subject": email["Subject"],
+            },
+        },
+        "receipt": {
+            "timestamp": (email["Date"].datetime + timedelta(seconds=1)).isoformat(),
+            "processingTimeMillis": 1001,
+            "recipients": [addr.addr_spec for addr in email["To"].addresses],
+            "spamVerdict": {"status": "PASS"},
```
groovecoder, line 173: question (non-blocking): does this need to be `"status": "FAIL"` for the russian spam fixture?

emails/tests/views_tests.py
```diff
@@ -137,31 +139,110 @@ def create_email_from_notification(
...
+    base_url = "https://sns.us-east-1.amazonaws.example.com"
+    sns_notification = {
+        "Type": "Notification",
+        "MessageId": str(uuid4()),
+        "TopicArn": topic_arn,
+        "Subject": email["Subject"],
+        "Message": json.dumps(sns_message),
+        "Timestamp": (email["Date"].datetime + timedelta(seconds=2)).isoformat(),
+        "SignatureVersion": "1",
+        "Signature": "invalid-signature",
```
groovecoder, line 191: question (non-blocking): do we want the SNS to be hard-coded to an invalid signature?

emails/tests/views_tests.py
```diff
@@ -520,6 +606,112 @@ def test_invalid_from_header(self, mock_logger, mock_generate_from_header) -> No
...
+        test_sns_notification = create_notification_from_email(email_text)
+        _sns_notification(test_sns_notification)
+
+        self.mock_send_raw_email.assert_called_once()
+        sender, recipient, headers, email = self.get_details_from_mock_send_raw_email()
+        assert sender == "replies@default.com"
+        assert recipient == "user@example.com"
+
+        # Subject translates as "Invitation | Why and how to do business in Africa?"
+        # Because of Russian characters, it is UTF-8 text encoded in Base63
```
groovecoder, line 659:
```suggestion
        # Because of Russian characters, it is UTF-8 text encoded in Base64
```

groovecoder (approved):
no blockers. looks good!

emails/tests/views_tests.py
jwhitlock: Maybe, but we're not testing spam handling here. That's `spamVerdict_FAIL_email_sns_body.json`. If we can get a legitimate email with a lot of non-ASCII content, we should replace this fixture to remove confusion

emails/tests/views_tests.py
jwhitlock: Yes, that's why it is hard-coded to the string "invalid-signature". It is supposed to be acceptable to `_sns_notification`, but not `sns_inbound`. I think it makes sense to clarify this in the docstring.

I think the Relay architecture for email processing is flawed, with the AWS provider details remaining a part of processing from start to finish. It restricts us to the one provider, and makes it harder to experiment with other systems in the future. If I had my preferences, I would not need to wrap emails in SNS structures at all, but could test them directly. We don't need that feature yet, so it will need to wait.

jwhitlock (commented):
Thanks for the eagle-eyed review, @groovecoder! I've made some changes:

* Fixed the `Base63` typo
* Expanded the docs for `create_notification_from_email`, to clarify what it is _not_ designed to do (create a notification that would pass verification, fully simulate AWS receipts).
* Re-arrange the "Expected Outgoing Emails" section to (maybe) remove confusion
* Add a TODO to create a better fixture than the Russian spam fixture.

PR #3955, opened 2023-10-03
fix MPP-3420: handle and log NoReverseMatch error

groovecoder:
This PR fixes #MPP-3420.

How to test:
1. Run `pytest`
   * [ ] All tests should pass
2. Go to http://127.0.0.1:8000/accounts/logout/?next=fuzzymcfuzzface
   * [ ] No error; it should redirect to the home page
   * [ ] Check the `runserver` output for an error message in the log

### DIFF
--- a/api/tests/views_tests.py
+++ b/api/tests/views_tests.py
@@ -1,4 +1,6 @@
 from datetime import datetime
+import logging
+from allauth.account.models import EmailAddress
 import pytest
 from model_bakery import baker
 import responses
@@ -25,6 +27,7 @@ from api.tests.authentication_tests import (
 from api.views import FXA_PROFILE_URL
 from emails.models import Profile, RelayAddress
 from emails.tests.models_tests import make_free_test_user, make_premium_test_user
+from privaterelay.tests.utils import log_extra
 
 
 @pytest.fixture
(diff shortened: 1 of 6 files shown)


say-yawn (changes requested):
There are [three additional paths](https://mozilla.sentry.io/issues/?project=4503976951152641&query=is%3Aunresolved+NoReverseMatch&referrer=issue-list&statsPeriod=14d) that also have `NoReverseMatch` error. Since the `is_safe_url` is part of `AccountAdapter` for django-allauth I was hoping this would handle any `NoReverseMatch` error from `/accounts` but that was not the case. Some examples I tested were:
- `/accounts/fxa/login/callback/`
- `/accounts/logout/fuzzy`

Can you see if there is a way to expand the work to apply for these other cases? Otherwise, the `NoReverseMatch` error happening from invalid query param and value like:
- `/accounts/logout?next=fuzzy` - 
- `/accounts/logout?next='fuzzy'`
are handled with `is_safe_url`
are handled

api/views/__init__.py
```diff
@@ -234,12 +236,26 @@ def terms_accepted_user(request):
...
+        try:
+            complete_social_login(request, social_login)
+            # complete_social_login writes ['account_verified_email', 'user_created', '_auth_user_id', '_auth_user_backend', '_auth_user_hash']
+            # on request.session which sets the cookie because complete_social_login does the "login"
+            # The user did not actually log in, logout to clear the session
+            if request.user.is_authenticated:
+                get_account_adapter(request).logout(request)
+        except NoReverseMatch as e:
+            # TODO: use this logging to fix the underlying issue
+            # https://mozilla-hub.atlassian.net/browse/MPP-3473
```
groovecoder, line 248: @say-yawn

api/tests/views_tests.py
```diff
@@ -571,3 +573,48 @@ def test_fxa_responds_with_no_fxa_uid_returns_404_and_cache_returns_404(self):
...
+        json=profile_json,
+    )
+
+    response = client.post("/api/v1/terms-accepted-user/")
+
+    assert response.status_code == 500
+    (rec1,) = caplog.records
+    assert "socialaccount_signup" in rec1.message
+    assert "fxa_uid" in rec1.message
+    assert "social_login.state" in rec1.message
```
groovecoder, line 620: @jwhitlock : this `caplog` doesn't seem to capture the logs from [the `logger.error` call ](https://github.com/mozilla/fx-private-relay/pull/3955/files#diff-c8f346bb08a272787ee05a4840681436e18932b82216fac4fb8b0530bb839379R250-R256)? Any idea why not?

api/views/__init__.py
```diff
@@ -234,12 +236,26 @@ def terms_accepted_user(request):
...
+            if "socialaccount_signup" in e.args[0]:
+                logger.error(
+                    str(e),
+                    extra={
+                        "fxa_uid": fxa_uid,
+                        "social_login.state": social_login.state,
+                    },
+                )
```
jwhitlock, line 256: The varied message will make it hard to find this in the logs. The period in `social_login.state` will also cause headaches. I suggest a static string for the message, and the full exception in the extra:

```suggestion
                logger.error(
                    'socialaccount_signup_error'
                    extra={
                        "exception": str(e),
                        "fxa_uid": fxa_uid,
                        "social_login_state": social_login.state,
                    },
                )
```

api/tests/views_tests.py
jwhitlock, line 620: Try this:
```suggestion
    assert rec1.fxa_uid == uid
    assert getattr(rec1, 'social_login.state') == {}
```

The `getattr` is needed because you used a period in the name, I suggest you don't do that.

jwhitlock (changes requested):
@groovecoder here's some suggested changes for the logger error

jwhitlock (approved):
👍 looks good, let's get some data!

say-yawn (approved):
👍 Sentry issue cleanup and logging to better handle other issue!

PR #3959, opened 2023-10-04
MPP-3407: Change how From: header is parsed

jwhitlock:
In some cases, `email.utils.parseaddr` does not correctly parse the `From:` header passed from AWS. Three cases from yesterday:

* A value encoded with `windows-1252`, like `=?windows-1252?Q?Place_de_march=E9?= <email@example.com>`
* A value with an unquoted comma, `Like, This! <email@example.com>`
* A value with embedded brackets, `Like This <Like This <email@example.com>>`

This PR adds a util function`parse_email_header`, which uses the lower-level `email.headerregistry.AddressHeader` to parse the raw header, and return only the results with an email in the `addr_spec` part (or at least a string with one `@`). It improves handling of two of these cases:

* Decodes the `windows-1252` string correctly
* Mangles the display name to `This!`, but finds the email `email@example.com`
* Can not parse the embedded brackets

This PR updates `_handle_received` to use `parse_email_header`. If no email is found, it logs an error and returns 400, stopping processing.

How to test:


emails/tests/fixtures/emperor_norton_incoming.email
```diff
@@ -0,0 +1,20 @@
...
+virtue of the authority thereby in me vested, do hereby order and direct the
+representatives of the different States of the Union to assemble in Musical
+Hall, of this city, on the 1st day of February next, then and there to make
+such alterations in the existing laws of the Union as may ameliorate the evils
+under which the country is laboring, and thereby cause confidence to exist,
+both at home and abroad, in our stability and integrity.
+
+NORTON I.,
+Emperor of the United States.
+17th September, 1859
```
groovecoder, line 20: dang, sounds like I missed a rockin' party.

emails/tests/utils_tests.py
```diff
@@ -145,6 +146,69 @@ def test_generate_from_header_raises(address) -> None:
...
+            (
+                "sos accessoire \\(Place de marché Cdiscount\\) ",
+                "vendeur@sn.example.com",
+            ),
+        ],
+    },
+}
+
+
+@pytest.mark.parametrize(
```
groovecoder, line 202: praise: I do love some `parametrize`

emails/tests/views_tests.py
```diff
@@ -711,6 +711,82 @@ def test_plain_text(self) -> None:
...
+        """
+        A From: header with commas in an unquoted display is forwarded.
+
+        AWS parses these headers as a single email, Python as a list of emails.
+        One of the root causes of MPP-3407.
+        """
+        email_text = EMAIL_INCOMING["emperor_norton"]
+        test_sns_notification = create_notification_from_email(email_text)
+
+        # Python parses differently than AWS
```
groovecoder, line 724: question (non-blocking): do we need these `Python parses differently than AWS` blocks of assertions? It looks like they are effectively asserting that `create_notification_from_email` and `json.loads` work like we expect?

emails/views.py
```diff
@@ -479,7 +480,18 @@ def _handle_received(message_json: AWS_SNSMessageJSON) -> HttpResponse:
         return HttpResponse("Address does not exist", status=404)
 
     _record_receipt_verdicts(receipt, "relay_recipient")
-    from_address = parseaddr(common_headers["from"][0])[1]
+    from_addresses = parse_email_header(common_headers["from"][0])
```
groovecoder, line 483: question (non-blocking): should we also change the `parseaddr` calls in `_handle_bounce` and `_handle_complaint` to call `parse_email_header` ? (We're also using `parseaddr` in `_get_relay_recipient_from_message_json`, but since those are our own email addresses, we seem to send all those with well-formed addresses.

groovecoder (approved):
No blockers.

emails/tests/views_tests.py
jwhitlock: Hmm, now that I think about it, this implies these _should_ be SNS JSON fixtures, since the trip from a raw email doesn't work.

emails/views.py
jwhitlock: maybe yes, and the code that parses the `To:` and `CC:` fields as well. This may be bigger than a "quick tech-debt" issue.

groovecoder: I'm still happy to merge this as an incremental improvement on what we have now.

jwhitlock:
Force-push to rebase and convert the fixtures from `_incoming.email` to `_sns_email_body.json` fixtures. This makes the tests much cleaner. Nice suggestion, @groovecoder!

PR #3983, opened 2023-10-09
MPP-3284: Log country determination

jwhitlock:
This PR addresses MPP-3284. It updates `_get_cc_from_request` in `privaterelay/utils.py` to log the details of the user's country code selection. The details are extra fields on the "eventsinfo" log message "region_details":
    - selected - which method was selected, one of "cdn", "accept_lang", "fallback" (US)
    - region - the region code returned from the function
    - cdn_region - The region code returned from the `X-Client-Region` header, or omitted if none
    - accept_lang_region - The region code determined from the `Accept-Language` header, or omitted if none

This log message is omitted once per request, by adding a flag on the request.

I originally wanted to do this logging in the `ResponseMetrics` middleware, by adding the region details as an attribute of the `HttpRequest` object. However, Django REST Framework doesn't use the original request, but recreates it as [its own Request](https://www.django-rest-framework.org/api-guide/requests/), losing any attributes it doesn't know about. I've addressed this issue on past projects, but didn't want to add that to this PR. I did retain the test and type annotations for `ResponseMetrics`. 

## How to test:

Optional test steps to see the log message and metrics:

* Add to `.env`:
   ```sh
   DJANGO_STATSD_ENABLED=True
   STATSD_DEBUG=True
   ```
* Run `./manage.py runserver`
* Visit http://127.0.0.1:8000/api/v1/runtime_data/ . Among others, you'll see a log entry like:
   ```json
  {"Timestamp": 1696881837835036928, "Type": "eventsinfo", "Logger": "fx-private-relay", "Hostname": "jwhitlock-MB-M2.local", "EnvVersion": "2.0", "Severity": 6, "Pid": 58553, "Fields": {"accept_lang_region": "US", "selected": "accept_lang", "region": "US", "msg": "region_details"}}
   ```
* Visit http://127.0.0.1:800/ , log in. Amoung others, you'll see log entries like:
   ```json
   {"Timestamp": 1696884254750748928, "Type": "markus", "Logger": "fx-private-relay", "Hostname": "jwhitlock-MB-M2.lan", "EnvVersion": "2.0", "Severity": 6, "Pid": 60055, "Fields": {"msg": "METRICS|2023-10-09 20:44:14|timing|fx.private.relay.response|372.32184410095215|#status:200,view:api.views.RelayAddressViewSet,method:GET"}}
[...]

PR #3984, opened 2023-10-09
MPP-3284: Log details of region determination

jwhitlock:
This PR addresses MPP-3284. It updates `_get_cc_from_request` in `privaterelay/utils.py` to log the details of the user's country code selection. The details are extra fields on the "eventsinfo" log message "region_details":
    - ~`selected`~ `region_method` - which method was selected, one of "cdn", "accept_lang", "fallback" (US)
    - `region` - the region code returned from the function
    - `cdn_region` - The region code returned from the `X-Client-Region` header, or omitted if none
    - `accept_lang` - The user's `Accept-Language` header
    - `accept_lang_region` - The region code determined from the `Accept-Language` header, or omitted if none

This log message is omitted once per request, by adding a flag on the request.

I originally wanted to do this logging in the `ResponseMetrics` middleware, but that had some problems. See PR #3983. I may pick up that idea again in the future.

## How to test:

Optional test steps to see the log message:

* Run `./manage.py runserver`
* Visit http://127.0.0.1:8000/api/v1/runtime_data/ . Among others, you'll see a log entry like:
   ```json
  {"Timestamp": 1696881837835036928, "Type": "eventsinfo", "Logger": "fx-private-relay", "Hostname": "jwhitlock-MB-M2.local", "EnvVersion": "2.0", "Severity": 6, "Pid": 58553, "Fields": {"accept_lang": "en-US, en", "accept_lang_region": "US", "region_method": "accept_lang", "region": "US", "msg": "region_details"}}
   ```

### DIFF
--- a/privaterelay/utils.py
+++ b/privaterelay/utils.py
@@ -2,6 +2,7 @@ from decimal import Decimal
 from functools import wraps
 from string import ascii_uppercase
 from typing import Callable, TypedDict
+import logging
 import random
 
 from django.conf import settings
@@ -17,6 +18,8 @@ from waffle.utils import (
 )
 
 from .plans import PlanCountryLangMapping, CountryStr
+
+info_logger = logging.getLogger("eventsinfo")
 
 
 class CountryInfo(TypedDict):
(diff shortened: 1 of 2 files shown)


privaterelay/utils.py
```diff
@@ -55,11 +58,37 @@ def get_countries_info_from_lang_and_mapping(
...
+    """Determine the user's region / country code."""
+
+    log_data: dict[str, str] = {}
+    cdn_region = None
+    region = None
     if "X-Client-Region" in request.headers:
-        return request.headers["X-Client-Region"].upper()
+        cdn_region = region = request.headers["X-Client-Region"].upper()
+        log_data["cdn_region"] = cdn_region
+        log_data["selected"] = "cdn"
```
groovecoder, line 69: suggestion (non-blocking): `selected` is ambiguous. When I first saw it I expected `selected` to mean "Which region was selected" - e.g., `selected: DE`. Can we change it to `region_selection_method` to be more explicit?

privaterelay/utils.py
```diff
@@ -55,11 +58,37 @@ def get_countries_info_from_lang_and_mapping(
...
+            region = accept_language_region
+            log_data["selected"] = "accept_lang"
+
+    if region is None:
+        region = "US"
+        log_data["selected"] = "fallback"
+    log_data["region"] = region
+
+    # MPP-3284: Log details of region selection. Only log once per request, since some
+    # endpoints, like /api/v1/runtime_data, call this multiple times.
```
groovecoder, line 86: praise: cool, thanks for catching and preventing the double-log.

groovecoder (changes requested):
change (blocking): We seem to be missing a test for `test_get_countries_info_cdn_and_accept_language_uses_cdn`? I.e., to test that a request with BOTH `X-Client-Region` and `Accept-Language` header values will use the region from `X-Client-Region`?

privaterelay/utils.py
jwhitlock: is `region_method` OK? I was mostly avoiding `method`, which is ambiguous with HTTP request method (`POST`, `GET`, etc)

jwhitlock:
Thanks for the review, @groovecoder! I've changed the name from `selected` to `region_method` (code and tests) and added the test for both headers set (test only).

groovecoder (approved):
Looks good! Thanks!

PR #3988, opened 2023-10-10
Fix date for ADR 0001

jwhitlock:
This was done in 2022, not 2023 🙈


pdehaan (approved):
Yep, that's a `3` alright!

groovecoder (approved):

PR #3989, opened 2023-10-11
MPP-3107: get PHONE_RATE_LIMIT value from environment

groovecoder:
This PR is for #MPP-3107.

How to test:
1. `RELAY_CHANNEL="prod" PHONE_RATE_LIMIT="20/minute" python manage.py shell`
2. `from django.conf import settings`
3. `settings.PHONE_RATE_LIMIT`
   * [x] Should return `'20/minute'`
      * Note: On `main` branch this returns `'5/minute'`

image

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -614,9 +614,11 @@ SPECTACULAR_SETTINGS = {
     "SERVE_INCLUDE_SCHEMA": False,
 }
 
-PHONE_RATE_LIMIT = "5/minute"
 if IN_PYTEST or RELAY_CHANNEL in ["local", "dev"]:
-    PHONE_RATE_LIMIT = "1000/minute"
+    _DEFAULT_PHONE_RATE_LIMIT = "1000/minute"
+else:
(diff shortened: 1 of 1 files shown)


jwhitlock (approved):
Thanks @groovecoder!

This is the fastest way to avoid 429's in production (other than disabling rate limiting at all). I still think the right thing would be to rate-limit the `PATCH` requests rather than all phone methods.

If we are configuring from the environment, we should remove the overrides for local and dev. One way to do this incrementally:

```python
if RELAY_CHANNEL in ["local", "dev"]:
   _DEFAULT_PHONE_RATE_LIMIT = "1000/minute"
else:
   _DEFAULT_PHONE_RATE_LIMIT = "5/minute"
PHONE_RATE_LIMIT = config("PHONE_RATE_LIMIT", _DEFAULT_PHONE_RATE_LIMIT)
```

We can then add environment overrides in Heroku and update `.env_dist`. Or, jump straight to removing those overrides.

rafeerahman (approved):

groovecoder:
I added the `_DEFAULT_PHONE_RATE_LIMIT` technique for now.

PR #3992, opened 2023-10-12
MPP-3434: Handle an empty JSON for account profile

jwhitlock:
When creating a user from a `POST` to `/api/v1/terms_accepter_user`, we request the user's profile. If this call fails, then log the profile request status code and text and return 500. Previously, an empty response was returned, and the call failed due to a JSON decode error.

Also:

* Add a TODO to split up this file. This can be done after the "first_forwarded_email" endpoint is merged.
* Use [.json() from the requests library](https://requests.readthedocs.io/en/latest/user/quickstart/#json-response-content) to parse the profile JSON

## How to test:

### DIFF
--- a/api/tests/views_tests.py
+++ b/api/tests/views_tests.py
@@ -475,9 +475,8 @@ class TermsAcceptedUserViewTest(TestCase):
         _setup_fxa_response(200, {"active": True, "sub": self.uid, "exp": exp_time})
         # FxA profile server is down
         responses.add(responses.GET, FXA_PROFILE_URL, status=502, body="")
-
-        # get fxa response with 201 response for new user and profile created
         response = self.client.post(self.path)
+
         assert response.status_code == 500
         assert response.json()["detail"] == (
             "Did not receive a 200 response for account profile."
--- a/api/tests/views_tests.py
+++ b/api/tests/views_tests.py
@@ -466,6 +466,23 @@ class TermsAcceptedUserViewTest(TestCase):
         assert responses.assert_call_count(self.fxa_verify_path, 2) is True
         assert responses.assert_call_count(FXA_PROFILE_URL, 1) is True
 
+    @responses.activate()
+    def test_failed_profile_fetch_for_new_user_returns_500(self):
+        user_token = "user-123"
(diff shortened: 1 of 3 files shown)


api/tests/views_tests.py
```diff
@@ -466,6 +466,23 @@ def test_201_new_user_created_and_202_user_exists(
         assert responses.assert_call_count(self.fxa_verify_path, 2) is True
         assert responses.assert_call_count(FXA_PROFILE_URL, 1) is True
 
+    @responses.activate()
+    def test_failed_profile_fetch_for_new_user_returns_500(self):
+        user_token = "user-123"
+        self._setup_client(user_token)
+        now_time = int(datetime.now().timestamp())
+        exp_time = (now_time + 60 * 60) * 1000
+        _setup_fxa_response(200, {"active": True, "sub": self.uid, "exp": exp_time})
```
groovecoder, line 475: quibble (non-blocking): I thought this was unneccessary at first, because we add the response for `FXA_PROFILE_URL` after this. But I see that this `_setup_fxa_response` might be better-named `_setup_fxa_introspection_response`?

api/tests/views_tests.py
```diff
@@ -466,6 +466,23 @@ def test_201_new_user_created_and_202_user_exists(
...
+        # FxA profile server is down
+        responses.add(responses.GET, FXA_PROFILE_URL, status=502, body="")
+
+        # get fxa response with 201 response for new user and profile created
```
groovecoder, line 479: change (blocking): this is copy-pasta from line 448.
```suggestion
        # assert response with 500 with details
```

groovecoder (changes requested):

api/tests/views_tests.py
jwhitlock: Filed [MPP-3527](https://mozilla-hub.atlassian.net/browse/MPP-3527) and added a `TODO` comment

jwhitlock (commented):
Thanks for the quick review @groovecoder! I've removed the comment, to make the test look more like the sad path tests below it.

groovecoder (approved):

PR #4012, opened 2023-10-17
e2e fix needed after the playwright version bump

mozrokafor:
Fix for the relay e2e failure that appeared after the version bump. This is a temporally fix as it doesn't address the root problem with nightly and playwright just yet.


e2e-tests/specs/relay-premium-upgrade.spec.ts
```diff
@@ -5,6 +5,8 @@ import { checkAuthState, defaultScreenshotOpts } from '../e2eTestUtils/helpers';
 test.use({ storageState: 'state.json' })
 test.describe.configure({ mode: 'parallel' });
 test.describe('Premium Relay - Purchase Premium Flow, Desktop', () => {
+  test.skip(({ browserName }) => browserName === 'firefox', 'FxA Issue with with authenication');
```
pdehaan, line 8: also, if we have a jira ticket for this, we should add it here of in a comment/annotation.

```suggestion
  test.skip(({ browserName }) => browserName === 'firefox', 'FxA Issue with with authentication');
```

groovecoder (approved):

Issue #3126, opened 2023-02-21
Upgrade to Django 4.2


jwhitlock:
Django 4.2 is the next Long Term Support (LTS) version. Key dates:

* January 16, 2023 - Django 4.2 alpha; feature freeze.
* February 20, 2023 - Django 4.2 beta; non-release blocking bug fix freeze.
* March 20, 2023 - Django 4.2 RC 1; translation string freeze.
* April 3, 2023 - Django 4.2 final
* April 2024 - Django 3.2 no longer supported.

The release notes for the minor versions describe the changes:

* https://docs.djangoproject.com/en/4.2/releases/4.0/
* https://docs.djangoproject.com/en/4.2/releases/4.1/
* https://docs.djangoproject.com/en/4.2/releases/4.2/

PR #3110 attempted to update to 4.1.7, which fails unless PR #3111 is applied.

One possible process:
* Add a Django 4.2 build to CircleCI, to see test failures and warnings related to the release
* Eliminate deprecation warnings when possible
  * `USE_L10N` default changes from `False` to `True` in 4.0, and is removed in 5.0. We set it to `True`, but can't remove until the update
* Eliminate other warnings when possible (such as warnings when generating swagger docs)

jwhitlock:
PR #4082 upgraded the Django version, and will go out with tag [2023.11.08](https://github.com/mozilla/fx-private-relay/releases/tag/2023.11.08). We'll watch stage for new issues from the update.

There are some additional changes needed to eliminate the warnings, but these need to wait for that to be deployed to production.

PR #4014, opened 2023-10-17
Update Django from 3.2.20 to 3.2.22

jwhitlock:
This update Django to the latest in the 3.2 series. It appears Dependabot is no longer suggesting 3.2 updates, which is another reason to update to 4.2 as soon as possible (#3126)

[Django 3.2.21](https://docs.djangoproject.com/en/3.2/releases/3.2.21/) was released September 4, 2023. It fixes a denial-of-service vulnerability with `django.utils.encoding.uri_to_iri()`.

**I do not believe we are vulnerable to this issue**, because our code does not use `uri_to_iri()`, and Django code does not appear to use it either.

[Django 3.2.22](https://docs.djangoproject.com/en/3.2/releases/3.2.22/) was released October 4, 2023. It fixed a denial-of-service vulnerability with `django.utils.text.Truncator`’s `chars()` and `words()` methods with `html=True`. 

**I do not believe we are vulnerable to this issue**, because we do not use the HTML mode. We use `Truncator` to create our `From` header for forwarded emails:

https://github.com/mozilla/fx-private-relay/blob/8041a14c6d958c05ae21b5d9e25451ef3b93519d/emails/utils.py#L298

We do not use the affected template filters `truncatechars_html` or `truncatewords_html`

## How to test

I believe the continuous integration tests should be sufficient to validate this update.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 boto3==1.28.62
 codetiming==1.4.0
 cryptography==41.0.4
-Django==3.2.20
+Django==3.2.22
 dj-database-url==2.1.0
 django-allauth==0.54.0
 django-cors-headers==4.2.0


say-yawn (approved):
Thanks for the thorough review on the Django update. Also noting that [Django 3.2 end of life ](https://docs.djangoproject.com/en/4.2/releases/3.2/#django-3-2-release-notes)is April 2024 and we should prioritize on getting #3126.

PR #4017, opened 2023-10-19
Restore overlay bug workaround for news menu

Vinnl:
So... This is a bit embarrassing. I was just trying to inspect something in the accessibility inspector, only to find out I couldn't reach it. It turned out that the <OverlayProvider> set aria-hidden by default, and git bisect turned up the following commit as the culprit: 44d2e15cf94c05c58ebb5e20b79dd4880a7e56d1.

Thus, this partially reverts that commit.

How to test: On `main`, if you're on a page that displays the News menu, and you right-click and select "Inspect Accessibility Properties", you'll find that you can't really select anything, and the accessibility tree is suspiciously short:

image

Then in the regular inspector, you'll see that the `div` with ID `overlayProvider` has `aria-hidden="true"`:

image

In this branch, everything should be accessible again.


lloan (approved):
Works as intended.

PR #4035, opened 2023-10-24
MPP-3477: Log header issues

jwhitlock:
This PR addresses:

* MPP-3477 - `ValueError: Header values may not contain linefeed or carriage return characters` - This exact traceback no longer occurs, but may have moved to a different issue.
* MPP-3503 - `IndexError: list index out of range` - should fix
* MPP-3540 - `IndexError: list index out of range` - should also fix

This PR extends Python's [email.policy.default](https://docs.python.org/3/library/email.policy.html) as `relay_policy` to change how emails headers are parsed.  The changes are:

* Each email header has a new attribute `.as_unstructured` which represents the header if parsed as an unstructured header, rather than as an email address, date, etc. This allows logging something like the original value when the header is not RFC compliant.
* The `Message-ID` header parser now catches `IndexError` and treats the header as an unstructured header.

This allows loading test emails with non-compliant headers, so they don't need to be stored as SNS JSON fixtures.

It also enables logging defect data for headers when forwarding an email, under warning message `"_handle_received: forwarding issues"`. The new code uses try / except blocks when reading and writing headers, and logs any exceptions. This will mean a few more emails are processed without killing the email task, but in other cases the error will be logged but the email will fail later in the process. The main benefit is to be able to see which headers are non-compliant and plan on how to handle them.

This PR also serializes the email before calling `ses_client.send_raw_email`, which will hopefully make serialization errors easier to distinguish from other SES client errors.

## How to test:

This code adds extra logging for emails with headers that are not RFC compliant. I have not found a way to create these emails for interactive testing, so you'll have to be content with code inspection and automated testing.


emails/views.py
```diff
@@ -799,23 +808,24 @@ def _convert_to_forwarded_email(
...
     """
-    email = message_from_bytes(incoming_email_bytes, policy=policy.default)
+    email = message_from_bytes(incoming_email_bytes, policy=relay_policy)
     # python/typeshed issue 2418
     # The Python 3.2 default was Message, 3.6 uses policy.message_factory, and
     # policy.default.message_factory is EmailMessage
     assert isinstance(email, EmailMessage)
 
-    _replace_headers(email, headers)
+    header_issues = _replace_headers(email, headers)
```
groovecoder, line 828: suggestion (non-blocking): code-smell: this function changes variable state. 😞 Could we change it to return, and change the calling code to handle, a tuple of `email_with_replaced_headers, header_issues` ?
```suggestion
    email_with_replaced_headers, header_issues = _replace_headers(email, headers)
```

groovecoder (approved):
A few suggestions, but none of them are blocking.

jwhitlock (commented):
Thanks for the quick review @groovecoder! I considered the suggestions, I'm not changing the code for this round.

emails/policy.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+        as_unstructured = as_unstructured_cls(name, value)
+        header_instance.as_unstructured = as_unstructured
+        return header_instance
+
+
+relay_header_factory = RelayHeaderRegistry()
+relay_header_factory.registry["message-id"] = cast(
+    type["_HeaderParser"], RelayMessageIDHeader
+)
```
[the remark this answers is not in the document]
groovecoder: Great, thanks for checking into that!

emails/views.py
groovecoder: Oh, good point. I keep forgetting the memory concern. I'll try to remember, but we could also throw a code comment into the functions, or before the function calls, that say why we're changing the object in-memory.

PR #4059, opened 2023-10-30
MPP-3477: Handle value=None

jwhitlock:
If the exception handler sets the value to `None`, it will not have the defects parameter.

This bug was introduced by PR #4035, but the only header I know that triggers an exception on reading was `Message-ID`, before the fix that was also in PR #4035.

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -901,7 +901,7 @@ def _replace_headers(
         except Exception as e:
             issues["incoming"][header] = {"exception_on_read": str(e)}
             value = None
-        if value.defects:
+        if getattr(value, "defects"):
             issues["incoming"][header] = {
                 "defect_count": len(value.defects),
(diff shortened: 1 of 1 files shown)


groovecoder (approved):
suggestion (non-blocking): Should we add a test for the error case?

jwhitlock:
> suggestion (non-blocking): Should we add a test for the error case?

Yes, if I can think how to trigger it... This is a blocker for the next tag, so if I can't figure out a test I may have to defer to the next PR or when I capture a failure in production.

jwhitlock:
I was unable to create a test email that exercised this code path. I do not think it is impossible, we'll just need the craziness of the internet to create the test cases.

I did take the opportunity to add a docstring requested on the previous code review:

https://github.com/mozilla/fx-private-relay/pull/4035#discussion_r1377897847

PR #4065, opened 2023-11-02
Add case-insensitive email indexes

jwhitlock:
The django-allauth code searches for existing, case insensitive emails. This search can use this index instead of a full table scan.

This index was created manually in production with:

```sql
CREATE INDEX CONCURRENTLY account_emailaddress_email_upper ON account_emailaddress (upper(email));
CREATE INDEX CONCURRENTLY auth_user_email_upper ON auth_user (upper(email));
```

The option `CONCURRENTLY` is not allowed in a Django migration. Additionally, the SQL uses `IF NOT EXISTS`, to avoid an error in production.

## How to test:

CircleCI will verify that the migrations run on SQLite and PostgreSQL. If you want to validate locally:

```sh
sqlite3 db.sqlite3 .schema | grep _email_upper  
# Empty response

./manage.py migrate privaterelay
# Applying privaterelay.0006_add_email_indexes... OK

sqlite3 db.sqlite3 .schema | grep _email_upper 
# CREATE INDEX account_emailaddress_email_upper ON account_emailaddress (upper(email));
# CREATE INDEX auth_user_email_upper ON auth_user (upper(email));

./manage.py migrate privaterelay 0005
# Unapplying privaterelay.0006_add_email_indexes... OK

sqlite3 db.sqlite3 .schema | grep _email_upper  
# Empty response
```


groovecoder (approved):
Spot-check looks good.

I suppose if `django` or `allauth` add their own migrations to "fix" this issue, their migration might fail, and we'll get a CI failure. But it's better to have this in sync and in our repo now.

jwhitlock:
> I suppose if django or allauth add their own migrations to "fix" this issue, their migration might fail, and we'll get a CI failure. But it's better to have this in sync and in our repo now.

It is fixed in later versions of allauth:

https://github.com/pennersr/django-allauth/blob/main/allauth/account/migrations/0005_emailaddress_idx_upper_email.py

Their index name 'account_emailaddress_upper' will not collide with our index named `account_emailaddress_email_upper`. We'll have two indexes doing the same thing, slowing down writes slightly.

We have work to do before updating django-allauth, see MPP-3390.

PR #4069, opened 2023-11-03
Avoid distortion in app logos, using object-fit

dholbert:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1863049 .

Before this commit, some of the app logos in Relay's top-right app-picker-menu were distorted, due to displaying a rectangular image file in a square img element.  "object-fit: contain" (added here) lets us preserve the image's aspect ratio when displaying it via an img element.

### DIFF
--- a/frontend/src/components/layout/navigation/AppPicker.module.scss
+++ b/frontend/src/components/layout/navigation/AppPicker.module.scss
@@ -110,6 +110,7 @@
         width: fit-content;
         box-sizing: content-box;
         border-radius: $border-radius-md;
+        object-fit: contain;
       }
 
       &:hover {


Vinnl (approved):
Whoops, nice fix, thank you!

Before/after:

image


image

dholbert:
Thanks for merging, @Vinnl ! Out of curiosity, do you know how soon I should expect to see this change make it to the live site? (I'm aiming to close https://bugzilla.mozilla.org/show_bug.cgi?id=1863049 as fixed-by-this-PR once I can verify the fix in the wild.)

Vinnl:
@dholbert Deploys are once a week on Tuesdays, first to stage, and then to production. So it should go to stage today, and then to production next week, barring unforeseen issues.

PR #4086, opened 2023-11-06
Apply migrations for MSW 2.x

Vinnl:
Following this migration guide:
https://mswjs.io/docs/migrations/1.x-to-2.x

(We use Mock Service Worker (MSW) to simulate API responses in the preview deployment and when running `npm run dev:mocked`.)

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -106,6 +106,27 @@ module.exports = withBundleAnalyzer({
       type: "asset/source",
     });
 
+    // msw/browser is not exported when compiling for Node (see
+    // https://github.com/mswjs/msw/blob/461a1885280451ec0837fdce897f49521b8ff260/package.json#L24)
+    // and msw/node is not exported when compiling for the browser (see
+    // https://github.com/mswjs/msw/blob/461a1885280451ec0837fdce897f49521b8ff260/package.json#L17),
+    // so to avoid build errors, we have to prevent Webpack from trying to
+    // traverse them when building.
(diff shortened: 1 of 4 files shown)


frontend/next.config.js
```diff
@@ -106,6 +106,27 @@ module.exports = withBundleAnalyzer({
...
+      }
+    } else {
+      if (Array.isArray(config.resolve.alias)) {
+        config.resolve.alias.push({ name: "msw/node", alias: false });
+      } else {
+        config.resolve.alias["msw/node"] = false;
+      }
+    }
+
     return config;
```
groovecoder, line 130: suggestion (totally-optional non-blocking): I treat every PR with an `else` statement like a little code-kata for me to practice refactoring into an early-return function instead.

```suggestion
    if (options.isServer) {
      if (Array.isArray(config.resolve.alias)) {
        config.resolve.alias.push({ name: "msw/browser", alias: false });
        return config;
      }
      config.resolve.alias["msw/browser"] = false;
      return config;
    }
    if (Array.isArray(config.resolve.alias)) {
      config.resolve.alias.push({ name: "msw/node", alias: false });
      return config;
    }
    config.resolve.alias["msw/node"] = false;
    return config;
```

groovecoder (approved):

frontend/next.config.js
Vinnl: I usually lean towards early-return as well, but in this case, modifying `config.resolve.alias` is not the only purpose of this function, and returning early could hamper future additions to the config.

It could work if I extracted this into a separate function, but... That seems like the indirection might harm readability more?

PR #4088, opened 2023-11-06
Fix responses decorator, update responses

jwhitlock:
The update of responses 0.23.3 to 0.24.0 broken on `mypy` in PR #4080, because that release removed some overrides that handled calling `@responses.activate()` with no parameters. This may be fixed in the next version, but the syntax `@responses.activate` does the same thing and will work with 0.24.0


groovecoder (approved):
praise: nice fix!

PR #4109, opened 2023-11-13
Update to Node 18.18

Vinnl:
Recent version of Next.js specify 18.17.0 as the minimum version, as indicated when running `npm install`:

> npm WARN EBADENGINE Unsupported engine {
> npm WARN EBADENGINE   package: 'next@14.0.2',
> npm WARN EBADENGINE   required: { node: '>=18.17.0' },
> npm WARN EBADENGINE   current: { node: 'v18.12.1', npm: '8.19.2' }
> npm WARN EBADENGINE }

This upgrade should be relatively safe, since Node is only used in the build process.

I also pinned npm to the version bundled with this version of Node, as suggested by Volta after pinning the Node version:

>   note: this version of Node includes npm@9.8.1, which is higher than your pinned version (8.19.2).
>      To use the version included with Node, run `volta pin npm@bundled`


jwhitlock (approved):
Thanks @Vinnl! The update looks good in CircleCI, and `volta` downloaded the new `node` when I ran `npm install`.

PR #4110, opened 2023-11-13
MPP-3583: Test header read exception, allow multiple headers with same name

jwhitlock:
PR #4059 did not actually fix the problem. I still can't write a text how I'd like, because I don't know what causes the issue, but `Mock` can raise the exception as needed.

While I'm touching the code, I ~added some new bug~ adjusted the output:

* Use `repr(exception)`, which includes the exception name
* Use a list of (name, issue) pairs instead of a dict of name → issue, so that if multiple headers with the same name have problems we capture them all.

## How to test:

### DIFF
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -33,7 +33,7 @@ from emails.models import (
     address_hash,
     get_domains_from_settings,
 )
-from emails.types import AWS_SNSMessageJSON
+from emails.types import AWS_SNSMessageJSON, OutgoingHeaders
 from emails.utils import (
     b64_lookup_key,
     decrypt_reply_metadata,
@@ -48,6 +48,7 @@ from emails.views import (
     _get_address,
     _get_keys_from_headers,
     _record_receipt_verdicts,
+    _replace_headers,
     _set_forwarded_first_reply,
     _sns_message,
     _sns_notification,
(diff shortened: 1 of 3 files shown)


jwhitlock:
MPP-3583 is fixed by PR #4121. This PR still adds a test for that code path, and changes the logging format to allow recording multiple header issues. This may be useful if we start parsing the [ARC](https://en.wikipedia.org/wiki/Authenticated_Received_Chain) headers and adding our own.

emails/tests/views_tests.py
```diff
@@ -1872,3 +1879,37 @@ def test_get_keys_from_headers_references_reply_dne():
...
+    email = message_from_string(email_text, policy=relay_policy)
+    assert isinstance(email, EmailMessage)
+    new_headers: OutgoingHeaders = {
+        "Subject": "Error Handling Test",
+        "From": "from@example.com",
+        "To": "to@example.com",
+    }
+    for name, value in new_headers.items():
+        assert email[name] != value
+    email["X-Fail"] = "I am OK"  # Do not fail initial parsing
```
groovecoder, line 1896: question (non-blocking): To help make sure I'm understanding this correctly ... is this comment on the right line? It looks like this first `for` loop is the logic that checks for no failure on the initial parsing. Then this line is the line that starts to set up the failing case? Should this comment be above the `for` loop?

groovecoder (approved):
Just 1 clarifying question.

emails/tests/views_tests.py
jwhitlock: No, the first `for` loop verifies that the headers are different than the existing headers. That way we can verify that `_replace_headers` caught the error and kept working later.

I'll add some comments about what is going on, because it isn't obvious. Maybe something like:

```suggestion
    # Add a header for testing
    # This header is a valid unstructured header, so we can add it to email
    # We'll override header access to raise an exception, to emulate an
    # exception raised when parsing a header.
    email["X-Fail"] = "I am OK"
```

PR #4111, opened 2023-11-13
MPP-3390: Auto-verify emails

jwhitlock:
`django-allauth` 0.55.0 changes the unique index on emails to a unique index on emails where `verified=True`. Our code relies on unique emails, so we need to start marking emails coming from Mozilla accounts as verified.

Upgrading to `django-allauth` 0.55.0 and later will require two steps over two production pushes.

* **Step 1** (This PR) - Mark emails as verified going forward.
* **Step 2** - Mark existing emails as verified

### DIFF
--- a/api/tests/views_tests.py
+++ b/api/tests/views_tests.py
@@ -604,7 +604,7 @@ def test_duplicate_email_logs_details_for_debugging(caplog: pytest.LogCaptureFix
     caplog.set_level(logging.ERROR)
     uid = "relay-user-fxa-uid"
     email = "user@email.com"
-    baker.make(EmailAddress, email=email)
+    baker.make(EmailAddress, email=email, verified=True)
     user_token = "user-123"
     client = _setup_client(user_token)
     now_time = int(datetime.now().timestamp())
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -517,6 +517,7 @@ SOCIALACCOUNT_PROVIDERS = {
         "PROFILE_ENDPOINT": config(
             "FXA_PROFILE_ENDPOINT", "https://profile.accounts.firefox.com/v1"
         ),
+        "VERIFIED_EMAIL": True,  # Assume FxA primary email is verified
     }
 }
 
--- a/privaterelay/tests/views_tests.py
+++ b/privaterelay/tests/views_tests.py
@@ -114,11 +114,13 @@ class UpdateExtraDataAndEmailTest(TestCase):
         extra_data = json.loads('{"test": "test"}')
 
         user = baker.make(User, email="user@example.com")
-        baker.make(EmailAddress, user=user, email="user@example.com")
+        baker.make(EmailAddress, user=user, email="user@example.com", verified=True)
         baker.make(SocialAccount, user=user, provider="fxa", extra_data=extra_data)
 
         user2 = baker.make(User, email="user2@example.com")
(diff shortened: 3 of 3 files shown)


say-yawn (approved):
This PR is required before #4112 because the updated `django-allauth` package results in change on uniqueness on email field in `EmailAddress` to require `verified=True`. Thus, when updating `django-allauth` package and applying migrations to update old `EmailAddress`, entries with `verified=True`, all `EmailAddress` will be set with `verified=True` and have email unique constraint applied.

(Suggestion, non-blocking) I recommend adding a test on creating two `EmailAddress` with duplicate email value resulting in unique constraint error. This will fail when #4112 updates the django-allauth and will need to be fixed to include `verified=True` attribute while creating the `EmailAddress`. Detailed explanation:

In `django-allauth=0.54.0`, regardless of `verified=True` is set on `EmailAddress`, adding new `EmailAddress` with the same email address results in
```
django.db.utils.IntegrityError: UNIQUE constraint failed: account_emailaddress.email
```

However, updating the the allauth package to `0.58.2` and applying following migrations:
```
  Applying account.0003_alter_emailaddress_create_unique_verified_email... OK
  Applying account.0004_alter_emailaddress_drop_unique_email... OK
  Applying account.0005_emailaddress_idx_upper_email... OK
  Applying socialaccount.0004_app_provider_id_settings... OK
  Applying socialaccount.0005_socialtoken_nullable_app... OK
  Applying socialaccount.0006_alter_socialaccount_extra_data... OK
```
we only get the unique constraint error if both `EmailAddress` were created with `verified=True` as explained in [django-allauth 0.55.0 notes](https://docs.allauth.org/en/latest/release-notes/recent.html#id21).

jwhitlock:
Thanks for the review @say-yawn!

> I recommend adding a test on creating two EmailAddress with duplicate email value resulting in unique constraint error.

I'm pretty sure the tests `test_update_all_data_conflict` tests this scenario. It attempts to update a user's email to `user@example.com`, except there is an existing user with this email. With the current django-allauth 0.54.0, this test succeeds because the `IntegrityError` due to the unique index, and it is turned into a 409.

This PR updates those tests to verify the existing emails. The test still succeeds, because of the unique index. Without that, the test will fail with django-allauth 0.56.0, which drops the unique index. If the existing emails were unverified, there would be no integrity error, and the function would return a 202 like the following test.

The reason I update the tests in this PR is because of the migrations tests. For those, we run the production release's tests against the migrations in the branch. This emulates how the production code will run during a deployment (the last release's code running with the new release's migrations). Attempts to update to django-allauth 0.56.0 will fail until those tests are updated.

So, yes we are testing the integrity error for colliding email addresses, but in a way that they will continue to collide as we update django-allauth versions.

say-yawn:
Thanks @jwhitlock 
> I'm pretty sure the tests test_update_all_data_conflict tests this scenario. It attempts to update a user's email to user@example.com
I agree that the test is similar but I was thinking testing the create instead of the update. The test I tried was doing:
```python
baker.make(EmailAddress, user=user, email="user@example.com", verified=True)

ea2 = baker.make(
            EmailAddress, user=user2, email="user@example.com", verified=True
        )
```
Same email on both `EmailAddress` during create and checking that during `baker.make()` for `ea2` the `IntegrityError` is thrown and maybe better on `emails/tests/models_tests.py` since it's on create.

PR #4121, opened 2023-11-15
Fixes in 2023.11.08.8, further changes for CSS sourcemap URLs

jwhitlock:
**Update 1**: I've identified Django plus caching as the cause of this issue. This further changes the code from the hotfix.

This PR applies fixes made to [2023.11.08.8](https://github.com/mozilla/fx-private-relay/releases/tag/2023.11.08.8), which probably did not fix the issue but are still useful. It then adds further changes that will fix the issue.

The breaking change is that with Django 3.2, the CSS file has a sourcemap comment like this:

```css
/*# sourceMappingURL=b8dfe6b69b0b05f4.css.map*/
```

Django 4.2 adds sourcemap "fixes", which changed the comment to:

```css
/*# sourceMappingURL=b8dfe6b69b0b05f4.css.map */
```

This change was introduced in Django 4.1, in [django.contrib.staticfiles](https://docs.djangoproject.com/en/4.2/releases/4.1/#django-contrib-staticfiles) "[ManifestStaticFilesStorage](https://docs.djangoproject.com/en/4.2/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage) now replaces paths to CSS source map references with their hashed counterparts.". We're preventing the storage backend from changing the filename in `RelayStaticFilesStorage`, but this new behavior still changed the comment. This would be OK, except that we are also adding the caching header:

```
Cache-Control: max-age=3600,public,immutable
```

This instructs intermediate caches (like the CDN and nginx) that they should cache the file (`public`) and that they don't need to check if the content changed (`immutable`), because we'll change the filename if the content changes. If `next.js` is unlucky enough to fetch the CSS from a caching server, then the content won't match the latest hash.

This probably means that none of the changes in `2023.11.08.8` actually fixes the problem, we just needed to wait for caches to invalidate. However, some may _still_ have old versions of the file, so we could use a real fix for this issue.

This code leans into the Django 4.1 changes to add the absolute path to the sourcemap. This will allow the developer tools to find the sourcemap, even when next.js injects it as an inline style, avoiding a 404. It would be nicer to do this on the JS side, but [Dart Sass](https://sass-lang.com/documentation/cli/dart-sass/#source-map-urls) doesn't have this option, and I don't know JS well enough to implement the local changes suggested in https://github.com/vercel/next.js/issues/27672.

This further adds a persistent comment to two SCSS files:

* `frontend/src/styles/globals.scss` - `/*! gv=1 */`
* `frontend/src/styles/tokens.scss` - `/*! v=1 */`

This is the [Sass syntax](https://sass-lang.com/documentation/syntax/comments/) for a comment that should remain even in compressed mode. This allows changing two places to change all the derived CSS and force a cache invalidation. This should only be necessary when Django staticfiles processing changes, and even then only if it changes the output. So, we may never need this again, but it is there and commented.

Other changes:

* Use `sha256` instead of `sha512` hashes. These are shorter, so the `Content-Security-Policy` header is shorter ~isn't extended as much for the double number of hashes~. It also lines up with the CSP error messages from Chrome, which include the `sha256` hash digest of the rejected content.
* Add `sha256` hash digest for an empty string. `next.js` first injects an empty `<style>` element, and this hash prevents a spurious CSP error.
* Sort hashes alphabetically. This helps human scanning for hashes.
* Fix MPP-3583. PR #4110 also includes the tests, and will need a rebase.

# How to test

The testing is similar to https://github.com/mozilla/fx-private-relay/pull/2704, and requires reconfiguring your local development mode to be more like production. Which is a pain.

Locally, I watched the output of `cd frontend; npm run build`, and noted that all the CSS filenames changed. You could inspect the CircleCI output in `main` versus this PR to see the same results.


Vinnl (commented):
Thanks for the sleuthing, and for sharing with other Next.js users as well. I didn't perform the painful local testing steps, but the code seems sensible to me, and nicely commented. Great job!

groovecoder (approved):
I didn't spot-check but like @Vinnl the code looks good to me!

PR #4145, opened 2023-11-21
WIP: MPP-3390: Update to django-allauth 0.55, prepare for 0.56.0

jwhitlock:
This is step 2 of MPP-3390, the django-allauth update. 

``django-allauth`` 0.55.0 changes the unique index on emails to a unique index on emails where ``verified=True``. Our code relies on unique emails. This PR marks existing emails as verified, and updates to django-allauth 0.55.0.

* Step 1 (PR #4111) - Mark emails as verified going forward. Shipped in [2023.11.17](https://github.com/mozilla/fx-private-relay/releases/tag/2023.11.17).
* Step 2 - Mark existing emails as verified, update to 0.55.0. Include `AccountMiddleware` if available, to prepare for 0.56.0
* Step 3 - Update to 0.58.2 (or latest)
* Step 4 - Always include `AccountMiddleware`


jwhitlock:
I'll need to re-work how the migrations tests run in order to better simulate a production deploy:

* Currently, we run the old tests with the **new** requirements and the new migrations
* Next, we need to run the old tests with the **old** requirements and the new migrations.

I think [--reuse-db](https://pytest-django.readthedocs.io/en/latest/database.html#reuse-db-reuse-the-testing-database-between-test-runs) will allow me to do this, but it will be a new PR to update the CirceCI test config.

PR #4146, opened 2023-11-21
MPP-3588: 2023 Holiday promo banner

rafeerahman:
This PR fixes MPP-3588.

# New feature description

A banner for our 2023 holiday promo was added for non signed in users on /premium and the relay home page.

Feature flag: holiday_promo_2023

# Demo

[Demo](https://github.com/mozilla/fx-private-relay/assets/59676643/4c63178f-6ece-47a3-933e-e00fc82c4ba5)

# How to test
1. Give yourself the waffle flag `holiday_promo_2023`
2. Go to the relay home page as a non-signed in user and verify the following:
- A user that is not signed in can see the banner on the relay home page (“/”) and on the premium page (“/premium”).
- A user that is signed in does not see the banner.
- Clicking the get 1 year of premium button takes the user to the Email 1 year plan and with the coupon code.
- Banner should be responsive to all screens.
- Banner is sticky, stays with the navbar.

- [x] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.


frontend/pendingTranslations.ftl
```diff
@@ -91,3 +91,14 @@ profile-free-onboarding--addon-item-headline-1 = They work anywhere you need an
 profile-free-onboarding--addon-item-description-1 = Anywhere you need to enter an email, you can use a { -brand-name-relay } email mask. You can even generate new masks directly from your { -brand-name-firefox } password manager. 
 profile-free-onboarding--addon-item-headline-2 = The { -brand-name-relay } extension makes it easy
 profile-free-onboarding--addon-item-description-2 = Whenever you see an email address field, the { -brand-name-firefox-relay } browser extension will autofill any email mask you want to use. No need to remember each mask!
+
+# 2023 Holiday promo
+# Variables:
+# $couponCode (string) - the coupon code.
+# $couponExpiryDate (string) - the coupon expiry date, for example, Dec 31, 2023.
```
Vinnl, line 98: I think you need to put these on the line before the strings that actually use those variables, otherwise localisers won't see them at the right time.

Tip: [check out Pontoon](https://pontoon.mozilla.org/es-MX/firefox-relay-website/), then you'll have an idea of how the localisers work. They don't edit FTL files directly :)

frontend/src/styles/tokens.scss
```diff
@@ -25,6 +25,7 @@ $buttonFocusOutline:
 
 // Overwrite the Protocol colours with Nebula colours:
 // Purple:
+$color-purple-100: #5015cd;
```
Vinnl, line 28: Just verifying: design is also using the term "Purple-100" for this colour? Because we didn't come up with these colours ourselves - they were aligned with design, so we'd have a consistent colour palette.

If not, then I'd either contact design and ask for a colour from the Nebula colour palette, or just pick the closed colour that already has a token available. (Or both :) )

frontend/src/components/layout/topmessage/HolidayPromoBanner.tsx
```diff
@@ -0,0 +1,72 @@
...
+type Props = {
+  profile?: ProfileData;
+  runtimeData?: RuntimeData;
+};
+
+export const HolidayPromoBanner = (props: Props) => {
+  const l10n = useL10n();
+  const [loadingProfile, setLoadingProfile] = useState(true);
+  const coupon = "HOLIDAY20";
+  const couponExpiry = "Dec 31, 2023";
```
Vinnl, line 22: This won't work for localisation; that's not how dates are formatted (or abbreviated) in every locale. You could use [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), but I'd actually recommend just "hardcoding" the date in the string, and let the localisers format it in a way that makes sense for their locale. It's not like we'll be changing this date, right?

frontend/src/components/layout/Layout.tsx
```diff
@@ -138,6 +140,13 @@ export const Layout = (props: Props) => {
           runtimeData={props.runtimeData}
         />
         <header className={styles["header-outer"]}>
+          {isFlagActive(props.runtimeData, "holiday_promo_2023") &&
```
Vinnl, line 143: Maybe also add "is today no later than Dec 31, 2023" as a condition? So it auto-hides when it's no longer available.

frontend/src/components/layout/topmessage/HolidayPromoBanner.tsx
```diff
@@ -0,0 +1,72 @@
...
+      </div>
+      <div className={styles["promo-container"]}>
+        <Link href={subscriberLinkWithCoupon} className={styles["cta-button"]}>
+          {l10n.getString("holiday-promo-banner-cta-button")}
+        </Link>
+        <p className={styles["promo-code-expiry"]}>
+          {l10n.getString("holiday-promo-banner-promo-expiry", {
+            couponExpiryDate: couponExpiry,
+          })}
+        </p>
```
Vinnl, line 68: Minor thing, but a [`<small>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/small) might be more appropriate here:

```suggestion
        <small className={styles["promo-code-expiry"]}>
          {l10n.getString("holiday-promo-banner-promo-expiry", {
            couponExpiryDate: couponExpiry,
          })}
        </small>
```

frontend/src/components/layout/topmessage/HolidayPromoBanner.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const [loadingProfile, setLoadingProfile] = useState(true);
+  const coupon = "HOLIDAY20";
+  const couponExpiry = "Dec 31, 2023";
+  const subscribeLink = isPeriodicalPremiumAvailableInCountry(props.runtimeData)
+    ? getPeriodicalPremiumSubscribeLink(props.runtimeData, "yearly")
+    : null;
+
+  useEffect(() => {
+    setLoadingProfile(false);
+  }, [props.profile]);
```
Vinnl, line 29: It's probably fine, because I think this works, but it feels kinda weird to have this. We're passing in `profile.data`, but [there's also `profile.isLoading`](https://swr.vercel.app/docs/api#return-values) which sounds like it's made for this use case. But fine to leave as-is.

Vinnl (approved):
That banner's looking really good!

frontend/src/components/layout/topmessage/HolidayPromoBanner.tsx
rafeerahman: Actually I realized the useEffect doesn't seem to solve the problem, but this isLoading property works, thanks for mentioning it!

frontend/src/styles/tokens.scss
rafeerahman: Thanks for the context, hard coded this color and removed it from tokens are talking with design. Left a comment saying its not in Nebula.

frontend/src/components/layout/topmessage/HolidayPromoBanner.tsx
rafeerahman: Originally did this incase we ever re use this banner in the future, we can re use the l10n string as well by just changing this variable. 

Hardcoded it into l10n now instead though.

PR #4167, opened 2023-11-27
MPP-3601: In free user onboarding, no `onBlur` for modal

jwhitlock:
The `onBlur` handler was submitting the email, which could then be submitted again if the user clicked Submit. Instead, only submit when the user submits.

### DIFF
--- a/frontend/src/components/dashboard/EmailForwardingModal.tsx
+++ b/frontend/src/components/dashboard/EmailForwardingModal.tsx
@@ -1,5 +1,4 @@
 import {
-  FocusEventHandler,
   FormEventHandler,
   ReactElement,
   ReactNode,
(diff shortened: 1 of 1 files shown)


frontend/src/components/dashboard/EmailForwardingModal.tsx
```diff
@@ -108,14 +101,13 @@ const ConfirmModal = (props: Props) => {
...
             onChange={(e) => setInputValue(e.target.value)}
-            onBlur={onBlur}
             aria-label={l10n.getString("profile-label-edit-2")}
             ref={inputRef}
             className={styles["label-input"]}
             placeholder={l10n.getString(
               "profile-free-onboarding--copy-mask-placeholder-relay-email-mask",
             )}
-            type="text"
+            type="email"
```
jwhitlock, line 110: I'm not sure about this one. It seems like an email input to me, but that may have undesired behavior.

lloan (approved):
Clean. Looks good - thanks @jwhitlock ! 

Must have copied another component and never addressed it on our end.

frontend/src/components/dashboard/EmailForwardingModal.tsx
lloan: It will require them to input an email instead of any text - the built-in validation of an html input field.

lloan: ONLY thing I can maybe foresee is the keyboard some devices popup might be optimized for email input, but I think that's a good thing here. Users are expected to copy/paste the email anyway. Oh, and autofill.

PR #4174, opened 2023-11-30
MPP-3390: Fix profile API PATCH tests

jwhitlock:
The Profile API uses the profile ID as the identifier, not the user ID. These tests work because the API tests run first, when one user is created for each profile, so the `user.id` is equal to the `profile.id`.

If you reuse a test postgres database (`pytest --reuse-db`), IDs will start at the end of the old test run, and `user_id` will be out of sync with `profile_id`. 

SQLite3 resets sequences to 0 and doesn't have this issue by default. See PR #4175 for a way to trigger it.

The new migrations tests will check out the production code and run those tests, so this PR needs to ship to production before the new migrations tests are deployed.

How to test:

### DIFF
--- a/api/tests/views_tests.py
+++ b/api/tests/views_tests.py
@@ -118,7 +118,7 @@ def test_patch_premium_user_subdomain_cannot_be_changed(
 
     new_subdomain = "helloworldd"
     response = prem_api_client.patch(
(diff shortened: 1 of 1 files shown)


say-yawn (approved):
Thanks John for pointing out that the tests were using the wrong id. This would definitely be a problem too if we created any superusers using the Django admin command.

PR #4222, opened 2023-12-12
Remove @next/bundle-analyzer

Vinnl:
We spend more time managing dependency updates (that, as far as I know, hardly ever actually change how `@next/bundle-analyzer` works), than that we actually look at the analysis - let alone act on it. Thus, it probably makes more sense to just re-add it when we do feel the need to analyse the make-up of our frontend bundles.


groovecoder (approved):
Good idea!

PR #4177, opened 2023-11-30
MPP-3390: CircleCI config refactor

jwhitlock:
While working on MPP-3390, I found that the current `.circleci/config.yaml` was fighting my changes. This makes a number of changes to the configuration.

*Update 1*: I added `.circleci/python_job.bash` to house the bash functions. I also updated the "django 4.2" job to "django 5.0", and added "python with postgres [15.5 / 16.1]" tests

* Add comments linking to the relevant CircleCI documentation for orbs, docker images, and configuration references.
* Use YAML anchors and aliases to repeat the docker authentication and tag filters
* In `python_job`, call a bash function rather than pass the command as a job parameter
* Expand `python_job` to replace `build_email_tracker_list_job`
* Expand `python_job` to replace `python_test_postgres`

More changes are coming once PR #4174 and PR #4175 are deployed to production, and I can update how the migrations tests run.

## How to test:

- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] The "build_frontend" job runs in CircleCI. This is similar to the job with the same name in main. The "Build Code" step builds the Next.js code without errors. Warnings are similar to the build step on main. The job on main has an extra "Install Node.js 18" step, which updates to Node v18.19.0 or later.
- [ ] The "build email tracker lists" job runs in CircleCI. This is similar to the job with the same name in main. The "Run build email tracker lists" builds the two lists an copies them to `/tmp/workspace/email-trackers`. This is similar to the "build_email_tracker_list_job" on main, combining steps "Build level one email tracker list" (with less warnings), "Build level two email tracker list" (with less warnings), and "Copy email tracker lists to persisted workspace".
- [ ] The "build_test_backend" job runs in CircleCI. This is nearly identical "build_test_backend" job in main. The main job includes saving the docker image, the PR job does not. The job runs 910 tests in both. The "Test Code" step shows "895 passed, 15 skipped, 2 warnings" in both. 
- [ ] The "black style check" job runs in CircleCI. This is similar to the job with the same name in main. The "Run black" step is similar to the "black --check ." step in the main branch, showing "203 files would be left unchanged."
- [ ] The "mypy type check" job runs in CircleCI. This is similar to the job with the same name in main. Both show 1 Tests ran. The "Run mypy" step is similar to the "mypy --no-incremental --junit-xml job-results/mypy.xml ." step in main, showing "Success: no issues found in 203 source files". 
[...]

PR #4246, opened 2023-12-18
MPP-3390: Update django-allauth to 0.60.0

jwhitlock:
This updates django-allauth from 0.54.0 to 0.59.0 for MPP-3390.

Changes needed for upgrades:

* Set older emails as verified. New emails are auto-verified with PR #4111
* [0.55](https://github.com/pennersr/django-allauth/blob/main/ChangeLog.rst#0550-2023-08-22) - This update made changes to allow adding OpenID providers in the Django admin rather than only in the settings.
    - Two fields, `provider_id` and `settings`, were added to `socialaccount_socialapp` to support OpenID providers in the Django admin. These are both non-nullable with Django-level defaults. A new migration adds database-level defaults for our migration tests.
    - There was a change in how a Socialauth Provider is created. This required changes in the `terms_accepted_view`, which was manually creating a `FirefoxAccountsProvider` in order to create a new account.
    - A working `SocialApp` is now needed for some tests which previously did not need them.
* [0.56](https://github.com/pennersr/django-allauth/blob/main/ChangeLog.rst#0560-2023-09-07) - This update captures the `request` in a global context, so the code doesn't need to pass it around as much.
    - Add a required middleware to load the request into a global context
    - This update added a case-insensitive index on email, so we can drop the one we added.
* [0.57](https://github.com/pennersr/django-allauth/blob/main/ChangeLog.rst#0570-2023-09-24) - No Relay changes required
* [0.58](https://github.com/pennersr/django-allauth/blob/main/ChangeLog.rst#0580-2023-10-26) - No Relay changes required
* [0.59](https://github.com/pennersr/django-allauth/blob/main/ChangeLog.rst#0590-2023-12-13) - No Relay changes required
* [0.60](https://github.com/pennersr/django-allauth/blob/main/ChangeLog.rst#0600-2024-01-05) - No Relay changes required


jwhitlock:
Force-push to rebase and add 0.60.0 update.

jwhitlock:
With PR #4177 merged, this is ready for review.

api/tests/views_tests.py
```diff
@@ -60,6 +61,12 @@ def prem_api_client(premium_user: User) -> APIClient:
     return client
 
 
+@pytest.fixture
+def fxa_social_app(db) -> SocialApp:
```
groovecoder, line 65: praise: nice new fixture!

api/tests/views_tests.py
```diff
@@ -598,9 +606,10 @@ def _setup_client(token):
     return client
 
 
-@pytest.mark.django_db
 @responses.activate
-def test_duplicate_email_logs_details_for_debugging(caplog: pytest.LogCaptureFixture):
+def test_duplicate_email_logs_details_for_debugging(
+    caplog: pytest.LogCaptureFixture, fxa_social_app: SocialApp
```
groovecoder, line 611: question (non-blocking): This test invokes the `fxa_social_app` fixture by adding it to the parameters of the test, but the test doesn't reference the `fxa_social_app` value. Why use a parameter here instead of using `@pytest.mark.usefixtures` like above?

api/views/__init__.py
```diff
@@ -253,7 +253,7 @@ def terms_accepted_user(request):
 
         # this is not exactly the request object that FirefoxAccountsProvider expects, but
         # it has all of the necssary attributes to initiatlize the Provider
-        provider = FirefoxAccountsProvider(request)
+        provider = get_social_adapter().get_provider(request, "fxa")
```
groovecoder, line 256: praise: nice clean-up.

privaterelay/migrations/0007_set_verified_email.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("privaterelay", "0006_add_email_indexes"),
+        ("account", "0002_email_max_length"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql="UPDATE account_emailaddress SET verified=True;",
+            reverse_sql="UPDATE account_emailaddress SET verified=False;",
```
groovecoder, line 15: concern (non-blocking): This worries me. I know we never plan to migrate backwards, but this one seems particularly destructive? What's the effect of setting all the `account_emailaddress` values to `false` later? Are we locking everyone out?

privaterelay/migrations/0009_remove_duplicate_index.py
```diff
@@ -0,0 +1,49 @@
+# Generated by Django 4.2.7 on 2023-11-13 21:03
+# This index was added in 0006_add_email_indexes.py
+# The same index was added by django-allauth 0.56.0
```
groovecoder, line 3: concern (blocking): Do we know that `python manage.py migrate` will add the index in the `django-allauth` migration before this `privaterelay` migration removes it? What happens if there's temporarily no index on the `account_emailaddress.email` field?

groovecoder: E.g., I checked out the branch and I pip installed the new requirements. Then I ran `python manage.py migrate` but I didn't see the new migration run?
```
 lcrouch  (e) relay-python-3.10  …  fx-private-relay   e496846b ⚑ 17  python manage.py migrate
Not using django_ftl autoreloader because pyinotify is not installed. Set `Bundle.auto_reload` or `AUTO_RELOAD_BUNDLES` to `False` to disable this warning.
System check identified some issues:

WARNINGS:
?: (dockerflow.health.W001) Unapplied migration privaterelay.0008_socialaccount_provider_add_default
?: (dockerflow.health.W001) Unapplied migration privaterelay.0009_remove_duplicate_index
Operations to perform:
  Apply all migrations: account, admin, auth, authtoken, contenttypes, emails, phones, privaterelay, sessions, silk, sites, socialaccount, waffle
Running migrations:
  Applying privaterelay.0008_socialaccount_provider_add_default... OK
  Applying privaterelay.0009_remove_duplicate_index... OK
 lcrouch  (e) relay-python-3.10  …  fx-private-relay   e496846b ⚑ 17  pip freeze | grep allauth
django-allauth==0.60.0
```

I *DO* see index though:
```
    "account_emailaddress_upper" btree (upper(email::text))
```

groovecoder (changes requested):
Code looks almost 100% good but I have a blocking concern/question on the migration that removes the index.

Spot-checks: sign-up with a new account works, and sign into existing account works too.

api/tests/views_tests.py
jwhitlock: Interesting idea, I've only used it on classes. I'll try it.

privaterelay/migrations/0007_set_verified_email.py
jwhitlock: No, users log in by Mozilla accounts ID, not by email, so no one is locked out. Other than that, a lot of nothing happens. You'd need to verify the email before logging in by email, but we don't provide log in by email. If there is some collision between a new Mozilla account's email and an existing one, the new one will get redirected to "socialaccount_signup" (and get a `NoReverseMatch` 500), but that happens whether the email is verified or unverified.

It's mostly for database consistency. All users created before https://github.com/mozilla/fx-private-relay/releases/tag/2023.11.17 was released on November 21st have `verified=False`. All the users after have `verified=True`. If you run the reverse migration, they will all get `verified=False`, but the next new user will have `verified=True`.

Maybe someday we'll use `verified` to support multiple real emails per user. Until then, it doesn't really impact Relay at all.

jwhitlock (commented):
Thanks @groovecoder. I've made the suggested test changes and answered some questions about the migrations.

api/tests/views_tests.py
jwhitlock: It worked, this may be the new pattern for "apply but don't use this fixture"

privaterelay/migrations/0009_remove_duplicate_index.py
groovecoder: Yup, looks like I had already run the migrations in the past. All good then.

groovecoder (approved):

PR #4253, opened 2023-12-21
MPP-3657 Fix relay address validation against deleted address check

say-yawn:
This PR fixes MPP-3657.

How to test:

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -665,7 +665,7 @@ class RelayAddress(models.Model):
                 locked_profile = Profile.objects.select_for_update().get(user=self.user)
                 check_user_can_make_another_address(locked_profile)
                 while True:
-                    if valid_address(self.address, self.domain):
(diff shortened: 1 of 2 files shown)


jwhitlock (approved):
Thanks @say-yawn, that fixes the issue.

Do you want to include some type hints, or should we do that in a new PR?

say-yawn:
@jwhitlock yes, it was actually part of the acceptance criteria and I forgot. Thanks for bringing it up.

jwhitlock (changes requested):
OK, add some type hints please!

jwhitlock:
Can you add a return value from `RelayAddress.save` so mypy will check the function?

```python
    def save(self, *args, **kwargs) -> None:
```

jwhitlock (approved):
Thanks @say-yawn!

PR #4291, opened 2024-01-09
MPP-3390: Convert mustache sections to conditionals

jwhitlock:
As [suggested in PR #4177](https://github.com/mozilla/fx-private-relay/pull/4177#discussion_r1445455959):

> Add comment on the [mustache conditionals](https://support.circleci.com/hc/en-us/articles/4417604103835-Using-Mustache-Conditionals-in-Config-File) and/or consider using steps and condition.

This uses the `when` / `unless` conditions which seems to be better documented by CircleCI, but requires more lines.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,13 +129,13 @@ commands:
           steps:
             - run:
                 name: Parameters - << parameters.comment_if_true >>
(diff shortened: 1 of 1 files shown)


jwhitlock:
@say-yawn, feel free to reopen when you have time to review this.

PR #4295, opened 2024-01-11
Avoid single tap being handled by both the deletion and confirmation button

Vinnl:
This makes React Aria handle touch events for the button (via `useButton`), making sure it doesn't handle touch events from before the dialog opens, resulting in the deletion of a mask without confirmation.

This PR fixes MPP-3687.

How to test: on an Android device, open a mask card, then tap "Delete". Take note where the confirmation button (the one that says "Delete" _in the dialog_, the button with the red background) is, and hover a finger over it. Then close the dialog, scroll the page so that the regular Delete button (i.e. the one _outside_ the dialog) is under your finger. Tap it.

Before this PR, the confirmation button would immediately be tapped as well, and the mask would disappear. With this PR, the dialog appears as usual.

- [ ] I've added a unit test to test for potential regressions of this bug. - Handed over to external library and dependent on coordinates on screen, so hard to unit test.

### DIFF
--- a/frontend/src/components/dashboard/aliases/AliasDeletionButtonPermanent.tsx
+++ b/frontend/src/components/dashboard/aliases/AliasDeletionButtonPermanent.tsx
@@ -9,7 +9,7 @@ import {
   AriaOverlayProps,
 } from "react-aria";
 import { useOverlayTriggerState } from "react-stately";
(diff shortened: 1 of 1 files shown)


lloan (approved):
Looks like you're putting back things like they were before. Approving this so we can test on stage.

PR #4318, opened 2024-01-17
fix MPP-3700: empty DomainAddress.used_on|description fields when pro…

groovecoder:
…file.server_storage=False

The ServerStorageCleaner task cleans this data out, but this change means the task has less work to do each time it runs.

This PR fixes MPP-3700.

## How to test:
1. `pytest`
4. Go to http://127.0.0.1:8000/api/v1/docs/#/domainaddresses/domainaddresses_create
5. Make a domain address with values in `description` and `used_on` fields. E.g.:
    ```
    {
      "enabled": true,
      "description": "some description",
      "block_list_emails": true,
      "used_on": "some_description.com",
      "address": "somedesc"
    }
    ```
   * [ ] You should receive a `201` response
6. Go to http://127.0.0.1:8000/accounts/settings/
7. UNCHECK the "Enable account names for email masks" setting
8. Click "Save"
9. Go to http://127.0.0.1:8000/accounts/profile/
   * [ ] The label/description value from Step 5 should be gone
10. Go to http://127.0.0.1:8000/api/v1/domainaddresses/
   * [ ] Both the `description` and `used_on` values from Step 5 should be gone

## Checklist

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -162,6 +162,8 @@ class Profile(models.Model):
         if not self.server_storage:
             relay_addresses = RelayAddress.objects.filter(user=self.user)
             relay_addresses.update(description="", generated_for="", used_on="")
+            domain_addresses = DomainAddress.objects.filter(user=self.user)
+            domain_addresses.update(description="", used_on="")
         if settings.PHONES_ENABLED:
             # any time a profile is saved with store_phone_log False, delete the
             # appropriate server-stored InboundContact records
@@ -831,10 +833,11 @@ class DomainAddress(models.Model):
             self.block_list_emails = False
             if update_fields:
                 update_fields = {"block_list_emails"}.union(update_fields)
-        if (not user_profile.server_storage) and self.description:
+        if (not user_profile.server_storage) and (self.description or self.used_on):
             self.description = ""
+            self.used_on = ""
             if update_fields:
-                update_fields = {"description"}.union(update_fields)
+                update_fields = {"description", "used_on"}.union(update_fields)
         super().save(
             force_insert=force_insert,
             force_update=force_update,
--- a/emails/tests/models_tests.py
+++ b/emails/tests/models_tests.py
@@ -773,6 +773,18 @@ class ProfileSaveTest(ProfileTestCase):
             used_on=self.TEST_USED_ON,
         )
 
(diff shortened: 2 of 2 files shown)


emails/models.py
```diff
@@ -162,6 +162,8 @@ def save(
         if not self.server_storage:
             relay_addresses = RelayAddress.objects.filter(user=self.user)
             relay_addresses.update(description="", generated_for="", used_on="")
+            domain_addresses = DomainAddress.objects.filter(user=self.user)
+            domain_addresses.update(description="", used_on="")
```
jwhitlock, line 166: 👍 nice job adding it here to clear it when `server_storage` is cleared

emails/models.py
```diff
@@ -833,6 +835,7 @@ def save(
                 update_fields = {"block_list_emails"}.union(update_fields)
         if (not user_profile.server_storage) and self.description:
             self.description = ""
+            self.used_on = ""
             if update_fields:
                 update_fields = {"description"}.union(update_fields)
```
jwhitlock, line 840: need to update `update_fields` for `used_on`:

```suggestion
        if (not user_profile.server_storage) and (self.description or self.used_on):
            self.description = ""
            self.used_on = ""
            if update_fields:
                update_fields = {"description", "used_on"}.union(update_fields)
```

emails/tests/models_tests.py
```diff
@@ -1293,6 +1309,7 @@ def test_clear_storage_with_update_fields(self) -> None:
         domain_address = DomainAddress.objects.create(
             user=self.storageless_user, address="no-storage"
         )
+        assert domain_address.used_on == None
```
jwhitlock, line 1312: There is a later `asserts` in this test that need to change as well, looks like like 1333. The `domain_address.used_on` should be cleared after running `domain_address.save(update_fields={"last_used_at"})`

jwhitlock (changes requested):
Thanks @groovecoder! The manual steps worked for me. There's one change needed in the `DomainAddress.save()` method, to clear `used_on` when `update_fields` is used, which should make `test_clear_storage_with_update_fields` fail.

jwhitlock (approved):
Thanks @groovecoder, this fix looks complete to me!

PR #4320, opened 2024-01-17
MPP-3636: Emit glean pings as logs

jwhitlock:
For MPP-3636:

* Add `glean_parser` to requirements
* Add a prototype Glean metrics YAML
* Add prototype interface code to emit a ping as a log
* Add metrics pings for mask creation events:
  - Random mask created via the API
  - Domain mask created via the API
  - Domain mask created via an incoming email

This has related changes that could be moved to their own PR(s):

* Wrap code and comments that exceed 88 columns, making `flake8` happy in files touched by this PR
* Force `settings.RELAY_CHANNEL` to be one of "local", "dev", "stage", "prod". Previously, other values could be set in the environment.
* Fix an issue where recreating a domain mask raised an exception with a nonsensical message. This code now has tests.
* Refactor loading version information from `version.json`, created by CircleCI, so the version number can be used in the Glean code as well. This code is used by the `/__version__` view, and now has tests.
* Add type hints to `emails.models.Profile.fxa`, to say that it can return `None` or the user's `SocialAccount`. Add checks and assertions for the code that uses this property to check that the value is not `None`.
* Generate UUIDs for FxA IDs in tests, rather than very long random strings, to more closely match read FxA IDs.

## Glean Integration

This PR does not send data to Glean. That requires [several steps](https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/index.html):

* Complete the glean metrics YAML (in progress)
* Add a `python_server` outputter to `glean_parser` (using this PR as a guide)
* Add `glean_parser` to the build system
* Go through the [data review process](https://wiki.mozilla.org/Data_Collection)
* Add Relay to the probe scraper
* Forward glean logs to the Glean backend injestion system

This PR is similar to the initial Glean integration PRs on other projects:

* https://github.com/mozilla/fxa/pull/15659 (Node.js)
* https://github.com/MozillaSocial/mastodon/pull/36 (Ruby on Rails)

The prototype glean metrics YAML is at `telemetry/glean/relay-server-metrics.yaml`. This directory structure allows for future front-end metrics YAML, as well as other telemetry files for logs, metrics, etc. It has a single event, `email.generate_mask`, for mask generation and related data. The complete file will have many other events, and this event may change significantly.

The prototype interface file is at `privaterelay/glean/server_events.py`. It is a Python translation of the `javascript_server` format. First, I generated the JS version:

```sh
glean_parser translate -f javascript_server -o privaterelay/glean telemetry/glean/relay-server-metrics.yaml
```

[...]

PR #4341, opened 2024-01-22
MPP-3636: Fix domain address conflict handling

jwhitlock:
This code was extracted from PR #4320. The `DomainAddressViewSet` has special handling for -re-creating a domain address via the API, and turning it into a `409 Conflict` response. However, it passed a `dict` to the exception rather than an exception method, and was untested. This PR adds tests ~and returns a proper exception~, checks for a duplicate in the `save()` method, and returns a translated exception. It also wraps comments and other lines that exceed the 88 character limit.

## How to test:

- ~[ ] l10n changes have been submitted to the l10n repository, if any.~
- ~[ ] I've added or updated relevant docs in the docs/ directory.~
- ~[ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).~

### DIFF
--- a/api/exceptions.py
+++ b/api/exceptions.py
@@ -25,7 +25,38 @@ class ErrorData(OptionalErrorData):
 
 
 class RelayAPIException(APIException):
-    """Base class for exceptions that may be returned through API"""
+    """
+    Base class for exceptions that may be returned through the API.
+
+    Derived classes should set `default_code` to a unique string identifying the
+    exception. There should be a matching Fluent string with an `api-error-`
+    prefix. For example, the Fluent string "api-error-free-tier-limit" matches
+    the exception class with the default_code "free-tier-limit". These Fluent
+    strings are in misc.ftl.
+
+    Derived classes can set `default_detail` to a human-readable string, or they
+    can set `default_detail_template` to dynamically create the string from extra
+    context. When using default_detail_template, the Fluent string should use the same
+    variable names.
+
+    Derived classes can set `status_code` to the HTTP status code, or accept the
+    APIException default of 500 for a Server Error.
+    """
+
+    default_code: str
+    default_detail: str
+    status_code: int
+
+    def __init__(self, *args, **kwargs) -> None:
+        """Check that derived classes have set the required data."""
+        assert isinstance(self.default_code, str)
+        assert isinstance(self.status_code, int)
+        if hasattr(self, "default_detail_template"):
+            context = self.error_context()
+            assert context
+            self.default_detail = self.default_detail_template.format(**context)
+        assert isinstance(self.default_detail, str)
+        super().__init__(*args, **kwargs)
 
     def error_context(self) -> ErrorContextType:
         """Return context variables for client-side translation."""
@@ -34,7 +65,7 @@ class RelayAPIException(APIException):
     def error_data(self) -> ErrorData:
         """Return extra data for API error responses."""
 
-        # For RelayAPIException classes, this should always be a string
+        # For RelayAPIException classes, this is the default_code and is a string
         error_code = self.get_codes()
         assert isinstance(error_code, str)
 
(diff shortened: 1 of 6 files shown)


api/views/__init__.py
```diff
@@ -153,16 +154,19 @@ class DomainAddressViewSet(SaveToRequestUser, viewsets.ModelViewSet):
...
             domain_address = DomainAddress.objects.filter(
                 user=self.request.user, address=serializer.validated_data.get("address")
             ).first()
-            raise ConflictError(
-                {"id": domain_address.id, "full_address": domain_address.full_address}
-            )
+            existing_id = existing_address = None
+            if domain_address:
+                existing_id = domain_address.id
+                existing_address = domain_address.full_address
```
say-yawn, line 168: (suggestion, blocking) If there is an `IntegrityError` that is not from conflicting DomainAddress, I think we should just raise the `IntegrityError` instead. From reading `IntegrityError` [description on the Django doc](https://code.djangoproject.com/wiki/IntegrityError), a good test case would be trying to create a `DomainAddress` with a user ID that does not exist.

api/exceptions.py
```diff
@@ -10,6 +10,17 @@ class ConflictError(APIException):
     default_code = "conflict_error"
 
 
+class DomainAddressConflictError(ConflictError):
+    default_detail = "This domain address already exists."
```
say-yawn, line 14: (question, non-blocking) I forget what our rules are around adding translation strings to exceptions being raised. Wondering if you have experience and if this needs to be added to translation strings? My initial thought is that if this is user-facing we should add string for translation.

say-yawn (changes requested):
Thanks @jwhitlock for fixing flake8 issues, typos, and using more standard way to pass exception! I added some suggestions and questions that I would like to be addressed.

api/views/__init__.py
jwhitlock: I don't think that is possible through this API endpoint. We first check that the user is authorized, and if the user doesn't exist, we'd fail way before we get to creating a new `DomainAddress`.

I agree we should be raising the `IntegrityError`, which means the duplicate check needs to be somewhere else.

api/exceptions.py
jwhitlock: The rules are the ones we make! I think we worked to make exceptions translated for those that would get to the Firefox integration. Since we don't have a way to create domain addresses, this one is in the grey area. Still, I think it would be good to get into the habit of making these translated. That would mean basing it on `RelayAPIException`, not `ConflictError`.

emails/models.py
```diff
@@ -411,7 +414,9 @@ def update_abuse_metric(
         email_forwarded=False,
         forwarded_email_size=0,
     ):
-        #  TODO: this should be wrapped in atomic to ensure race conditions are properly handled
+        # TODO MPP-3720: This should be wrapped in atomic or select_for_update to ensure
```
jwhitlock, line 417: I opened MPP-3720 to track this TODO as (low priority) tech debt.

jwhitlock:
From a Slack conversation, made some more changes:

* Move code from `relay_exception_handler` to new `RelayAPIException.error_data()`
* Add documentation for `RelayAPIException`
* Check that derived classes implemented the required variables for `RelayAPIException`
* Handle `default_detail_template` in `RelayAPIException.__init__()` instead of derived classes

say-yawn (approved):
Thanks for better handling when the domain address is in conflict and refactoring the API exception so it's more obvious on how the error strings are translated.

(suggestion, non-blocking) Update `RelaySMSException` to also inherit from `RelayAPIException`.

jwhitlock:
> (suggestion, non-blocking) Update RelaySMSException to also inherit from RelayAPIException.

I think this is out of scope for this PR. I've filed MPP-3722 as tech debt.

PR #4346, opened 2024-01-24
Fix e2e tests for Mozilla Account changes

jwhitlock:
The end-to-end tests have been broken since January 19th, 2024. This may be due to a "Full prod rollout of react signup" on January 18th:

https://github.com/mozilla/fxa/commit/0881a3931a3df2fa6f61dfa60eac05df09699fbf

This PR fixes the end-to-end tests for me.


rafeerahman:
Thanks, confirming this fixes the failing tests with the new account creation auth page. Will only work on dev right now since there is maintenance going on with stage, once that maintenance is done it will pass.

Ran locally and here https://github.com/mozilla/fx-private-relay/actions/runs/7657716934.

PR #4321, opened 2024-01-17
for MPP-3467: add googlecloudprofiler to PrivateRelayConfig.ready

groovecoder:
# Google Cloud Profiler
https://cloud.google.com/profiler/docs
https://cloud.google.com/profiler/docs/profiling-python

# Screenshot (if applicable)
image

# How to test
Follow the steps in https://github.com/mozilla/fx-private-relay/pull/4321/files#diff-01748e8486974d80818eb9b8c9f623769f0ddb0821f867644d6e113f0dc008d3 for local server setup.

# Checklist (Definition of Done)
- [x] ~Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.~
- [x] ~All acceptance criteria are met.~
- [x] ~Jira ticket has been updated (if needed) to match changes made during the development process.~
- [x] I've added or updated relevant docs in the docs/ directory
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.
- [x] ~All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).~
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] ~l10n changes have been submitted to the l10n repository, if any.~

PR #4347, opened 2024-01-26
MPP-3698: Enable `mypy --check-untyped-def --no-implicit-reeexport` across code

jwhitlock:
For MPP-3698, I updated the code so both `--check-untyped-def` is used across our code. This means that code without type annotations is checked by `mypy` rather than skipped, and finds errors like not considering `None` or other alternate return types.

This is a big PR, but the changes themselves are small.

## Bugs Fixed

* `api/authentication.py`
  -  `introspect_token` - Error handler referred to `fxa_resp`, was not initialized (Sentry `RELAY-J0`, `RELAY-J2`)
* `api/permissions.py`
  - `HasPremium` called a `bool` (unused in code)
* `emails/management/commands/process_emails_from_sqs.py`
  - `Command.handle` - Use different name for exception inside of an exception handler
  
## Code changes
* Moved app config helper functions to app file. For example `emails_config()` is now in `emails.app.py`.
* Lots of new annotations
* Add `assert` to ensure a value is not `None` when `mypy` complained
* Add `isinstance` to ensure `request.user` is not an anonymous user in API code
* Add `hasattr` checks as needed for dynamically added fields
* Comments and long code lines wrapped to 88 chars (in changed files). Long function names have a `# noqa: E501` comment.
* Used walrus operator (`:=`) to assign and assert (usually not `None`) in one line
* Changes `Optional[X]` to Python 3.9 format `X | None`
* More use of `from __future__ import annotations` for [PEP-563](https://peps.python.org/pep-0563/), Postponed Evaluation of Annotations, to allow more forward annotations without quotes.
* `api/tests/authentication_tests.py`
  - Used instance of `FxaTokenAuthentication` instead of class in `FxaTokenAuthenticationTest`
  - Used Django 4.2 header-setting style in `FxaTokenAuthenticationTest`
* `privaterelay/settings.py` 
  - `RELAY_CHANNEL` is now restricted to "local", "dev", "stage", or "prod"
  - Added many annotations to `privaterelay/settings.py` for "circular import" issues. 
  - Removed `cast=str` when it is the default.
  - Set `AUTO_RELOAD_BUNDLES = False` to avoid `django-ftl` warning


jwhitlock:
PR #4321 is merged, this is ready for review.

privaterelay/settings.py
```diff
@@ -12,7 +12,7 @@
 
 from __future__ import annotations
 from pathlib import Path
-from typing import Any, Optional, TYPE_CHECKING
+from typing import Any, Literal, Optional, TYPE_CHECKING, cast, get_args
```
groovecoder, line 15: suggestion (non-blocking): change L719 to `sentry_release: str | None` and remove import of `Optional`

privaterelay/settings.py
```diff
@@ -226,7 +233,7 @@
 
 DJANGO_STATSD_ENABLED = config("DJANGO_STATSD_ENABLED", False, cast=bool)
 STATSD_DEBUG = config("STATSD_DEBUG", False, cast=bool)
-STATSD_ENABLED = DJANGO_STATSD_ENABLED or STATSD_DEBUG
+STATSD_ENABLED: bool = DJANGO_STATSD_ENABLED or STATSD_DEBUG
 STATSD_HOST = config("DJANGO_STATSD_HOST", "127.0.0.1")
 STATSD_PORT = config("DJANGO_STATSD_PORT", "8125")
 STATSD_PREFIX = config("DJANGO_STATSD_PREFIX", "fx.private.relay")
```
groovecoder, line 239: question (non-blocking): I can't figure out why we declare the type of `STATSD_ENABLED` but none of these other ones? Does the default value in the `config` call effectively declare the type for `mypy`? I can't tell what the pattern is of when we do/don't declare the type of a `settings` variable.

privaterelay/tests/fxa_utils_tests.py
```diff
@@ -65,16 +63,13 @@ def test_get_phone_subscription_dates_subscription_not_in_data_no_free_phone(
     )
 
 
+@override_flag("free_phones", active=True)
```
groovecoder, line 66: praise: nice clean-up.

privaterelay/tests/mgmt_sync_phone_related_dates_on_profile_tests.py
```diff
@@ -85,13 +86,12 @@ def test_free_phone_user_gets_only_date_phone_subscription_reset_field_updated(
...
     expected_now = patch_datetime_now
     mocked_dates.return_value = (None, None, None)
-    account = baker.make(SocialAccount, provider="fxa")
+    account: SocialAccount = baker.make(SocialAccount, provider="fxa")
     profile = Profile.objects.get(user=account.user)
     baker.make(Flag, name="free_phones")
-    free_phones_flag = Flag.objects.filter(name="free_phones").first()
+    free_phones_flag = Flag.objects.get(name="free_phones")
     free_phones_flag.users.add(profile.user)
     free_phones_flag.save()
```
groovecoder, line 94: suggestion (non-blocking): this block is used at least twice. Can we move this logic into a `get_user_free_phones()` helper function?

pyproject.toml
```diff
@@ -55,40 +55,3 @@ module = [
...
-    "emails.tests.models_tests",
-    "emails.tests.views_tests",
-    "emails.utils",
-    "phones.models",
-    "phones.tests.models_tests",
-    "privaterelay.tests.fxa_utils_tests",
-    "privaterelay.tests.mgmt_sync_phone_related_dates_on_profile_tests",
-    "privaterelay.tests.mgmt_update_phone_remaining_stats_tests",
-    "privaterelay.tests.views_tests",
-]
```
groovecoder, line 94: praise: woohoo!! look at all these modules with better type-checking now!

groovecoder (changes requested):
One blocking question about returning `401` status code that's very important to get right because Firefox clients change their behavior based on that status code.

jwhitlock (commented):
Thanks for the quick review @groovecoder! I've answered some questions, and made two suggested changes:

* Removed `Optional` from `privaterelay/settings.py`
* Added `create_free_phones_flag_for_user` helper

api/authentication.py
```diff
@@ -85,18 +86,18 @@ def get_fxa_uid_from_oauth_token(token, use_cache=True):
         raise APIException("Did not receive a 200 response from FXA.")
 
     if not fxa_resp_data["json"].get("active"):
-        raise AuthenticationFailed("FXA returned active: False for token.", code=401)
+        raise AuthenticationFailed("FXA returned active: False for token.")
```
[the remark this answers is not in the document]
groovecoder: Thanks for the detailed info! That, and the fact that `test_inactive_token_responds_with_401` test makes sure the `401` is returned in the right case, and that test is still passing, makes me non-blocking on this again!

groovecoder (approved):

api/authentication.py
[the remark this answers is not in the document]
say-yawn: I found out that `code` is returned in response data like:
```python
> response.data
{'detail': ErrorDetail(string='FXA returned active: False for token.', code='authentication_failed')}
```
And I checked in the [Firefox code](https://searchfox.org/mozilla-central/rev/3c72de9280ec57dc55c24886c6334d9e340500e8/toolkit/components/satchel/integrations/FirefoxRelay.sys.mjs#108) @groovecoder mentioned and it does not refer to `code` in the data details. So I also think we are good.

PR #4366, opened 2024-01-29
create Google app creds file from env var as needed

groovecoder:
This PR fixes MPP-3467.

It changes the profiling code to automatically create the necessary GCP key JSON file from the value of the `GOOGLE_CLOUD_PROFILER_CREDENTIALS_B64` environment variable. This should work identically across local, dev, stage, and prod environments.

How to test:
1. Follow [the doc](https://github.com/mozilla/fx-private-relay/pull/4366/files#diff-01748e8486974d80818eb9b8c9f623769f0ddb0821f867644d6e113f0dc008d3)!
2. When you `runserver` you should see output lines: `HTTPAdapter.get_connection url: http://metadata/computeMetadata/v1/instance/zone` which indicates that the profiler code is working


privaterelay/apps.py
```diff
@@ -29,30 +30,44 @@ def get_profiler_startup_data() -> tuple[str | None, str | None]:
...
+        ):
             # Set up Google Cloud Profiler
             service, version = get_profiler_startup_data()
             if service != None:
-                import googlecloudprofiler
-
-                # Make sure the expect gcp_key.json file exists
                 gcp_key_json_path = Path(settings.GOOGLE_APPLICATION_CREDENTIALS)
-                if gcp_key_json_path.exists() and gcp_key_json_path.is_file():
+                if not gcp_key_json_path.exists() or not gcp_key_json_path.is_file():
```
jwhitlock, line 53: If you set `GOOGLE_APPLICATION_CREDENTIALS` to `"privaterelay"`, this fails with `IsADirectoryError: [Errno 21] Is a directory: 'privaterelay'`. I think the line should be:

```suggestion
                if not gcp_key_json_path.exists() and gcp_key_json_path.is_file():
```

At the same time, you'll get a `IsADirectoryError` in a couple of lines when you try to read it. So, maybe drop the `is_file()` check and let Relay crash on startup.

jwhitlock (changes requested):
Thanks @groovecoder, this looks a little safer. Please consider what should happen if `GOOGLE_APPLICATION_CREDENTIALS` is a directory - raise an exception, or skip loading the profiler?

privaterelay/apps.py
groovecoder: I removed the `is_file` check and expanded the `try/except` block around the code for the file so that the Relay app can still start up and run even if google cloud profiler can't.

jwhitlock (approved):
Thanks, this worked with my tests (folder, empty json). One suggestion you could take or leave.

Issue #2731, opened 2022-10-31
Upgrade to Python 3.11


jwhitlock:
[Python 3.11.0](https://www.python.org/downloads/release/python-3110/) was released October 24, 2022. ~We'll want to upgrade to Python 3.10 first (#2452)~ _done!_, and then ensure that our dependencies support Python 3.11.

![](https://user-images.githubusercontent.com/11718525/197611877-583a0bb2-a8fb-4275-8827-39f2f06ade6c.png)

Some of the new major new features and changes in Python 3.11 are:

## General changes

* [PEP 657](https://www.python.org/dev/peps/pep-0657/) -- Include Fine-Grained Error Locations in Tracebacks
* [PEP 654](https://www.python.org/dev/peps/pep-0654/) -- Exception Groups and `except*`
* [PEP 680](https://www.python.org/dev/peps/pep-0680/) -- tomllib: Support for Parsing TOML in the Standard Library
* [gh-90908](https://github.com/python/cpython/issues/90908) -- Introduce task groups to asyncio
* [gh-34627](https://github.com/python/cpython/issues/34627/) -- Atomic grouping `((?>...))` and possessive quantifiers `(*+, ++, ?+, {m,n}+)` are now supported in regular expressions.
* The [Faster CPython Project](https://github.com/faster-cpython/) is already yielding some exciting results. Python 3.11 is up to 10-60% faster than Python 3.10. On average, we measured a 1.22x speedup on the standard benchmark suite. See [Faster CPython](https://docs.python.org/3.11/whatsnew/3.11.html#faster-cpython) for details.

## Typing and typing language changes

* [PEP 673](https://www.python.org/dev/peps/pep-0673/) -- Self Type
* [PEP 646](https://www.python.org/dev/peps/pep-0646/) -- Variadic Generics
* [PEP 675](https://www.python.org/dev/peps/pep-0675/) -- Arbitrary Literal String Type
* [PEP 655](https://www.python.org/dev/peps/pep-0655/) -- Marking individual TypedDict items as required or potentially-missing
* [PEP 681](https://www.python.org/dev/peps/pep-0681/) -- Data Class Transforms
 
```[tasklist]
### Tasks
- [ ] https://github.com/mozilla/fx-private-relay/issues/2452
- [ ] https://github.com/mozilla/fx-private-relay/issues/3126
```

jwhitlock:
[...]

PR #4368, opened 2024-01-30
Update to Python 3.11.7

jwhitlock:
For MPP-2530 / Issue #2731, Update from Python 3.10 to 3.11.7. Basic tests have been running in CircleCI with no issues.


jwhitlock:
Everything passes, but we do not have capacity to deploy this in our current two-week sprint. We'll revisit in the near future.

PR #4372, opened 2024-02-01
MPP-3726 Set and get when user profile phone was registered

say-yawn:
# New feature description
To supply helpful data to users on Relay add-on's phone tab, provide when the user set registered their phone.
`/api/v1/profiles` has a new key `date_phone_registered` that:
- If user does not have RealPhone set return None
- If user does have RealPhone set return verified_date
- If user does have RelayNumber and created_at date is not None, return created_at

# How to test
Go to `/api/v1/profiles` and check that there is `date_phone_registered` that returns the values described in above section

- [ ] Commits in this PR are minimal and [have descriptive commit m


emails/models.py
```diff
@@ -390,6 +393,20 @@ def joined_before_premium_release(self):
...
+            return None
+
+        try:
+            real_phone = RealPhone.objects.get(user=self.user)
+            relay_number = RelayNumber.objects.get(user=self.user)
+        except RealPhone.DoesNotExist:
+            return None
+        except RelayNumber.DoesNotExist:
+            return real_phone.verified_date
+        return relay_number.created_at or real_phone.verified_date
```
jwhitlock, line 408: Clever, I wondered how you were going to do this without populating `created_at`

emails/tests/models_tests.py
```diff
@@ -677,6 +695,78 @@ class ProfileHasPhoneTest(ProfileTestCase):
     def test_default_False(self) -> None:
         assert self.profile.has_phone is False
 
+    def test_phone_subscription_returns_True(self) -> None:
+        self.upgrade_to_phone_premium()
+        assert self.profile.has_phone is True
+
+
+@pytest.mark.skipif(not settings.PHONES_ENABLED, reason="PHONES_ENABLED is False")
```
jwhitlock, line 703: 👍 good use of this decorator

emails/tests/models_tests.py
```diff
@@ -677,6 +695,78 @@ class ProfileHasPhoneTest(ProfileTestCase):
...
+        datetime_now = datetime.now(timezone.utc)
+        RealPhone.objects.create(
+            user=self.profile.user,
+            number="+12223334444",
+            verified=True,
+            verified_date=datetime_now,
+        )
+        assert self.profile.date_phone_registered == datetime_now
+
+    @override_settings(PHONES_NO_CLIENT_CALLS_IN_TEST=True)
```
jwhitlock, line 735: Please move this up as a decorator for the class, and remove the `setUp` and `tearDown` methods

phones/models.py
```diff
@@ -182,7 +182,7 @@ def mark_verified(self):
         return self
 
 
-@receiver(post_save, sender=RealPhone)
+@receiver(post_save, sender=RealPhone, dispatch_uid="realphone_post_save")
```
jwhitlock, line 185: Even though we don't need it for this change, a `dispatch_uid` is still a good idea, so please keep this.

jwhitlock (changes requested):
Thanks @say-yawn! This looks good, and I like the tests as well.

I think `@override_settings(PHONES_NO_CLIENT_CALLS_IN_TEST=True)` is a better solution than disconnecting and reconnecting the signal, and makes it unnecessary to disconnect the signal. Let's simplify the code and just use `@override_settings` as a class decorator.

emails/tests/models_tests.py
say-yawn: Done 677eed23b0eaf1d2e599d63dab5b8bd12d83eeba

jwhitlock (approved):
Thanks, looks great!

PR #4453, opened 2024-02-28
MPP-3762: Handle order issue with `useGoogleAnalytics`

jwhitlock:
In some cases (such as `npn dev:mocked`, or occasionally in the browser), the function `initGoogleAnalytics` is run before `useGoogleAnalytics` has replaced `globalEnableGoogleAnalytics`. This changes `globalEnableGoogleAnalytics` to start with a default of `null`. When it is not yet a function, `gaIsInitialized` is changed instead, to provide a default of `true` for the `singletonHook`.


rafeerahman (approved):
Looks good, this fixes the inconsistency

PR #4455, opened 2024-02-29
MPP-3644: Update more uses of `useIsLoggedIn`

jwhitlock:
Previously, this returned `true` or `false`, but now it returns "unknown", "logged-in", or "logged-out". Update some more checks that expected a boolean.

This is (one of?) the causes of the failing Playwright tests, and was caused by #4392. One indication is that the logo in the upper left always links to `/accounts/profile`, even when logged-out.


rafeerahman (approved):
Thanks, confirmed that the tests pass locally, and double checked the uses of useIsLoggedIn. 

(non-blocking)

I believe `profilesData.isLoading` (isLoading comes from useSWR) could be used [here](https://github.com/mozilla/fx-private-relay/blob/d9780cddbfdcff878e63396eafed8220def6bf64/frontend/src/hooks/session.ts#L15) instead of `checking`. I think they accomplish the same thing so it doesn't really matter.

jwhitlock:
> I believe profilesData.isLoading (isLoading comes from useSWR) could be used [here](https://github.com/mozilla/fx-private-relay/blob/d9780cddbfdcff878e63396eafed8220def6bf64/frontend/src/hooks/session.ts#L15) instead of checking. I think they accomplish the same thing so it doesn't really matter.

Good suggestion, this worked for me locally and will be easier to find in the documentation.

PR #4477, opened 2024-03-04
Switch to cimg/base for Docker commands

jwhitlock:
These images should have the same functionality of the Docker-provided images, but should be heavily cached and download quicker. 🤞


groovecoder (approved):

jwhitlock:
One run was fast, the next run was slow, so I don't think this will speed up builds in general. I think getting on the CircleCI images will help in the future.

PR #4538, opened 2024-03-25
CircleCI Fix: Remove merged strings from pending.ftl

rafeerahman:
This fixes failing CircleCI tests by removing pending.ftl strings that were merged in https://github.com/mozilla/fx-private-relay/commit/c61dbeda5ddc5ee62413a5ce41d91ca6a491343f, which caused the tests to break.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ If you're not yet ready to submit some strings for translation, you can
 tentatively add them to `frontend/pendingTranslations.ftl`. Strings in that file
 will show up until strings with the same ID are added to the l10n repository.
 
+Similarly, there is a `pending_locales/pending.ftl` where temporary backend locales strings can be stored. Once the strings from the pull request in [the l10n repo](https://github.com/mozilla-l10n/fx-private-relay-l10n) has been merged into the Relay repo, these respective strings need to be removed from `pending_locales/pending.ftl` to avoid failing CircleCI tests.
+
 #### Commit translations for release
 
 To commit updates to the app's translations (e.g., before a release), we need
--- a/privaterelay/pending_locales/en/pending.ftl
+++ b/privaterelay/pending_locales/en/pending.ftl
@@ -3,8 +3,3 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 # This is the Django equivalent of frontend/pendingTranslations.ftl
-
-# Variables:
-#   $duplicate_address (string) - User-set email address that already exists
-api-error-duplicate-address = “{ $duplicate_address }” already exists. Please try again with a different mask name.
-api-error-address-not-editable = You cannot edit an existing domain address field.


say-yawn (approved):
Changes look good, the strings that were temporarily in the pending.ftl files have been [merged in the l10n repo](https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/169). 

(suggestion, non-blocking) The [instructions on README for pending strings](https://github.com/mozilla/fx-private-relay?tab=readme-ov-file#working-with-translations) only mentions:
> If you're not yet ready to submit some strings for translation, you can tentatively add them to frontend/pendingTranslations.ftl. Strings in that file will show up until strings with the same ID are added to the l10n repository. 

Should we add another line in README and on `pending.ftl` indicating pending strings should be removed after string is merged in the [l10n repo](https://github.com/mozilla-l10n/fx-private-relay-l10n)?

rafeerahman:
> Changes look good, the strings that were temporarily in the pending.ftl files have been [merged in the l10n repo](https://github.com/mozilla-l10n/fx-private-relay-l10n/pull/169).
> 
> (suggestion, non-blocking) The [instructions on README for pending strings](https://github.com/mozilla/fx-private-relay?tab=readme-ov-file#working-with-translations) only mentions:
> 
> > If you're not yet ready to submit some strings for translation, you can tentatively add them to frontend/pendingTranslations.ftl. Strings in that file will show up until strings with the same ID are added to the l10n repository.
> 
> Should we add another line in README and on `pending.ftl` indicating pending strings should be removed after string is merged in the [l10n repo](https://github.com/mozilla-l10n/fx-private-relay-l10n)?

Yes, this will be a good addition. I recall this same issue happened before because pending strings were kept. Do you want me to add that change in this PR?

say-yawn:
(comment) Interesting to note that the [frontend/pendingTranslations.ftl](https://github.com/mozilla/fx-private-relay/blob/main/frontend/pendingTranslations.ftl) still has old strings that are merged but does not error. I wonder if there is a configuration to use the l10n repo strings if the string is found in both pending and in translations repo.

say-yawn:
@rafeerahman, adding the comments in this PR would be helpful.

rafeerahman:
> (comment) Interesting to note that the [frontend/pendingTranslations.ftl](https://github.com/mozilla/fx-private-relay/blob/main/frontend/pendingTranslations.ftl) still has old strings that are merged but does not error. I wonder if there is a configuration to use the l10n repo strings if the string is found in both pending and in translations repo.

Yeah, the issue seems to be in https://github.com/mozilla/fx-private-relay/blob/main/privaterelay/ftl_bundles.py#L29-L40, where Bundle() causes an error if there are duplicate strings from the files listed. Would be nice if we could have a configuration to ignore/fix this.

Vinnl:
> (comment) Interesting to note that the [frontend/pendingTranslations.ftl](https://github.com/mozilla/fx-private-relay/blob/main/frontend/pendingTranslations.ftl) still has old strings that are merged but does not error. I wonder if there is a configuration to use the l10n repo strings if the string is found in both pending and in translations repo.

Yep, the frontend only uses the pending strings if the IDs aren't found among the l10n repo's strings, but otherwise ignores them - i.e. they're a fallback only.

PR #4544, opened 2024-03-26
MPP-2539 - Update to Python 3.11

jwhitlock:
For MPP-2539 / Issue #2731 , this PR updates from Python 3.10 to 3.11.

We may release an update as `2024.03.21.1`, to isolate any issues caused by 3.11.


say-yawn (approved):
Changes LGTM especially since we have been running tests against 3.11 and will be deploying this in a contained release with upgraded Python as the only change.

jwhitlock:
Pushed to production as `2024.03.21.1`, works there as well.

PR #4519, opened 2024-03-19
MPP-3753: Disable DomainAddress address field updates

rafeerahman:
This PR fixes MPP-3753.

# New feature description

Disables updates to the address field for DomainAddress'.

# Screenshot (if applicable)

Not applicable.

# How to test

1. Log in or create account with premium emails and a domain locally

2. Create a new domain mask

3. Go to api/v1/docs/#/domainaddresses/domainaddresses_list, select GET /api/v1/domainaddresses/, click “Try it out”, scroll down and then click “Execute”, to get the list of domain addresses. Get the ID for the new domain mask.

4. Still on  api/v1/docs/#/domainaddresses/domainaddresses_list, select PATCH /api/v1/domainaddresses/{id}/. Click “Try it out”, enter the ID for the new domain mask, and enter JSON that changes the address, such as {"address": "Huh-You-Can-Change-This"}.

# Checklist (Definition of Done)
- [x] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [x] Customer Experience team has seen or waived a demo of functionality.
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.
- [x] I've added or updated relevant docs in the docs/ directory
- [x] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.
- [x] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] l10n changes have been submitted to the l10n repository, if any.

PR #4546, opened 2024-03-26
MPP-3753: More test changes for write-once DomainAddress address

jwhitlock:
This is follow-on work for PR #4519:

* Remove the TODO and the `address` test case for `test_patch_domainaddress_read_only`. If a caller tries to change the `address`, it is a 400 error, not a 200, as tested in `test_patch_domainaddress_address_fails`
* Wrap a comment to under 88 lines. This will be enforced soon with `ruff`.


rafeerahman (approved):
Thanks! Looks good.

PR #4547, opened 2024-03-27
MPP-2530 - Update to Python 3.11 on 2024.03.21

jwhitlock:
These are the same changes as PR #4544, but applied to `2024.03.21`. When tests pass, this will become tag `2024.03.21.1`.


jwhitlock:
Released to stage as tag [2024.03.21.1](https://github.com/mozilla/fx-private-relay/releases/tag/2024.03.21.1).

PR #4548, opened 2024-03-27
MPP-2530 - Modernize code with `pyupgrade --py311-plus`

jwhitlock:
[pyupgrade](https://pypi.org/project/pyupgrade/) can reformat code to use update-to-date syntax and library usage. `--py311-plus` will use features in 3.11 and later. We'll be able to enforce this style with [ruff](https://github.com/astral-sh/ruff).

# How to test:
- CircleCI continues to pass


groovecoder (approved):
Lots of little code tweaks and tests all pass.

PR #4551, opened 2024-03-28
MPP-3780: Tune glean query

jwhitlock:
The glean logging looks for a user's first mask with `generated_for`, to guess when they started using the add-on or Firefox integration. This query was very slow in production until this index was added:

```sql
CREATE INDEX incident_glean_202403
 ON emails_relayaddress (user_id)
 INCLUDE (created_at)
 WHERE (generated_for != '');
```

This PR adds this partial index with the Django model syntax. The generated PostgreSQL index statement is:

```sql
CREATE INDEX "idx_ra_created_by_addon"
 ON "emails_relayaddress" ("user_id")
 INCLUDE ("created_at")
 WHERE NOT ("generated_for" = '');
```

SQLite3 does not allow `INCLUDE ("created_at")`, so it is omitted. There is a warning `"models.W040"` that is silenced.

The migration will also drop the `incident_glean_202403` index if found, after creating the new index.

Additionally, the query now only loads two fields, such as:

```sql
SELECT "emails_relayaddress"."id", "emails_relayaddress"."created_at"
 FROM "emails_relayaddress"
 WHERE ("emails_relayaddress"."user_id" = 1 AND
   NOT ("emails_relayaddress"."generated_for" = ''))
 ORDER BY "emails_relayaddress"."created_at" ASC
 LIMIT 1;
```

This will *hopefully* mean the index has all the required information for this query, making it even more efficient.


say-yawn (approved):
Looks good. Today I learned about partial index and learned how to add index via model's metadata.

PR #4566, opened 2024-04-01
Group boto and boto types

jwhitlock:
boto dependencies are generated from API specifications and change weekly. The typing modules are often modified on the same schedule. Grouping them will avoid mismatches and help the typing group update less often.

This was suggested by @groovecoder on February 12.


jwhitlock:
This is non-production code, so I'm self-reviewing.

PR #4518, opened 2024-03-19
E2E fixes for FxA auth page changes (2024-03-19)

rafeerahman:
In the FxA production environment, changes for account creation changes handled in https://github.com/mozilla/fx-private-relay/pull/4346 were reverted, causing failing tests for Relay prod. The changes were not reverted for the stage FxA auth page as of today, so no updates are needed there.

# Checklist (Definition of Done)
- [ ] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [ ] Customer Experience team has seen or waived a demo of functionality.
- [ ] All acceptance criteria are met.
- [ ] Jira ticket has been updated (if needed) to match changes made during the development process.
- [ ] I've added or updated relevant docs in the docs/ directory
- [ ] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.
- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [ ] l10n changes have been submitted to the l10n repository, if any.

PR #4570, opened 2024-04-02
E2E test fixes for stage auth changes

rafeerahman:
The changes done in https://github.com/mozilla/fx-private-relay/pull/4518 to fix tests on production are also needed in stage now since the stage FxA auth page for account creation was reverted.

Ran the workflow here and tests pass, https://github.com/mozilla/fx-private-relay/actions/runs/8526118812


jwhitlock (approved):
Thanks @rafeerahman!

PR #4573, opened 2024-04-05
MPP-3784: Add FxA ID to logs for bounces and complaints

jwhitlock:
This adds the user's FxA ID to the log when we receive a bounce or complaint. This is an aid in debugging an increase in either metric, allowing us to see if a particular user is responsible for the increase. It can also aid in customer support, to diagnose if a user's email is bouncing. If the user opts-out of Mozilla account metrics, their FxA ID will not be included in the data.

This also removes some legacy logs marked for deletion last year.

# How to Test

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -1474,7 +1474,7 @@ def _handle_bounce(message_json: AWS_SNSMessageJSON) -> HttpResponse:
             user = User.objects.get(email=recipient_address)
             profile = user.profile
             data["user_match"] = "found"
-            if fxa := profile.fxa:
+            if (fxa := profile.fxa) and profile.metrics_enabled:
                 data["fxa_id"] = fxa.uid
         except User.DoesNotExist:
             # TODO: handle bounce for a user who no longer exists
@@ -1577,7 +1577,7 @@ def _handle_complaint(message_json: AWS_SNSMessageJSON) -> HttpResponse:
             user = User.objects.get(email=recipient_address)
             profile = user.profile
             data["user_match"] = "found"
-            if fxa := profile.fxa:
+            if (fxa := profile.fxa) and profile.metrics_enabled:
                 data["fxa_id"] = fxa.uid
         except User.DoesNotExist:
             data["user_match"] = "missing"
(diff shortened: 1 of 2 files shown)


emails/views.py
```diff
@@ -1476,6 +1474,8 @@ def _handle_bounce(message_json: AWS_SNSMessageJSON) -> HttpResponse:
             user = User.objects.get(email=recipient_address)
             profile = user.profile
             data["user_match"] = "found"
+            if (fxa := profile.fxa) and profile.metrics_enabled:
+                data["fxa_id"] = fxa.uid
```
say-yawn, line 1478: (praise) Clever use of walrus operator!

say-yawn: (question) Does `fxa_uid` not being included when `data` variable is set might make the log hard to query/parse?

emails/views.py
```diff
@@ -1518,19 +1518,6 @@ def _handle_bounce(message_json: AWS_SNSMessageJSON) -> HttpResponse:
         )
         info_logger.info("bounce_notification", extra=data)
 
-        # Legacy log, can be removed Q4 2023
```
say-yawn, line 1521: (praise) YAY FOR DELETING LEGACY CODE/LOGS.

emails/views.py
```diff
@@ -1557,11 +1544,7 @@ def _handle_complaint(message_json: AWS_SNSMessageJSON) -> HttpResponse:
     * complaint_user_agent - identifies the client used to file the complaint
     * complaint_extra - Extra data from complainedRecipients data, if any
     * domain - User's domain, if an address was given
-
-    Emits a legacy log "complaint_received", with data:
-    * recipient_domains: list of extracted user domains
-    * subtype: 'onaccounsuppressionlist', or 'none'
-    * feedback: feedback from ISP or 'none'
+    * fxa_id - The Mozilla account ID of the user
```
say-yawn, line 1547:
```suggestion
    * fxa_id - The Mozilla account (previously known as Firefox Accounts) ID of the user
```
(suggestion, non-blocking) In case Relay survives long enough and Firefox Accounts have long been dropped for Mozilla Accounts.

say-yawn (approved):
Thanks for adding information to further diagnose complaints and bounces. Some question and suggestion that are non-blocking.

emails/views.py
jwhitlock: Good point. There's work-arounds, but it would be better to set it to an empty string instead.

emails/views.py
jwhitlock: and in 2027 we can change it to

```suggestion
    * fxa_id - The Firefox account (previously known as Mozilla account) ID of the user
```

PR #4587, opened 2024-04-08
Move Python configuration to `pyproject.toml`

jwhitlock:
Make several changes to centralize Python configuration in `pyproject.toml`:

* Set the minimum Python version to 3.11, which allows `black` to use [parenthesized context managers](https://docs.python.org/3.10/whatsnew/3.10.html#parenthesized-context-managers) from Python 3.10.
* Move config for [paul-mclendahand](https://github.com/willkg/paul-mclendahand) from `setup.cfg` to `pyproject.toml`
* Move config for `pytest` from `pytest.ini` to `pyproject.toml`
* Alphabetize `pyproject.toml` where appropriate


jwhitlock:
This is all non-production code (configuration and tests), so I'm skipping review.

PR #4589, opened 2024-04-09
MPP-3786: Settings additions for security and dockerflow configuration

jwhitlock:
This PR adds environment overrides for some security settings and dockerflow configuration that we may want to enable in production:

* [SECRET_KEY_FALLBACKS](https://docs.djangoproject.com/en/4.2/ref/settings/#secret-key-fallbacks) - Allows rotating `SECRET_KEY` without instantly invalidating sessions.
* [SECURE_HSTS_INCLUDE_SUBDOMAINS](https://docs.djangoproject.com/en/4.2/ref/settings/#secure-hsts-include-subdomains) - Adds `includeSubDomains` to `Strict-Transport-Security` header. See [HTTP Strict Transport Security](https://docs.djangoproject.com/en/4.2/ref/middleware/#http-strict-transport-security) for more info.
* [SECURE_HSTS_PRELOAD](https://docs.djangoproject.com/en/4.2/ref/settings/#secure-hsts-preload) - Adds `preload` to `Strict-Transport-Security` header
* [CSRF_COOKIE_SECURE](https://docs.djangoproject.com/en/4.2/ref/settings/#csrf-cookie-secure) - Adds 'secure' to the CSRF cookie.
* `DOCKERFLOW_REQUEST_ID_HEADER_NAME` - Sets the name for a header that dockerflow should use for an externally set request ID. Both Heroku and GCP set a request ID, but under different headers.
* [SILENCED_SYSTEM_CHECKS](https://docs.djangoproject.com/en/4.2/ref/settings/#silenced-system-checks) - A list of system check messages to ignore. Dockerflow consults this list for the healthcheck endpoint.

It also updates the logger config:

* Add `dockerflow` log to the stderr stream at `WARNING` level. This is where failed healthcheck messages go.
* Add dockerflow's `RequestIdLogFilter`. This will add a request ID to log messages, so we can correlate the logs for a single request.

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -731,6 +731,7 @@ if RELAY_CHANNEL == "local":
     # In local dev, next runs on localhost and makes requests to /accounts/
     CORS_ALLOWED_ORIGINS += [
         "http://localhost:3000",
+        "http://0.0.0.0:3000",
         "http://127.0.0.1:8000",
     ]
     CORS_URLS_REGEX = r"^/(api|accounts)/"
@@ -751,10 +752,6 @@ if RELAY_CHANNEL == "local":
     # origin and thus has access to the same cookies.
     CORS_ALLOW_CREDENTIALS = True
     SESSION_COOKIE_SAMESITE = None
-    CORS_ALLOWED_ORIGINS += [
-        "http://localhost:3000",
-        "http://0.0.0.0:3000",
-    ]
     CSRF_TRUSTED_ORIGINS += [
         "http://localhost:3000",
         "http://0.0.0.0:3000",
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -660,6 +660,11 @@ LOGGING = {
             "level": "DEBUG",
             "propagate": IN_PYTEST,
         },
+        "dockerflow": {
+            "handlers": ["console_err"],
+            "level": "WARNING",
+            "propagate": IN_PYTEST,
+        },
     },
 }
 
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -871,6 +871,7 @@ DOCKERFLOW_CHECKS = [
 ]
 if REDIS_URL:
     DOCKERFLOW_CHECKS.append("dockerflow.django.checks.check_redis_connected")
+DOCKERFLOW_REQUEST_ID_HEADER_NAME = config("DOCKERFLOW_REQUEST_ID_HEADER_NAME", None)
 SILENCED_SYSTEM_CHECKS = config("SILENCED_SYSTEM_CHECKS", "", cast=Csv())
 
 # django-ftl settings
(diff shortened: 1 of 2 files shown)


privaterelay/settings.py
```diff
@@ -95,9 +96,14 @@
     r"^__heartbeat__",
     r"^__lbheartbeat__",
 ]
+SECURE_HSTS_INCLUDE_SUBDOMAINS = config(
+    "DJANGO_SECURE_HSTS_INCLUDE_SUBDOMAINS", False, cast=bool
```
groovecoder, line 100: question (blocking): We're setting this to `False` here, but are we planning to set it to `True` in production? The warning in the docs ("Setting this incorrectly can irreversibly (for the value of [SECURE_HSTS_SECONDS](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-SECURE_HSTS_SECONDS)) break your site.") makes me nervous about setting this on any *.firefox.com subdomain. If a client sees this parameter on relay.firefox.com, could a client apply this to accounts.firefox.com, data.firefox.com, etc.?

privaterelay/settings.py
```diff
@@ -725,6 +743,7 @@ def set_index_cache_control_headers(
     # In local dev, next runs on localhost and makes requests to /accounts/
     CORS_ALLOWED_ORIGINS += [
         "http://localhost:3000",
+        "http://0.0.0.0:3000",
```
groovecoder, line 746: praise: nice cleanup.

groovecoder (changes requested):

privaterelay/settings.py
jwhitlock: No, this would only affect relay.firefox.com and subdomains under relay.firefox.com, such as www.relay.firefox.com. The specification at [RFC 6797 Section 8.2 Known HSTS Host Domain Name Matching](https://www.rfc-editor.org/rfc/rfc6797#section-8.2) has some examples. The domains accounts.firefox.com or data.firefox.com are not affected, because they are not subdomains of relay.firefox.com.

Our current header on relay.firefox.com is:

```
strict-transport-security: max-age=31536000
```

This tells browsers to remember to only use HTTPS on relay.firefox.com for 1 year. https://hstspreload.org/ recommends 2 years. We can lower this age while we are testing changes, such as reducing it to 300 (5 minutes)

The current header for firefox.com is:

```
strict-transport-security: max-age=0
```

According to [RFC 6797 Section 6.2 Examples](www.rfc-editor.org/rfc/rfc6797#section-6.2), this is interpreted as:

> The HSTS header field ... indicates that the UA must delete the entire HSTS Policy associated with the HSTS Host that sent the header field

This does not include `includeSubDomains`, so it says nothing about relay.firefox.com, data.firefox.com, etc.

accounts.firefox.com uses 1 year and includes subdomains:

```
strict-transport-security: max-age=31536000; includeSubDomains
```

data.firefox.com does not set the header.

groovecoder (approved):

PR #4609, opened 2024-04-16
MPP-3786: More settings changes for dockerflow

jwhitlock:
* Refactor `Content-Security-Policy` (CSP) settings:
  - Move all the CSP logic to the same section, rather than scattered over the settings.
  - Use leading underscore `_` for variables only used in `privaterelay/settings.py`
  - Base Mozilla Account URLS on `FXA_BASE_ORIGIN` instead of `FXA_PROFILE_ENDPOINT`
  - Use stage and production Mozilla account hosts for `connect_src`
  - If `'unsafe-inline'` is used for styles, do not include hashes in `style-src`. If they are included, browsers will ignore `unsafe-inline`. 
  - Allow setting a report URI, such as https://docs.sentry.io/product/security-policy-reporting/
* Ensure `DJANGO_SECURE_SSL_REDIRECT` is converted to bool for `SECURE_SSL_REDIRECT`
* Fix setting `SILENCED_SYSTEM_CHECKS` from the environment

# How to test
I pushed this branch to the dev server to test. With the proper environment settings, https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net/__heartbeat__ shows `{"status": "ok"}`, and no CSP issues on the dashboard or API docs.

The CSP changes are targeted at local development and the dev server. The `/api/v1/docs` URL looks like this on main:

main-csp-blocked-doc

With this PR, inline style is allowed, so there is a grey background applied:

pr-csp-allowed

For `/api/v1/docs/redoc`, the sidebar floats a little from the edge:

main-csp-blocked-redoc

With this PR, the sidebar hugs the edge:

PR-csp-allowed-redoc


rafeerahman (approved):
Not too familiar with CSP and this work, a secondary review might be helpful. 

Did not run into any errors with these changes and confirmed that the inline-styles applied.

groovecoder (commented):
changes LGTM too!

PR #4613, opened 2024-04-19
Avoid python-ipware 2.0.4

jwhitlock:
python-ipware 2.0.4 breaks the test [test_request_data_routable_ip_is_extracted](https://github.com/mozilla/fx-private-relay/blob/380a13732c56b323f0c7e9407bab4307f9c0f43f/privaterelay/tests/glean_tests.py#L66-L75). It works with 2.0.3.
 
 Filed upstream as https://github.com/un33k/django-ipware/issues/117.


rafeerahman (approved):
Thanks for the fix!

PR #4657, opened 2024-04-29
MPP-3799: Split API views by emails and privaterelay

jwhitlock:
For MPP-3799, I'm going to add schema definitions to make `drf_spectacular.W0002` warnings go away, which will expand the definitions of API serializers and view functions, and might change some of the implementation as well. Before I do that, I wanted to make the `emails` and `privaterelay` code mirror the `phones` code, without any deeper changes.

The changes in `api/urls.py` is a decent overview of the changes:

* `api.views.emails.py` has the email-specific views:
   - `DomainAddressViewSet`
   - `RelayAddressViewSet`
   - `first_forwarded_email` view function
* `api.views.privaterelay.py` has the general service views:
   - `FlagViewSet`
   - `ProfileViewSet`
   - `UserViewSet`
   - `report_webcompat_issue`
   - `runtime_data`
   - `terms_accepted_user`
   
I also moved the tests and serializers to separate files.

If this is too many changes, I can instead have view-specific PRs, that move the view and add the schema annotations. I'm open to other suggestions as well.

# How to test:

- `pytest` runs without issues


groovecoder (approved):
Good call. I didn't go line-by-line thru the code, but I looked at the number of lines moved between files and then investigated the diffs and it looks good.

PR #4658, opened 2024-04-29
Move coverage to pyproject

jwhitlock:
I'm getting errors when attempting to upload coverage for `privaterelay/locales/de/banner.ftl`. This PR should fix this, while also moving `coverage` configuration to `pyproject.toml` and adding coverage of test files.


jwhitlock:
I can go either way on coverage for test files.

The argument for excluding them is that it makes the coverage percentage higher that "deserved", since it included both production and test code.

The argument for including them is that it exposes issues like test code that doesn't run as expected.

So, I'm relying on reviewers to decide which way to go...

groovecoder (approved):

PR #4661, opened 2024-05-02
MPP-3799: Add schemas for privaterelay APIs

jwhitlock:
# How to test:

In the browsable API at http://127.0.0.1:8000/api/v1/docs/, views are grouped, and each view function has documentation and examples:

  - privaterelay
    - The CRUD views for `/api/v1/flags/`, `/api/v1/profiles/`, and `/api/v1/users/` are unchanged, except for a basic description.
    - `POST /api/v1/report_webcompat_issue/`
    - `GET /api/v1/runtime_data/`
    - `POST /api/v1/terms-accepted-user/` - Added documentation, but still can't test in the browsable API due to restrictions on `Authentication` field, [part of the spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields-10). We'd have to implement an authorization scheme instead, could be tricky.

When running `./manage.py check --deploy`, no views in `api/views/privaterelay.py` have a `drf_spectacular.W002` warning.

### DIFF
--- a/api/views/privaterelay.py
+++ b/api/views/privaterelay.py
@@ -11,7 +11,7 @@ from allauth.socialaccount.adapter import get_adapter as get_social_adapter
 from allauth.socialaccount.helpers import complete_social_login
 from allauth.socialaccount.models import SocialAccount
 from django_filters.rest_framework import FilterSet
-from drf_spectacular.utils import OpenApiResponse, extend_schema
+from drf_spectacular.utils import OpenApiExample, OpenApiResponse, extend_schema
 from rest_framework.authentication import get_authorization_header
 from rest_framework.decorators import (
     api_view,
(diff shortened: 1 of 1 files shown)


api/views/privaterelay.py
```diff
@@ -148,6 +258,17 @@ def runtime_data(request):
...
 
 @extend_schema(
+    tags=["privaterelay"],
+    parameters=[
+        OpenApiParameter(
+            name="Authorization",
+            required=True,
+            location="header",
+            examples=[OpenApiExample("bearer", "Bearer XXXX-ZZZZ")],
+            description="FXA Bearer Token. Can not be set in browsable API.",
```
groovecoder, line 268: praise: thanks for adding this to make it clear to api doc readers they can't use this endpoint in the docs.

groovecoder (approved):

PR #4684, opened 2024-05-07
fix MPP-3802: stop ignoring bandit security checks 

groovecoder:
This PR fixes #MPP-3802.

How to test:
1. `pytest`
2. `ruff check .`

### DIFF
--- a/api/authentication.py
+++ b/api/authentication.py
@@ -29,7 +29,7 @@ def get_cache_key(token):
 
 def introspect_token(token: str) -> dict[str, Any]:
     try:
-        fxa_resp = requests.post(INTROSPECT_TOKEN_URL, json={"token": token})
+        fxa_resp = requests.post(INTROSPECT_TOKEN_URL, json={"token": token}, timeout=1)
     except Exception as exc:
         logger.error(
             "Could not introspect token with FXA.",
--- a/api/authentication.py
+++ b/api/authentication.py
@@ -29,7 +29,11 @@ def get_cache_key(token):
 
 def introspect_token(token: str) -> dict[str, Any]:
     try:
-        fxa_resp = requests.post(INTROSPECT_TOKEN_URL, json={"token": token}, timeout=1)
+        fxa_resp = requests.post(
+            INTROSPECT_TOKEN_URL,
+            json={"token": token},
+            timeout=settings.FXA_REQUESTS_TIMEOUT_SECONDS,
+        )
     except Exception as exc:
         logger.error(
             "Could not introspect token with FXA.",
--- a/api/exceptions.py
+++ b/api/exceptions.py
@@ -58,13 +58,17 @@ class RelayAPIException(APIException):
         self, detail: _APIExceptionInput = None, code: str | None = None
     ) -> None:
         """Check that derived classes have set the required data."""
-        assert isinstance(self.default_code, str)
-        assert isinstance(self.status_code, int)
+        if not isinstance(self.default_code, str):
+            raise TypeError("default_code must be type str")
+        if not isinstance(self.status_code, int):
+            raise TypeError("self.status_code must be type int")
         if hasattr(self, "default_detail_template"):
             context = self.error_context()
-            assert context
+            if not context:
+                raise ValueError("error_context is required")
             self.default_detail = self.default_detail_template.format(**context)
-        assert isinstance(self.default_detail, str)
+        if not isinstance(self.default_detail, str):
+            raise TypeError("self.default_detail must be type str")
         super().__init__(detail, code)
 
     def error_context(self) -> ErrorContextType:
@@ -76,7 +80,8 @@ class RelayAPIException(APIException):
 
         # For RelayAPIException classes, this is the default_code and is a string
         error_code = self.get_codes()
-        assert isinstance(error_code, str)
+        if not isinstance(error_code, str):
+            raise TypeError("error_code must be type str")
 
         # Build the Fluent error ID
         ftl_id_sub = "api-error-"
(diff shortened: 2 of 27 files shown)


api/authentication.py
```diff
@@ -29,7 +29,7 @@ def get_cache_key(token):
 
 def introspect_token(token: str) -> dict[str, Any]:
     try:
-        fxa_resp = requests.post(INTROSPECT_TOKEN_URL, json={"token": token})
+        fxa_resp = requests.post(INTROSPECT_TOKEN_URL, json={"token": token}, timeout=1)
```
jwhitlock, line 32: I'm not sure if 1 second is the right value for FxA. Maybe move it to a setting that defaults to 1? That way we can adjust after deploying without a code change.

api/exceptions.py
```diff
@@ -58,13 +58,17 @@ def __init__(
         self, detail: _APIExceptionInput = None, code: str | None = None
     ) -> None:
         """Check that derived classes have set the required data."""
-        assert isinstance(self.default_code, str)
-        assert isinstance(self.status_code, int)
+        if not isinstance(self.default_code, str):
+            raise TypeError("default_code must be type str")
+        if not isinstance(self.status_code, int):
+            raise TypeError("self.status_code must be type int")
```
jwhitlock, line 64: Thank you for converting all of these `assert`s!

api/views/privaterelay.py
```diff
@@ -330,7 +330,7 @@ def terms_accepted_user(request):
     except SocialAccount.DoesNotExist:
         # User does not exist, create a new Relay user
         fxa_profile_resp = requests.get(
-            FXA_PROFILE_URL, headers={"Authorization": f"Bearer {token}"}
+            FXA_PROFILE_URL, headers={"Authorization": f"Bearer {token}"}, timeout=1
```
jwhitlock, line 333: This could be the same FxA setting

emails/utils.py
```diff
@@ -85,7 +85,7 @@ def get_trackers(level):
 
 def download_trackers(repo_url, category="Email"):
     # email tracker lists from shavar-prod-list as per agreed use under license:
-    resp = requests.get(repo_url)
+    resp = requests.get(repo_url, timeout=10)
```
jwhitlock, line 88: this one does seem safer, but @say-yawn may know if it is slow

privaterelay/apps.py
```diff
@@ -81,7 +81,9 @@ def ready(self) -> None:
 
         import privaterelay.signals
 
-        assert privaterelay.signals  # Suppress "imported but unused" warnings
+        assert (  # noqa S101
```
jwhitlock, line 84: Nit: alternate fix is to add `# noqa F401` to the import and drop the `assert`:

```
# Register signals by importing but not using the module
import privaterelay.signals  # noqa F401
```

privaterelay/plans.py
```diff
@@ -557,15 +557,18 @@ def _cached_country_language_mapping(
 
     mapping: PlanCountryLangMapping = {}
     for relay_country in relay_maps.get("by_country", []):
-        assert relay_country not in mapping
+        if relay_country in mapping:
+            raise ValueError("relay_country should not be in mapping.")
```
jwhitlock, line 561: Nit: these will be harder to debug, since we won't know _which_ `relay_county` was already in the mapping. However, we can expand the error message when we get to that point, instead of making every `ValueError` include the value.

privaterelay/settings.py
```diff
@@ -40,7 +40,7 @@
     # https://github.com/jazzband/django-silk
     import silk
 
-    assert silk  # Suppress "imported but unused" warning
+    assert silk  # Suppress "imported but unused" warning # noqa S101
```
jwhitlock, line 43: Nit: use `# noqa F401` on the import, drop the `assert`

pyproject.toml
```diff
@@ -97,11 +97,6 @@ testpaths = [
 ]
 
 [tool.ruff.lint]
-ignore = [
-    # TODO MPP-3802: Enable more bandit security checks
-    "S101",  # https://docs.astral.sh/ruff/rules/assert/
-    "S113",  # https://docs.astral.sh/ruff/rules/request-without-timeout/
-]
```
jwhitlock, line 104: 🎉

jwhitlock (changes requested):
Thanks @groovecoder! Only required change from me is creating a setting for the FxA profile timeout that can be configured via the environment.

emails/utils.py
say-yawn: I agree that 10 seconds is enough for us to get response back from GitHub for the tracker list. This request also only gets called during CircleCI jobs so it should be safe for the response time to take longer.

privaterelay/plans.py
groovecoder: This was the change that made me wonder: should I go back thru this whole PR and include the value in every `ValueError` message while I'm in here? I don't mind doing that now rather than later.

privaterelay/settings.py
groovecoder: Thanks. I like the `noqa F401` on the single line rather than `noqa S101` on the additional line.

api/authentication.py
groovecoder: Updated to use `settings.FXA_REQUESTS_TIMEOUT_SECONDS`

api/views/privaterelay.py
groovecoder: Updated to use `settings.FXA_REQUESTS_TIMEOUT_SECONDS`

groovecoder:
Made updates, but @jwhitlock if you think [I should go ahead and include values in `ValueError` messages](https://github.com/mozilla/fx-private-relay/pull/4684#discussion_r1595592686), let me know and I'll add that while I'm in this code- and head-space.

jwhitlock (approved):
Thanks @groovecoder!

No, I think we're OK without the detailed messages. If one of these fires in production, we'll add 1) a test to trigger the exception and 2) more information to exception message. 95% should never need it.

PR #4689, opened 2024-05-10
MPP-3815: Handle bad things in email processing

jwhitlock:
This PR makes several changes to `./manage.py process_emails_from_sqs`, now that it is being used to process the dead letter queue (DLQ) as well as the incoming emails:

* Messages are now processed in a subprocess, which allows updating the healthcheck file and also aborting processing when the processing seems stuck. A new setting `PROCESS_EMAIL_MAX_SECONDS_PER_MESSAGE`, default 120, controls how many seconds is too many. This will stop email tasks and the DLQ task from getting killed by Kubernetes after 120 seconds.
* The processor now captures exceptions and sends them to Sentry. This will stop the email tasks and the DLQ tasks from existing on an unhandled exception.
* A new setting `PROCESS_EMAIL_LOG_FAILED_MESSAGES` (default False) controls if notifications for failed messages are sent to logs. When True, the notification (as a JSON string) is truncated to `PROCESS_EMAIL_LOG_FAILED_MESSAGES_SIZE` bytes (default 20k), compressed with zlib at highest settings, and encoded as base64. This will hopefully make it easier to replicate failures in the development environment, making them easier to fix.

Additional changes:
* Remove `process_delayed_emails_from_sqs.py`, which is now unused and had no tests
* Add dependencies `mypy-boto3-sns`, `mypy-boto3-sqs`, and `mypy-boto3-s3` for more boto3 type hints
* Add type hints to all code in `process_emails_from_sqs.py` and its tests
* Add tests for above functionality

# How to test

If you have your local environment setup for email processing from a queue, enable your AWS credentials, run `./manage.py process_emails_from_sqs` and send some emails. You can also push the branch to Heroku and test it there.


jwhitlock:
I'm going to refactor to remove the `PROCESS_EMAIL_LOG_FAILED_MESSAGES` bit. Our [current policy](https://www.mozilla.org/en-US/privacy/subscription-services/) is:

> To send and forward your email messages from your masked email address(es) to your primary email address, Firefox Relay processes your email messages. We do not read or store the content of any of your messages. In the event that an email cannot be delivered to you, we will keep it on our servers and delete it after it has been delivered (in no event will we hold onto it for more than three days). If you use the feature to block promotional emails, the service will check email headers to determine whether the email should be blocked.

The `PROCESS_EMAIL_LOG_FAILED_MESSAGES` setting would store the email headers in our logs, which are retained for 90 days, which is more than 3 days.

To avoid any suspicion that this code was running in production, I'm going to rebase to remove this code, so that it doesn't appear in the commit history.

PR #4702, opened 2024-05-13
MPP-3815: Handle broken email processing

jwhitlock:
This PR makes several changes to `./manage.py process_emails_from_sqs`, now that it is being used to process the dead letter queue (DLQ) as well as the incoming emails. This PR replaces PR #4689. The changes to processing are:

* Messages are now processed in a subprocess, which allows updating the healthcheck file and also aborting processing when the processing seems stuck. A new setting `PROCESS_EMAIL_MAX_SECONDS_PER_MESSAGE`, default 120, controls how many seconds is too many. This will stop email tasks and the DLQ task from getting killed by Kubernetes after 120 seconds.
* The processor now captures exceptions and sends them to Sentry. This will stop the email tasks and the DLQ tasks from existing on an unhandled exception.

Additional changes:
* Remove `process_delayed_emails_from_sqs.py`, which is now unused and had no tests
* Add dependencies `mypy-boto3-sns`, `mypy-boto3-sqs`, and `mypy-boto3-s3` for more boto3 type hints
* Add type hints to all code in `process_emails_from_sqs.py` and its tests
* Add tests for above functionality

# How to test

If you have your local environment setup for email processing from a queue, enable your AWS credentials, run `./manage.py process_emails_from_sqs` and send some emails. You can also push the branch to Heroku and test it there.


emails/tests/mgmt_process_emails_from_sqs_tests.py
```diff
@@ -284,62 +361,124 @@ def test_no_body_deleted(mock_sqs_client, caplog, test_settings):
...
+    assert summary["failed_messages"] == 1
+    msg.delete.assert_not_called()
+    rec2 = caplog.records[1]
+    assert rec2.msg == "Message processed"
+    rec2_extra = log_extra(rec2)
+    assert rec2_extra["success"] is False
+    assert rec2_extra["error"] == "Timed out after 120.0 seconds."
+    assert rec2_extra["message_process_time_s"] >= 120.0
+    assert rec2_extra["subprocess_setup_time_s"] == 1.0
+    assert mock_process_pool_future._timeouts == [1.0] * 60
```
groovecoder, line 476: question (non-blocking): Again I don't understand this `_timeouts` property: why is it a list of 60 `1.0` values when the message timed out after 120 seconds? That seems like there should be 120 `1.0` values?

emails/management/commands/process_emails_from_sqs.py
```diff
@@ -398,68 +408,77 @@ def process_message(self, message):
...
+            results["subprocess_setup_time_s"] = round(setup_time, 3)
+
+            message_start_time = time.monotonic()
+            message_duration = 0.0
+            while message_duration < self.max_seconds_per_message:
+                self.write_healthcheck()
+                future.wait(1.0)
+                message_duration = time.monotonic() - message_start_time
+                if future.ready():
+                    break
```
groovecoder, line 453: question (non-blocking): Maybe related to how I don't understand `future._timeouts` property ... why do `wait(1.0)` rather than `wait(self.max_seconds_per_message)`? Is it to call `write_healthcheck` every second?

groovecoder (approved):
Nice code; looks good; spot-check works well; tests pass. Just a couple clarifying questions.

emails/management/commands/process_emails_from_sqs.py
jwhitlock: Yes, exactly. If we don't write the healthcheck, then k8s will kill the process, and we won't have a chance to delete the stalled email from S3. This is especially bad in the DLQ - we just keep processing the same stalled email until it expires in 3 days. With this, we'll at least process one every three minutes, about 500 per day.

emails/tests/mgmt_process_emails_from_sqs_tests.py
```diff
@@ -284,62 +361,124 @@ def test_no_body_deleted(mock_sqs_client, caplog, test_settings):
...
+    summary = summary_from_exit_log(caplog)
+    assert summary["total_messages"] == 1
+    msg.delete.assert_called()
+    rec2 = caplog.records[1]
+    assert rec2.msg == "Message processed"
+    rec2_extra = log_extra(rec2)
+    assert rec2_extra["success"] is True
+    assert rec2_extra["message_process_time_s"] < 120.0
+    assert rec2_extra["subprocess_setup_time_s"] == 1.0
+    assert mock_process_pool_future._timeouts == [1.0]
```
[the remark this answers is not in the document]
jwhitlock: Good question! I got the mock backwards. This line:

`mock_process_pool_future._is_stalled.side_effect = [False, False, True]`

should be:

`mock_process_pool_future._is_stalled.side_effect = [True, True, False]`

`_timeouts` is `[1.0]` because it was called one. When you call `future.wait(1.0)` on the mock future, it runs:

```python
            def call_wait(timeout: float) -> None:
                mocked_clocks(timeout)
                mock_future._timeouts.append(timeout)
                if not mock_future._is_stalled():
                    mock_future._ready = True
                    try:
                        ret = func(*args)
                    except BaseException as e:
                        if error_callback:
                            error_callback(e)
                    else:
                        if callback:
                            callback(ret)
```

So, `_timeouts` *always* has a value when `future.wait()` is called.

emails/tests/mgmt_process_emails_from_sqs_tests.py
jwhitlock: Good question! The loop includes a call to `time.monotonic`, which also increments the mocked clock, so each loops is 2 fake seconds. However, it makes more sense if `future.wait` does _not_ increment the clock. I'll make that change.

PR #4706, opened 2024-05-13
MPP-3815: Fix email processing tests

jwhitlock:
These are a couple of test improvements suggested by @groovecoder's review of PR #4702:

* Change `test_ses_slow` to actually test slow email processing
* Change the mocked `future.wait(1.0)` call to _not_ increment the timer, and instead delegate to the `time.monotonic` call in the same loop, so the loop takes the expected 1 simulated second instead of 2 simulated seconds.

### DIFF
--- a/emails/tests/mgmt_process_emails_from_sqs_tests.py
+++ b/emails/tests/mgmt_process_emails_from_sqs_tests.py
@@ -30,7 +30,7 @@ TEST_SNS_MESSAGE = EMAIL_SNS_BODIES["s3_stored"]
 
 
 @pytest.fixture(autouse=True)
-def mocked_clocks() -> Iterator[Callable[[float], float]]:
+def mocked_clocks() -> Iterator[None]:
     """
     Mock time functions, so tests run faster than real time.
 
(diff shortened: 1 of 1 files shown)


emails/tests/mgmt_process_emails_from_sqs_tests.py
```diff
@@ -30,7 +30,7 @@
 
 
 @pytest.fixture(autouse=True)
-def mocked_clocks() -> Iterator[Callable[[float], float]]:
+def mocked_clocks() -> Iterator[None]:
```
groovecoder, line 33: you somehow resisted the dark urge to rename this `mlocks` again.

emails/tests/mgmt_process_emails_from_sqs_tests.py
```diff
@@ -447,7 +446,8 @@ def test_ses_slow(
     assert rec2_extra["success"] is True
     assert rec2_extra["message_process_time_s"] < 120.0
     assert rec2_extra["subprocess_setup_time_s"] == 1.0
-    assert mock_process_pool_future._timeouts == [1.0]
+    # future.wait(1.0) was called 3 times, was ready after the 3rd call.
```
groovecoder, line 449: praise: nice comment. now I see how this correlates to the 3-item `_is_stalled.side_effect` list.

groovecoder (approved):

PR #4708, opened 2024-05-13
MPP-3641: Call probe-scraper when glean YAML updates

jwhitlock:
This action was copied from:

https://mozilla.github.io/glean/book/user/adding-glean-to-your-project/enable-data-ingestion.html#validate-and-publish-metrics

I couldn't help tuning it, but we can use the origin version if we want to try `probe-scraper` on all merges.


rafeerahman (approved):
LGTM, thanks!

PR #4709, opened 2024-05-16
for MPP-3817: prevent all Relay operations when user.is_active = False

groovecoder:
This PR fixes #MPP-3817.

How to test:
1. Run new `python manage.py deactivate_user` a number of ways (See `--help` for details)
2. After deactivating a user, make sure they cannot perform any of the following:
   * [ ] Existing sessions are signed out
   * [ ] Cannot sign back in
   * [ ] Cannot make any API requests
   * [ ] Cannot receive any emails
   * [ ] Cannot reply to any previous emails
   * [ ] Cannot receive any text messages
   * [ ] Cannot receive any forwarded phone calls

Note: some tests may be easiest on the dev server.


privaterelay/management/commands/deactivate_user.py
```diff
@@ -0,0 +1,56 @@
...
+            try:
+                user = SocialAccount.objects.get(uid=uid).user
+            except SocialAccount.DoesNotExist:
+                msg = "ERROR: Could not find user with that FXA UID."
+                self.stderr.write(msg)
+                return msg
+
+        user.is_active = False
+        user.save()
+        msg = "SUCCESS: deactivated user."
```
say-yawn, line 54: (suggestion, blocking) Descriptive success message with the identifier key and value for better logging.

say-yawn: (question) Should we consider a place to log why the account was deactivated as well?

privaterelay/management/commands/deactivate_user.py
```diff
@@ -0,0 +1,56 @@
...
+
+class Command(BaseCommand):
+    help = "Deactivates a user to effectively block all usage of Relay."
+
+    def add_arguments(self, parser: ArgumentParser) -> None:
+        parser.add_argument("--key", type=str, help="User API key")
+        parser.add_argument("--email", type=str, help="User email address")
+        parser.add_argument("--uid", type=str, help="User FXA UID")
+
+    def handle(self, *args: Any, **options: Any) -> str:
```
say-yawn, line 20: (suggestion) Unit test for this command.

say-yawn (changes requested):
Thanks @groovecoder for the extensive code to prevent Relay operations for users that are not active. Changes look good with some minor suggestions in review. Also some additional suggestions:
- (blocking) We are missing the `is_active` check on `_reply_allowed`
- (non-blocking) Can we refactor to add the `is_active` check in the `has_premium` check instead?

api/authentication.py
```diff
@@ -148,6 +148,12 @@ def authenticate(self, request):
             )
         user = sa.user
 
+        if not user.is_active:
+            raise PermissionDenied(
+                "Authenticated user does not have an active Relay account."
+                " Have they been deactivated?"
+            )
+
```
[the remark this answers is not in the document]
say-yawn: Resolving this conversation as I see it's following the convention used for authentication.

privaterelay/management/commands/deactivate_user.py
groovecoder: Updated message with identifier key and value.

privaterelay/management/commands/deactivate_user.py
groovecoder: Added tests.

say-yawn (changes requested):
Code changes look good. While manually running tests I noted:
- Instruction uses `python manage.py deactivate_user_by_token` it should be `python manage.py deactivate_user` instead
- (blocking) After deactivating the user and while signed-in I get the following error: 
```
NoReverseMatch at /accounts/fxa/login/callback/

Reverse for 'account_inactive' not found. 'account_inactive' is not a valid view function or pattern name.
```
Seems like allauth tries to send the user to the "account_inactive" page and fails because we have no view setup for that.

groovecoder:
>     * (blocking) After deactivating the user and while signed-in I get the following error:
> 
> 
> ```
> NoReverseMatch at /accounts/fxa/login/callback/
> 
> Reverse for 'account_inactive' not found. 'account_inactive' is not a valid view function or pattern name.
> ```
> 
> Seems like allauth tries to send the user to the "account_inactive" page and fails because we have no view setup for that.

Were you using http://127.0.0.1:8000 or http://localhost:3000?

jwhitlock:
I've only enabled some of the allauth endpoints:

https://github.com/mozilla/fx-private-relay/blob/b961ce074965dacbb0a2d1885adaf1dcc5932f04/privaterelay/urls.py#L30-L38

You'll need to enable more of them, or all of them. I disabled most because we have proper Django-side template to render them, or support features like a second social account. This way the user would get a 404 instead of a 500.

say-yawn:
> Were you using http://127.0.0.1:8000 or http://localhost:3000?
@groovecoder, I was using http://127.0.0.1:8000. Does it work for you? For me, adding `path("accounts/logout/", allauth_views.logout, name="account_inactive"),` on [privaterelay/urls.py](https://github.com/mozilla/fx-private-relay/blob/b961ce074965dacbb0a2d1885adaf1dcc5932f04/privaterelay/urls.py#L30-L38) did the trick too.

groovecoder:
In https://github.com/mozilla/fx-private-relay/pull/4709/commits/59a8b4f41ce1bb1b46b893ab4e0826e5bab5afac I added a new `accounts/account_inactive` nextjs page to display the error message in the Relay layout. I made our custom allauth adapter redirect deactivated users to that url.

PR #4718, opened 2024-05-20
MPP-3815: Reset the database at start of each email

jwhitlock:
This is a hotfix for [tag 2024.05.16](https://github.com/mozilla/fx-private-relay/releases/tag/2024.05.16). The email process runs for too long and the database connection can become unusable. Test for a usable database connection for each email processed.

I've tested this against my local postgres server. It appears to happen in Heroku as well as stage, but it is unclear if it is a timeout or other server action that results in a failed connection.

# How to test


jwhitlock:
I'm merging this. If it fixes the issue on dev, then I'll tag this branch as `2024.05.16.1`.

PR #4724, opened 2024-05-23
MPP-3487 - Fix tests, handle missing sqlcommenter in Python 3.12, 

jwhitlock:
This makes changes needed for tests to run under Python 3.12, with no deprecation warnings.

## Mock that doesn't do what was intended
For `test_grab_keyfile_checks_cert_url_origin`, the line 

```python
assert mock_urlopen.called_once_with(cert_url)
```
raises an exception:

```
AttributeError: 'called_once_with' is not a valid assertion. Use a spec for the mock if 'called_once_with' is meant to be an attribute.. Did you mean: 'assert_called_once_with'?
```

This is a new check for mock functions which are similar to the mock assertion functions. The developer probably intended:

```python
mock_urlopen.assert_called_once_with(cert_url)
```
However, this also fails. It would be a bit of work to make this pass, requiring mocking the return of a valid PEM. After a bit of reading, I re-wrote as:

```python
mock_urlopen.assert_not_called()
```
and called it after the `SuspiciousOperation` was raised. 

## LogRecord has new attribute `taskName`

This attribute was added to log the asyncio task name, https://github.com/python/cpython/issues/91513 . The `log_extra` helper needs to ignore it.

## `utcnow()` is finally going away

[datetime.utcnow()](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) is one of the gotcha methods. It returns a `datetime` in UTC time, but _without_ timezone information. Django will catch this if you try to store in a model, but some of our test code still uses it. It now raises a `DeprecationWarning`. I've replaced this code with `datetime.now(UTC)`, which returns a timezone-aware datetime.

## google-cloud-sqlcommenter doesn't like Python 3.12

This module prints a bunch of `DeprecationWarning`s: in Python 3.11

```
/private/tmp/workspace/.py311/lib/python3.11/site-packages/google/__init__.py:17: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').declare_namespace(__name__)
/private/tmp/workspace/.py311/lib/python3.11/site-packages/google/__init__.py:17: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google')`.
Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
  __import__('pkg_resources').declare_namespace(__name__)
```

In Python 3.12, `pkg_resources` is no longer installed automatically in virtualenvs, and these become an error:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/private/tmp/workspace/.py312/lib/python3.12/site-packages/google/__init__.py", line 17, in <module>
    __import__('pkg_resources').declare_namespace(__name__)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'pkg_resources'
```

You can install it via `setuptools`, but installing this unconditionally will cause issues when you are _not_ in a virtualenv, like our Docker image. I went with skipping it for Python 3.12, and not installing the middleware if the import fails (either not installed, or installed in virtualenv without `setuptools`). I opened https://github.com/google/sqlcommenter/issues/275 to track the issue.

### DIFF
--- a/emails/tests/sns_tests.py
+++ b/emails/tests/sns_tests.py
@@ -9,12 +9,7 @@ from ..sns import _grab_keyfile
 class GrabKeyfileTest(TestCase):
     @patch("emails.sns.urlopen")
     def test_grab_keyfile_checks_cert_url_origin(self, mock_urlopen):
-        cert_url = (
-            "https://sns.us-east-1.amazonaws.com/"
-            "SimpleNotificationService-7ff5318490ec183fbaddaa2a969abfda.pem"
-        )
-        assert mock_urlopen.called_once_with(cert_url)
-
+        cert_url = "https://attacker.com/cert.pem"
         with self.assertRaises(SuspiciousOperation):
-            cert_url = "https://attacker.com/cert.pem"
             _grab_keyfile(cert_url)
+        mock_urlopen.assert_not_called()
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -45,6 +45,13 @@ except ImportError:
     HAS_SILK = False
 
 try:
+    import google.cloud.sqlcommenter  # noqa: F401
+
+    HAS_SQLCOMMENTER = True
+except ImportError:
+    HAS_SQLCOMMENTER = False
+
+try:
     from privaterelay.glean.server_events import GLEAN_EVENT_MOZLOG_TYPE
 except ImportError:
     # File may not be generated yet. Will be checked at initialization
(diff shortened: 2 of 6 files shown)


say-yawn (approved):
Changes have very good documentation and commit message. LGTM 🚀

PR #4725, opened 2024-05-24
Remove Sentry logger for SocialAccount does not exist

say-yawn:
Fix MPP-3767 and add debugging log instead.

How to test:
No more events sent to [RELAY-GB](https://mozilla.sentry.io/issues/4716503385/?referrer=jira_integration).

### DIFF
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -140,15 +140,6 @@ def fxa_rp_events(request: HttpRequest) -> HttpResponse:
     try:
         social_account = _get_account_from_jwt(authentic_jwt)
     except SocialAccount.DoesNotExist:
-        if settings.DEBUG:
-            info_logger.info(
-                "fxa_profile_update",
-                extra={
-                    "success": False,
-                    "reason": "SocialAccount does not exist.",
-                    "jwt": authentic_jwt,
-                },
-            )
         # Don't error, or FXA will retry
         return HttpResponse("202 Accepted", status=202)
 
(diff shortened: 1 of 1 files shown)


privaterelay/views.py
```diff
@@ -139,9 +139,17 @@ def fxa_rp_events(request: HttpRequest) -> HttpResponse:
...
-        # capture an exception in sentry, but don't error, or FXA will retry
-        sentry_sdk.capture_exception(e)
+    except SocialAccount.DoesNotExist:
+        if settings.DEBUG:
+            info_logger.info(
+                "fxa_profile_update",
+                extra={
+                    "success": False,
+                    "reason": "SocialAccount does not exist.",
+                    "jwt": authentic_jwt,
```
jwhitlock, line 149: I was going to ask what a `authentic_jwt` looks like, and if it has user info, but it is only active in `DEBUG`, so it is unlikely to be run in production unless we've really messed up. 

But also, will we ever be in a situation with `DEBUG=True` and FxA is sending relying party events? The logging seems like dead code, could be removed with no impact.

say-yawn: `authentic_jwt` is a dictionary objected called FxAEvent from https://github.com/mozilla/fx-private-relay/blob/b961ce074965dacbb0a2d1885adaf1dcc5932f04/privaterelay/views.py#L188-L202

The user's FxA UID is also stored there, which is parsed in https://github.com/mozilla/fx-private-relay/blob/b961ce074965dacbb0a2d1885adaf1dcc5932f04/privaterelay/views.py#L269-L271

>but it is only active in DEBUG, so it is unlikely to be run in production unless we've really messed up.

Agreed, I was following the logging behavior in the later part of the code https://github.com/mozilla/fx-private-relay/blob/b961ce074965dacbb0a2d1885adaf1dcc5932f04/privaterelay/views.py#L147-L156

Happy to remove the logs all together since the likelihood of this being used is low.

say-yawn: Updated with logs removed in 5cf91b5

jwhitlock (approved):

PR #4744, opened 2024-05-28
Use the events logger for warning or error levels

jwhitlock:
The GCP logging ingester treats all logs from `eventsinfo` as `INFO`-level, even if logged at `WARNING` or `ERROR`. Use `events` so that these issues stand out in the logs.


say-yawn (approved):
Nice catch! Manually using different loggers is not ideal but it worked at the time when we were trying to get certain logs to stdout versus stderror on GCP logs. I created https://github.com/mozilla/fx-private-relay/issues/4745 to take a look at this again.

PR #4746, opened 2024-05-30
For `ResponseMetrics` middleware, improve `view` tag and add tests

jwhitlock:
The `ResponseMetrics` middleware emits the timing metric `fx.private.relay.response`. This gives us the application's view of the response time, bucketed by the hour. We use the `view` tag to segment by views, to determine trends over time and slow periods.

This PR improve the tag `view` in a few cases:

* `/api/v1/runtime_data`, other API view functions. - The tag was `api.views.privaterelay.view`. This is now `api.views.privaterelay.runtime_data`. Other API view functions also had '.view' for the last component, and now use their function name.
* `/__heartbeat__`, other Dockerflow views - The tag was `<unknown_view>`, and is now `dockerflow.django.views.heartbeat`. Similar changes are made for `/__version__` and `/__lbheartbeat__`.
* Frontend files - The tag was `<unknown_view>`, and is now `<static_file>`. 

This PR also includes tests for `ResponseMetrics`.

## How to test

### DIFF
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -60,6 +60,7 @@ class ResponseMetrics:
 
     def __init__(self, get_response: Callable[[HttpRequest], HttpResponse]) -> None:
         self.get_response = get_response
+        self.middleware = RelayStaticFilesMiddleware()
 
     def __call__(self, request: HttpRequest) -> HttpResponse:
         if not settings.STATSD_ENABLED:
@@ -89,6 +90,8 @@ class ResponseMetrics:
             return f"{view.__module__}.{view.__name__}"
         if match := self.re_dockerflow.match(request.path_info):
             return f"dockerflow.django.views.{match[1]}"
+        if self.middleware.is_staticfile(request.path_info):
+            return "<static_file>"
         return "<unknown_view>"
 
 
@@ -130,3 +133,16 @@ class RelayStaticFilesMiddleware(WhiteNoiseMiddleware):
             return True
         else:
             return super().immutable_file_test(path, url)
+
+    def is_staticfile(self, path_info: str) -> bool:
+        """
+        Returns True if this file is served by the middleware.
+
+        This uses the logic from whitenoise.middleware.WhiteNoiseMiddleware.__call__:
+        https://github.com/evansd/whitenoise/blob/220a98894495d407424e80d85d49227a5cf97e1b/src/whitenoise/middleware.py#L117-L124
+        """
+        if self.autorefresh:
+            static_file = self.find_file(path_info)
+        else:
+            static_file = self.files.get(path_info)
+        return static_file is not None
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -1,3 +1,4 @@
+import re
 import time
 from collections.abc import Callable
 from datetime import UTC, datetime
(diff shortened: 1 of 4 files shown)


privaterelay/middleware.py
```diff
@@ -120,3 +133,16 @@ def immutable_file_test(self, path, url):
...
             return super().immutable_file_test(path, url)
+
+    def is_staticfile(self, path_info: str) -> bool:
+        """
+        Returns True if this file is served by the middleware.
+
+        This uses the logic from whitenoise.middleware.WhiteNoiseMiddleware.__call__:
+        https://github.com/evansd/whitenoise/blob/220a98894495d407424e80d85d49227a5cf97e1b/src/whitenoise/middleware.py#L117-L124
+        """
+        if self.autorefresh:
```
say-yawn, line 144: (suggestion, non-blocking): Since we are using data that can come from user input I recommend using `shlex.quote`.

privaterelay/settings.py
```diff
@@ -349,16 +349,7 @@
...
-def _get_initial_middleware() -> list[str]:
-    if STATSD_ENABLED:
-        return [
-            "privaterelay.middleware.ResponseMetrics",
-        ]
-    return []
-
-
-MIDDLEWARE = _get_initial_middleware()
+MIDDLEWARE = ["privaterelay.middleware.ResponseMetrics"]
```
say-yawn, line 352: (suggestion, non-blocking) I agree with more simple logic on initial list of middlewares but not sure if the `ResponseMetrics` should be used in all environment. `STATSD_ENABLED` is `False` in local:
https://github.com/mozilla/fx-private-relay/blob/2d8973e9483bbaa876d46befda23bd9d8dbc1dde/.env-dist#L14-L15

```suggestion
if RELAY_CHANNEL == "local":
    MIDDLEWARE = []
else:
    MIDDLEWARE = ["privaterelay.middleware.ResponseMetrics"]
```

say-yawn (approved):
Thanks @jwhitlock for making changes that help us get more detailed data on responses. I've made some suggestions but are not required.

privaterelay/middleware.py
```diff
@@ -51,24 +54,22 @@ def __call__(self, request):
...
 
-class ResponseMetrics:
-    def __init__(self, get_response):
+    def __init__(self, get_response: Callable[[HttpRequest], HttpResponse]) -> None:
         self.get_response = get_response
+        self.middleware = RelayStaticFilesMiddleware()
+
+    def __call__(self, request: HttpRequest) -> HttpResponse:
+        if not settings.STATSD_ENABLED:
+            return self.get_response(request)
```
jwhitlock, line 67: This is where the middleware is disabled at run-time, if `STATSD_ENABLED` is False.

privaterelay/middleware.py
jwhitlock: No, that would change the logic from `.__call__` and break the function. See https://github.com/evansd/whitenoise/blob/220a98894495d407424e80d85d49227a5cf97e1b/src/whitenoise/middleware.py#L117-L124.

privaterelay/settings.py
jwhitlock: I retained the spirit of the old logic by making the `ResponseMetrics` middleware do nothing if `STATSD_ENABLED` is False. This reminds me, I have a TODO to enable `STATSD_DEBUG` locally..

PR #4763, opened 2024-06-08
MPP-3827: Move Profile and RegisteredSubdomain models to privaterelay

jwhitlock:
This PR moves Django models and related code from `emails` to `privaterelay`. 

It does _not_ update the database. The moved models have the [`Meta.db_table` property](https://docs.djangoproject.com/en/4.2/topics/db/models/#meta-options) set to ensure the table names do not change (still `emails_profile` and `emails_registeredsubdomain`. The migrations use the [SeparateDatabaseAndState](https://docs.djangoproject.com/en/4.2/ref/migration-operations/#separatedatabaseandstate) operation to avoid SQL.  Deployment will work, and that the terraform SQL will continue to work. Thanks to https://davit.hashnode.dev/django-move-model for explaining this technique.

It may be possible to smoothly rename the tables as well. PostgreSQL [views](https://www.postgresql.org/docs/current/sql-createview.html) are updatable, so we could create a `emails_profile` view that allows existing code to continue to query, insert, and update to `emails_profile` while the table is named `privaterelay_profile`. There are two issues. One is the SQLite does not have updatable views, so I'd need to craft triggers to make tests pass. The second is that if a test needs to run in a transaction, `TRUNCATE` is used and further complicates things. Both could be handled. SQLite triggers are possible to write, just hard. I eliminated all the transaction tests, and an upstream contribution to `pytest_django` is possible. We could also just say "dev and migration tests are broken for a bit". In any case, I'd like to avoid any table renames in this PR.

* Models: 
  - Moved `Profile`, since it contains a mix of email, phone, and account information for a Relay user
  - Moved `RegisteredSubdomain`, since it is concerned with `Profile.subdomain`, with a validator that uses it. I could also argue that this should remain in `emails`, since subdomains are an email feature.
  - Did not move `AbuseMetrics`, since it only has email usage and doesn't have a relationship to the `Profile`. I could also argue that this should move to `privaterelay`, since it could gain phone abuse metrics in the future.
* Validators:
  - Moved `valid_available_subdomain` to `privaterelay`, to follow `RegisteredSubdomain`
* Signals:
  - Moved all signals to `privaterelay`, to follow `Profile`
* Exceptions:
  - Moved `CannotMakeSubdomainException` to `privaterelay`, to follow `RegisteredSubdomain`
* Tests, Admin - also moved to stay with related code  

# How to test

* Get a hot beverage. 
* Review the code.
* Think about these questions:
   - Should we instead break up `Profile` into `RelayProfile`, `EmailsProfile`, and `PhoneProfile`? If so, are those separate tables or combined using [abstract base classes](https://docs.djangoproject.com/en/4.2/topics/db/models/#abstract-base-classes)?
   - Which app should have `RegisteredSubdomain`?
   - Which app should have `AbuseMetrics`?

CircleCI does all of the validation that this works. The migrations tests are especially important. If you want to verify they run no SQL:

```sh
./manage.py sqlmigrate privaterelay 0010
./manage.py sqlmigrate emails 0062
```
both return:
```sql
BEGIN;
--
-- Custom state/database change combination
--
-- (no-op)
COMMIT;
```

# Previous PRs

The initial spike was broken up into smaller, focused PRs. Thanks to @groovecoder and @say-yawn for the reviews and making this a better change.

* [x] #4798 - Prefer `user.profile` to other methods of getting the profile
* [x] #4799 - Move exceptions and validators to own files
* [x] #4801 - Use APIClient in IQ tests, and test without DB transactions
* [x] #4802 - Move test user creation to test utils file
* [x] #4822 - Optimize bad word checks, testing if it makes a significant difference


emails/migrations/0062_move_profile_and_registered_subdomain_models.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.13 on 2024-06-07 22:08
+# Move Profile and RegisteredSubdomain to privaterelay app
+# Other half is
+# privaterelay/migrations/0010_move_profile_and_registered_subdomain_models.py
+# See https://davit.hashnode.dev/django-move-model
```
say-yawn, line 5: (praise) Glad to see this documentation!
(suggestion, non-blocking) Add that it uses Solution # 2 of the referenced documentation that uses the model's Meta option [db_table](https://docs.djangoproject.com/en/3.0/ref/models/options/#db-table).

privaterelay/models.py
```diff
@@ -0,0 +1,561 @@
...
+    onboarding_free_state = models.PositiveIntegerField(default=0)
+    auto_block_spam = models.BooleanField(default=False)
+    forwarded_first_reply = models.BooleanField(default=False)
+    # Empty string means the profile was created through relying party flow
+    created_by = models.CharField(blank=True, null=True, max_length=63)
+    sent_welcome_email = models.BooleanField(default=False)
+    last_engagement = models.DateTimeField(blank=True, null=True, db_index=True)
+
+    class Meta:
+        db_table = "emails_profile"
```
say-yawn, line 97: (suggestion, non-blocking) Add comment here that also points to migrations used and documentation on how the model was moved from `emails/models.py`

privaterelay/models.py
```diff
@@ -0,0 +1,561 @@
...
+
+class RegisteredSubdomain(models.Model):
+    subdomain_hash = models.CharField(max_length=64, db_index=True, unique=True)
+    registered_at = models.DateTimeField(auto_now_add=True)
+
+    def __str__(self):
+        return self.subdomain_hash
+
+    class Meta:
+        db_table = "emails_registeredsubdomain"
```
say-yawn, line 557: (suggestion, non-blocking) Add comment here that also points to migrations used and documentation on how the model was moved from `emails/models.py`

emails/migrations/0062_move_profile_and_registered_subdomain_models.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ("emails", "0061_relayaddress_idx_ra_created_by_addon"),
+    ]
+
+    state_operations = [
+        migrations.RemoveField(model_name="profile", name="user"),
+        migrations.DeleteModel(name="RegisteredSubdomain"),
+        migrations.DeleteModel(name="Profile"),
+    ]
```
say-yawn, line 20: (praise) No `migrations.AlterField` operation needed since no models directly uses `Profile` and instead uses `User` 🤯 
(question, blocking) Why is the [RemoveField](https://docs.djangoproject.com/en/5.0/ref/migration-operations/#removefield) operation there? Wouldn't this result in dropping the `user` column in the `emails_profile` table?

privaterelay/settings.py
```diff
@@ -445,10 +445,12 @@
...
-MAX_REPLIES_PER_DAY = config("MAX_REPLIES_PER_DAY", 100, cast=int)
-MAX_FORWARDED_PER_DAY = config("MAX_FORWARDED_PER_DAY", 1000, cast=int)
-MAX_FORWARDED_EMAIL_SIZE_PER_DAY = config(
+MAX_ADDRESS_CREATION_PER_DAY: int = config(
+    "MAX_ADDRESS_CREATION_PER_DAY", 100, cast=int
+)
+MAX_REPLIES_PER_DAY: int = config("MAX_REPLIES_PER_DAY", 100, cast=int)
+MAX_FORWARDED_PER_DAY: int = config("MAX_FORWARDED_PER_DAY", 1000, cast=int)
+MAX_FORWARDED_EMAIL_SIZE_PER_DAY: int = config(
     "MAX_FORWARDED_EMAIL_SIZE_PER_DAY", 1_000_000_000, cast=int
```
say-yawn, line 454: (praise) Added types!

privaterelay/views.py
```diff
@@ -25,14 +25,15 @@
...
 
-# from silk.profiling.profiler import silk_profile
-from emails.exceptions import CannotMakeSubdomainException
 from emails.models import DomainAddress, RelayAddress
 from emails.utils import incr_if_enabled
-from emails.validators import valid_available_subdomain
 
 from .apps import PrivateRelayConfig
+
+# from silk.profiling.profiler import silk_profile
```
say-yawn, line 33: (suggestion, non-blocking) either delete this line or use local environment check and debug mode to enable/disable this.

say-yawn (changes requested):
In this first pass I:
✅ models `Profile` and `RegisteredSubdomain` were moved from `emails.models` to `privaterelay.models` with appropriate migrations are created by using the model's [db_table](https://docs.djangoproject.com/en/3.0/ref/models/options/#db-table) Meta option

In the second pass I will check the the following moved codes:
⚪ valid_available_subdomain
⚪ all signals
⚪ CannotMakeSubdomainException
⚪ associated tests and admin code
and that the migration works as expected (no table dropped or created)

emails/apps.py
```diff
@@ -60,9 +60,6 @@ def _load_terms(self, filename: str) -> list[str]:
                 terms.append(word.strip())
         return terms
 
-    def ready(self):
-        import emails.signals  # noqa: F401 (imported but unused warning)
-
```
say-yawn, line 65: (question, non-blocking) Any idea why we imported the signals during the ready? Wondering if we need to do this in the `privaterelay` app since the signals have been moved?

say-yawn (commented):
✅ [Models diff](https://www.diffchecker.com/JHuDWMns/)
✅ `valid_available_subdomain` [validator diff](https://www.diffchecker.com/rYr9632K/)
✅ [Signals diff](https://www.diffchecker.com/8mfz4fB1/)
✅ `CannotMakeSubdomainException` [diff](https://www.diffchecker.com/WeFPTo3s/)
✅ Admin code

(question, non-blocking) I noticed that on the models and signals move from `emails` to `privaterelay` app resulted in the method/functions being stripped from type annotations. Was this intentional and can you elaborate why?

jwhitlock (commented):
Thanks @say-yawn! I've rebased and added some commits to address your feedback so far.

emails/migrations/0062_move_profile_and_registered_subdomain_models.py
say-yawn: Thanks @jwhitlock for this write up. Here's my understood summary: 
because we separated the model state and the database using  `state_operations(migrations.SeparateDatabaseAndState(state_operations=state_operations))` the `RemoveField` will only change the model state and will not result in DB changes (which I can confirm when running the migration locally).

jwhitlock: yep, you've got it!

say-yawn (approved):
✅ [Models tests diff](https://www.diffchecker.com/NG4jcHSv/)
✅ [Signals tests diff](https://www.diffchecker.com/iGE1voh5/)
✅ [Validators tests diff](https://www.diffchecker.com/H1oM19kp/)
✅ Checked sql for the migrations, ran migration, and confirmed in the tables that the values were not dropped and that the table names stayed the same.

One minor suggestion, not required for re-review.

jwhitlock:
Rebased to pick up changes from PR #4885.

PR #4798, opened 2024-06-20
MPP-4763 - Prefer `user.profile` to `Profile.objects.get(user=user)`

jwhitlock:
The pattern `profile = Profile.objects.get(user=user)` was useful when `user` had a `ForiegnKey` relationship. Since #2749, it has been unnecessary. Switching these to `user.profile` has some benefits:

* In most cases, it reduces the lines of code
* It removes a possible mismatch between `user.profile` and the stand-alone `Profile` object, which could introduce bugs
* In some cases, importing `Profile` is no longer needed, which will reduce the code changes when moving `Profile` to `privaterelay`.

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -656,7 +656,6 @@ class RelayAddress(models.Model):
         )
         deleted_address.save()
         profile = self.user.profile
-        profile.refresh_from_db()
         profile.address_last_deleted = datetime.now(UTC)
         profile.num_address_deleted += 1
         profile.num_email_forwarded_in_deleted_address += self.num_forwarded
--- a/emails/models.py
+++ b/emails/models.py
@@ -822,10 +822,6 @@ class DomainAddress(models.Model):
             update_fields=update_fields,
         )
 
-    @property
-    def user_profile(self):
-        return Profile.objects.get(user=self.user)
-
     @staticmethod
     def make_domain_address(
         user: User, address: str | None = None, made_via_email: bool = False
@@ -852,7 +848,7 @@ class DomainAddress(models.Model):
         # TODO: create hard bounce receipt rule in AWS for the address
         deleted_address = DeletedAddress.objects.create(
             address_hash=address_hash(
-                self.address, self.user_profile.subdomain, self.domain_value
+                self.address, self.user.profile.subdomain, self.domain_value
             ),
             num_forwarded=self.num_forwarded,
             num_blocked=self.num_blocked,
--- a/emails/models.py
+++ b/emails/models.py
@@ -8,7 +8,7 @@ from collections import namedtuple
 from collections.abc import Iterable
 from datetime import UTC, datetime, timedelta
 from hashlib import sha256
-from typing import Literal, cast
+from typing import Any, Literal, cast
 
 from django.conf import settings
 from django.contrib.auth.models import User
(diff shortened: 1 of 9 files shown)


emails/models.py
```diff
@@ -655,7 +655,8 @@ def delete(self, *args, **kwargs):
             num_spam=self.num_spam,
         )
         deleted_address.save()
-        profile = Profile.objects.get(user=self.user)
+        profile = self.user.profile
+        profile.refresh_from_db()
```
say-yawn, line 659: (question, non-blocking) Why was the `profile.refresh_from_db()` needed?

emails/models.py
```diff
@@ -851,17 +848,15 @@ def delete(self, *args, **kwargs):
...
             num_forwarded=self.num_forwarded,
             num_blocked=self.num_blocked,
             num_replied=self.num_replied,
             num_spam=self.num_spam,
         )
         deleted_address.save()
-        # self.user_profile is a property and should not be used to
-        # update values on the user's profile
-        profile = Profile.objects.get(user=self.user)
+        profile = self.user.profile
```
say-yawn, line 859: (praise) 🎊

privaterelay/tests/mgmt_sync_phone_related_dates_on_profile_tests.py
```diff
@@ -103,16 +102,16 @@ def create_free_phones_flag_for_user(user: User) -> None:
...
-    mocked_dates, patch_datetime_now, db
-):
+    mocked_dates: Mock, patch_datetime_now: Mock, db: None
+) -> None:
     expected_now = patch_datetime_now
     mocked_dates.return_value = (None, None, None)
     account: SocialAccount = baker.make(SocialAccount, provider="fxa")
-    profile = Profile.objects.get(user=account.user)
-    create_free_phones_flag_for_user(profile.user)
+    create_free_phones_flag_for_user(account.user)
```
say-yawn, line 110:
```suggestion
    user = account.user_set.get()
    create_free_phones_flag_for_user(account.user)
```
(blocking) since `user` is still a foreign key of `SocialAccount`, see [code](https://github.com/pennersr/django-allauth/blob/4746fb58bc3c1621fddecf1119887703ac6f04d7/allauth/socialaccount/models.py#L92-L93)

privaterelay/tests/mgmt_sync_phone_related_dates_on_profile_tests.py
```diff
@@ -103,16 +102,16 @@ def create_free_phones_flag_for_user(user: User) -> None:
...
 
     num_profiles_updated = sync_phone_related_dates_on_profile("free")
 
+    profile = account.user.profile
```
say-yawn, line 114:
```suggestion
    profile = user.profile
```
If above change to use `user = account.user_set.get()` is accepted.

say-yawn (changes requested):
Most of the changes look good and relieved to see we can directly use `profile` instead doing the `profile_set.get()` thanks to the changes from #2749 that makes `Profile` and `User` one-to-one relationship.

emails/models.py
jwhitlock: I removed it, tests were OK, so let's remove it.

A `refresh_from_db()` would be a good idea if we use `F` expressions to avoid race conditions here:

https://docs.djangoproject.com/en/4.2/ref/models/expressions/#avoiding-race-conditions-using-f

privaterelay/tests/mgmt_sync_phone_related_dates_on_profile_tests.py
jwhitlock: It's a to-one relation, so there is no `user_set`, just `user`

emails/models.py
say-yawn: Whoa! I didn't know about `F` expressions. Good to know going forward we can use this especially for incrementing!

say-yawn (approved):
[Changes look good. Learned about F expression](https://docs.djangoproject.com/en/4.2/ref/models/expressions/#avoiding-race-conditions-using-f)!

PR #4799, opened 2024-06-20
MPP-3827: Move non-model code out of `emails/models.py`

jwhitlock:
This PR moves code around, in preparation for moving `Profile` (and other models) from the `emails` app to the `privaterelay` app.

* `emails/exceptions.py` - Defines the exceptions raised in the `emails` app
* `emails/validators.py` - Defines the validator functions used in the `emails` app.
* `emails/models.py`
   - The remaining functions have moved to the top of the file, and have type hints.
   - `DomainAddress.make_domain_address` used to take a `Profile`, and now takes a `User`.

This also involved moving tests around, and some test changes for the new interfaces.

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -675,7 +675,7 @@ class RelayAddress(models.Model):
         if self._state.adding:
             with transaction.atomic():
                 locked_profile = Profile.objects.select_for_update().get(user=self.user)
-                check_user_can_make_another_address(locked_profile)
+                check_user_can_make_another_address(locked_profile.user)
                 while True:
                     address_is_allowed = not is_blocklisted(self.address)
                     address_is_valid = valid_address(self.address, self.domain_value)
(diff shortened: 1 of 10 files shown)


emails/models.py
```diff
@@ -1014,9 +827,9 @@ def user_profile(self):
 
     @staticmethod
     def make_domain_address(
-        user_profile: Profile, address: str | None = None, made_via_email: bool = False
+        user: User, address: str | None = None, made_via_email: bool = False
```
groovecoder, line 830: question (non-blocking): This is the only change I noticed that seems to do more than simply move code from one module to another? This changes the method signature of `make_domain_address`. Though it looks like all the appropriate calling code in `views.py` and `models_tests.py` are likewise updated correctly too. What was the motivation to change this parameter?

groovecoder (approved):

emails/models.py
jwhitlock: To be honest, a big reason is it looks better to me. A `DomainAddress` needs a `User`, so why create it with a `Profile`? One answer is `check_user_can_make_domain_address` takes a `Profile`, but then why wasn't it named `check_profile_can_make_domain_address`? Passing a `User` makes the interface make more sense to me. The fact that `user.profile` is called by `check_user_can_make_domain_address` now an implementation detail, not the interface.

A more practical concern is that when `Profile` moves to `privaterelay`, tricks will be needed to avoid circular imports. With the old version, this would be something like:

```python
class DomainAddrress(models.Model):
   ...

    @staticmethod
    def make_domain_address(
        user_profile: "Profile", address: str | None = None, made_via_email: bool = False
    ) -> DomainAddress:
        from privaterelay.models import Profile
        if not isinstance(user_profile, Profile):
            raise TypeError("user_profile is not a Profile")
        check_user_can_make_domain_address(user_profile)
        ...
```

With this change, we can skip the manual type checking:

```python

...

class DomainAddrress(models.Model):
   ...

    @staticmethod
    def make_domain_address(
        user: User, address: str | None = None, made_via_email: bool = False
    ) -> DomainAddress:
        check_user_can_make_domain_address(user)
        ...
```

PR #4796, opened 2024-06-18
MPP-2822: add gtag to Layout and update gaEvent hook to ping both GA endpoints

groovecoder:
# Add GA4
## Screenshot

image

## How to test

1. Checkout this branch
2. In `frontend/` run `npm run dev`
3. Go to http://localhost:3000/
   * [ ] In the network inspector, verify a request to www.googletagmanager.com is sent
5. Go to https://analytics.google.com/analytics/web/#/p314403930/realtime/overview
   * [ ] Verify you see your traffic in the realtime report
6. In `frontend/` run `npm run build`
7. In app root, run `python manage.py collectstatic`
8. In app root, run `python manage.py runserver`
9. Go to http://127.0.0.1:8000/
   * [ ] In the network inspector, verify a request to www.googletagmanager.com is sent
10. Go to https://analytics.google.com/analytics/web/#/p314403930/realtime/overview
    * [ ] Verify you see your traffic in the realtime report

# Checklist (Definition of Done)
- [x] Product Owner accepted the User Story (demo of functionality completed) or waived the privilege.
- [x] ~Customer Experience team has seen or waived a demo of functionality.~
- [x] All acceptance criteria are met.
- [x] Jira ticket has been updated (if needed) to match changes made during the development process.
- [x] I've added or updated relevant docs in the docs/ directory
- [x] Jira ticket has been updated (if needed) with suggestions for QA when this PR is deployed to stage.
- [x] ~All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).~
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).
- [x] ~l10n changes have been submitted to the l10n repository, if any.~

PR #4800, opened 2024-06-20
Update to django-csp 4.0

jwhitlock:
This release is in beta, but we can be an early adopter to help work out the bugs.

## How to test

* Install the requirements
* Run `./manage.py check`, no issues reported
* Push to Heroku Dev, go to API docs, look for CSP errors in console.

### DIFF
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -186,7 +186,6 @@ else:
 
 CONTENT_SECURITY_POLICY: CONTENT_SECURITY_POLICY_T = {
     "DIRECTIVES": {
-        "report-uri": config("CSP_REPORT_URI", ""),
         "default-src": [SELF],
         "connect-src": [
             SELF,
@@ -219,6 +218,8 @@ if _CSP_STYLE_INLINE:
     CONTENT_SECURITY_POLICY["DIRECTIVES"]["script-src"].append(UNSAFE_INLINE)
 if _API_DOCS_CSP_WORKER_SRC:
     CONTENT_SECURITY_POLICY["DIRECTIVES"]["worker-src"] = _API_DOCS_CSP_WORKER_SRC
+if _CSP_REPORT_URI := config("CSP_REPORT_URI", ""):
+    CONTENT_SECURITY_POLICY["DIRECTIVES"]["report-uri"] = _CSP_REPORT_URI
 
 REFERRER_POLICY = "strict-origin-when-cross-origin"
 
--- a/privaterelay/settings.py
+++ b/privaterelay/settings.py
@@ -26,6 +26,7 @@ import dj_database_url
 import django_stubs_ext
 import markus
 import sentry_sdk
+from csp.constants import NONE, SELF, UNSAFE_INLINE
 from decouple import Choices, Csv, config
 from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.logging import ignore_logger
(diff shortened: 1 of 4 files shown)


privaterelay/types.py
```diff
@@ -1,5 +1,59 @@
...
+        "require-sri-for": list[str],
+        # Trusted Types Directives
+        "require-trusted-types-for": list[str],
+        "trusted-types": list[str],
+        # Other Directives
+        "webrtc": list[str],
+        "worker-src": list[str],
+        # Directives Defined in Other Documents
+        "upgrade-insecure-requests": list[str],
+        "block-all-mixed-content": list[str],  # Deprecated.
```
robhudson, line 50: These last 2 are actually booleans.

privaterelay/settings.py
```diff
@@ -183,34 +184,42 @@
...
+
+CONTENT_SECURITY_POLICY: CONTENT_SECURITY_POLICY_T = {
+    "DIRECTIVES": {
+        "default-src": [SELF],
+        "connect-src": [
+            SELF,
+            "https://www.google-analytics.com/",
+            "https://location.services.mozilla.com",
+            "https://api.stripe.com",
+            BASKET_ORIGIN,
```
groovecoder, line 196: question (non-blocking): Why do we include this constant `BASKET_ORIGIN` inline here, but then use `.extend()` to add the other `_ACCOUNT_CONNECT_SRC` constant?

requirements.txt
```diff
@@ -5,7 +5,7 @@ Django==4.2.13
 dj-database-url==2.2.0
 django-allauth[socialaccount]==0.63.3
 django-cors-headers==4.3.1
-django-csp==3.8
+django-csp==4.0b1
```
groovecoder, line 8: question (non-blocking): We don't seem to use any other beta versions of any libraries. When this is out of beta, I assume we'll be able to update this from `4.0b1` to just `4.0`?

jwhitlock: I assume so too, but that's a Dependabot question

privaterelay/settings.py
jwhitlock: `BASKET_ORIGIN` is a single string in every environment, while the others are lists that change content and size (for example, 0 length in one environment, 2 in others).

I tried to make the changes small, since there is no way to validate a CSP outside of running it on a server. but we could certainly re-arrange the logic. The setup code above `CONTENT_SECURITY_POLICY` could move to conditionals below it.

privaterelay/types.py
jwhitlock: I'm going to review the spec, I think some of the other types are wrong as well.

@robhudson If you want me to add typing annotations upstream (https://github.com/mozilla/django-csp/issues/198), this could be the seed. 28 files should take me about a day to annotate.

jwhitlock:
@groovecoder We may want to wait for https://github.com/mozilla/fx-private-relay/pull/4796, since that also changes the CSP rules. Or get those changes into this PR as well.

privaterelay/settings.py
groovecoder: Nah, that's fine. So long as it was intentional.

groovecoder:
> @groovecoder We may want to wait for #4796, since that also changes the CSP rules. Or get those changes into this PR as well.

Either way is fine. The CSP changes in #4796 will be easy to change to this new syntax.

groovecoder (approved):

privaterelay/settings.py
```diff
@@ -183,36 +184,42 @@
...
-    + _CSP_STYLE_HASHES
-)
-CSP_REPORT_URI = config("CSP_REPORT_URI", "")
+CONTENT_SECURITY_POLICY: CONTENT_SECURITY_POLICY_T = {
+    "DIRECTIVES": {
+        "default-src": [SELF],
+        "connect-src": [
+            SELF,
+            "https://www.google-analytics.com/",
+            "https://www.googletagmanager.com/",
```
groovecoder, line 193: praise: thanks!

groovecoder (approved):

PR #4802, opened 2024-06-21
MPP-4763: Move user creation methods to `privaterelay.tests.utils`

jwhitlock:
Helper functions like `make_free_test_user` are used all over the code. This moves them from `emails.tests.model_tests` to `privaterelay.tests.utils`, which will make even more sense when `Profile` moves to `privaterelay.models`.


groovecoder (approved):
Checks pass; code looks good.

PR #4820, opened 2024-06-27
MPP-3835: Workaround next/image's inline style for SVG images

jwhitlock:
Next.js's `Image` from `next/image` adds an inline style `style="color:transparent"` to some images, such as the logo image:

```html
<img
  alt=""
  loading="lazy"
  width="42"
  height="44"
  decoding="async"
  data-nimg="1"
  style="color:transparent"
  src="/_next/static/media/relay-logo.cabccc81.svg"
/>
```

This is a known issue, with a good explanation by an issue writer at https://0xdbe.github.io/NextJS-CSP-NextImage/ .

Since we do not use image optimization, we can replace this with an `` element. This PR extracts the alt text and image parameters from the `<Image>` declaration, turns an inline `style="color:transparent"` into a `className` with an equivalent CSS rule, and returns the new `` element.

# How to test:

**To see the current behavior**
In `main` or PR #4819, run `cd frontend; npm run build`. Look for `<img` or `style="color:transparent"` in that document. If desired, make a copy for diff later.

In Firefox, open the developer tools, console, and load / reload https://relay.firefox.com/. See many Content-Security-Policy violations.

**To see the new behavior**
In this branch, run `cd frontend; npm run build`. The string "color:transparent" should not be in that document. Some of the `<img` tags will have a new `className` entry, like `Image_transparent__sZTo9`. A visual diff with the old version should show that the `alt=` tag is still present, and the `data-nimg="1"` element is dropped.


frontend/src/components/Image.module.scss
```diff
@@ -0,0 +1,3 @@
+.transparent {
+  color: transparent;
```
rafeerahman, line 2: 👍

frontend/src/components/Image.tsx
```diff
@@ -0,0 +1,37 @@
...
+// If this is fixed in a future next.js release, replace:
+// import Image from "../path/to/Image";
+// with
+// import Image from next/image;
+
+import NextImage, { getImageProps } from "next/image";
+import { ComponentProps } from "react";
+import styles from "./Image.module.scss";
+
+export default function Image(props: ComponentProps<typeof NextImage>) {
```
rafeerahman, line 15: LGTM, tested it out and the build outputs html 's with no inline styles.

frontend/src/components/Image.tsx
```diff
@@ -0,0 +1,37 @@
+// Work around next.js adding inline style color:transparent
+// This breaks our CSP rules about inline styles.
+// More context and some suggested solutions from:
+// https://0xdbe.github.io/NextJS-CSP-NextImage/
```
rafeerahman, line 4: Was a good read, learned some more about CSP and the Image issue with Next.js 👍

rafeerahman (approved):
Thanks for the changes! 

Not required before merging but I get this warning when running a production build,

` ⚠ For production Image Optimization with Next.js, the optional 'sharp' package is strongly recommended. Run 'npm i sharp', and Next.js will use it automatically for Image Optimization.
Read more: https://nextjs.org/docs/messages/sharp-missing-in-production`

Wondering if we should install it or try and remove the warning, as we're not using Image Optimization.

jwhitlock:
> Wondering if we should install sharp or try and remove the warning, as we're not using Image Optimization.

I read https://nextjs.org/docs/messages/sharp-missing-in-production, but I don't see a way to remove the warning, or an existing issue. This will need some more research. It may be worth installing `sharp` and seeing if it changes the SVGs at all, and if it is compatible with this change. I'll spend a little time on this before merging.

jwhitlock:
The output is identical with `sharp` installed, which is what we'd expect since we've disabled optimization. Let's ignore the warning for now, and maybe revisit disabling after the next.js 15 upgrade.

PR #4823, opened 2024-06-28
MPP-3834: Allow inline script, but keep API docs from requiring them

jwhitlock:
*Update - PR #4838 included some of this work, so I've rebased to a single commit.*

The [FAQ](https://drf-spectacular.readthedocs.io/en/latest/faq.html#my-swagger-ui-and-or-redoc-page-is-blank) notes that `SpectacularSwaggerSplitView` can be used to avoid the inline script, but also notes that URL re-writing may break it. This PR switches to `SpectacularSwaggerSplitView`, so only [silk](https://pypi.org/project/django-silk/) is needed for inline scripts (as well as GA4, which we hope to fix soon).

# How to test:

* Load the website, change to different pages. The site continues to load.
* Go to http://127.0.0.1:8000/docs. The page loads, and you can run an API call like `GET /api/v1/profiles/`.

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -4,7 +4,7 @@ from django.urls import URLPattern, URLResolver, include, path, register_convert
 from drf_spectacular.views import (
     SpectacularAPIView,
     SpectacularRedocView,
-    SpectacularSwaggerView,
+    SpectacularSwaggerSplitView,
 )
 from rest_framework import routers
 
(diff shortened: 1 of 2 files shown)


groovecoder:
Now that I've got https://github.com/mozilla/fx-private-relay/pull/4854 up, I'll work on merging these changes into that branch too.

groovecoder:
I combined this change into https://github.com/mozilla/fx-private-relay/pull/4854

PR #4856, opened 2024-07-11
MPP-3807 Remove extension related queries on Glean mentrics

say-yawn:
The DB utilization randomly spikes of 100% DB usage that can result into service interruption or outage. During the spikes the slowest queries are caused while querying if the user has extension and deriving when the extension was installed for the Glean metric pings. For now default `has_extension` and `date_got_extension` to `False` and `-2`

How to test:

### DIFF
--- a/privaterelay/glean_interface.py
+++ b/privaterelay/glean_interface.py
@@ -26,7 +26,11 @@ EmailBlockedReason = Literal[
 
 def _opt_dt_to_glean(value: datetime | None) -> int:
     """Convert an optional datetime to an integer timestamp."""
-    return -1 if value is None else int(value.timestamp())
+    if value == datetime.min:
+        return -2  # datetime was not checked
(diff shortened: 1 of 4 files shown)


privaterelay/glean_interface.py
```diff
@@ -83,19 +87,10 @@ def from_user(cls, user: User) -> UserData:
...
-            has_extension = False
-            date_got_extension = None
-        else:
-            has_extension = True
-            date_got_extension = earliest_mask.created_at
-
+        # Until more accurate date_got_extension is calculated (MPP-3765)
+        # do not check for when the user got extension
+        has_extension = False
+        date_got_extension = datetime.min
```
jwhitlock, line 93: Clever!

jwhitlock (approved):
Thanks @say-yawn!

PR #4880, opened 2024-07-15
MPP-3846: Allow percent greater than 100%, and run in transaction

jwhitlock:
For one of the cleaners (most likely `emails.cleaner.ServerStorageCleaner`, one of the "parts" was greater than the "whole". One way this could happen is if something on the live API server updated the database in the middle of a cleaning command. For example, if the number of `RelayAddress` with `server_storage=False` was 100,000, and then someone created a new one, the number of `RelayAddress` with `server_storage=False` and no stored data would be 100,001.

This removes the check that the whole is greater than or equal to the part, allowing a percentage that is above 100%. I've added tests, along with a comment why.

I also refactored the `./manage.py cleanup_data` command to run each task in a transaction. This should ensure that the data is consistent through the run. This was a pure refactor, no test changes.

# How to test

Run both, see that they are successful:

```sh
./manage.py cleanup_data -v 2
./manage.py cleanup_data -v 2 --clean
```

### DIFF
--- a/privaterelay/cleaner_task.py
+++ b/privaterelay/cleaner_task.py
@@ -827,8 +827,6 @@ class DataIssueTask:
             raise ValueError(f"whole ({whole}) can not be less than 0")
         if part < 0:
             raise ValueError(f"part ({part}) can not be negative")
-        if part > whole:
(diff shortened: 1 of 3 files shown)


groovecoder (approved):
Code and spot-checks look good.

PR #4879, opened 2024-07-15
MPP-4841: Handle trailing newline in headers

jwhitlock:
The Python `email` library does not expect newlines in header values. It rejects newlines with a `ValueError`, except when the newline is at the end of the value. A newline can be present in the incoming email due to encoding. This newline is injected into the forwarded email. The mail client interprets this as the start of the email, and stops header processing.

This PR strips any trailing newlines from the header value, fixing issue #4841.

# How to test:

- [x] I've added a unit test to test for potential regressions of this bug.
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

Issue #4841, opened 2024-07-07
HTML in mail not rendered properly


MarcDirven:
If I receive an e-mail in my "Other" mail or "Non prioritized", HTML is not rendered properly:
![image](https://github.com/mozilla/fx-private-relay/assets/31624073/72029291-8e31-490c-9144-f35a11817627)
![image](https://github.com/mozilla/fx-private-relay/assets/31624073/beee7df7-4d55-4d38-b257-32290b8cbf75)

This only happens in that specific tab, not in my main e-mail tab.

jwhitlock:
What email client is this? 

The email headers seem correct:

```
Content-Type: text/html; charset="utf-8"
Content-Transfer-Encoding: quoted-printable
```
The content looks like [quoted-printable](https://en.wikipedia.org/wiki/Quoted-printable) content. Since it only happens in one tab, this seems like an email client issue.

MarcDirven:
This is [outlook.live.com](url). I wasn't sure whether this was a FFR problem or a mail client problem, problem only happens with FFR in "Other" e-mail tab.

jwhitlock:
I created an outlook.com account and a paired Relay account. I have not been able to reproduce this yet, but I also haven't been able to send a similar email with `Content-Type: text/html; charset="utf-8"`. 
Mails from Gmail and Yahoo end up with a a header `Content-Type: multipart/alternative; boundary="000..."`, with the HTML and text content in MIME multipart sections.

Can you tell me more about the source of this email? If not, I can move on to simulating the email in tests.

MarcDirven:
Well the source is info@alloverpiercings.com and the other one (the correctly rendered one that is) is some cryptic icloud account (info_at_draaimolen_nu_2v4ff4tvbb2230_dadb0052@icloud.com).

It is not the first time this happened by the way. This happened to me once earlier but their website was very yanky, so I thought: "oh well if their website is yanky then so are their e-mails probably", but maybe I guessed wrong. The sender/source of that very first incident was vinylminded-321@mail.common-ground.io. The actual mail I do not have anymore since I deleted that a while ago.

jwhitlock:
As a backend dev, that makes some sense. Promotional emails are often sent by mass email systems that are tuned to work as widely as possible and get into your inbox. Order confirmation emails and transactional emails are sent by backend systems and the developer stops when it works for them...

[...]

PR #4881, opened 2024-07-16
MPP-3839: Update email header logging

jwhitlock:
This is a reporting follow-on for PR #4879 / issue #4841. There were a few issues with the header logging when diagnosing this issue:

* Some parsed headers (from the Python [headerregistry](https://docs.python.org/3.11/library/email.headerregistry.html) set `defects` as documented. Some, like the `Subject` header, do not. However, they are parsed as a `TokenList` with an [all_defects](https://github.com/python/cpython/blob/263c7e611bb24715e513d457a3477a61fff15162/Lib/email/_header_value_parser.py#L136-L138) property that returns any issues. The encoded newline was reported this way, and this may be a root cause of other header errors.
* The previous log entry used the "unstructured" value, which bypassed any special header processing. This was useful for the `Message-ID` bug back in PR #4035, but is less useful than the original raw value for other headers. This switches the log to the raw value. The unstructured value can be derived locally from this.
* The log entries `_handle_received: forwarding issues` are being discarded when routing log entries to BigQuery datasets. I don't know the exact cause, so I'm guessing and changing the log to match others that are getting through. I've used log exports to get a sample of the issues, but I can't determine long-term trends this way. I've changed the log level from `WARNING` to `INFO`, and used a list of dictionaries / objects instead of deeply nested objects with the header name as a key. I hope this will allow me to use BQ to further analyze header issues. 

# How to test:
The unit tests should be sufficient to show this is implemented without errors.

standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).

### DIFF
--- a/emails/types.py
+++ b/emails/types.py
@@ -34,7 +34,7 @@ class EmailHeaderExceptionOnWriteIssue(TypedDict):
 class EmailHeaderDefectIssue(TypedDict):
     defect_count: int
     parsed_value: str
-    unstructured_value: str
+    raw_value: str
 
 
 EmailHeaderIssue = (
--- a/emails/types.py
+++ b/emails/types.py
@@ -23,15 +23,21 @@ AWS_MailJSON = dict[str, Any]
 
 
 class EmailHeaderExceptionOnReadIssue(TypedDict):
+    header: str
+    direction: Literal["in"]
     exception_on_read: str
 
 
 class EmailHeaderExceptionOnWriteIssue(TypedDict):
+    header: str
+    direction: Literal["out"]
     exception_on_write: str
     value: str
 
 
 class EmailHeaderDefectIssue(TypedDict):
+    header: str
+    direction: Literal["in", "out"]
     defect_count: int
     parsed_value: str
     raw_value: str
@@ -43,8 +49,6 @@ EmailHeaderIssue = (
     | EmailHeaderDefectIssue
 )
 
-EmailHeaderIssues = dict[
-    Literal["incoming", "outgoing"], list[tuple[str, EmailHeaderIssue]]
-]
+EmailHeaderIssues = list[EmailHeaderIssue]
 
 EmailForwardingIssues = dict[Literal["headers"], EmailHeaderIssues]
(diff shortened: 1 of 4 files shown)


emails/types.py
```diff
@@ -23,18 +23,24 @@
...
 class EmailHeaderExceptionOnWriteIssue(TypedDict):
+    header: str
+    direction: Literal["out"]
     exception_on_write: str
     value: str
 
 
 class EmailHeaderDefectIssue(TypedDict):
+    header: str
+    direction: Literal["in", "out"]
```
groovecoder, line 40: question (non-blocking): These common fields make me wonder: could the `EmailHeaderExceptionOnReadIssue` and `EmailHeaderExceptionOnWriteIssue` classes extend from this `EmailHeaderDefectIssue` class, and then we might not need the additional `EmailHeaderIssue` type? So we could have `EmailHeaderIssues = list[EmailHeaderDefectIssue]` ?

emails/views.py
```diff
@@ -784,7 +783,7 @@ def _handle_received(message_json: AWS_SNSMessageJSON) -> HttpResponse:
     if has_text:
         incr_if_enabled("email_with_text_content", 1)
     if issues:
-        info_logger.warning(
+        info_logger.info(
```
groovecoder, line 786: question (non-blocking): Looks like this line should have been logging at the info level all along?

groovecoder (approved):
Looks good. Couple non-blocking questions.

emails/views.py
jwhitlock: there's some rules about the loggers that don't seem to be documented anywhere. Only logging 'info` might be one of them.

emails/types.py
jwhitlock: It sounds like you are asking for this:

```python
class EmailHeaderIssue(TypedDict):
    direction: Literal["in", "out"]
    header: str
    defect_count: NotRequired[int]
    parsed_value: NotRequired[str]
    raw_value: NotRequired[str]
    exception_on_read: NotRequired[str]
    value: NotRequired[str]
    exception_on_write: NotRequired[str]
    
EmailHeaderIssues = list[EmailHeaderDefectIssue]
```

That would allow issues like:

```python
issues.append({"direction": "in", "header": "Header"})
issues.append({"direction": "in", "header": "Header", "value": "foo"})
issues.append({"direction": "in", "header": "Header", "defect_count": 3})
issues.append({"direction": "in", "header": "Header", "exception_on_write": "ValueError"})
```

No thanks, I want to type checker to help write correct code.

PR #4900, opened 2024-07-25
MPP-3849: Add Puerto Rico to supported regions

jwhitlock:
[Puerto Rico](https://en.wikipedia.org/wiki/Puerto_Rico) is an unincorporated territory in the United States with the region code `PR`. Puerto Rico is supported by Mozilla VPN.

This adds Puerto Rico to the allowed regions for Relay Plans. Twilio supports [Puerto Rico](https://www.twilio.com/en-us/guidelines/pr/sms), and requires a 10 DLC campaign, like US numbers.

### DIFF
--- a/phones/models.py
+++ b/phones/models.py
@@ -35,6 +35,7 @@ LAST_CONTACT_TYPE_CHOICES = [
     ("call", "call"),
     ("text", "text"),
 ]
+DEFAULT_REGION = "US"
 
 
 def verification_code_default():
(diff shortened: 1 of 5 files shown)


phones/models.py
```diff
@@ -291,9 +292,9 @@ def save(self, *args, **kwargs):
         # as realphone
         self.country_code = realphone.country_code.upper()
 
-        # Add US numbers to the Relay messaging service, so it goes into our
-        # US A2P 10DLC campaign
-        if use_twilio and self.country_code == "US":
+        # Add numbers to the Relay messaging service, so it goes into our
+        # US A2P 10DLC campaigns
```
groovecoder, line 296: quibble/question: Does this comment also need to have "US" removed from it? Not sure if A2P 10DLC campaigns are country-specific?
```suggestion
        # A2P 10DLC campaigns
```

groovecoder (approved):

phones/models.py
jwhitlock: Yep, Twilio just calls them A2P 10DLC - https://www.twilio.com/en-us/a2p-10dlc

PR #5000, opened 2024-09-04
MPP-3864: Add more container-level (C2) diagrams

jwhitlock:
This adds a "top-level" container (c2) diagram, which makes a smoother transition from the system context (c1) diagram, by collecting containers into "container collections". There are further diagrams that "zoom in" to these container collections:

* User Interfaces - The four front-ends for Relay, plus the phone and email services
* Periodic Tasks - The Django management commands scheduled as cronjobs
* Email Processing - The AWS services used to implement email processing
* Other Managed Services - The GCP services other than the database used to implement Relay

The "All details" diagram remains. For this and the deployment diagrams, I've adjusts some of the relationship lines so they do not overlap as much, but instead remain distinct traces.

The easiest way to review is to view the `docs/system_diagrams.md` document with the inline SVG images.


groovecoder (approved):
Whew that is some great documentation and diagrams! Just a single question/suggestion and a praise!

PR #5002, opened 2024-09-06
Update locales for `api-error-account-is-inactive`

jwhitlock:
The id `api-error-account-is-inactive` has been merged upstream, so remove it from pending translations.


jwhitlock:
Self-merging, so the auto-update and Monday dependencies will not break.

PR #5044, opened 2024-09-16
MPP-3897: Add managers to RealPhone to handle multiple phones per user

jwhitlock:
This PR adds managers with helper methods to `RealPhone`, and updates the code to use them. The managers and methods:

* `RealPhone.objects` - The original, can still be used in tests or when working with all `RealPhone` objects.
* `RealPhone.verified_objects` - `RealPhone` objects that the user has verified via a texted verification code.
   - `.get_for_user(user)` - Return the verified `RealPhone` object for the user, or raise `DoesNotExist`. Used to get the `RealPhone` for a requesting user.
   - `.exists_for_number(number)` - Returns True if there is a verified `RealPhone` object for the phone number. Used to determine if a number is already claimed.
   - `.country_code_for_user(user)` - Returns the country code of the user's verified `RealPhone`. Used to limit the suggested Relay numbers to the user's country code.
* `RealPhone.recent_objects` - `RealPhone` objects where a verification code was recently sent
  - `.get_for_user_number_and_verification_code(user, number, code)` - Used to check that a verification code is the one sent by the authorized user to their phone number, and that it hasn't been too long since it was sent. To keep double submissions from being errors, the `RealPhone` can already be verified.
* `RealPhone.pending_objects` - Similar to `RealPhone.recent_objects`, except only the unverified `RealPhone` objects.
  - `exists_for_number(number)` - Check if a pending verification code is currently active for this number, to prevent having two pending codes at the same time and a potential abuse vector.
* `RealPhone.expired_objects` - The opposite of `RealPhone.recent_objects` - `RealPhone` objects where the verification code was sent a long time ago and is no longer valid.
  - `delete_for_number(number)` - Delete any expired records for a particular number. Used in the `RealPhone` save method for ... reasons.
  
This PR fixes a few bugs around `RealPhone` queries when a user has multiple `RealPhone` records, like a verified `RealPhone` and some unverified `RealPhone` for other numbers they tried:

* `/api/v1/profiles/` returns a 500 error, trying to determine when they subscribed to  The code now only queries for the verified `RealPhone`, and there should be one per user (enforced by code, not the database)
* Calls to `send_welcome_message`, to send them their Relay number and contact info, fail. The codenow only queries for the verified `RealPhone`.
* Using `./manage.py delete_phone_data` will fail. Now it deletes all the user's `RealPhone` records, and reports one per line.

How to test: Unit tests should be sufficient. If you want to try live tests, ensure you add an unverified `RealPhone` before a verified `RealPhone`. The code will not allow you to register a new `RealPhone` if you have a verified one.

### DIFF
--- a/api/views/phones.py
+++ b/api/views/phones.py
@@ -160,10 +160,12 @@ class RealPhoneViewSet(SaveToRequestUser, viewsets.ModelViewSet):
         verification_code = serializer.validated_data.get("verification_code")
         if verification_code:
             try:
-                valid_record = RealPhone.recent_objects.get(
-                    user=request.user,
-                    number=serializer.validated_data["number"],
-                    verification_code=verification_code,
+                valid_record = (
+                    RealPhone.recent_objects.get_for_user_number_and_verification_code(
+                        user=request.user,
+                        number=serializer.validated_data["number"],
+                        verification_code=verification_code,
+                    )
                 )
             except RealPhone.DoesNotExist:
                 incr_if_enabled("phones_RealPhoneViewSet.create.invalid_verification")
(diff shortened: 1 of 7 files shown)


groovecoder (approved):
(priase) Nice changes! I had about 3-4 comments lined up to submit, but as I went thru the code each question or suggestion I had was already addressed.

PR #5045, opened 2024-09-19
MPP-3119: complaint notification disables mask

groovecoder:
# Disable mask on complaint notification
See https://mozilla-hub.atlassian.net/browse/MPP-3119 for more context.

When a user (or their email provider) marks a Relay email as spam, we receive a complaint notification. To prevent further spam from reaching the user (causing more complaints), we want to disable the email mask that forwarded the spam email, and notify the user.

# Screenshot (if applicable)
image

# How to test
This will be hard to test until it's on a server with fully operational AWS SES. When it is:

0. Sign up for Relay with an FXA that has a gmail, yahoo, or outlook primary email address
1. In the django admin site, activate the new `disable_mask_on_complaint` flag for the user
2. Create an email mask
3. Send an email to that mask
4. In the final destination mailbox, mark that email as spam
   * Note: It's unclear if all email providers send these complaint notifications, but gmail, yahoo, outlook, and other popular providers seem to do so.
   * Note 2: It's also unclear how long it takes for AWS to process complaint notifications and send them to Relay
5. At some point, Relay should receive the complaint notification
   * [ ] Verify the mask is disabled
   * [ ] Verify the user received an email notification that their mask was disabled

- [ ] Customer Experience team has seen or waived a demo of functionality.
- ~[x] I've added or updated relevant docs in the docs/ directory~

### DIFF
--- a/emails/templates/emails/disabled_mask_for_spam.html
+++ b/emails/templates/emails/disabled_mask_for_spam.html
@@ -5,29 +5,28 @@
 {% load ftl %}
 {% load email_extras %}
 {% withftl bundle='privaterelay.ftl_bundles.main' language=language %}
+
 {% include "emails/direct_email_header.html" %}
+
     <table role="presentation" border="0" cellpadding="0" cellspacing="10px" style="padding: 60px 30px 120px 30px;" align="center">
         <tr>
             <td style="max-width:850px; padding-top: 0px; padding-bottom: 0px; text-align: center;">
+                {% with mask|striptags|urlencode as mask_url %}
                 <h2 style="font-family: inter medium, Arial, system-ui, sans-serif;"> 
                     <img width="18" src="{{ SITE_ORIGIN }}/static/images/email-images/warning.png" style="margin: 0 5px;" alt="warning icon"/>
-                    {% if forwarded == True %}
-                        {% ftlmsg 'replies-not-included-in-free-account-header' %}
-                    {% else %}
-                        {% ftlmsg 'reply-not-sent-header' %}
-                    {% endif %}
+                    {% ftlmsg 'relay-disabled-your-mask' %}
                 </h2>
                 <p style="line-height: 200%; margin-bottom: 30px; padding: 20px 60px 20px 60px">
-                    {% if forwarded == True %}
-                        {% ftlmsg 'first-reply-forwarded' sender=sender %}
-                    {% else %}
-                        {% ftlmsg 'other-reply-not-forwarded-2' sender=sender %}
-                    {% endif %}
-                    {% ftlmsg 'replies-only-available-with-premium' %}
+                {% ftlmsg 'relay-received-spam-complaint-html' mask=mask %} {% ftlmsg 'relay-disabled-your-mask-detail-html' mask=mask %}
                 </p> 
-                <a href="{{ SITE_ORIGIN|add:"/premium/?utm_campaign=email_replies&utm_source=email&utm_medium=email" }}" style="margin: 10px 0;">{% ftlmsg 'upgrade-to-reply-to-future-emails' %}</a>
+                <a href="{{ SITE_ORIGIN }}/accounts/profile/#{{ mask_url }}">
+                    {% ftlmsg 're-enable-your-mask' %}
+                </a>
+                {% endwith %}
             </td> 
         </tr>
     </table>
-{% include "emails/direct_email_footer.html" %}
+
+    {% include "emails/direct_email_footer.html" %}
+
 {% endwithftl %}
--- a/emails/tests/fixtures/disabled_mask_for_spam_expected.email
+++ b/emails/tests/fixtures/disabled_mask_for_spam_expected.email
@@ -1,8 +1,7 @@
-Subject: Replies are not included with your free account
-From: replies@default.com
-To: w41fwbt4q@test.com
-In-Reply-To: <relay-user-msg-id-456@usersemail.com>
-References: <relay-user-msg-id-456@usersemail.com>
+Subject: =?utf-8?q?=E2=81=A8Firefox_Relay=E2=81=A9?= has disabled one of your
+ email masks.
+From: reply@relay.example.com
+To:
 MIME-Version: 1.0
 Content-Type: multipart/alternative;
  boundary="==[BOUNDARY0]=="
@@ -14,24 +13,23 @@ Content-Transfer-Encoding: quoted-printable
 
 
 
-Upgrade for more protection
+=E2=81=A8Firefox Relay=E2=81=A9 has disabled one of your email masks.
 
-Replies are not included with your free account
+=E2=81=A8Firefox Relay=E2=81=A9 received a spam complaint for an email sent t=
+o =E2=81=A8w41fwbt4q@test.com=E2=81=A9. This usually happens if you or your e=
+mail provider mark an email as spam. To prevent further spam, =E2=81=A8Firefo=
+x Relay=E2=81=A9 has disabled your =E2=81=A8w41fwbt4q@test.com=E2=81=A9 mask.
 
+Visit your =E2=81=A8Firefox Relay=E2=81=A9 dashboard to re-enable this mask.
+http://127.0.0.1:8000/accounts/profile/#w41fwbt4q%40test.com
 
-We=E2=80=99ve sent this reply to =E2=81=A8external_sender@test.com=E2=81=A9. =
-But moving forward, your replies will not be sent.
-
-Replying to forwarded emails from your masked email is only available with =
-=E2=81=A8Firefox Relay Premium=E2=81=A9.
-http://127.0.0.1:8000/premium/?utm_campaign=3Demail_replies&amp;utm_source=3D=
-email&amp;utm_medium=3Demail
 
 
 --==[BOUNDARY0]==
 Content-Type: text/html; charset="utf-8"
 Content-Transfer-Encoding: quoted-printable
 MIME-Version: 1.0
+
 
 
 
@@ -225,35 +223,38 @@ ail" style=3D"color: white;">Upgrade for more protection</a>=20
         </tr>
     </table> =20
 
+
     <table role=3D"presentation" border=3D"0" cellpadding=3D"0" cellspacing=
 =3D"10px" style=3D"padding: 60px 30px 120px 30px;" align=3D"center">
         <tr>
             <td style=3D"max-width:850px; padding-top: 0px; padding-bottom: 0=
 px; text-align: center;">
+               =20
                 <h2 style=3D"font-family: inter medium, Arial, system-ui, san=
 s-serif;">=20
                     <img width=3D"18" src=3D"http://127.0.0.1:8000/static/ima=
 ges/email-images/warning.png" style=3D"margin: 0 5px;" alt=3D"warning icon"/>
-                   =20
-                        Replies are not included with your free account
-                   =20
+                    =E2=81=A8Firefox Relay=E2=81=A9 has disabled one of your =
+email masks.
                 </h2>
                 <p style=3D"line-height: 200%; margin-bottom: 30px; padding: =
 20px 60px 20px 60px">
-                   =20
-                        We=E2=80=99ve sent this reply to =E2=81=A8external_se=
-nder@test.com=E2=81=A9. But moving forward, your replies will not be sent.
-                   =20
-                    Replying to forwarded emails from your masked email is on=
-ly available with =E2=81=A8Firefox Relay Premium=E2=81=A9.
+                Firefox Relay received a spam complaint for an email sent to =
+<strong>w41fwbt4q@test.com</strong>. This usually happens if you or your emai=
+l provider mark an email as spam. To prevent further spam, Firefox Relay has =
+disabled your <strong>w41fwbt4q@test.com</strong> mask.
                 </p>=20
-                <a href=3D"http://127.0.0.1:8000/premium/?utm_campaign=3Demai=
-l_replies&amp;utm_source=3Demail&amp;utm_medium=3Demail" style=3D"margin: 10p=
-x 0;">Upgrade now to send replies</a>
+                <a href=3D"http://127.0.0.1:8000/accounts/profile/#w41fwbt4q%=
+40test.com">
+                    Visit your =E2=81=A8Firefox Relay=E2=81=A9 dashboard to r=
+e-enable this mask.
+                </a>
+               =20
             </td>=20
         </tr>
     </table>
 
+   =20
 
     <!-- temporary footer -->
     <table id=3D"relay-email-footer" width=3D"100%" bgcolor=3D"#3D3D3D" style=
@@ -281,6 +282,7 @@ dium=3Demail" style=3D"color: white;">Manage your masks</a>=20
     </table> =20
   </body>
 </html>
+
 
 
 
--- a/emails/tests/fixtures/disabled_mask_for_spam_expected.email
+++ b/emails/tests/fixtures/disabled_mask_for_spam_expected.email
@@ -1,7 +1,7 @@
 Subject: =?utf-8?q?=E2=81=A8Firefox_Relay=E2=81=A9?= has disabled one of your
  email masks.
 From: reply@relay.example.com
-To:
+To: testreal@email.com
 MIME-Version: 1.0
 Content-Type: multipart/alternative;
  boundary="==[BOUNDARY0]=="
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -1192,7 +1192,7 @@ class ComplaintHandlingTest(TestCase):
         self.ra.refresh_from_db()
 
     def test_build_disabled_mask_for_spam_email(self):
-        free_user = make_free_test_user()
+        free_user = make_free_test_user("testreal@email.com")
         test_mask_address = "w41fwbt4q"
         relay_address = baker.make(
             RelayAddress, user=free_user, address=test_mask_address, domain=2
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -24,6 +24,7 @@ from botocore.exceptions import ClientError
 from markus.main import MetricsRecord
 from markus.testing import MetricsMock
 from model_bakery import baker
+from waffle.testutils import override_flag
 
 from emails.models import (
     DeletedAddress,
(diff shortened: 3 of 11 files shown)


emails/views.py
```diff
@@ -1634,6 +1645,22 @@ def _handle_complaint(message_json: AWS_SNSMessageJSON) -> HttpResponse:
         profile.auto_block_spam = True
         profile.save()
 
+    for destination_address in message_json.get("mail", {}).get("destination", []):
+        try:
+            address = _get_address(destination_address, False)
```
groovecoder, line 1650: Note: I changed the function call signature on this to accept a new `bool` parameter to control if the function will/won't create the necessary `DomainAddress` object.

There was enough logic in the `_get_address` function that I didn't want to re-write it. I considered re-factoring into 2 separate `_get_address` and `_get_or_create_address` functions, but that would require changing [all 17 calls `_get_address`](https://github.com/search?q=repo%3Amozilla%2Ffx-private-relay+_get_address&type=code), which seemed like a more disruptive change.

emails/views.py
```diff
@@ -1634,6 +1729,17 @@ def _handle_complaint(message_json: AWS_SNSMessageJSON) -> HttpResponse:
...
 
+    disable_mask_on_complaint_flag, _ = get_waffle_flag_model().objects.get_or_create(
+        name="disable_mask_on_complaint",
+        defaults={
+            "note": (
+                "MPP-3119: When a Relay user marks an email as spam, disable the mask."
+            )
+        },
+    )
+    if disable_mask_on_complaint_flag.is_active_for_user(user):
```
jwhitlock, line 1740: This code should be inside the `for recipient in complained_recipients:` loop. `user` is not guaranteed to be set outside of it.

jwhitlock, line 1740: I'd rather not use `is_active_for_user` until https://github.com/jazzband/django-waffle/pull/498 is released. Please use our override for a while longer:

```suggestion
    if flag_is_active_in_task("disable_mask_on_complaint_flag", user):
```

emails/tests/fixtures/disabled_mask_for_spam_expected.email
```diff
@@ -0,0 +1,289 @@
...
+                <h2 style=3D"font-family: inter medium, Arial, system-ui, san=
+s-serif;">=20
+                    <img width=3D"18" src=3D"http://127.0.0.1:8000/static/ima=
+ges/email-images/warning.png" style=3D"margin: 0 5px;" alt=3D"warning icon"/>
+                    =E2=81=A8Firefox Relay=E2=81=A9 has disabled one of your =
+email masks.
+                </h2>
+                <p style=3D"line-height: 200%; margin-bottom: 30px; padding: =
+20px 60px 20px 60px">
+                Firefox Relay received a spam complaint for an email sent to =
```
jwhitlock, line 242: 🤔 interesting, if the ftl_id ends with `-html`, FTL does not add the unicode FSI and PDI codepoints. Maybe we need a `{ -brand-name-firefox-relay-html }` for these cases?

emails/tests/fixtures/disabled_mask_for_spam_expected.email
```diff
@@ -0,0 +1,289 @@
...
+<strong>w41fwbt4q@test.com</strong>. This usually happens if you or your emai=
+l provider mark an email as spam. To prevent further spam, Firefox Relay has =
+disabled your <strong>w41fwbt4q@test.com</strong> mask.
+                </p>=20
+                <a href=3D"http://127.0.0.1:8000/accounts/profile/#w41fwbt4q%=
+40test.com">
+                    Visit your =E2=81=A8Firefox Relay=E2=81=A9 dashboard to r=
```
jwhitlock, line 249: Here's what the FSI and PDI codepoints look like in the HTML. The `convert_fsi_to_span` filter changes these to `<span dir="auto">...</span>`, but it can be a pain to use.

emails/templates/emails/disabled_mask_for_spam.html
```diff
@@ -0,0 +1,32 @@
...
+{% load ftl %}
+{% load email_extras %}
+{% withftl bundle='privaterelay.ftl_bundles.main' language=language %}
+
+{% include "emails/direct_email_header.html" %}
+
+    <table role="presentation" border="0" cellpadding="0" cellspacing="10px" style="padding: 60px 30px 120px 30px;" align="center">
+        <tr>
+            <td style="max-width:850px; padding-top: 0px; padding-bottom: 0px; text-align: center;">
+                {% with mask|striptags|urlencode as mask_url %}
```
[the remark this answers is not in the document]
groovecoder: The honest answer is copy-pasta from [this line](https://github.com/mozilla/fx-private-relay/blob/2e576168efeb946737fd0da0ee7c07ca3742588d/emails/templates/emails/wrapped_email.html#L134) of `wrapped_email.html` The answer for that one is probably to prevent XSS if someone were to put XSS into the localpart of a domain address? Highly unlikely, but better safe than sorry?

emails/tests/views_tests.py
```diff
@@ -1073,14 +1082,20 @@ def setUp(self):
...
         }
         self.complaint_body = {"Message": json.dumps(complaint)}
+        ses_client_patcher = patch(
+            "emails.apps.EmailsConfig.ses_client",
+            spec_set=["send_raw_email"],
+        )
+        self.mock_ses_client = ses_client_patcher.start()
+        self.addCleanup(ses_client_patcher.stop)
 
     def test_notification_type_complaint(self):
```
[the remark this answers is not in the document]
groovecoder: Hmm, but I like that this test tests the no-flag behavior to make sure that all the existing functionality is the same. What do you think? I'm tempted to just mark this resolved and not worry about it.

emails/tests/views_tests.py
```diff
@@ -1125,6 +1140,77 @@ def test_complaint_log_with_optout(self) -> None:
...
         assert not log_data["fxa_id"]
 
+    @override_flag("disable_mask_on_complaint", active=True)
+    def test_complaint_disables_mask(self):
+        """
+        A notificationType of complaint:
+            1. sets enabled=False on the mask, and
+            2. returns 200.
+        """
+        self.ra = baker.make(
```
say-yawn, line 1150:
```suggestion
        ra = baker.make(
```
(suggestion, non-blocking) The relay address does not get used anywhere outside of this test.

emails/tests/views_tests.py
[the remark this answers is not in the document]
say-yawn: If adding the `override_flag` I recommend also testing that the mask is not disabled as well.

emails/tests/views_tests.py
```diff
@@ -1125,6 +1140,77 @@ def test_complaint_log_with_optout(self) -> None:
...
+        assert destinations == [self.ra.user.email]
+        assert "To prevent further spam" in data_without_newlines
+        assert self.ra.full_address in data_without_newlines
+
+        # re-enable the mask for other tests
+        self.ra.enabled = True
+        self.ra.save()
+        self.ra.refresh_from_db()
+
+    def test_build_disabled_mask_for_spam_email(self):
```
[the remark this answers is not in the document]
groovecoder: It's using `@override_settings(RELAY_FROM_ADDRESS="reply@relay.example.com")` which is tiny, but 🤷 at least this way they're all grouped together.

emails/tests/views_tests.py
```diff
@@ -1125,6 +1140,77 @@ def test_complaint_log_with_optout(self) -> None:
...
+
+        self.ra.refresh_from_db()
+        source = self.mock_ses_client.send_raw_email.call_args.kwargs["Source"]
+        destinations = self.mock_ses_client.send_raw_email.call_args.kwargs[
+            "Destinations"
+        ]
+        raw_message = self.mock_ses_client.send_raw_email.call_args.kwargs["RawMessage"]
+        data_without_newlines = raw_message["Data"].replace("\n", "")
+
+        assert self.ra.enabled is False
```
say-yawn, line 1179:
```suggestion
        assert self.user.profile.auto_block_spam is True
        assert self.ra.enabled is False
```
(suggestion, non-blocking) To make sure that the auto_block_spam still works as expected.

emails/views.py
groovecoder: Addressed in https://github.com/mozilla/fx-private-relay/pull/5045/commits/5d2e79df59aa660422e44018406d5793651e26e2

groovecoder: Addressed in https://github.com/mozilla/fx-private-relay/pull/5045/commits/5d2e79df59aa660422e44018406d5793651e26e2

emails/views.py
```diff
@@ -1414,12 +1444,12 @@ def _get_domain_address(local_portion: str, domain_portion: str) -> DomainAddres
...
 
-def _get_address(address: str) -> RelayAddress | DomainAddress:
+def _get_address(address: str, create: bool = True) -> RelayAddress | DomainAddress:
     """
     Find or create the RelayAddress or DomainAddress for an email address.
 
-    If an unknown email address is for a valid subdomain, a new DomainAddress
-    will be created.
+    If an unknown email address is for a valid subdomain, and create is True,
+    a new DomainAddress will be created.
```
say-yawn, line 1452: (question, blocking) Can you explain why this was added? When will the `create` be set to `False`?

(suggestion, blocking) Also, if the `create` parameter is only used to get/create `DomainAddress` logic, I recommend making the `create` parameter more descriptive like `create_domainaddress`.

emails/views.py
```diff
@@ -1568,10 +1600,73 @@ def _handle_bounce(message_json: AWS_SNSMessageJSON) -> HttpResponse:
...
+
+
 def _handle_complaint(message_json: AWS_SNSMessageJSON) -> HttpResponse:
     """
     Handle an AWS SES complaint notification.
 
+    Sets the user's auto_block_spam flag to True.
+
+    Disables the mask thru which the spam mail was forwarded, and sends an email to the
+    user to notify them the mask is disabled and can be re-enabled on their dashboard.
```
say-yawn, line 1679:
```suggestion
    If `disable_mask_on_complaint` flag enabled,
    disables the mask thru which the spam mail was forwarded, and sends an email to the
    user to notify them the mask is disabled and can be re-enabled on their dashboard.
```

emails/views.py
say-yawn: Note, to move the `get_waffle_flag_model` higher in the `_handle_complaint` code when this is done.

emails/views.py
```diff
@@ -1568,10 +1600,73 @@ def _handle_bounce(message_json: AWS_SNSMessageJSON) -> HttpResponse:
...
+        )
+    except ClientError as e:
+        logger.error("reply_not_allowed_ses_client_error", extra=e.response["Error"])
+    incr_if_enabled("free_user_reply_attempt", 1)
+
+
+def _disable_masks_for_complaint(message_json: dict) -> None:
+    for destination_address in message_json.get("mail", {}).get("destination", []):
+        try:
+            address = _get_address(destination_address, False)
```
say-yawn, line 1654: (question, blocking)  Why would the `create` for `DomainAddress` be set to `False`? Is this handling for getting a complaint on a `DomainAddress` that does not exist? If so, I think a different error should be raised from `ObjectDoesNotExist`.

say-yawn (changes requested):
Thanks for getting this code worked on so quickly. I have some questions on why the `create` parameter was added. I have some additional recommendations that will definitely scope creep:
- Add a field on the mask that the mask was blocked due to complaint. This can later on be used to filter mask for better UX and/or measuring how many masks were set to block in this way.
- Consider using a threshold to decide complaint resulting in blocking all emails. FxA tallies the number of complaint and bounces before they stop sending emails.

emails/views.py
```diff
@@ -1729,16 +1740,7 @@ def _handle_complaint(message_json: AWS_SNSMessageJSON) -> HttpResponse:
...
-        name="disable_mask_on_complaint",
-        defaults={
-            "note": (
-                "MPP-3119: When a Relay user marks an email as spam, disable the mask."
-            )
-        },
-    )
-    if disable_mask_on_complaint_flag.is_active_for_user(user):
-        _disable_masks_for_complaint(message_json)
+    _disable_masks_for_complaint(message_json)
```
say-yawn, line 1743: (suggestion, non-blocking) I recommend this be inside the for-loop where the the auto block spam is being set.

emails/tests/views_tests.py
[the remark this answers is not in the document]
say-yawn: The default behavior will change based on the flag being defaulted to enabled or disabled right? If that's the case I think adding the override here to make it explicit would be better. This would also help when the flag is removed to update this test.

emails/tests/fixtures/disabled_mask_for_spam_expected.email
```diff
@@ -0,0 +1,289 @@
+Subject: =?utf-8?q?=E2=81=A8Firefox_Relay=E2=81=A9?= has disabled one of your
+ email masks.
+From: reply@relay.example.com
+To:
```
jwhitlock, line 4: 🔴 The blank "To" looks like trouble. Could be test setup, could be a logic error.

emails/views.py
jwhitlock: OK, I misunderstood before when I suggested moving this inside the loop. The new code does this:

1. Loop through the `complained_recipients`, matching emails to Relay user's real email. Set the `auto_block_spam` flag for those users
2. Loop through the `destination` field in the mail message, looking for masks. Set those masks to blocking, and send an email
3. Log the actions taken for `complained_recipients`
4. Return an HTTP response like it is 2021

I think the problem is adding part 2 between parts 1 and 3. Instead, you should move the `complained_recipients` code to a new method, and have `_handle_complaint` call both of them. Unless, you want to ensure the masks belong to the complaining users, in which case it should move inside the loop.

jwhitlock (changes requested):
Thanks for cleaning up the code. I still think `_handle_complaint` could be better. I'd support the `get_address` rename instead of the new parameter, if you are ambitious.

Also, the expected email goes to no one, which is either a bug in the new code or the test setup, and should be fixed.

emails/views.py
groovecoder: Yes, it's meant to handle if we ever receive a complaint for a DomainAddress that doesn't exist. E.g.:
1. User creates a domain address
2. User receives some email thru the address
3. User deletes the domain address from Relay
4. User sends a spam complaint about a past email sent thru the domain address

In this scenario, we don't want to re-create the old domain address.

It's raising `ObjectDoesNotExist` just like `_handle_received` and `_reply_allowed` and the other code in `_get_domain_address`. What other error do you think it should raise?

groovecoder:
I made https://github.com/mozilla/fx-private-relay/pull/5045/commits/75e4d85aaa7c1ff2f2d4a3ded5f1fb05b5aed4aa to address https://github.com/mozilla/fx-private-relay/pull/5045#discussion_r1774118156.

Need to dig into https://github.com/mozilla/fx-private-relay/pull/5045#discussion_r1774112661.

emails/views.py
```diff
@@ -1568,10 +1600,107 @@ def _handle_bounce(message_json: AWS_SNSMessageJSON) -> HttpResponse:
...
+    email_regex = rf"[\w\.-]+@(?:{domain_pattern})"
+    matches = re.findall(email_regex, source)
+    if not matches:
+        return
+    for mask_address in matches:
+        try:
+            address = _get_address(mask_address, False)
+            # ensure the mask belongs to the user for whom Relay received a complaint
+            if address.user != user:
+                return
```
jwhitlock, line 1677: maybe go through the loop again?
```suggestion
                continue
```

emails/tests/fixtures/disabled_mask_for_spam_expected.email
jwhitlock: 🔴 this is still blank

emails/views.py
```diff
@@ -1634,6 +1763,8 @@ def _handle_complaint(message_json: AWS_SNSMessageJSON) -> HttpResponse:
         profile.auto_block_spam = True
         profile.save()
 
+        _disable_masks_for_complaint(message_json, user)
```
jwhitlock, line 1766: 👍 this works for doing both

jwhitlock (approved):
I think the blank "To" is problematic, but we're also behind a feature flag, so I'm OK getting some code into `main` and iterating.

emails/tests/fixtures/disabled_mask_for_spam_expected.email
groovecoder: Found and fixed the blank `To:` bug. It was a bug in the test. https://github.com/mozilla/fx-private-relay/pull/5045/commits/53855a3d12dcede9d3b026bdd42e2ec125c9766f

emails/views.py
```diff
@@ -1568,10 +1600,107 @@ def _handle_bounce(message_json: AWS_SNSMessageJSON) -> HttpResponse:
...
+    domain_pattern = "|".join(re.escape(domain) for domain in email_domains)
+    email_regex = rf"[\w\.-]+@(?:{domain_pattern})"
+    matches = re.findall(email_regex, source)
+    if not matches:
+        return
+    for mask_address in matches:
+        try:
+            address = _get_address(mask_address, False)
+            # ensure the mask belongs to the user for whom Relay received a complaint
+            if address.user != user:
```
say-yawn, line 1676:
```suggestion
            if address.user != user or address.enabled == False:
```
(suggestion, non-blocking) Should we send the disabled mask email if the mask was already set to blocked? This can happen when when complaints was filed after the mask was already set to block by us or by the user.

emails/views.py
say-yawn: (suggestion, non-blocking) `DomainAddress` no longer re-creates deleted DomainAddress: https://github.com/mozilla/fx-private-relay/blob/4cde8436e29149a8fc24073e48b2725b6da6810e/emails/validators.py#L75

If we are handling for getting a complaint on a `DomainAddress` or `RelayAddress` that never existed I think throwing a [SuspiciousOperations](https://docs.djangoproject.com/en/5.1/ref/exceptions/#suspiciousoperation) or [ValidationError](https://docs.djangoproject.com/en/5.1/ref/exceptions/#validationerror) so it is not responding with bad request and separately handle and log in `_disable_masks_for_complaint` as:
```
            logger.error(
                "Received a complaint from a destination address that never existed as "
                "a Relay address.",
            )
```

say-yawn (approved):
I agree with what @jwhitlock said about this being behind a feature flag, so I'm OK getting some code into main and iterating. But I made some suggestions that can be included now or in the future. 

Scope creep UX idea: 
I am wondering if it would be helpful to mention how many emails were blocked for the address when the service automatically blocks the email so the user can know they are getting more emails from that mask and make decision to re-enable it later.

jwhitlock (approved):

PR #5073, opened 2024-10-03
MPP-3905: Filter more settings / environment variables

jwhitlock:
Extend the Django default filter to prevent showing sensitive values when `DEBUG=True`. This is enabled on local development machines and occasionally in a development deployment.

# How to test

Run the service with `DEBUG=True`, and go to a URL that will raise an exception like http://127.0.0.1:8000/__debug__/render_panel/. The debug exception handler should run, and the META and SETTINGS sections should have more entries with `'********************'` values.

### DIFF
--- a/privaterelay/tests/debug_tests.py
+++ b/privaterelay/tests/debug_tests.py
@@ -1,4 +1,6 @@
 from django.conf import settings
+from django.http import HttpRequest
+from django.test import RequestFactory
 from django.views.debug import get_default_exception_reporter_filter
 
 import pytest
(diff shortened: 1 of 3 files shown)


Aghied1:
reviewd

privaterelay/tests/debug_tests.py
```diff
@@ -0,0 +1,126 @@
...
+        "SECRET_KEY_FALLBACKS",
+        "TWILIO_ACCOUNT_SID",
+        "TWILIO_ALLOWED_COUNTRY_CODES",
+        "TWILIO_AUTH_TOKEN",
+        "TWILIO_MAIN_NUMBER",
+        "TWILIO_MESSAGING_SERVICE_SID",
+        "TWILIO_NEEDS_10DLC_CAMPAIGN",
+        "TWILIO_SMS_APPLICATION_SID",
+        "TWILIO_TEST_ACCOUNT_SID",
+        "TWILIO_TEST_AUTH_TOKEN",
```
groovecoder, line 68: praise: 34 more tests!

groovecoder (approved):

PR #5089, opened 2024-10-09
MPP-3931: Enable pgcrypto

jwhitlock:
Enable the [pgcrypto](https://www.postgresql.org/docs/current/pgcrypto.html) extension for PostgreSQL, which allows using server-side cryptographic hashes. This is considered "trusted", so the `CREATE` privilege is enough to enable it.

I also adjusted the database configuration so that it can get `DATABASE_URL` from the `.env` file as well as the environment. This allowed me to confirm the migration runs on both SQLite and PostgreSQL.

# How to test
CircleCI tests should be sufficient to test the migration.


groovecoder (changes requested):
question (blocking): This seems fine, but I'm curious why we're enabling this? It's not obvious that the code or application are using it in any way?

groovecoder (approved):

PR #5090, opened 2024-10-11
MPP-3932: Add flag 'developer_mode', use to simulate complaint and log notifications

jwhitlock:
Add a new waffle flag 'developer_mode', that enables additional logging for the user. This logging goes beyond our user privacy policy, so it should only be enabled for Relay staff with consent. In addition, special mask labels are used to limit logging for forwarded emails.

For forwarded mail, the mask label is used to determine developer actions. If the mask contains `DEV:simulate_complaint` (no spaces), then the mail is forwarded to the [AWS complaint simulator address](https://docs.aws.amazon.com/ses/latest/dg/send-an-email-from-console.html#send-email-simulator) instead of the user's real address. If the mask contains `DEV:` (which includes (`DEV:simulate_complaint`)), the received notification itself is logged, using [zlib.compress](https://docs.python.org/3/library/zlib.html#zlib.compress) and [base64.a85encode](https://docs.python.org/3/library/base64.html#base64.a85encode) to reduce the size. If the encoded notification is more than 1024 bytes, the message is split over several log lines.

For complaints, any complaint notification for the user is logged. When we can reliably detect the mask, we may change this to use the mask label to further limit logging.

## How to test:
I wrote tests for the new functionality in handling forwards and complaints.
I refactored the `ComplaintHandlingTest` to add the `mail` element to all test complaint notifications, and to use the forwarded "russian spam" fixture instead of the incoming version. This required some other changes to mock a `Date` header, added by SES when forwarding the email.

I did not manually test this code. I plan to test it on the dev server, where we can set flags and debug the results.

### DIFF
--- a/emails/views.py
+++ b/emails/views.py
@@ -14,7 +14,7 @@ from email.utils import parseaddr
 from io import StringIO
 from json import JSONDecodeError
 from textwrap import dedent
-from typing import Any, Literal
+from typing import Any, Literal, NamedTuple
 from urllib.parse import urlencode
 
 from django.conf import settings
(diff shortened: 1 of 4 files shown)


emails/views.py
```diff
@@ -939,6 +955,66 @@ def _get_email_bytes(
...
+    Relay staff accounts with prior permission.
+
+    The notification JSON will be compressed, Ascii85-encoded with padding, and broken
+    into 1024-bytes chunks. This will ensure it fits into GCP's log entry, which has a
+    64KB limit per label value.
+    """
+
+    notification_gza85 = base64.a85encode(
+        zlib.compress(json.dumps(notification).encode()), wrapcol=1024, pad=True
+    ).decode("ascii")
```
groovecoder, line 1003: quesgestion (non-blocking): can this sequence of `base64`, `zlib`, and `json` calls be moved into a helper function `format_notification_for_log` that can help with re-using it exactly the same way each time? E.g., in the test code?

groovecoder (approved):

jwhitlock:
Added `emails.utils.encode_dict_gza85` and `emails.utils.decode_dict_gza85`. I'll merge after tests pass.

PR #5108, opened 2024-10-15
MPP-3831: Send metrics_event data to influxdb, logs instead of UA

jwhitlock:
For MPP-3831, send the `/metrics-event` data to InfluxDB (tagged counter metric) and logs instead of Universal Analytics, which was retired July 1, 2024. The metric is `fx.private.relay.metrics_event`, and has the UA data as tags. The info-level log has the message `metrics_event`, with the UA data as extra JSON-encoded data. The log includes a truncated hash of the UUID4 generated by the extension, so that we can continue to estimate active installations. We could just log the UUID4 as well, but this ensures we can't easily go from logs to an extension.

I removed the thread code that was talking to Universal Analytics. Since we're talking to in-cluster services instead of a remote service, the data collection should be fast again. It was a cool hack.

### DIFF
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -106,8 +106,6 @@ def metrics_event(request: HttpRequest) -> JsonResponse:
         return JsonResponse({"msg": "Could not decode JSON"}, status=415)
     if "ga_uuid" not in request_data:
         return JsonResponse({"msg": "No GA uuid found"}, status=404)
-
-        # "real_address": sha256(new_emddail.encode("utf-8")).hexdigest(),
     event_data = {
         "ga_uuid_hash": sha256(request_data["ga_uuid"].encode()).hexdigest()[:16],
         "category": request_data.get("category", None),
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -1,6 +1,5 @@
 import json
 import logging
-import threading
 from collections.abc import Iterable
 from datetime import UTC, datetime
 from functools import cache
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -19,7 +19,6 @@ dockerflow==2024.4.2
 drf-spectacular==0.27.2
 drf-spectacular-sidecar==2024.7.1
 glean_parser==15.0.1
-google-measurement-protocol==1.1.0
```
groovecoder, line 22: praise: nice clean-up.

privaterelay/views.py
```diff
@@ -91,40 +90,40 @@ def profile_subdomain(request):
...
-        dimension7=request_data.get("dimension7", "website"),
-    )
-    t = threading.Thread(
-        target=send_ga_ping,
-        args=[settings.GOOGLE_ANALYTICS_ID, request_data.get("ga_uuid"), event_data],
-        daemon=True,
-    )
-    t.start()
+
+        # "real_address": sha256(new_emddail.encode("utf-8")).hexdigest(),
```
groovecoder, line 110: question (non-blocking): where did this comment line come from? What's `new_emddail`?

groovecoder (approved):

privaterelay/views.py
jwhitlock: It's from here:

https://github.com/mozilla/fx-private-relay/blob/d878ad537cff5518036deaf61e5b9a0485471b76/privaterelay/views.py#L304

`emddail` is `email` and trying to run the vim delete line command `dd` when you are instead in insert mode.

I'll remove it.

PR #5114, opened 2024-10-15
MPP-3932: Move developer mode logging inside complaint loop

jwhitlock:
Heroku is raising an error ([RELAY-ZZ](https://mozilla.sentry.io/issues/5993744206/)) during processing that isn't found in the tests:

```
UnboundLocalError('cannot access local variable 'user' where it is not associated with a value')
```

Moving the logging inside the loop should fix this.


groovecoder:
Are you planning to test this on dev heroku server before merging?

jwhitlock:
I was not, but I can

jwhitlock:
No error, but no match on the email either. I'm going to iterate a bit on dev.

Issue #5107, opened 2024-10-14
Update to Python 3.13


jwhitlock:
[Python 3.13](https://www.python.org/downloads/release/python-3130/) is released. Our pattern has been:

* Add CircleCI jobs to test the new Python
* Update to previous version (Python 3.12, #3982)
* Wait a bit

In H2 2024, we have a new role of [Base Load Engineer](https://github.com/mozilla/fx-private-relay/blob/main/docs/base-load-engineer-playbook.md), so it may be more natural to fit this into the normal workflow. We'd want to do this before or after the Django 5.2 update, which is planned for release around April 2025. For example, we may update to Python 3.13 in Q1 2025, and Django 5.2 in Q2 2025.

Issue #3982, opened 2023-10-09
Upgrade to Python 3.12


jwhitlock:
[Python 3.12.0](https://www.python.org/downloads/release/python-3120/) was released October 2, 2023. ~We'll want to upgrade to Python 3.11 first (issue #2731)~ (_done_), then ensure our dependencies work with 3.12.

## New features
* [More flexible f-string parsing](https://docs.python.org/3.12/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings), allowing many things previously disallowed ([PEP 701](https://peps.python.org/pep-0701/)).
* [Support for the buffer protocol](https://docs.python.org/3.12/whatsnew/3.12.html#pep-688-making-the-buffer-protocol-accessible-in-python) in Python code ([PEP 688](https://peps.python.org/pep-0688/)).
* [A new debugging/profiling API](https://docs.python.org/3.12/whatsnew/3.12.html#pep-669-low-impact-monitoring-for-cpython) ([PEP 669](https://peps.python.org/pep-0669/)).
* [Support for isolated subinterpreters](https://docs.python.org/3.12/whatsnew/3.12.html#pep-684-a-per-interpreter-gil) with separate Global Interpreter Locks ([PEP 684](https://peps.python.org/pep-0684)).
* [Even more improved error messages](https://docs.python.org/3.12/whatsnew/3.12.html#improved-error-messages). More exceptions potentially caused by typos now make suggestions to the user.
* [Support for the Linux perf profiler](https://docs.python.org/3.12/howto/perf_profiling.html) to report Python function names in traces.
* [Many large and small performance improvements](https://docs.python.org/3.12/whatsnew/3.12.html#optimizations) (like [PEP 709](https://peps.python.org/pep-0709/) and support for the BOLT binary optimizer), delivering an estimated 5% overall performance improvement.

## Type annotations
* [New type annotation syntax](https://docs.python.org/3.12/whatsnew/3.12.html#pep-695-type-parameter-syntax) for generic classes ([PEP 695](https://peps.python.org/pep-0695/)).
* [New override decorator](https://docs.python.org/3.12/whatsnew/3.12.html#pep-698-override-decorator-for-static-typing) for methods ([PEP 698](https://peps.python.org/pep-0698)).

## Deprecations
* The deprecated `wstr` and `wstr_length` members of the C implementation of unicode objects were removed, per [PEP 623](https://peps.python.org/pep-0623/).
[...]

PR #5167, opened 2024-11-01
MPP-3487: Update to Python 3.11.10, prepare for Python 3.12, 3.13

jwhitlock:
**DRAFT**

Prepare for the next round of Python updates.

* Update from Python 3.11.8 to 3.11.10
* Set Python 3.12 tests as required to pass (#3982 / MPP-3487)
* Add Python 3.13 tests that are allowed to fail (#5107 / MPP-3933)


.circleci/config.yml
```diff
@@ -681,11 +681,18 @@ workflows:
 
       - python_job:
           name: python 3.12 test
-          python_version: "3.12"
+          python_version: "3.12.7"
           command: pytest
-          pytest_fail_fast: true
           test_results_filename: pytest-python-3-12.xml
+          filters: *default_filters
```
groovecoder, line 687: question (non-blocking): Should we alphabetize these? Or at least make their order consistent in each python job?

groovecoder (approved):

.circleci/config.yml
jwhitlock: OK, I'm ordering the `python_job` job definition, then ordering the workflow definitions:

* `python_job.executor` parameters
  - `name`
  - `postgres_version` (as a matrix for the postgres jobs)
  - `python_version`
  - `variant`
* `python_job` parameters
  - `allow_fail`
  - `command`
  - `executor`
  - `mypy_strict`
  - `production_with_new_migrations`
  - `pytest_fail_fast`
  - `pytest_phones_backend`
  - `test_results_filename`
  - `update_dependency`
* `workflow` parameters
  - `filters`
  - (matrix could go here since it is a workflow parameter)

PR #5179, opened 2024-11-07
MPP-3927: Pre-fetch profile for domain address

jwhitlock:
The API for domain address uses the subdomain on the Profile to construct the full domain address. Previously, this required 1 query to get all the domain addresses, then 2 queries _per address_ to get the subdomain. So, a user with 100 domain addresses would use 201 queries to get their list of addresses when they visited the dashboard. This may contribute to some users getting a 30 second timeout when viewing the dashboard.

With this change, the user and profile query is done once, so a user needs 1 query if they have no domain addresses, and 3 total queries for any number of domain addresses.

The relay addresses API does not use user or profile data, so it continues to use 1 query for any number of addresses.

## How to test:

### DIFF
--- a/api/tests/emails_views_tests.py
+++ b/api/tests/emails_views_tests.py
@@ -21,6 +21,29 @@ from privaterelay.tests.utils import (
 )
 
 
+@pytest.fixture
+def settings_without_sqlcommenter(settings: SettingsWrapper) -> SettingsWrapper:
+    """
+    Remove the sqlcommenter from the middleware.
+
+    For sqlite, it injects two queries into the recorded queries. The
+    first is a plain string, the second is the expected dictionary format.
+    This breaks the tests using django_assert_num_queries
+    First query: "SELECT id, ...
+    Second query: {"sql": "SELECT id, ..."}A
+    """
+
+    # The sqlcommenter middleware records queries twice for sqlite
+    try:
+        settings.MIDDLEWARE.remove(
+            "google.cloud.sqlcommenter.django.middleware.SqlCommenter"
+        )
+    except ValueError:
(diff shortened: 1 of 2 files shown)


jwhitlock:
Looking into the test failures

jwhitlock:
Figured it out. The `sqlcommenter` middleware was doubling the queries recorded for sqlite. After disabling it in the tests, the tests pass with both sqlite and psql.

api/tests/emails_views_tests.py
```diff
@@ -9,7 +9,7 @@
 
 import pytest
 from model_bakery import baker
-from pytest_django.fixtures import SettingsWrapper
+from pytest_django.fixtures import DjangoAssertNumQueries, SettingsWrapper
```
groovecoder, line 12: praise: TIL `DjangoAssertNumQueries` !

api/tests/emails_views_tests.py
```diff
@@ -21,6 +21,29 @@
...
+    Remove the sqlcommenter from the middleware.
+
+    For sqlite, it injects two queries into the recorded queries. The
+    first is a plain string, the second is the expected dictionary format.
+    This breaks the tests using django_assert_num_queries
+    First query: "SELECT id, ...
+    Second query: {"sql": "SELECT id, ..."}A
+    """
+
+    # The sqlcommenter middleware records queries twice for sqlite
```
groovecoder, line 36: suggestion (non-blocking): this comment is redundant after the doc-block above
```suggestion
    # The sqlcommenter middleware records queries twice for sqlite
```

groovecoder (approved):
praise: dang, nice investigation and fix! thanks for aligning the request IDs across all the services too.

PR #5234, opened 2024-12-02
MPP-3852: Test `verify_from_sns`

jwhitlock:
This PR adds tests for `verify_from_sns`, which verifies the signature in an SNS JSON body against the signing certificate. Previously, this code was untested, and mocked whenever called. This hides that `OpenSSL.crypto.verify` has been removed in 24.3.0 (see https://github.com/mozilla/fx-private-relay/pull/5228). When running the tests, `pytest` will now report:

```
DeprecationWarning: verify() is deprecated. Use the equivalent APIs in cryptography.
```

My plan is to rebase PR #5228 after this merges, and we should then see test failures. I'll then replace `pyopenssl` and `pem` with the `cryptography` equivalents. The change will be the exception raised when verification fails.

### DIFF
--- a/emails/sns.py
+++ b/emails/sns.py
@@ -64,6 +64,13 @@ SUPPORTED_SNS_TYPES = [
 
 
 def verify_from_sns(json_body):
+    """
+    Check that the SNS message was signed by the cetificate.
+
+    https://docs.aws.amazon.com/sns/latest/dg/sns-verify-signature-of-message.html
+
+    TODO MPP-3852: Stop using OpenSSL.crypto
+    """
     pemfile = _grab_keyfile(json_body["SigningCertURL"])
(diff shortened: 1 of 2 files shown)


jwhitlock:
Here's the warning in CircleCI (`build_test_backend`):
```
=============================== warnings summary ===============================
emails/tests/sns_tests.py::test_verify_from_sns_notification_with_subject_ver1
emails/tests/sns_tests.py::test_verify_from_sns_notification_with_subject_ver1_fails
emails/tests/sns_tests.py::test_verify_from_sns_notification_without_subject_ver1
emails/tests/sns_tests.py::test_verify_from_sns_subscription_ver1
  /app/emails/sns.py:83: DeprecationWarning: verify() is deprecated. Use the equivalent APIs in cryptography.
    crypto.verify(
```

groovecoder (approved):
praise: very nice tests!

PR #5245, opened 2024-12-06
MPP-3487: Revert Python 3.12.7 to 3.11.8

jwhitlock:
This reverts commit cb75394b6e358978ea147952e619eb22cb8302d8, which is suspected to slowdown email processing.


jwhitlock:
Using +r on PR #5244.

PR #5257, opened 2024-12-10
Fix types, CSS import in frontend

jwhitlock:
When attempting to fix PR #5248, I found several typescript errors, and continued to get style warnings. This PR fixes those errors, most of which are in test code.

* Convert `next.config.js` to `next.config.ts`, pass type checking
* Add missing object members to mocked API calls, re-arrange object members to match type definition order
* Switch `fxaFlowTracker.test.ts` to use a number for the time parameter instead of a string, matching the code
* Switch to importing as `import "react-toastify/dist/ReactToastify.css"`, to avoid dart sass warnings. This has the same declarations but with some additional space, which caused the CSS hash to change.
* Reformat touched files

## How to test:

* `cd frontend`
* [x] `npx tsc --noEmit` runs with no warnings or errors
* [x] `npm run test` runs with all tests passing

### DIFF
--- a/frontend/src/components/layout/Layout.tsx
+++ b/frontend/src/components/layout/Layout.tsx
@@ -7,7 +7,7 @@ import {
 } from "react";
 import Link from "next/link";
 import { ToastContainer, Slide } from "react-toastify";
-import "react-toastify/scss/main.scss";
+import "react-toastify/dist/ReactToastify.css";
 import styles from "./Layout.module.scss";
 import logoTypeLight from "./images/fx-private-relay-logotype-light.svg";
 import logoTypeDark from "./images/fx-private-relay-logotype-dark.svg";
--- a/frontend/src/components/layout/navigation/whatsnew/WhatsNewDashboard.test.tsx
+++ b/frontend/src/components/layout/navigation/whatsnew/WhatsNewDashboard.test.tsx
@@ -158,12 +158,17 @@ describe("The 'What's new' dashboard", () => {
       />,
     );
 
-    expect(screen.queryByText(allEntries[1].content)).not.toBeInTheDocument();
+    expect(typeof allEntries[1].content).toBe("string");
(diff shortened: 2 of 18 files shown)


frontend/src/components/layout/Layout.tsx
```diff
@@ -7,7 +7,7 @@ import {
 } from "react";
 import Link from "next/link";
 import { ToastContainer, Slide } from "react-toastify";
-import "react-toastify/scss/main.scss";
+import "react-toastify/dist/ReactToastify.css";
```
groovecoder, line 10: praise: ooo, this might also fix the CSS for the react-toastify 11.0.2 update.

frontend/src/components/layout/navigation/whatsnew/WhatsNewDashboard.test.tsx
```diff
@@ -158,12 +158,17 @@ describe("The 'What's new' dashboard", () => {
       />,
     );
 
-    expect(screen.queryByText(allEntries[1].content)).not.toBeInTheDocument();
+    expect(typeof allEntries[1].content).toBe("string");
+    if (typeof allEntries[1].content !== "string") {
+      return;
+    }
```
groovecoder, line 164: TIL (via Copilot):

> The if (typeof ...) { return; } block is not strictly necessary if the expect(typeof ...).toBe("string") call already ensures that allEntries[1].content is a string. However, the if block provides a safeguard to exit the function early if the type check fails, preventing any further execution that relies on the assumption that allEntries[1].content is a string. This can be useful to avoid runtime errors and make the test more robust.

groovecoder (approved):
Some non-blocking questions/comments. Code looks good. Tests and spot-checks pass.

groovecoder:
Yup. I rebased the react-toastify branch on this one and the build error is fixed AND the styling looks right.

PR #5272, opened 2024-12-20
WIP MPP-3505: RewriteAccount Bearer Token Authentication

jwhitlock:
*Work In Progress - publishing in case I get run over by a reindeer*

While working on MPP-3505 (An `IntegrityError` while setting up a Relay account for a Mozilla account user through Firefox), I found several issues with the Bearer Token Authentication implementation:

* The cache for the Accounts introspection response uses `hash(token)` as a cache key. This give a consistent number for a given Python instance, but a different number on a different Python instance, like a different pod or possibly a different gunicorn thread on the same pod. This means these responses were almost never cached.
* For some endpoints, we make a fresh request (`use_cache=False`), but do not cache the response
* `/api/v1/terms-accepted-user` returns other error codes for some situations, like a 404 if the introspection result does not have a `sub` field. However, the Firefox integration is looking [for a 401 response](https://searchfox.org/mozilla-central/source/toolkit/components/satchel/integrations/FirefoxRelay.sys.mjs#154-157) or a [403 response](https://searchfox.org/mozilla-central/source/toolkit/components/satchel/integrations/FirefoxRelay.sys.mjs#154-157). We should use a 5xx response, like `502 Service Unavailable`, for these unexpected responses, to signal to the client to retry later.
* We should log the time it takes to call a separate service in general. Specifically, the `/api/v1/terms-accepted-user` endpoint calls the Account introspection API and the profile API. Tracking these timers will allow us to determine when Accounts is having an issue that affects Relay.
* The `FxaTokenAuthentication` could use Django REST Framework's [TokenAuthentication](https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication) to standardize token processing and errors.
* `/api/v1/terms-accepted-user` shares code with `FxaTokenAuthentication` but does not use it, requiring some duplicate code.
* `FxaTokenAuthentication` includes some permissions checks that could be implemented as permissions classes and make it more generic.

There was enough to justify a re-write. However, it mostly works for the current Firefox integration, so the new authentication should be shipped along with the old authentication, and the code path selected by a env-backed setting. This will allow testing on stage, and a quick rollback if it fails in production.


groovecoder:
> /api/v1/terms-accepted-user returns other error codes for some situations, like a 404 if the introspection result does not have a sub field. However, the Firefox integration is looking [for a 401 response](https://searchfox.org/mozilla-central/source/toolkit/components/satchel/integrations/FirefoxRelay.sys.mjs#154-157) or a [403 response](https://searchfox.org/mozilla-central/source/toolkit/components/satchel/integrations/FirefoxRelay.sys.mjs#154-157). We should use a 5xx response, like 502 Service Unavailable, for these unexpected responses, to signal to the client to retry later.

Do we know how many `5xx` responses we would return with this change? Those turn into ugly error messages to Firefox clients and end-users. I wonder if we should spend the time to prevent the errors completely rather than surface more up to end-users?

api/authentication.py
```diff
@@ -1,160 +1,414 @@
...
+    fxa_resp: IntrospectionResponse | IntrospectionError | None = None
+    if use_cache:
+        fxa_resp = load_introspection_result_from_cache(cache, token)
+    if fxa_resp is None:
+        fxa_resp = introspect_token(token)
+
+    # If the response is an error, raise an exception
+    if isinstance(fxa_resp, IntrospectionError):
+        if not fxa_resp.from_cache:
+            fxa_resp.save_to_cache(cache, token, default_cache_timeout)
```
groovecoder, line 332: question (non-blocking): If I understand correctly, this means we will cache introspection errors for just 60s? I like that. When I first saw the cache + error code, I thought it might cache something like a timeout error for a long time. This looks like it will prevent a thundering herd AND recover from temporary errors in a reasonable amount of time.

api/authentication.py
```diff
@@ -1,160 +1,414 @@
...
+            from_cache=True,
+        )
+    return IntrospectionResponse(data=cached.get("data"), from_cache=True)
+
+
+def introspect_token(token: str) -> IntrospectionResponse | IntrospectionError:
+    """
+    Validate an Accounts OAuth token with the introspect API.
+
+    If it is a valid token for a Accounts user, returns IntrospectionResponse.
```
groovecoder, line 268:
```suggestion
    If it is a valid token for an Accounts user, returns IntrospectionResponse.
```

api/authentication.py
```diff
@@ -1,160 +1,414 @@
...
+    """
+    Introspect a Mozilla account OAuth token, to get data like the FxA UID.
+
+    If anything goes wrong, raise an exception.
+    """
+    default_cache_timeout = 60
+
+    fxa_resp: IntrospectionResponse | IntrospectionError | None = None
+    if use_cache:
+        fxa_resp = load_introspection_result_from_cache(cache, token)
```
groovecoder, line 325: question (non-blocking): What happens if the cached response is an expired token? It looks like we save the response with a cache expiration time to match the token expiration time, but does the `cached = cache.get(cache_key)` call return `None` in this scenario? I didn't see a test for that.

api/authentication.py
```diff
@@ -1,160 +1,414 @@
...
         method = request.method
-        if method in ["POST", "DELETE", "PUT"]:
-            use_cache = False
-            if method == "POST" and request.path == "/api/v1/relayaddresses/":
-                use_cache = True
-        fxa_uid = get_fxa_uid_from_oauth_token(token, use_cache)
+        path = request.path
+        self.use_cache = (method == "POST" and path == "/api/v1/relayaddresses/") or (
+            method not in ("POST", "DELETE", "PUT")
+        )
```
groovecoder, line 376: suggestion (non-blocking): This logic seems harder to follow than nested `if` statements. Can we revert that? (I also don't remember why we force `use_cache=True` when `POST`ing to `/api/v1/relayaddresses` but I'm sure there's a reason.)

api/views/privaterelay.py
```diff
@@ -314,84 +316,134 @@ def terms_accepted_user(request):
...
+        existing_sa = True
+        action = "found_existing"
+
+    if not existing_sa:
+        # Get the user's profile
+        fxa_profile, response = _get_fxa_profile_from_bearer_token(token)
+        if response:
+            return response
+
+        # Since this takes time, see if another request created the SocialAccount
```
groovecoder, line 341: question (non-blocking): do you suspect this might be the cause of the `SocialAccount.DoesNotExist` errors? That separate requests are creating/retrieving the `SocialAccount` object?

groovecoder (commented):
I didn't look thru the tests yet, but I left some first comments on the first draft of the functional code.

api/authentication.py
jwhitlock: There's a few systems involved, and different answers for each: the cache, the FxA endpoint, and Relay.

The Redis backend checks the expire time on a GET, and if the key is expired, it deletes it and returns None. The tests exercise cache misses and hits, trying both. I'm not sure what the Django in-memory cache does, which is used in tests, but it probably doesn't matter.

If you attempt to introspect an expired token, I expect FxA to return an error. I'd need to research what that error is.

If a user presents an expired token that we've cached, we treat it as authorized. I don't think we check the expiration of cached tokens. There's a chance this has never happened in production, since using `hash()` meant every call is a cache miss. But this new code would be a cache hit, so we may want to check for expiration or near-future expiration.

api/authentication.py
jwhitlock: Yes, I believe the previous code also cached errors for 60 seconds. Except, because of `hash()`, it is possible that cached errors were never loaded again.

api/authentication.py
jwhitlock: Sure, I can revert to the nested `if` statements. I'll see if I can determine from the development PR why we use the cache for creating a new mask. Maybe the Firefox code assumes if the token was good to get the list of masks, it will still be good when creating the mask, and has less error handling.

jwhitlock (commented):
Thanks for the review @groovecoder! This one is still a work in progress. My notes from last year:

* Add the token to `IntrospectionResult` / `IntrospectionError`. Either omit from the cache or require it to be the same.
* Remove token processing from `terms_accepted_user`
* Try permissions classes instead of permission checks in app
* Try always returning anon user, use permission check to reject
* Add timing for introspection API, profile API, and log it and stat it
* Are we done?
* Go back to main and implement as a parallel auth system, throw away this branch

So I do not think this will be the final PR. I'll answer some questions if I can before that final PR.

Additional research and tasks from this review:

* How does the existing code handle expired tokens (if it does)?
* What does FxA's introspect do for an expired token?
* Test and handle cached but expired tokens
* Why does `POST /api/v1/relayaddresses` use the cache, while `POST` to other endpoints skips it?

jwhitlock:
The decision for skipping a cache read on `POST /api/v1/relayaddresses` is on MPP-3156. There's not a justification, but I'm guessing that the `POST` always comes after a `GET`? I'd be OK to remove this exception.

Remaining items:

* Add timing stat for profile fetch
* (Maybe) add tests for logs and stats
* Add checks and tests for expired tokens
* Try sending an expired token to FxA introspect
* Throw away and restart on main

api/views/privaterelay.py
jwhitlock: That is my best guess. But, it could be something weirder!

jwhitlock:
@groovecoder this is ready for review. I was able to implement the flag without starting over from main.

This is the "big switch" version of this change. There are other ways to make these changes. For example, I could extract smaller bits, like checking for a social account again after fetching a profile. Let me know if you'd like a different strategy.

groovecoder:
> This is the "big switch" version of this change. There are other ways to make these changes. For example, I could extract smaller bits, like checking for a social account again after fetching a profile. Let me know if you'd like a different strategy.

We can keep it like this. Though, I'm in no hurry to throw any big Fx integration switches while we're working to expand the Fx integration 10-100x, unless we think we need something like the caching to handle the 10-100x load? But then, it seems like if we fix the caching, that's the change that could have the most unknown effect on the Fx integration flow?

jwhitlock:
> We can keep it like this. Though, I'm in no hurry to throw any big Fx integration switches while we're working to expand the Fx integration 10-100x, unless we think we need something like the caching to handle the 10-100x load? But then, it seems like if we fix the caching, that's the change that could have the most unknown effect on the Fx integration flow?

I agree, I'd want to keep the proven code running during the launch of the integration. The proven code is responsible for 50% of errors, but it is likely these do not impact signup flow, since they haven't shown up in QA testing or customer service reports.

Ideally, I'd be able to run in prod before the ship date, to collect some data in that configuration. The next best is to run in stage, assuming there is a way to test the Firefox integration in stage. That should allow collecting data without affecting Relay users.

Do you know if there is a way to test the Firefox integration in stage?

PR #5299, opened 2025-01-08
Update to latest node 18.x LTS

jwhitlock:
I've been having to restart jobs due to network failures between CircleCI and the node package repository. This _probably_ won't fix that, but it will get us to the latest in the node 18 line. This LTS version is in maintenance and will be supported until around April 2025, when node 24 is released.


groovecoder (approved):
Code looks good. And the netlify deploy job [worked](https://deploy-preview-5299--fx-relay-demo.netlify.app/) on this branch.

PR #5311, opened 2025-01-14
Update to Django 4.2.18

jwhitlock:
This is a [security update](https://www.djangoproject.com/weblog/2025/jan/14/security-releases/). We do not appear to use the feature, so I'm not proposing a patch to the production release.


jwhitlock:
Dependabot opened #5312, removing this one

PR #5373, opened 2025-02-11
MPP-3936: Make AcceptLanguageError picklable, handle for upgrade link

jwhitlock:
For [MPP-3936](https://mozilla-hub.atlassian.net/browse/MPP-3936), make `AcceptLanguageError` picklable, so that if the email handling code raises it, it can be passed from the subprocess to the cronjob's main process.

The most likely cause of the exception is `get_subplat_upgrade_link_by_language`. This PR adds tests and uses `US` as the default if the `Accept-Language` header stored in Mozilla accounts is invalid.

[MPP-3936]: https://mozilla-hub.atlassian.net/browse/MPP-3936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/privaterelay/tests/utils_tests.py
+++ b/privaterelay/tests/utils_tests.py
@@ -21,6 +21,7 @@ from ..utils import (
     AcceptLanguageError,
     flag_is_active_in_task,
     get_countries_info_from_request_and_mapping,
+    get_subplat_upgrade_link_by_language,
     get_version_info,
     guess_country_from_accept_lang,
 )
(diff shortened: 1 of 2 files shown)


privaterelay/tests/utils_tests.py
```diff
@@ -311,6 +313,20 @@ def test_guess_country_from_accept_lang_short_primary_lang_fails(
     assert exc_info.value.accept_lang == accept_lang
 
 
+def test_accept_language_error_is_pickleable_mpp_3936() -> None:
+    bad_header = "en-gb;q=1.0000"
```
groovecoder, line 317: question (non-blocking): I'm curious - what makes this an invalid accept-language header value?

groovecoder (approved):

privaterelay/tests/utils_tests.py
jwhitlock: I copied this from earlier in the file:

https://github.com/mozilla/fx-private-relay/blob/51bb3d0bb2bc9a1d2ecd7eaa96e079a75a11ff0e/privaterelay/tests/utils_tests.py#L245-L246

which I copied from [Django's test suite](https://github.com/django/django/blob/14b46c1b848d846fad55c8634daeabd8787cf0d6/tests/i18n/tests.py#L1557-L1558) two years ago in PR #3682 - don't you remember?

You have to go to [RFC 2616](https://www.rfc-editor.org/rfc/rfc2616#section-3.9) to see why it is invalid - a "q value" should have no more that 3 digits after the decimal.

PR #5376, opened 2025-02-14
MPP-4054: ADR 0005: Document the metrics backend changes

jwhitlock:
This PR adds ADR 5, which documents the changes to metrics platform that started in H2 2024 and are expected to complete in H1 2024. It also updates the system diagrams to reflect the platform at the end of the transition period.


groovecoder (approved):
question (non-blocking): Do we need to mention the technical details of the metrics backend in so many different diagrams? Are there sufficient diagrams for the new Monarch-centered metrics backend in SRE docs somewhere to which we could link from our diagrams?

jwhitlock:
> Do we need to mention the technical details of the metrics backend in so many different diagrams?

I'm not sure I understand this question. Maybe you can give me an example of a change you would make?

> Are there sufficient diagrams for the new Monarch-centered metrics backend in SRE docs somewhere to which we could link from our diagrams?

There's a diagram in the "Proposal: Evolving metrics storage at Mozilla" document, but that is private to Mozilla. I don't think it would be helpful enough to extract and make public. Take a look and see if you think we should pursue that.

I'm not aware of any SRE diagrams of the old or new metrics systems. @pattersonbl2 or @bwells-moz, do you know any?

docs/workspace.dsl
```diff
@@ -26,8 +26,8 @@ workspace "${SERVICE_NAME}" "Mozilla's service providing email and phone masks."
                 description "The Mozilla Telemetry Platform"
                 tags "Other Software System"
             }
-            metrics_system = softwareSystem "Operational Metrics Platform" {
-                description "The Mozilla operational metrics and reporting platform"
+            metrics_system = softwareSystem "Google Managed Prometheus" {
+                description "Stores metrics for queries and reports"
```
groovecoder, line 30: question (non-blocking): E.g., Do we need to change this label from "Operational Metrics Platform", "The Mozilla operational metrics and reporting platform" to the specific "Google Managed Prometheus" label? Insofar as someone needs to understand the Relay architecture, do they need to know that Mozilla's operational metrics & reporting platform is Google Managed Prometheus? Or can we keep this as the more generic "Operational Metrics Platform" and link it to another document that has the technical details?

docs/workspace.dsl
```diff
@@ -68,7 +68,7 @@ workspace "${SERVICE_NAME}" "Mozilla's service providing email and phone masks."
                     description "Collects metrics from other containers"
                     tags "Optional Application",in_other_managed_services
                     technology telegraf
-                    -> metrics_system "Sends metrics" "Telegraf"
+                    -> metrics_system "Pulls metrics" "Prometheus"
```
groovecoder, line 71: comment (non-blocking): same comment here: does a Relay ENGR need to know that metrics are pulled vs. sent and that the underlying system is prometheus vs. telegraf? Or can that detail be in another doc and linked from our docs?

docs/workspace.dsl
```diff
@@ -240,7 +240,7 @@ workspace "${SERVICE_NAME}" "Mozilla's service providing email and phone masks."
                     -> ga "Forwards extension's UI events" "Measurement Protocol"
                     -> email_sender "Sends email" "SES API"
                     -> sentry "Sends exceptions" "HTTPS API"
-                    -> metrics "Sends metrics" "HTTP"
+                    -> metrics "Sends metrics" "UDP"
```
groovecoder, line 243: comment (non-blocking): same comment here: to understand the Relay architecture, does a Relay ENGR need to know that metrics are sent via UDP instead of HTTP?

groovecoder (commented):
It looks like some of the `.svg` files were changed just because they were re-generated - i.e., there weren't any content changes, but the re-generation moved some pixels around.

But apart from that, I gave a few example comments to clarify what I mean by including the technical details of the metrics backend in so many different places.

PR #5456, opened 2025-03-19
for MPP-4020: add configurable SP3 urls

groovecoder:
This PR fixes MPP-4020.

TODO:

This adds plan mappings (prices and urls) for SubPlat3.

* Adds new settings for SubPlat3 configuration.
* Introduces sp3_plans.py to return SP3 plan pricing, subscription URLs, and supported countries.
* Updates `runtime_data` endpoint to dynamically select between SubPlat3 and the legacy system based on `USE_SUBPLAT3` config.
* Modifies frontend logic to support direct subscription URLs from SubPlat3 when available.
* Moves country detection logic into `country_utils.py` for better separation of concerns and  fix a circular dependency.

How to test:

1. Go to http://127.0.0.1:8000/premium/#pricing
   * [ ] Verify the existing prices & purchase links still work as expected
2. Set env var `USE_SUBPLAT3=True`
3. Go to http://127.0.0.1:8000/premium/#pricing
   * [ ] Verify the new prices & purchase links work as expected and go to SubPlat3


privaterelay/country_utils.py
```diff
@@ -0,0 +1,273 @@
+import logging
```
groovecoder, line 1: Note: this code is all un-changed code moved into this new module from `privaterelay.utils` to avoid a circular import.

privaterelay/sp3_plans.py
```diff
@@ -0,0 +1,250 @@
+from functools import lru_cache
```
groovecoder, line 1: When we're moved to SP3, we can delete the existing `privaterelay.plans` module.

frontend/src/functions/getPlan.ts
```diff
@@ -74,7 +83,12 @@ export const getBundleSubscribeLink = (
...
 ) => {
   const plan = getPlan(runtimeData.BUNDLE_PLANS, "yearly");
-  return `${runtimeData.FXA_ORIGIN}/subscriptions/products/${runtimeData.BUNDLE_PRODUCT_ID}?plan=${plan.id}`;
+  if (plan.id) {
+    return `${runtimeData.FXA_ORIGIN}/subscriptions/products/${runtimeData.BUNDLE_PRODUCT_ID}?plan=${plan.id}`;
+  } else if (plan.url) {
+    return plan.url;
+  }
+  return "";
 };
```
vpremamozilla, line 92: could also write it like this:

> if (plan.id) {
>     return `${baseUrl}/subscriptions/products/${productId}?plan=${plan.id}`;
>   }
>   return plan.url ?? "";
> };

frontend/src/functions/getPlan.ts
```diff
@@ -35,8 +35,12 @@ export const getPeriodicalPremiumSubscribeLink = (
...
   const plan = getPlan(runtimeData.PERIODICAL_PREMIUM_PLANS, billingPeriod);
-
-  return `${runtimeData.FXA_ORIGIN}/subscriptions/products/${runtimeData.PERIODICAL_PREMIUM_PRODUCT_ID}?plan=${plan.id}`;
+  if (plan.id) {
+    return `${runtimeData.FXA_ORIGIN}/subscriptions/products/${runtimeData.PERIODICAL_PREMIUM_PRODUCT_ID}?plan=${plan.id}`;
+  } else if (plan.url) {
+    return plan.url;
+  }
+  return "";
 };
```
vpremamozilla, line 44: in the ticket it says:

'As SubPlat upgrades towards SubPlat 3.0 we are updating the URLs that take users from our Marketing website towards the Checkout flow (managed by SubPlat). The new URLs don’t hardcode a specific plan ID but rather necessitate only an interval (monthly, daily, yearly). The rest (locale + country) is determined on SubPlat’s end.'

but we still use `plan.id` to return `${baseUrl}/subscriptions/products/${productId}?plan=${plan.id}`

is your fix a way to keep both old and new functionality?

groovecoder: Right - we'll keep both the old and new functionality in the code while we test the new functionality on our stage server. So we'll leave the production server configured to use old SubPlat 2, and configure the stage server to use new SubPlat 3.0.

When we've tested Relay stage + SubPlat 3.0 enough to verify everything works, we'll configure production to use SubPlat 3.0, verify everything works in production, and then we can remove the SubPlat 2 code.

groovecoder:
> @groovecoder It looks like the benefit of the SP3 stuff is that we don't need to keep track of countries and languages. Shouldn't you send the old structure for old code and a new structure better tuned to the new code, and let the frontend use that instead? Otherwise, you'll be explaining for years why the structure is so complex.

Halfway right. We still need to keep track of countries to adjust the frontend to "Join waitlist" for people from countries in which a paid plan is not available.

Re: languages structure, in the doc-block for `sp3_plans.py`, I added this to explain the complexity:

> The second-level keys are the languages for that country. When all languages in that
> country have the same plan, the single entry is `"*"`. In SubPlat3, all countries have
> `"*"`, because Relay does not need to distinguish between the languages in a country:
> SubPlat3 does that for us. We have kept the second-level structure for backwards
> compatibility with SP2 code while we migrate to SP3. When we have migrated to SP3, we
> could refactor the data structure to remove the unneeded 2nd-level language keys.

frontend/src/functions/getPlan.test.ts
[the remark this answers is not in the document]
vpremamozilla: overall i think the tests are great. you may want to add a test for the edge case - if runtimeData is undefined or missing. not sure if that is a possibility so could be worth adding a test for that

privaterelay/sp3_plans.py
```diff
@@ -0,0 +1,301 @@
...
+* phones - premium, plus a phone mask
+* bundle - premium and phones, plus Mozilla VPN
+
+get_sp3_country_language_mapping gets the details of the paid plans in this structure:
+
+{
+  "AT": {
+    "*": {
+      "monthly": {
+        "id": "price_1LYC79JNcmPzuWtRU7Q238yL",
```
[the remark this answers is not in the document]
groovecoder: Oops, no. I removed this from the doc block.

frontend/src/functions/getPlan.test.ts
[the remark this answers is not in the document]
groovecoder: I tried making a test for it, but all of the functions declare their `runtimeData` parameters to be `RuntimeDataWithPlanAvailable`, which will prevent them from breaking on `undefined` or missing values. So it's not actually possible to call the functions with `undefined` or missing value that I can tell.

vpremamozilla (approved):

frontend/src/functions/getPlan.test.ts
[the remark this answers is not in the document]
Vinnl: Yeah looks good to me. You might be interested in [the `it.each` API](https://jestjs.io/docs/api#each), but that's not a blocker at all.

PR #5470, opened 2025-03-26
MPP-3976 'Get Relay' Button Loop

vpremamozilla:
This PR fixes #<MPP-3976>.

How to test:
- navigate to landing page
- notice the free tier on the matrix has a blue button (active) that says 'Get Relay'
- log into an account 
- navigate to https://relay.firefox.com/premium/
- notice the same button is now disabled and read 'Your Plan'

- [X] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

# Screenshot (if applicable)
Screenshot 2025-03-26 at 5 09 07 PM


frontend/src/components/Button.module.scss
```diff
@@ -67,3 +67,26 @@
...
+
+.button.is-destructive.disabled {
+  background-color: $color-red-10;
+}
+
+.button.is-secondary.disabled {
+  background-color: transparent;
+  color: $color-blue-50;
+  opacity: 0.4;
+}
```
groovecoder, line 92: question (non-blocking): Are these just for completeness? I didn't see any destructive or secondary buttons that would be disabled. But I assume these styles are for possible future buttons?

frontend/src/components/Button.tsx
```diff
@@ -46,29 +47,32 @@ export const LinkButton = forwardRef<
   HTMLAnchorElement,
   Props & AnchorHTMLAttributes<HTMLAnchorElement>
 >((props, ref) => {
-  if (props.href?.startsWith("/")) {
-    const propsWithoutHref = { ...props };
-    delete propsWithoutHref.href;
+  const { disabled, className, variant, href, children, ...rest } = props;
```
groovecoder, line 50: question (non-blocking): I like this `...rest` destructuring pattern, but this is the first and only place I see it in our code-base. Is it a different style of coding React components? Could this confuse someone (probably me) less familiar with React and Typescript in the future? Should we adopt it across the whole code-base or not-at-all?

frontend/src/components/landing/PlanMatrix.tsx
```diff
@@ -30,6 +30,7 @@ import { useL10n } from "../../hooks/l10n";
 import { Localized } from "../Localized";
 import { LinkButton } from "../Button";
 import { VisuallyHidden } from "../VisuallyHidden";
+import { useUsers } from "../../hooks/api/user";
```
groovecoder, line 33: joke 🤡 (non-blocking): how many users would users use if users could `useUsers`? 🔃

frontend/src/components/landing/PlanMatrix.tsx
```diff
@@ -167,8 +170,13 @@ export const PlanMatrix = (props: Props) => {
                   href={getRuntimeConfig().fxaLoginUrl}
                   onClick={() => countSignIn("plan-matrix-free-cta-desktop")}
                   className={styles["primary-pick-button"]}
+                  disabled={
+                    typeof userData.data?.[0] === "object" && !userData.error
```
groovecoder, line 174: suggestion (blocking): Rather than this custom `userData` check from `useUsers`, I think we should re-use the `useIsLoggedIn` hook instead?

groovecoder (changes requested):

frontend/src/components/Button.module.scss
vpremamozilla: yep, i wanted to go ahead and allow other buttons to be able to be disabled as well

frontend/src/components/landing/PlanMatrix.tsx
vpremamozilla: lollll

frontend/src/components/landing/PlanMatrix.tsx
vpremamozilla: good call, ill look at this

frontend/src/components/Button.tsx
vpremamozilla: ive noticed a lot in our code base we are not de structuring at all, ie using a lot of `props.something`
i personally dont care for it but im cool to revert if that is better for consistency. or we can make a tech debt ticket address this across the app

frontend/src/components/Button.module.scss
vpremamozilla: yes this is for completeness. there is no code for disabled buttons across other types so i figured this would be good for the future

vpremamozilla:
had to make a new PR, something got messed up during rebase

PR #5608, opened 2025-05-30
MPP-4227 add google analytics tags megabundle (also fix e2e tests)

vpremamozilla:
- added ga tags for megabundle components
- fix for e2e tests

### DIFF
--- a/e2e-tests/specs/relay-e2e.spec.ts
+++ b/e2e-tests/specs/relay-e2e.spec.ts
@@ -66,35 +66,33 @@ test.describe("Email forwarding and trackers removal", () => {
   });
 });
 
-test.describe("Subscription flows @health_check", () => {
+test.describe("Subscription flows with PlanGrid @health_check", () => {
   /**
-   * Verifies that all plans correctly redirect to its corresponding subscriptions page.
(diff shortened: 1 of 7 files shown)


e2e-tests/specs/relay-e2e.spec.ts
```diff
@@ -66,122 +66,106 @@ test.describe("Email forwarding and trackers removal", () => {
...
 
-  test('Verify that the monthly emails and phones bundle plan "Sign Up" button works correctly, C1818792', async ({
+  test('Verify that the monthly "Relay + Phone" bundle plan works correctly, C1818792', async ({
     landingPage,
     subscriptionPage,
   }) => {
-    test.skip(
-      process.env.E2E_TEST_ENV === "dev",
-      "Dev environment will redirect to the yearly subscription plan.",
-    );
```
groovecoder, line 160: question (non-blocking): so does this test work on the dev server now?

vpremamozilla: no so this just has to do with certain links not existing on the dev server. there was similar logic in place for the plan matrix previously

PR #5647, opened 2025-06-12
MPP-4153 - increase frontend test coverage for mask management

vpremamozilla:
# [MPP-4153](https://mozilla-hub.atlassian.net/browse/MPP-4153) - Part One

Test code coverage improved on the following: 
- waitlist
- phone components
- infomodal
- infotooltip
- phone page

### **Part Two will contain test coverage improvement for components/dashboard & pages/accounts**

# OLD COVERAGE
Screenshot 2025-06-11 at 10 53 33 PM

# NEW COVERAGE (still a WIP)
- waitlist --> 100%
- infomodal --> 100%
- infotooltip --> 100%
- phone components (onboarding) --> 86.12%
- phone components (dashboard) --> 65.55%
- phone (page) --> 86.27%

[MPP-4153]: https://mozilla-hub.atlassian.net/browse/MPP-4153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/components/InfoModal.test.tsx
+++ b/frontend/src/components/InfoModal.test.tsx
@@ -1,0 +1,41 @@
+import "@testing-library/jest-dom";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { InfoModal } from "./InfoModal";
+
+const defaultProps = {
+  isOpen: true,
+  onClose: jest.fn(),
+  modalTitle: "Test Title",
+  modalBodyText: "This is the modal body",
+};
+
+// Helper function to render InfoModal with custom props if needed
+function renderInfoModal(props = {}) {
+  return render(<InfoModal {...defaultProps} {...props} />);
+}
+
+describe("InfoModal", () => {
+  it("renders modal title and body when isOpen is true", () => {
+    renderInfoModal();
+    expect(screen.getByRole("dialog")).toBeInTheDocument();
+    expect(screen.getByText(defaultProps.modalTitle)).toBeInTheDocument();
+    expect(screen.getByText(defaultProps.modalBodyText)).toBeInTheDocument();
+  });
+
+  it("calls onClose when dismiss button is clicked", async () => {
+    const user = userEvent.setup();
+    renderInfoModal();
+    await user.click(screen.getByRole("button"));
+    expect(defaultProps.onClose).toHaveBeenCalled();
+  });
+
+  it("renders custom React elements for title and body", () => {
+    renderInfoModal({
+      modalTitle: <span data-testid="custom-title">Custom Title</span>,
+      modalBodyText: <div data-testid="custom-body">Custom Body</div>,
+    });
+    expect(screen.getByTestId("custom-title")).toBeInTheDocument();
+    expect(screen.getByTestId("custom-body")).toBeInTheDocument();
+  });
+});
--- a/frontend/src/components/InfoTooltip.tsx
+++ b/frontend/src/components/InfoTooltip.tsx
@@ -26,6 +26,7 @@ export const InfoTooltip = (props: Props) => {
         ref={triggerRef}
         {...tooltipTrigger.triggerProps}
         className={styles.trigger}
+        data-testid="trigger"
       >
         <InfoIcon
           alt={props.alt}
--- a/frontend/src/components/phones/dashboard/PhoneDashboard.tsx
+++ b/frontend/src/components/phones/dashboard/PhoneDashboard.tsx
@@ -291,6 +291,7 @@ export const PhoneDashboard = (props: Props) => {
         ) : (
           // Caller and SMS Senders Panel
           <SendersPanelView
+            data-testid="senders-panel-view"
             type={getSendersPanelType()}
             back_btn={toggleSendersPanel}
           />
--- a/frontend/src/components/phones/onboarding/PhoneOnboarding.test.tsx
+++ b/frontend/src/components/phones/onboarding/PhoneOnboarding.test.tsx
@@ -1,0 +1,99 @@
+import { render, screen } from "@testing-library/react";
+import { PhoneOnboarding } from "./PhoneOnboarding";
+import * as realPhoneHook from "../../../hooks/api/realPhone";
+import {
+  mockedProfiles,
+  mockedRuntimeData,
+  mockedRealphones,
+} from "frontend/src/apiMocks/mockData";
+import { RuntimeDataWithPhonesAvailable } from "../../../functions/getPlan";
+import { useL10n } from "../../../hooks/l10n";
+
+// Mocks
+jest.mock("../../../hooks/l10n");
+jest.mock("../../../hooks/api/realPhone");
+
+const mockUseRealPhonesData = realPhoneHook.useRealPhonesData as jest.Mock;
+
+const l10nMock = {
+  bundles: [{ locales: ["en"] }],
+  getString: jest.fn((key, vars) =>
+    vars
+      ? `l10n string: [${key}], with vars: ${JSON.stringify(vars)}`
+      : `l10n string: [${key}], with vars: {}`,
+  ),
+  getFragment: jest.fn((key) => `l10n fragment: [${key}]`),
+};
+
+const renderComponent = (profileKey: keyof typeof mockedProfiles) =>
+  render(
+    <PhoneOnboarding
+      onComplete={jest.fn()}
+      profile={mockedProfiles[profileKey]}
+      runtimeData={mockedRuntimeData as RuntimeDataWithPhonesAvailable}
+    />,
+  );
+
+describe("PhoneOnboarding", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    (useL10n as jest.Mock).mockReturnValue(l10nMock);
+  });
+
+  it("renders nothing if realPhoneData is undefined", () => {
+    mockUseRealPhonesData.mockReturnValue({ data: undefined });
+
+    const { container } = renderComponent("empty");
+
+    expect(container).toBeEmptyDOMElement();
+  });
+
+  it("renders RealPhoneSetup if no verified phones", async () => {
+    mockUseRealPhonesData.mockReturnValue({
+      data: mockedRealphones.empty,
+      error: null,
+      requestPhoneVerification: jest.fn(),
+      requestPhoneRemoval: jest.fn(),
+      submitPhoneVerification: jest.fn(),
+    });
+
+    renderComponent("empty");
+
+    expect(
+      await screen.findByText((text) =>
+        text.includes("phone-onboarding-step2-headline"),
+      ),
+    ).toBeInTheDocument();
+  });
+
+  it("renders RealPhoneSetup if realPhoneData has error", async () => {
+    mockUseRealPhonesData.mockReturnValue({
+      data: [],
+      error: true,
+      requestPhoneVerification: jest.fn(),
+      requestPhoneRemoval: jest.fn(),
+      submitPhoneVerification: jest.fn(),
+    });
+
+    renderComponent("empty");
+
+    expect(
+      await screen.findByText((text) =>
+        text.includes("phone-onboarding-step2-headline"),
+      ),
+    ).toBeInTheDocument();
+  });
+
+  it("renders RelayNumberIntro if there are verified phones", async () => {
+    mockUseRealPhonesData.mockReturnValue({
+      data: mockedRealphones.full as realPhoneHook.VerifiedPhone[],
+      error: null,
+    });
+
+    renderComponent("full");
+
+    expect(
+      await screen.findByTestId("relay-number-intro-cta"),
+    ).toBeInTheDocument();
+  });
+});
--- a/frontend/src/components/phones/onboarding/PurchasePhonesPlan.test.tsx
+++ b/frontend/src/components/phones/onboarding/PurchasePhonesPlan.test.tsx
@@ -1,0 +1,83 @@
+import { render, screen, fireEvent } from "@testing-library/react";
+import { PurchasePhonesPlan } from "./PurchasePhonesPlan";
+import { RuntimeDataWithPhonesAvailable } from "../../../functions/getPlan";
+import { mockedRuntimeData } from "frontend/src/apiMocks/mockData";
+
+jest.mock("../../../hooks/l10n", () => ({
+  useL10n: () => ({
+    getString: (id: string, vars?: Record<string, string>) =>
+      vars
+        ? `l10n string: [${id}] with vars: ${JSON.stringify(vars)}`
+        : `l10n string: [${id}]`,
+  }),
+}));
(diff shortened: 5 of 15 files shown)


frontend/src/components/InfoModal.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+        onClose={onClose}
+        modalTitle={<span data-testid="custom-title">Custom Title</span>}
+        modalBodyText={<div data-testid="custom-body">Custom Body</div>}
+      />,
+    );
+
+    expect(screen.getByTestId("custom-title")).toBeInTheDocument();
+    expect(screen.getByTestId("custom-body")).toBeInTheDocument();
+  });
+});
```
groovecoder, line 41: suggestion (non-blocking): via GitHub Copilot to reduce code duplication (similar to the `renderTooltip` in the other file.):

```
import "@testing-library/jest-dom";
import { render, screen } from "@testing-library/react";
import userEvent from "@testing-library/user-event";
import { InfoModal } from "./InfoModal";

const defaultProps = {
  isOpen: true,
  onClose: jest.fn(),
  modalTitle: "Test Title",
  modalBodyText: "This is the modal body",
};

// Helper function to render InfoModal with custom props if needed
function renderInfoModal(props = {}) {
  return render(<InfoModal {...defaultProps} {...props} />);
}

describe("InfoModal", () => {
  it("renders modal title and body when isOpen is true", () => {
    renderInfoModal();
    expect(screen.getByRole("dialog")).toBeInTheDocument();
    expect(screen.getByText(defaultProps.modalTitle)).toBeInTheDocument();
    expect(screen.getByText(defaultProps.modalBodyText)).toBeInTheDocument();
  });

  it("calls onClose when dismiss button is clicked", async () => {
    const user = userEvent.setup();
    renderInfoModal();
    await user.click(screen.getByRole("button"));
    expect(defaultProps.onClose).toHaveBeenCalled();
  });

  it("renders custom React elements for title and body", () => {
    renderInfoModal({
      modalTitle: <span data-testid="custom-title">Custom Title</span>,
      modalBodyText: <div data-testid="custom-body">Custom Body</div>,
    });
    expect(screen.getByTestId("custom-title")).toBeInTheDocument();
    expect(screen.getByTestId("custom-body")).toBeInTheDocument();
  });
});
```

frontend/src/components/InfoTooltip.tsx
```diff
@@ -26,6 +26,7 @@ export const InfoTooltip = (props: Props) => {
         ref={triggerRef}
         {...tooltipTrigger.triggerProps}
         className={styles.trigger}
+        data-testid="trigger"
```
groovecoder, line 29: praise: yay more `data-testid` !!

frontend/src/components/phones/dashboard/PhoneDashboard.tsx
```diff
@@ -291,6 +291,7 @@ export const PhoneDashboard = (props: Props) => {
         ) : (
           // Caller and SMS Senders Panel
           <SendersPanelView
+            data-testid="senders-panel-view"
```
groovecoder, line 294: praise: 🤩 `data-testid`!!

frontend/src/components/phones/onboarding/PhoneOnboarding.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    renderComponent("empty");
+
+    expect(
+      await screen.findByText((text) =>
+        text.includes("phone-onboarding-step2-headline"),
+      ),
+    ).toBeInTheDocument();
+  });
```
groovecoder, line 85: question (non-blocking): hmm ... when does this happen? What does `error: true` here mean? Should it be showing the phone onboarding step 2 or should it be showing an error?

frontend/src/components/phones/onboarding/PhoneOnboarding.test.tsx
```diff
@@ -0,0 +1,99 @@
+import { render, screen } from "@testing-library/react";
+import { PhoneOnboarding } from "./PhoneOnboarding";
+import * as realPhoneHook from "../../../hooks/api/realPhone";
+import {
+  mockedProfiles,
+  mockedRuntimeData,
+  mockedRealphones,
+} from "frontend/src/apiMocks/mockData";
+import { RuntimeDataWithPhonesAvailable } from "../../../functions/getPlan";
+import { useL10n } from "../../../hooks/l10n";
```
groovecoder, line 10: suggestion (non-blocking): There's 16 instances of `useL10n` in this PR alone. Unless we have a ticket specifically for mocking `useL10n`, it's probably time to figure how out to mock this hook in `__mocks__` rather than re-mock it individually in every test file.

frontend/src/components/phones/onboarding/PurchasePhonesPlan.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+jest.mock("../../../functions/trackPurchase", () => ({
+  trackPlanPurchaseStart: jest.fn(),
+}));
+
+jest.mock("../../../functions/getPlan", () => {
+  const actual = jest.requireActual("../../../functions/getPlan");
+  return {
+    ...actual,
+    getPhonesPrice: jest.fn(() => "$9.99"),
+    getPhoneSubscribeLink: jest.fn((_, period) => `/subscribe/${period}`),
```
groovecoder, line 32: question (non-blocking): do we need to mock these functions? Seems like this test doesn't assert against these values, and I thought our test framework had fixtures for the runtimedata and billing periods so these could work?

frontend/src/components/phones/onboarding/PurchasePhonesPlan.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    // Switch to monthly tab
+    const monthlyTab = screen.getByText(
+      /phone-onboarding-step1-period-toggle-monthly/i,
+    );
+    fireEvent.click(monthlyTab);
+
+    const monthlyLink = screen.getByRole("link", {
+      name: /phone-onboarding-step1-button-cta-2/i,
+    });
+    expect(monthlyLink).toHaveAttribute("href", "/subscribe/monthly");
```
groovecoder: question (non-blocking): Why does only this test use `/i` case-insensitive matching? Should all (or none) of the tests use this consistently?

frontend/src/components/phones/onboarding/RealPhoneSetup.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+jest.mock("../../Localized", () => ({
+  Localized: ({ children }: React.PropsWithChildren) => <>{children}</>,
+}));
+
+jest.mock("../../../hooks/api/realPhone", () => {
+  const actual = jest.requireActual("../../../hooks/api/realPhone");
+  return {
+    ...actual,
+    useRealPhonesData: () => ({ data: [] }),
```
groovecoder, line 24: question (non-blocking): do we need to mock this `useRealPhonesData` function? I don't see any test code that asserts against its value.

frontend/src/components/phones/onboarding/RealPhoneSetup.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    expect(
+      screen.getByText(/phone-onboarding-step2-headline/),
+    ).toBeInTheDocument();
+
+    expect(
+      screen.getByRole("button", {
+        name: /phone-onboarding-step3-button-cta/,
+      }),
+    ).toBeInTheDocument();
```
groovecoder, line 91: question (non-blocking): a number of other tests expect the `phone-onboarding-step2-headline` element to be in the document, but they DO NOT expect this `step3-button-cta` element? Is that on purpose? Why expect it specifically in this test?

frontend/src/components/phones/onboarding/RealPhoneSetup.tsx
```diff
@@ -178,7 +178,11 @@ const RealPhoneForm = (props: RealPhoneFormProps) => {
           />
         </div>
 
-        <Button className={styles.button} type="submit">
+        <Button
+          className={styles.button}
+          type="submit"
+          data-testid="relay-number-intro-cta"
```
groovecoder, line 184: praise: `data-testid`

frontend/src/components/phones/onboarding/RelayNumberConfirmationModal.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+      ),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByText((text) => text.includes("234") && text.includes("8900")),
+    ).toBeInTheDocument();
+  });
+
+  it("calls onClose when close icon is clicked", () => {
+    renderWithProvider(<RelayNumberConfirmationModal {...defaultProps} />);
+    const closeButton = screen.getByRole("button", { name: /close/i });
```
groovecoder: issue (non-blocking): another usage of case-insensitive `/i` matching.

frontend/src/components/phones/onboarding/RelayNumberPicker.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+        name: "translated(phone-onboarding-step3-code-success-cta-2)",
+      }),
+    );
+
+    const showMoreBtn = screen.getByRole("button", {
+      name: "translated(phone-onboarding-step4-button-more-options)",
+    });
+
+    fireEvent.click(showMoreBtn);
+    expect(screen.getByRole("radio")).not.toBeChecked(); // radio reset
```
groovecoder: issue (blocking): this doesn't actually expect/verify the behavior described in the test description? I.e., the test description says it cycles suggestions when clicking 'more options' but this test only expects that the radio is not checked? It doesn't expect or verify that the suggestions have cycled.

frontend/src/components/phones/onboarding/RelayNumberPicker.tsx
```diff
@@ -71,7 +71,10 @@ const RelayNumberIntro = (props: RelayNumberIntroProps) => {
 
       {/* TODO: Add logic to display success message */}
 
-      <div className={`${styles["step-input-verificiation-code-success"]} `}>
+      <div
+        data-testid="confirmation-success-title"
```
groovecoder, line 75: praise: 🤩 `data-testid`

frontend/src/pages/phone/waitlist.page.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+    mockGetString.mockImplementation((id: string) => {
+      const strings: Record<string, string> = {
+        "waitlist-heading-phone": "Get a private phone number",
+        "waitlist-lead-phone": "Be the first to try Relay phone masking",
+        "waitlist-privacy-policy-use-phone":
+          "We will use your phone only to notify you.",
+      };
+      return strings[id] ?? id;
+    });
+  });
```
groovecoder, line 33: comment (non-blocking): Yeah, this kind of mocking of `useL10n` really makes me want to figure out our suite-wide technique for mocking `l10n`. It seems like we should be able to assert against message IDs instead of text, and/or we should assert against the real translations.

frontend/src/components/phones/onboarding/RelayNumberPicker.test.tsx
vpremamozilla: good callout; test is updated to capture and compare the set of suggestion labels before and after clicking "More Options", so it now explicitly verifies that the suggestions have cycled, not just that the radio was reset

vpremamozilla:
> 1 blocking issue.
> 
> And this PR really makes it clear to me that we need to establish our suite-wide technique for mocking l10n. Or possibly NOT mocking it, and making the real l10n functions work in the test suite.

- addressing this, i think i will prioritize fixing the test mocking system first before writing more tests for this epic. will pull that into this sprint (or next)

frontend/src/components/phones/onboarding/PhoneOnboarding.test.tsx
vpremamozilla: agreed. will pull in that ticket next! https://mozilla-hub.atlassian.net/browse/MPP-4229

frontend/src/components/phones/onboarding/PhoneOnboarding.test.tsx
vpremamozilla: this is a lil funky based on what the `PhoneOnboarding` component is doing: 
```

if (realPhoneData.data === undefined) {
  return null;
}
```

- The component renders nothing (returns null) if the data is still undefined (likely loading).

```
if (realPhoneData.error || verifiedPhones.length === 0) {
  step = <RealPhoneSetup ... />;
}
```

- If there's an error, or no verified phones, the user is sent to the RealPhoneSetup step.


TL;DR: the current behavior is intentional.

- Even if there's an error, the app still tries to render the onboarding step for verifying phones.

- This implies that:
     1 - The error might be non-fatal (e.g. a partial failure, or recoverable).
     2 - The product is optimized for getting users into verification flow even if data fetching isn’t perfect.

frontend/src/components/phones/onboarding/PurchasePhonesPlan.test.tsx
vpremamozilla: these are mocked for typing reasons, the tests fail without them in place

frontend/src/components/phones/onboarding/PurchasePhonesPlan.test.tsx
vpremamozilla: this is a good callout - i think they were in there previously (or some other file perhaps) but removing them doesnt break anything so ill remove them everywhere

frontend/src/components/phones/onboarding/RealPhoneSetup.test.tsx
vpremamozilla: same as previous comment - this is for typing

frontend/src/components/phones/onboarding/RealPhoneSetup.test.tsx
vpremamozilla: yes on purpose, to distinguish between form and verification step

frontend/src/components/phones/onboarding/RelayNumberPicker.test.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+    const newRadios = screen.getAllByRole("radio");
+    const newLabels = newRadios.map((radio) => {
+      const label = screen.getByLabelText(
+        radio.getAttribute("aria-label") || "",
+      );
+      return label.textContent?.trim();
+    });
+
+    expect(newLabels).not.toEqual(initialLabels);
```
groovecoder, line 148: praise: thanks for the fix!

groovecoder (approved):

PR #5731, opened 2025-07-19
e2e test subscription flows fix

vpremamozilla:
- fixes e2e tests 
- subscription page had issues with datatest-id being repeated

### DIFF
--- a/e2e-tests/pages/subscriptionPaymentPage.ts
+++ b/e2e-tests/pages/subscriptionPaymentPage.ts
@@ -47,7 +47,10 @@ export class SubscriptionPaymentPage {
     this.planDetails = page.locator("#plan-details-product");
     this.planDetails3 = page.locator("#product-details-heading");
     this.planType = page.locator(".plan-details-description");
-    this.planType3 = page.getByTestId("total-price");
+    // this is ugly but someone repeated the datatest-id="total-price" in the component
+    this.planType3 = page.locator(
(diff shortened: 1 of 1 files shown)


e2e-tests/pages/subscriptionPaymentPage.ts
```diff
@@ -47,31 +47,45 @@ export class SubscriptionPaymentPage {
     this.planDetails = page.locator("#plan-details-product");
     this.planDetails3 = page.locator("#product-details-heading");
     this.planType = page.locator(".plan-details-description");
-    this.planType3 = page.getByTestId("total-price");
+    // this is ugly but someone repeated the datatest-id="total-price" in the component
```
groovecoder, line 50: suggestion (non-blocking): bring this up with FXA team too? I'm okay to merge this as-is, but it seems like they should fix this part themselves, and if they do, we might have to revert this particular code?

vpremamozilla: yes i agree!

vpremamozilla:
> question (non-blocking): Why did we seem to remove the `isVersion3()` check from only a couple of the tests, but we left it in-place for the other one?

- this is actually a good point, do i have the go-ahead to remove all the version3 logic? it seems that they all use version3 values. i didnt want to remove it since it seemed like you put that in there intentionally a couple of months ago

PR #5742, opened 2025-07-22
MPP-4033: Fix Relay user replies to be case insensitve to their stored email

joeherm:
This PR fixes [MPP-4033](https://mozilla-hub.atlassian.net/browse/MPP-4033)

I've both reproduced, tested this fix with a local E2E email setup, and added a unit tests.
How to test:

1. Create a user with an uppercase email address in FXA and enable premium subscription
2. Create a mask for the user
3. Send an email to the mask, the user should receive it, but not be able to reply
4. After deploy, user can reply to the email

- [x] Reproduced in staging and can test this well after a deploy

Checklist:

[MPP-4033]: https://mozilla-hub.atlassian.net/browse/MPP-4033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -969,6 +969,15 @@ class SNSNotificationRepliesTest(SNSNotificationTestBase):
             expected_fixture_name="s3_stored_replies_with_emoji",
         )
 
+    def test_reply_case_insensitive_relay_user_email(self) -> None:
+        """Casing of a Relay user's email doesn't prevent outgoing replies"""
+        self.user.email = "sOuRcE@sender.com"
+        self.user.save()
+        self.successful_reply_test_implementation(
+            text="this is a text reply",
+            expected_fixture_name="s3_stored_replies",
+        )
+
     @patch("emails.views._reply_allowed")
     def test_reply_not_allowed(self, mocked_reply_allowed: Mock) -> None:
         mocked_reply_allowed.return_value = False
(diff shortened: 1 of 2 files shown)


emails/tests/views_tests.py
```diff
@@ -969,6 +969,15 @@ def test_reply_with_emoji_in_text(self) -> None:
             expected_fixture_name="s3_stored_replies_with_emoji",
         )
 
+    def test_reply_case_insensitive_relay_user_email(self) -> None:
+        """Casing of a Relay user's email doesn't prevent outgoing replies"""
+        self.user.email = "Source@sender.com"
```
groovecoder: quibble (non-blocking): can we try a few more variants here just to make sure it's not some weird case where a first capital letter works but a capital letter somewhere else doesn't?

groovecoder (approved):
Nice! The new test failed on `main` and then passed on this branch. 1 non-blocking quibble if you want to add a little extra diligence to the test before merging.

emails/tests/views_tests.py
joeherm: I'm just going to update this to use a different (non-capitalized) cased email, so that I don't have to extend `successful_reply_test_implementation` or add mostly redundant cases to the suite. It should make sure that we don't accidentally remove `.lower()` in favor of a potential `.capitalize()` logic.

PR #5795, opened 2025-08-14
fix(settings.py): use more consistent names between settings.py variables and envvars

joeherm:
This PR fixes MPP-4341, where we identify the goal of using more consistent variable names between `settings.py` and environment variables. Specifically, some variables are no longer used, others are inconsistent in their use of the `DJANGO_` prefix. We are motivated to do this now because we are changing our deployment strategy.

**I've made the opinionated choice to drop the `DJANGO_` prefix altogether, while maintaining backwards-compatibility with existing deployments.** We by no means have to drop its usage everywhere and could instead maintain more consistency between sections. Consider this a proposal, happy to iterate based on feedback.

How to test:


privaterelay/settings.py
```diff
@@ -89,19 +91,30 @@
...
 # Honor the 'X-Forwarded-Proto' header for request.is_secure()
 SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
-SECURE_SSL_HOST = config("DJANGO_SECURE_SSL_HOST", None)
-SECURE_SSL_REDIRECT = config("DJANGO_SECURE_SSL_REDIRECT", False, cast=bool)
+SECURE_SSL_HOST = config("SECURE_SSL_HOST", None) or config(
+    "DJANGO_SECURE_SSL_HOST", None
+)
+SECURE_SSL_REDIRECT = config("SECURE_SSL_REDIRECT", False, cast=bool) or config(
+    "DJANGO_SECURE_SSL_REDIRECT", False, cast=bool
+)
```
groovecoder, line 99: question (non-blocking): I forget - are these `or` expressions evaluated left-first or right-first? I'm wondering about the (small) amount of time where the app code may run on some pods with `DJANGO_SECURE_*` and some pods with `SECURE_*`. It looks like the logic here will work fine, but just want to ask the question to make sure.

joeherm: They're evaluated left-to-right. I do think it's a good question about overlap- I think ideally in MozCloud we never have pods with both variables set so should be a non-issue. As for GCP v1, assuming we don't set both vars we should always be using the right-hand side of the `or`. I'll double-check to make sure we don't have any extraneous variables in the existing deployment.

PR #5809, opened 2025-08-18
fix(ci): Build frontend in Dockerfile

jbuck:
Relay is now running in MozCloud at https://relay-dev.allizom.org/ . Not much of a front-end going on there!
This PR adds the frontend build to the Dockerfile. 

https://mozilla-hub.atlassian.net/browse/SVCSE-3549

### DIFF
--- a/.github/workflows/build-and-push-to-gar.yml
+++ b/.github/workflows/build-and-push-to-gar.yml
@@ -1,18 +1,34 @@
 name: Push container image to MozCloud GAR
 on:
+  pull_request: {}
   push:
     branches: [ main ]
     tags: ['*']
 
 jobs:
   build-and-push:
-    secrets: inherit
     permissions:
-      packages: write
       contents: read
       id-token: write
-    uses: mozilla-it/deploy-actions/.github/workflows/build-and-push.yml@main
-    with:
-      image_name: fx-private-relay
-      gar_name: ${{ vars.GCP_MOZCLOUD_GAR_NAME_PROD }}
-      project_id: ${{ vars.GCP_MOZCLOUD_PROJECT_ID_PROD }}
+      packages: write
+    
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+          submodules: true
+      - name: Build and Tag Container Image
+        id: build
+        uses: mozilla-it/deploy-actions/docker-build@v4.4.0
+        with:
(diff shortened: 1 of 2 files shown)


groovecoder:
[There's a previous PR that removed the frontend build from the `Dockerfile`](https://github.com/jbuck/fx-private-relay/commit/b1d2d5b5d294990a6a294de7b96bb50994c78f39#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557).

I'm not sure how much, if any, of those steps we might still need or want. @vpremamozilla and/or @Vinnl: can you look into the differences here and help us understand which (if any) of the previous frontend build steps we still need? I know we've dropped gulp so we don't need that? But I'm not familiar with the rest of the next.js build stack.

@jbuck , @joeherm : We can we reconcile this `Dockerfile` build of the frontend with the `.circleci/config.yml` build of the frontend? It seems like we should use a single job to build the frontend so we know the CI tests are running against the exact same build that is used in the app runtime.

jbuck:
> `npm ci && npm run build` should be sufficient, so this looks good to me. What I don't get is why this only matters for dev, unless that's the only environment that's been transitioned to MozCloud?

That's correct, the container images being built by GitHub Actions are only being used in MozCloud atm

joeherm (changes requested):
I think I have a better understanding of what @groovecoder is saying now that I've looked more closely.

This change modifies the `Dockerfile` which _is_ used by CircleCI and presumably Jenkins (I lack repo access to clarify how those deploy scripts work) for both tests and various deploys. I think rather than make changes to an image we are actively using, we should either:

1. Create a new Dockerfile than we invoke and push
2. Leave this Dockerfile alone and set `DEBUG=False` in the MozCloud env vars so that `collectstatic` runs and the front end works

groovecoder:
>     1. Create a new Dockerfile than we invoke and push
> 
>     2. Leave this Dockerfile alone and set `DEBUG=False` in the MozCloud env vars so that `collectstatic` runs and the front end works

Yeah, I'd prefer one of these.

Another option is to create a shell script that runs the steps of the front-end build, and then have both GHA and CircleCI use the same shell script.

Just any technique so that we're always using the same commands to build the frontend whether that's part of the test run or the deploy.

jbuck:
This is about as far as I can take it - it's still missing the email tracking lists as well

jbuck:
Reopened as #5821

PR #5904, opened 2025-09-26
Switch from `runtime.txt` to `.python-version`

edmorley:
Since the `runtime.txt` file is deprecated:
https://devcenter.heroku.com/changelog-items/3141


joeherm:
Thanks for the contribution! We are moving our dev environment off of Heroku, so I'm going to close this.

PR #5946, opened 2025-10-09
fix(relay): Identify GCPv2 dev vs heroku (MPP-4425)

jwhitlock:
Currently:
* https://relay-dev.allizom.org (dev in GCPv2) has `RELAY_CHANNEL="local"`
* https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net (Heroku) has `RELAY_CHANNEL="dev"`

After this PR:
* https://relay-dev.allizom.org will have `RELAY_CHANNEL="dev"`
* https://dev.fxprivaterelay.nonprod.cloudops.mozgcp.net will have `RELAY_CHANNEL="heroku"`

### DIFF
--- a/emails/utils.py
+++ b/emails/utils.py
@@ -150,7 +150,7 @@ def get_email_domain_from_settings() -> str:
     email_network_locality = str(urlparse(settings.SITE_ORIGIN).netloc)
     # on dev server we need to add "mail" prefix
     # because we can’t publish MX records on Heroku
-    if settings.RELAY_CHANNEL == "dev":
+    if settings.RELAY_CHANNEL == "heroku":
         email_network_locality = f"mail.{email_network_locality}"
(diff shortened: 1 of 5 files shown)


emails/utils.py
```diff
@@ -150,7 +150,7 @@ def get_email_domain_from_settings() -> str:
     email_network_locality = str(urlparse(settings.SITE_ORIGIN).netloc)
     # on dev server we need to add "mail" prefix
     # because we can’t publish MX records on Heroku
-    if settings.RELAY_CHANNEL == "dev":
+    if settings.RELAY_CHANNEL == "heroku":
```
jwhitlock, line 153: This function handles the slightly different email behavior in Heroku. This function will be a lot simpler when we drop Heroku.

PR #5984, opened 2025-10-20
fix: Keep querystring when redirecting for slash (MPP-3448)

jwhitlock:
When [WhiteNoise](https://whitenoise.readthedocs.io/en/latest/) sees a URL like `/accounts/profile`, it recognizes it as a static file (`accounts/profile/index.html`), and redirects to `/accounts/profile/` with a trailing slash. This change adds the UTM parameters to the redirect URL. This will help when a campaign URL has the URL version without a slash.

## Testing

With this branch:

* `curl -I "http://127.0.0.1:8000/accounts/profile` has header `Location: profile/` (existing behavior)
* `curl -I "http://127.0.0.1:8000/accounts/profile?utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=nav&utm_content=products"` has header `Location: profile/?utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=nav&utm_content=products` (fixed behavior)
* `curl -I "http://127.0.0.1:8000/accounts/profile?utm_source=www.mozilla.org&other=123abc"` has header `Location: profile/?utm_source=www.mozilla.org` (unknown headers are not retained)

### DIFF
--- a/privaterelay/middleware.py
+++ b/privaterelay/middleware.py
@@ -4,6 +4,7 @@ import re
 import time
 from collections.abc import Callable
 from datetime import UTC, datetime
+from urllib.parse import urlencode, urlsplit, urlunsplit
 
 from django.conf import settings
 from django.http import HttpRequest, HttpResponse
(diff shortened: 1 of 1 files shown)


privaterelay/middleware.py
```diff
@@ -200,6 +201,44 @@ def is_staticfile(self, path_info: str) -> bool:
...
+            "utm_term",
+        )
+    )
+
+    @staticmethod
+    def serve(static_file, request):
+        """Preserve some query params when redirecting."""
+        resp = WhiteNoiseMiddleware.serve(static_file, request)
+        if (
+            resp.status_code == 302
```
groovecoder, line 219: question (non-blocking): Is `302` too specific? Should we make this anything in the 3xx range?

groovecoder (approved):
Code looks good; spot-checks work. 1 non-blocking question.

privaterelay/middleware.py
jwhitlock: I was already poking around the whitenoise code, so I knew where to look. The status code is always 302 Found:

https://github.com/evansd/whitenoise/blob/c53a05e8adf56ba7f6a78c2dec9b29f4748b8d17/src/whitenoise/responders.py#L235-L239

This is when index file serving was added to whitenoise, and the comment makes it clear they picked 302 instead of 301.

https://github.com/evansd/whitenoise/commit/15e80f6f779b2acbf6dfec036a1dddae01b2d914

So I think we're safe being explicit.

Future dev: if I'm proven wrong, add a comment and I own you lunch.

Vinnl: Calling it that Whitenoise updates it to a 307 in a future release :P

PR #5988, opened 2025-10-22
task: Add more context for FxA event issues (MPP-4460)

jwhitlock:
This PR adds more context to sentry exceptions around the FxA relying party endpoint and other FxA profile updates:

* Include the body and parsed JWT token for FxA relying party events. This will help with handling of new events and identifying problematic FxA accounts.
* Create Sentry issue for new FxA events. This will help us discover when FxA is sending new events.
* Include calling arguments for FxA profile updates. This will help in the case when a user requests a profile refresh, rather than the refresh being triggered by an FxA event
* When the FxA profile request returns a non-JSON response, create a sentry issue including the status code and body. This will help determine what we should do for these non-JSON responses.

### DIFF
--- a/privaterelay/views.py
+++ b/privaterelay/views.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import json
 import logging
 from collections.abc import Iterable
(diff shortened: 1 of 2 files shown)


privaterelay/views.py
```diff
@@ -127,8 +129,21 @@ def metrics_event(request: HttpRequest) -> JsonResponse:
...
+        return fxa_rp_events_inner(request, sentry_context)
+
+
+def fxa_rp_events_inner(
+    request: HttpRequest, sentry_context: dict[str, Any]
+) -> HttpResponse:
+    """Augment previous version of fxa_rp_events with additional sentry data"""
+
     req_jwt = _parse_jwt_from_request(request)
     authentic_jwt = _authenticate_fxa_jwt(req_jwt)
```
groovecoder: quesggestion (non-blocking): Should we move this code back "up" as the first steps of the the "outer" `fxa_rp_events` so we know we're only setting up sentry scope context for valid fxa rp events?

privaterelay/views.py
```diff
@@ -127,8 +129,21 @@ def metrics_event(request: HttpRequest) -> JsonResponse:
 
 @csrf_exempt
 def fxa_rp_events(request: HttpRequest) -> HttpResponse:
+    """MPP-4460: Track more data for FxA relying party exceptions"""
+    with sentry_sdk.new_scope() as scope:
```
groovecoder, line 143: praise: TIL `new_scope` and [Sentry scopes](https://docs.sentry.io/platforms/python/enriching-events/scopes/).

groovecoder (approved):
Code looks good. 1 non-blocking question/suggestion.

privaterelay/views.py
jwhitlock: There are no existing Sentry issues around "_authenticate_fxa_jwt", so that makes sense. There are issues around `_parse_jwt_from_request`, but that's because someone hits the endpoint w/o a auth header. Seems like an error we should handle.

PR #6009, opened 2025-10-27
fix: Check front-end files in manifest mode (MPP-3730)

jwhitlock:
When setting a next URL, django allauth checks if the URL is a known path (`is_safe_url`), to avoid security issues with redirecting a newly logged-in user. Because of our next.js / Django chimera, we need to check if the URL is served by the static file middleware. This middleware works slightly differently in development, where it looks in the `frontend/out` folder, and deployments, where it uses a build-time manifest (see https://relay.firefox.com/staticfiles.json).

This PR updates the `is_safe_url()` call to use `RelayStaticFilesMiddleware.is_staticfile`, which handles the differences between development and deployment.

This is needed for MPP-4348 to work correctly, carrying the UTM parameters to the next URL.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -530,10 +530,12 @@ jobs:
           name: Set test defaults
           command: cp .env-dist .env
       - run:
-          name: Create empty staticfiles directories
+          name: Create test staticfiles directories
           command: |
-            mkdir -p staticfiles
-            mkdir -p frontend/out
+            mkdir -p staticfiles/accounts/profile
+            mkdir -p frontend/out/accounts/profile
+            touch frontend/out/accounts/profile/index.html
+            touch staticfiles/accounts/profile/index.html
       # Section 2: Setup optional steps
       # Setup postgres, if selected
       # This is done later in the process to give the postgres image more time to boot up.


.circleci/config.yml
```diff
@@ -530,10 +530,12 @@ jobs:
...
       - run:
-          name: Create empty staticfiles directories
+          name: Create test staticfiles directories
           command: |
-            mkdir -p staticfiles
-            mkdir -p frontend/out
+            mkdir -p staticfiles/accounts/profile
+            mkdir -p frontend/out/accounts/profile
+            touch frontend/out/accounts/profile/index.html
+            touch staticfiles/accounts/profile/index.html
```
groovecoder, line 538: question/suggestion (non-blocking): Are these just for the tests? Is there any risk these could somehow make their way into the docker image on stage or prod and conflict with the real urls?

jwhitlock: It don't think it is possible for these to get into the image. This step is part of the `python job` section:

https://github.com/mozilla/fx-private-relay/blob/3b43cd02940c38b6fc5c5c7f7170018ca3033e96/.circleci/config.yml#L384-L396

This is used for quick python jobs that don't rely on the frontend, such as linters:

https://github.com/mozilla/fx-private-relay/blob/3b43cd02940c38b6fc5c5c7f7170018ca3033e96/.circleci/config.yml#L613-L622

It is also used for running tests against the next Django, with the old code versus the newly migrated DB, etc.:

https://github.com/mozilla/fx-private-relay/blob/3b43cd02940c38b6fc5c5c7f7170018ca3033e96/.circleci/config.yml#L663-L670

The image is built in a separate step, `build-test-backend`

https://github.com/mozilla/fx-private-relay/blob/3b43cd02940c38b6fc5c5c7f7170018ca3033e96/.circleci/config.yml#L230-L233

This one requires the frontend to build and test successfully first, and pushes the image to dockerhub when needed. We have similar commands in the Dockerfile to create the `staticfiles` folder:

https://github.com/mozilla/fx-private-relay/blob/3b43cd02940c38b6fc5c5c7f7170018ca3033e96/Dockerfile#L47

It's worth keeping in mind, so we don't copy the wrong code when moving to GitHub.

PR #6010, opened 2025-10-27
fix(deploy): Properly set defaults for dev deploy from push

joeherm:
This PR fixes an issue with `--dev` not being properly added to the container image tag when we push to `main`.
Screenshot 2025-10-27 at 3 51 27 PM

I misunderstood how the `workflow_call` option is used. It doesn't need to be part of the "Deploy to MozCloud" action, the `job.uses` will be enough to run the `build-and-push-to-gar`, and we can set the default in there for the case where we haven't manually specified the environment during a `workflow_dispatch`.

I ran a temporary test by having the logic check this branch instead of `main`. Tagging now works properly:
Screenshot 2025-10-27 at 4 17 45 PM

How to test:

### DIFF
--- a/.github/workflows/deploy-mozcloud.yml
+++ b/.github/workflows/deploy-mozcloud.yml
@@ -3,12 +3,6 @@ on:
   push:
     branches: [ '**' ]
     tags: ['*']
-  workflow_call:
-    inputs:
-      environment:
-        required: true
-        default: dev
-        type: string
   workflow_dispatch:
     inputs:
       environment:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-mozcloud.yml
```diff
@@ -29,4 +23,4 @@ jobs:
       github.ref == 'refs/heads/main' ||
       github.event_name == 'workflow_dispatch'
     with:
-      image_tag_metadata: ${{ inputs.environment }}
\ No newline at end of file
+      image_tag_metadata: ${{ inputs.environment || 'dev' }}
```
joeherm, line 26: Note that this is needed here, despite the `default` setting. This is because `default` is only honored by `workflow_dispatch`, not the `on.push`.

jbuck (approved):

PR #5883, opened 2025-09-15
fix alert #59: remove social account logging

groovecoder:
How to test:
`pytest` should cover it.

- [x] ~l10n changes have been submitted to the l10n repository, if any.~
- [x] ~I've added a unit test to test for potential regressions of this bug.~
- [x] ~I've added or updated relevant docs in the docs/ directory.~
- [x] ~All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol / Nebula colors where applicable (see `/frontend/src/styles/colors.scss`).~
- [x] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).

PR #6049, opened 2025-11-14
fix(relay): Create alternate bearer token auth for FxA (MPP-3505)

jwhitlock:
This is a recreation of PR #5272. It includes the removal of debug logging by PR #5883. I don't quite understand that code, but I also am far away enough from this code that I don't fully understand it anymore. If this proves to be unreviewable, I should start over.

The original PR statement:

This PR adds a new implementation of Bearer Token Authentication, for creating a new user or authorizing with a Mozilla account user logged into Firefox. By default, the existing authentication implementation is used. Setting the environment variable `FXA_TOKEN_AUTH_VERSION=2025` picks the new implementation.

While working on MPP-3505 (An `IntegrityError` while setting up a Relay account for a Mozilla account user through Firefox), I found one important bug in the existing implementation. The cache for the Accounts introspection response uses `hash(token)` as a cache key. This give a consistent number for a given Python instance, but a different number on a different Python instance, like a different pod or possibly a different gunicorn thread on the same pod. This means these responses were almost never cached.

I'm reluctant to fix just this issue, because we don't know what will happen if we start using the cache reliably.

The new implementation uses a cache key that will be consistent across pods. It also has some potential improvements.:

* Accounts introspection API responses or error are cached. In the existing implementation, 'write' requests like a `POST` call the API and do not cache the results.
* `/api/v1/terms-accepted-user` more consistently returns a `401` or `403` response. The Firefox integration is looking [for a 401 response](https://searchfox.org/mozilla-central/source/toolkit/components/satchel/integrations/FirefoxRelay.sys.mjs#154-157) or a [403 response](https://searchfox.org/mozilla-central/source/toolkit/components/satchel/integrations/FirefoxRelay.sys.mjs#154-157). The existing implementation returns `404` in some instances.
* The new code returns `502 Service Unavailable` when the upstream Accounts API is unavailable. This could be tuned if the Firefox integration handles this poorly.
* The new code tracks the time it takes to call the Account introspection API and profile API. Tracking these timers will allow us to determine when Accounts is having an issue that affects Relay.
* The new `FxaTokenAuthentication` is based on Django REST Framework's [TokenAuthentication](https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication), uses permissions classes to add permission checks beyond the token check, and returns token details in `request.auth`. This makes the authentication look more like other DRF authentication, and allows better code sharing between `/api/v1/terms-accepted-user` and other endpoints like `/api/v1/relayaddresses` used by the Firefox integration.

### DIFF
--- a/api/authentication.py
+++ b/api/authentication.py
@@ -16,21 +16,13 @@ import sentry_sdk
 from allauth.socialaccount.models import SocialAccount
 from codetiming import Timer
 from markus.utils import generate_tag
-from rest_framework.authentication import (
-    BaseAuthentication,
-    TokenAuthentication,
-    get_authorization_header,
-)
-from rest_framework.exceptions import (
-    APIException,
-    AuthenticationFailed,
-    NotFound,
-    ParseError,
-    PermissionDenied,
-)
+from rest_framework.authentication import BaseAuthentication, TokenAuthentication
+from rest_framework.exceptions import APIException, AuthenticationFailed
 from rest_framework.request import Request
 
 from emails.utils import histogram_if_enabled
+
+from .authentication_2025 import FxaTokenAuthentication as FxaTokenAuthentication2025
 
 logger = logging.getLogger("events")
 INTROSPECT_TOKEN_URL = "{}/introspect".format(
@@ -39,25 +31,25 @@ INTROSPECT_TOKEN_URL = "{}/introspect".format(
 
 # Specify the version strings in FXA_TOKEN_AUTH_VERSION
 #
-# The older version ("2024") works, but has a few issues.
+# The older version ("2025") works, but has a few issues.
 # The cache key changes between Python instances, so little or no cache hits are used.
 # Fetching a profile takes a few seconds, in which time another process can create a
 # SocialAccount, leading to IntegrityError. Some of these are tracked in MPP-3505.
 #
-# The newer version ("2025") addresses these issues, works more like a standard DRF
+# The newer version ("2026") addresses these issues, works more like a standard DRF
 # authentication class, expands the logged data, and tracks the time to call Accounts
 # introspection and profile APIs. However, it is unproven, so we're using an
 # environment variable to be able to try it in stage before production, and to
 # revert with a config change only.
 #
 # The names are designed to be annoying so they will be removed. The old code has
-# the suffix _2024 and the new code _2025 (when needed). When the new code is
+# the suffix _2025 and the new code _2026 (when needed). When the new code is
 # proven, the old code can be removed with minimal name changes.
 #
-# ruff thinks the strings "2024" and "2025" are passwords (check S105 / S106).
+# ruff thinks the strings "2025" and "2026" are passwords (check S105 / S106).
 # These constants allow telling ruff to ignore them once.
-FXA_TOKEN_AUTH_OLD_AND_PROVEN = "2024"  # noqa: S105
-FXA_TOKEN_AUTH_NEW_AND_BUSTED = "2025"  # noqa: S105
+FXA_TOKEN_AUTH_OLD_AND_PROVEN = "2025"  # noqa: S105
+FXA_TOKEN_AUTH_NEW_AND_BUSTED = "2026"  # noqa: S105
 
 
 class CachedFxaIntrospectResponse(TypedDict, total=False):
@@ -92,11 +84,6 @@ class FxaIntrospectCompleteData(TypedDict):
     active: bool
     sub: str
     exp: int
-
-
-def get_cache_key_2024(token):
-    """note: hash() returns different results in different Python processes."""
-    return hash(token)
 
 
 def get_cache_key(token: str) -> str:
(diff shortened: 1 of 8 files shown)


privaterelay/settings.py
```diff
@@ -638,6 +638,9 @@ def set_index_cache_control_headers(
     "https://api.accounts.firefox.com/v1",
 )
 FXA_SUPPORT_URL = config("FXA_SUPPORT_URL", f"{FXA_BASE_ORIGIN}/support/")
+FXA_TOKEN_AUTH_VERSION = config(
+    "FXA_TOKEN_AUTH_VERSION", "2024", cast=Choices(["2024", "2025"], cast=str)
```
groovecoder: praise: good call to use a setting so we can quickly change without a full app code redeploy.

api/authentication.py
```diff
@@ -1,33 +1,353 @@
...
+# Fetching a profile takes a few seconds, in which time another process can create a
+# SocialAccount, leading to IntegrityError. Some of these are tracked in MPP-3505.
+#
+# The newer version ("2025") addresses these issues, works more like a standard DRF
+# authentication class, expands the logged data, and tracks the time to call Accounts
+# introspection and profile APIs. However, it is unproven, so we're using an
+# environment variable to be able to try it in stage before production, and to
+# revert with a config change only.
+#
+# The names are designed to be annoying so they will be removed. The old code has
```
groovecoder, line 45: praise: 😆

api/authentication.py
```diff
@@ -1,33 +1,353 @@
...
+    error_args: list[str]
+
+
+class FxaIntrospectData(TypedDict, total=False):
+    """Keys seen in the JSON returned from a Mozilla Accounts introspection request"""
+
+    active: bool
+    sub: str
+    exp: int
+    error: str
```
groovecoder, line 71: question (blocking): Will this handle a case where Mozilla Accounts introspection request returns JSON that doesn't match this structure? (We were recently stung by a situation where Twilio changed an API call without notice.)

api/authentication.py
```diff
@@ -1,33 +1,353 @@
...
+    """
+
+    active: bool
+    sub: str
+    exp: NotRequired[int]
 
-def get_cache_key(token):
+
+def get_cache_key_2024(token):
+    """note: hash() returns different results in different Python processes."""
```
groovecoder: comment (non-blocking): 😠  at `hash()` !

api/authentication.py
```diff
@@ -1,33 +1,353 @@
...
+        self.data: FxaIntrospectCompleteData = cast(FxaIntrospectCompleteData, data)
+        self.from_cache = from_cache
+        self.request_s = None if request_s is None else round(request_s, 3)
+
+    def __repr__(self) -> str:
+        params = [repr(self.token), repr(self.data)]
+        if self.from_cache:
+            params.append(f"from_cache={self.from_cache!r}")
+        if self.request_s is not None:
+            params.append(f"request_s={self.request_s!r}")
```
groovecoder: quibble (non-blocking): I assume this is meant to help with debugging, likely while reading logs? 

If that's the case, should we include the `from_cache` and `request_s` values even if they are default values? When someone is reading the log messages, they may not know or remember what are the application code's default values.

api/authentication.py
```diff
@@ -1,33 +1,353 @@
...
+    def is_expired(self) -> bool:
+        return self.time_to_expire <= 0
+
+    @property
+    def fxa_id(self) -> str:
+        return self.data["sub"]
+
+
+INTROSPECT_ERROR = Literal[
+    "Timeout",  # Introspection API took too long to response
```
groovecoder: quibble (non-blocking):
```suggestion
    "Timeout",  # Introspection API took too long to respond
```

api/authentication.py
```diff
@@ -1,33 +1,353 @@
...
+        )
+    response = IntrospectionResponse(token, data=data_maybe, from_cache=True)
+    if response.is_expired:
+        return IntrospectionError(
+            token,
+            "TokenExpired",
+            error_args=[str(response.time_to_expire)],
+            data=data_maybe,
+            from_cache=True,
+        )
```
groovecoder, line 344: question (blocking just so I don't forget to check into this before we merge): In previous (2024?) code, I recall we tried to re-fetch a new un-expired token in the case where the current token is expired. Can we still do that? Or maybe we're already doing that in the Firefox code?

api/authentication.py
```diff
@@ -1,33 +1,353 @@
...
+        "TokenExpired": 401,
+    }
+
+    def raise_exception(self, method: str, path: str) -> NoReturn:
+        if not self.from_cache and self.error in self._log_failure:
+            logger.error(
+                "accounts_introspection_failed",
+                extra={
+                    "error": self.error,
+                    "error_args": [shlex.quote(arg) for arg in self.error_args],
```
groovecoder: suggestion (non-blocking): I think [`shlex.quote` expects a string](https://docs.python.org/3/library/shlex.html#shlex.quote), and this `self.error_args` value can come from `data = fxa_resp.json()` which could have non-string values? Should we be sure to cast `arg` into a string? 

```suggestion
                    "error_args": [shlex.quote(str(arg)) for arg in self.error_args],
```

api/authentication.py
```diff
@@ -111,7 +431,146 @@ def get_fxa_uid_from_oauth_token(token: str, use_cache: bool = True) -> str:
...
+
+def introspect_and_cache_token(
+    token: str, read_from_cache: bool = True
+) -> IntrospectionResponse | IntrospectionError:
+    """
+    Introspect a Mozilla account OAuth token, to get data like the FxA UID.
+
+    If anything goes wrong, raise an exception.
+    """
+    default_cache_timeout = 60
```
groovecoder: quesggestion (non-blocking): Should we make this a configurable value so we can tweak it without a full application code change?

api/views/privaterelay.py
```diff
@@ -410,3 +431,173 @@ def terms_accepted_user(request):
...
+        fxa_prof_or_resp, profile_time_s = _get_fxa_profile_from_bearer_token(token)
+        if isinstance(fxa_prof_or_resp, Response):
+            return fxa_prof_or_resp
+        fxa_profile = fxa_prof_or_resp
+
+        # Since this takes time, see if another request created the SocialAccount
+        try:
+            socialaccount = SocialAccount.objects.get(uid=fxa_uid, provider="fxa")
+            existing_sa = True
+            action = "created_during_profile_fetch"
```
groovecoder, line 382: praise: woohoo!

api/views/privaterelay.py
```diff
@@ -410,3 +431,173 @@ def terms_accepted_user(request):
...
+
+    error: None | Literal["timeout", "bad_response"] = None
+
+    try:
+        with Timer(logger=None) as profile_timer:
+            fxa_profile_resp = requests.get(
+                FXA_PROFILE_URL,
+                headers={"Authorization": f"Bearer {token}"},
+                timeout=settings.FXA_REQUESTS_TIMEOUT_SECONDS,
+            )
```
groovecoder, line 463: question (non-blocking): I'm wondering - should we try to cache this as well? Maybe not, since this seems to only be called while creating a new user, and we likely want to make sure we're doing a live check in that code path.

api/authentication.py
jwhitlock: In `introspect_token` later in this file, there is code to check the structure of the returned data. Current version is at:

https://github.com/mozilla/fx-private-relay/blob/a5dc5cbc88b3a2c0315dd0630de4885850f7f7ce/api/authentication.py#L434

The checks include:

* request timed out
* other error making request
* response was not JSON
* response was JSON but not an object
* response had a 401 Not Authorized status code
* response had a different status code that wasn't 200 OK
* `active` is not a bool or was False
* `sub` is not a string or is an empty string

It then passes it to `IntrospectResponse`, which has further checks that raise `ValueError`. The current version is at:

https://github.com/mozilla/fx-private-relay/blob/a5dc5cbc88b3a2c0315dd0630de4885850f7f7ce/api/authentication.py#L103

The checks include:

* `active` is not a bool or was False (repeats)
* `sub` is not a string or is an empty string (repeats)
* `exp` is present but not integer

Looking at that code, there may be issues if FxA changes `exp` to a float or a string representation of a float. An unhandled `ValueError` would be raised. I could add code to handle that case.

On slack, you clarified that if the API changes:

> Hrm, I prefer the opposite - let people log in if possible. And log a “handled” exception into Sentry to review.

In that case, I think only these would block authorization:

* 401 Not Authorized
* `active` is exactly `False`
* `sub` is not included, so we don't have an FxA ID

That will require some changes.

api/views/privaterelay.py
jwhitlock: This code is mirroring the django-allauth `fxa` provider. When a user creates an account from our dashboard, that code calls the FxA profile endpoint, and adds the data to the `SocialAccount`:

https://github.com/pennersr/django-allauth/blob/8a91a09d5c4543d420e05165b49ae223862b43e2/allauth/socialaccount/providers/fxa/views.py#L17-L23

So, the data is cached in the database if it succeeds. A double cache is not needed, and is more likely in my opinion to make any issues worse and harder to debug.

api/authentication.py
jwhitlock: My understanding is that the Firefox bearer token is more privileged than the Relay oauth tokens, so I doubt we can request a token refresh from our code.

This is new code. The old code had this note:

https://github.com/mozilla/fx-private-relay/blob/6a232dff2259a6a3d7969eb271adbea3ae50cf48/api/authentication.py#L98-L99

I believe the desired effect was to consider a token valid up to 60 seconds after the expiration time. However, the result was to consider very old cached tokens to still be valid. This exposes Relay to token replay attacks, where a man-in-the-middle attacker uses expired Firefox tokens against Relay.

I'm OK with a 60 second grace period for clock skew, but not the infinite grace period of the old code.

api/authentication.py
jwhitlock: The should be strings, based on the type hints:

https://github.com/mozilla/fx-private-relay/blob/d7cfa26b6a59c693bffc22e44eb4df51003db224/api/authentication.py#L204

This required some type conversion:

https://github.com/mozilla/fx-private-relay/blob/d7cfa26b6a59c693bffc22e44eb4df51003db224/api/authentication.py#L343

So, `arg` should always be a string. However, runtime can still defeat static type checking. I'm going to enforce this in the `__init__` instead, with:

```python
        self.error_args = [str(arg) for arg in error_args or []]
```

api/authentication.py
```diff
@@ -1,33 +1,353 @@
...
+                "accounts_introspection_failed",
+                extra={
+                    "error": self.error,
+                    "error_args": [shlex.quote(arg) for arg in self.error_args],
+                    "status_code": self.status_code,
+                    "data": self.data,
+                    "method": method,
+                    "path": path,
+                    "introspection_time_s": self.request_s,
+                },
```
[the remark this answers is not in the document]
jwhitlock: I'm OK with this, because we restrict log access and it is data we need to debug issues. If others disagree, we could 1) log as base64 so it is not clear text, or 2) omit until we need it for debugging, or 3) encrypt with key available to maintainers.

api/authentication.py
jwhitlock: It was not configurable before, but config is cheap so I can add it

api/authentication.py
jwhitlock: I went to change the code, and was unable to write a test to exercise it.

I still think it is unnecessary, but let's go with your version.

api/authentication.py
groovecoder: Okay yeah that sounds good. Thanks for checking into that!

api/views/terms_accepted_user_2025.py
groovecoder, line 1: question (non-blocking): I assume this code is the unmodified current code?

api/authentication.py
groovecoder: question (blocking): So if the most likely failure state seems to be FxA changes `exp` to a string, then an unhandled `ValueError` is raised. Is it as simple as updating the type declaration to `exp: int | str | float` for that case?

And then if I remember python type enforcement - these types are enforced in our code by `mypy`, but not at runtime? So if FxA changes other values in a way that doesn't match this type declaration, it won't break the code at runtime?

api/views/terms_accepted_user_2025.py
jwhitlock: yes

api/authentication.py
jwhitlock: I thought I understood these requested changes, but I'm more confused now than I was. Maybe we should talk about this over video or in person?

I think it is really unlikely that the FxA response to [/v1/introspect](https://mozilla.github.io/ecosystem-platform/api#tag/OAuth-Server-API-Overview/operation/postIntrospect), even from an int to a float. Why would it change to a float? Is it now in seconds instead of microseconds? I have no idea how to handle an expiration that changes to "it's fine bro". I think it would make the code worse to do anything but fail authentication if the authentication token changes formats. If you think failing authentication means breaking the code, then we're far from agreement.

Yes, static type enforcement does not check at runtime. That's why `introspect_token` carefully examines the response from FxA before returning an `FxaIntrospectData`. `mypy` gives some assurances we correctly went from an unknown response from FxA to this firmer type definition. It offers a sharper line between "I think I know what this is" and "I'm sure what this is, for the parts that matter to me".

PR #6052, opened 2025-11-17
fix(auth): enforce relay scope on bearer token auth

joeherm:
This PR fixes MPP-4437.

Changes:
- Enforce Relay scope being needed for Bearer token auth.
- Add Relay scope to the default scopes that the application requests. Even though a request coming from the web application will ultimately use session authentication, we should still generate a token that _could_ be used for Bearer auth.

How to test:

### DIFF
--- a/api/authentication.py
+++ b/api/authentication.py
@@ -95,6 +95,12 @@ def get_fxa_uid_from_oauth_token(token: str, use_cache: bool = True) -> str:
         raise NotFound("FXA did not return an FXA UID.")
     fxa_uid = str(raw_fxa_uid)
 
+    scopes = fxa_resp_data.get("json", {}).get("scope", "").split()
+    if settings.RELAY_SCOPE not in scopes:
+        raise AuthenticationFailed(
+            "FXA token is missing scope: https://identity.mozilla.com/apps/relay."
+        )
+
     # cache valid access_token and fxa_resp_data until access_token expiration
     # TODO: revisit this since the token can expire before its time
     if isinstance(fxa_resp_data.get("json", {}).get("exp"), int):
--- a/api/tests/authentication_tests.py
+++ b/api/tests/authentication_tests.py
@@ -31,6 +31,7 @@ class FxaResponse(TypedDict, total=False):
     sub: str
     exp: int
     error: str
+    scope: str
 
 
 class CachedFxaResponse(TypedDict):
@@ -48,7 +49,10 @@ def _setup_fxa_response(
         json=json,
     )
     if json is None:
-        return {"status_code": status_code}
+        return {
+            "status_code": status_code,
+            "json": {"scope": "https://identity.mozilla.com/apps/relay"},
+        }
     return {"status_code": status_code, "json": json}
 
 
@@ -85,6 +89,7 @@ class AuthenticationMiscellaneous(TestCase):
             "active": True,
             "sub": self.uid,
             "exp": exp_time,
+            "scope": "https://identity.mozilla.com/apps/relay",
         }
         status_code = 200
         expected_fxa_resp_data = {"status_code": status_code, "json": json_data}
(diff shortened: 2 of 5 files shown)


api/authentication.py
```diff
@@ -95,6 +95,13 @@ def get_fxa_uid_from_oauth_token(token: str, use_cache: bool = True) -> str:
         raise NotFound("FXA did not return an FXA UID.")
     fxa_uid = str(raw_fxa_uid)
 
+    if "https://identity.mozilla.com/apps/relay" not in fxa_resp_data["json"].get(
+        "scope", ""
```
jwhitlock: Nit: I would write as:

```suggestion
    scopes = fxa_resp_data.get("json", {}).get("scope", "").split()
    if "https://identity.mozilla.com/apps/relay" not in scopes:
```

Note-to-self: I expected `scope` to be a list, but nope, it is a space-separated string:

https://mozilla.github.io/ecosystem-platform/api#tag/OAuth-Server-API-Overview/operation/postIntrospect

and that the format comes from the spec:

https://mozilla.github.io/ecosystem-platform/reference/oauth-details#oauth-scopes

api/tests/authentication_tests.py
```diff
@@ -48,7 +49,10 @@ def _setup_fxa_response(
         json=json,
     )
     if json is None:
-        return {"status_code": status_code}
+        return {
+            "status_code": status_code,
+            "json": {"scope": "https://identity.mozilla.com/apps/relay"},
```
jwhitlock, line 54: Nit: make a constant `RELAY_SCOPE = "https://identity.mozilla.com/apps/relay"` and use it everywhere in this file. It could be defined `authentication.py` and used there as well.

privaterelay/settings.py
```diff
@@ -607,7 +607,11 @@ def set_index_cache_control_headers(
 SOCIALACCOUNT_PROVIDERS = {
     "fxa": {
         # Note: to request "profile" scope, must be a trusted Mozilla client
-        "SCOPE": ["profile", "https://identity.mozilla.com/account/subscriptions"],
+        "SCOPE": [
+            "profile",
+            "https://identity.mozilla.com/account/subscriptions",
+            "https://identity.mozilla.com/apps/relay",
```
jwhitlock: Note to self: In my dev environment, I was able to create a new user with this scope against the stage server. When I logged in with an existing user, the user got the `../apps/relay` scope as well.

api/tests/authentication_tests.py
```diff
@@ -213,13 +224,63 @@ def test_get_fxa_uid_from_oauth_token_not_active_uses_cached_response_returns_er
...
+
+        assert cache.get(cache_key) is None
+
+        # get fxa response with no fxa uid for the first time
+        try:
+            get_fxa_uid_from_oauth_token(missing_scopes_token)
+        except AuthenticationFailed as e:
+            assert (
+                str(e.detail)
+                == "FXA missing is missing scope: https://identity.mozilla.com/apps/relay."
```
groovecoder: suggestion (non-blocking):

```suggestion
                == "FXA token is missing scope: https://identity.mozilla.com/apps/relay."
```

api/tests/authentication_tests.py
```diff
@@ -213,13 +224,63 @@ def test_get_fxa_uid_from_oauth_token_not_active_uses_cached_response_returns_er
...
+            )
+            assert responses.assert_call_count(self.fxa_verify_path, 1) is True
+            return
+        self.fail("Should have raised AuthenticationFailed")
```
groovecoder, line 268: question (non-blocking): do you want to test the caching behavior here too? Should we let the other tests test the cache and keep this test focused on the scope error? Or is there something about the scope error that we should test the cache here too?

api/authentication.py
```diff
@@ -95,6 +95,13 @@ def get_fxa_uid_from_oauth_token(token: str, use_cache: bool = True) -> str:
...
+    ):
+        raise AuthenticationFailed(
+            "FXA missing is missing scope: https://identity.mozilla.com/apps/relay."
```
groovecoder: suggestion (non-blocking):

```suggestion
            "FXA token is missing scope: https://identity.mozilla.com/apps/relay."
```

groovecoder (approved):

api/tests/authentication_tests.py
joeherm: I was just following a pattern where most of the other tests in this file assert that the cache is working. I'll remove.

api/tests/authentication_tests.py
joeherm: I moved it to `settings.py` to avoid circular imports

PR #6055, opened 2025-11-19
Use reusable workflow for Mozcloud build and push

jwhitlock:
This reopens PR #6054 as a local branch.

### DIFF
--- a/.github/workflows/build-and-push-to-gar.yml
+++ b/.github/workflows/build-and-push-to-gar.yml
@@ -47,3 +47,4 @@ jobs:
         GIT_BRANCH=${{ needs.determine-metadata.outputs.git_branch }}
         GIT_SHA=${{ github.sha }}
         GIT_TAG=${{ needs.determine-metadata.outputs.git_tag }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-push-to-gar.yml
```diff
@@ -9,20 +9,13 @@ on:
...
-
+    outputs:
+      git_branch: ${{ steps.git_env_vars.outputs.GIT_BRANCH }}
+      git_tag: ${{ steps.git_env_vars.outputs.GIT_TAG }}
     steps:
-      - name: Checkout Repository
-        uses: actions/checkout@v5
-        with:
-          persist-credentials: false
-          submodules: true
```
grahamalama, line 25: Ahh, we'll either need to enable this as an option on the reusable workflow or fetch submodules from within the workflow in the `prebuild_script` input. 

https://github.com/mozilla-it/deploy-actions/blob/a34e7a868025b983d7c948354ca3d1e887ed778d/.github/workflows/build-and-push.yml#L73-L78

PR #6093, opened 2025-12-01
Revert "fix(auth): enforce relay scope on bearer token auth"

joeherm:
This reverts commit 7bda4c0ad096e324d3507f4773915ddcc17bbb4a, a change to scope which is causing issues in staging QA.


privaterelay/settings.py
```diff
@@ -611,11 +607,7 @@ def set_index_cache_control_headers(
 SOCIALACCOUNT_PROVIDERS = {
     "fxa": {
         # Note: to request "profile" scope, must be a trusted Mozilla client
-        "SCOPE": [
-            "profile",
-            "https://identity.mozilla.com/account/subscriptions",
-            RELAY_SCOPE,
-        ],
+        "SCOPE": ["profile", "https://identity.mozilla.com/account/subscriptions"],
```
jwhitlock, line 610: Can we try just reverting this line?

PR #6103, opened 2025-12-03
fix(relay): Update to next 15.5.7, glob 10.5.0

jwhitlock:
Update two packages with `npm audit fix`:

* next 15.5.6 to 15.5.7
* glob from 10.4.5 to 10.5.0


frontend/package.json
Tibap, line 29: Can we update this to 19.2.1 as well?

jwhitlock: We're hitting some complications updating react, so we're going to merge this one and then work on the react update.

PR #6183, opened 2026-01-07
fix MPP-4464: chore(flag): remove custom_domain_management_redesign flag

groovecoder:
This PR fixes MPP-4464.

How to test:
1. You can still register a subdomain with the new design
2. When you generate a new mask, it auto-collapses other mask cards
3. You can still click copy icon to copy a mask


emails/validators.py
```diff
@@ -78,14 +76,9 @@ def valid_address(address: str, domain: str, subdomain: str | None = None) -> bo
 
     address_pattern_valid = valid_address_pattern(address)
     address_contains_badword = has_bad_words(address)
-    address_already_deleted = 0
-    # TODO MPP-4464: Remove custom_domain_management_redesign flag, assume on
-    if not subdomain or flag_is_active_in_task(
```
groovecoder, line 83: note: because the `flag_is_active_in_task` condition was always `True`, this `if` check is no longer needed.

jwhitlock: 🤦 yeah that's the way boolean logic works, I'm embarrassed I flagged this last time.

groovecoder: no worries I confused myself about it too! it wasn't until I looked at it again that I realized the same thing.

PR #6192, opened 2026-01-13
fix(tests): Add create_fxa_introspect_data helper function (MPP-3505)

jwhitlock:
The new `create_fxa_introspect_data` function returns a random, active FxA user by default. It has optional parameters for testing, such as setting `sub` (the FxA ID) or setting `exp` (expiration time) to the past. This eliminates duplicate setup code, lowering the new line count.

There are no app code changes, and the tests should be testing the same behavior.

Other changes:

* Remove `setUp` from `AuthenticationMiscellaneous`. The values appear to be copied from `FxaTokenAuthenticationTest`, and most unused. `self.uid` is used in one test, `test_get_fxa_uid_from_oauth_token_returns_cached_response`, and moved to there. `self.fxa_verify_path` is an unneeded alias for `INTROSPECT_TOKEN_URL` with the same number of characters.
* Remove `self.fxa_verify_path` from `FxaTokenAuthenticationTest`, for the same reasons.
* Avoid a long string compare in `test_get_fxa_uid_from_oauth_token_has_wrong_scope_returns_error_response`
* Fix docstring in `api/tests/privaterelay_views_tests.py`

### DIFF
--- a/api/tests/privaterelay_views_tests.py
+++ b/api/tests/privaterelay_views_tests.py
@@ -1,6 +1,4 @@
-"""Tests for api/views/email_views.py"""
-
-from datetime import datetime
+"""Tests for api/views/privaterelay_views.py"""
 
 from django.conf import LazySettings
 from django.contrib.auth.models import User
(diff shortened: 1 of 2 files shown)


api/tests/privaterelay_views_tests.py
```diff
@@ -1,6 +1,4 @@
-"""Tests for api/views/email_views.py"""
-
-from datetime import datetime
+"""Tests for api/views/privaterelay_views.py"""
```
groovecoder, line 1: 🦅 👁️

PR #6193, opened 2026-01-14
Switch from Webpack to Turbopack

Vinnl:
This PR fixes MPP-4535.

Turbopack is the default in Next.js now, so it's probably safest to align with that.

Some changes I had to do:
- Set up Turbopack to load FTL files as raw strings.
- Add the raw-loader dependency to do so.
- Compensate for Turbopack (at the moment) loading those strings on the `default` export property for FTL modules, rather than having the complete module be a string.
- Remove the MSW workaround — I no longer received build errors, so that no longer appears to be necessary. This seems to align with the current package.json of msw, which indeed now no longer sets any exports explicitly to `null`.
- Switch to the function-style syntax for :global, since Turbopack always uses LightningCSS, which only supports that.
- Hardcode the breakpoints we use, since Turbopack/LightningCSS won't support exporting variables from CSS: https://github.com/vercel/next.js/issues/88544
- Next.js itself updated next-env.d.ts

How to test: no behaviour should have changed.

- [ ] I've added a unit test to test for potential regressions of this bug. - N/A, existing e2e tests should cover this.


frontend/src/functions/getL10n.ts
```diff
@@ -23,10 +23,19 @@ export function getL10n(options: { deterministicLocales: boolean }) {
     // Filenames are formatted as `./<locale>/<module>.ftl`.
     // Example: ./en/bundle.ftl
     const locale = fileName.split("/")[1];
+    const sourceModule = translationsContext(fileName);
+    // Turbopack's behaviour is currently different from Webpack's behaviour,
+    // in that it returns an object with the file contents on a `default` property.
+    // This is considered a bug, so they might restore the Webpack behaviour
+    // in the future, so best to be compatible with both.
+    // See https://github.com/vercel/next.const loadedSource =
```
groovecoder: change (non-blocking): this url doesn't work for me?

Vinnl: Whoops, must've accidentally hit `p` here. I already missed that `const loadedSource` elsewhere :sweat_smile: 

About removing Webpack completely: what do you mean by that? It's not an explicit direct dependency, and I didn't find any particular references to it in the codebase.

groovecoder: the `webpack` section is still in `next.config.ts`

Vinnl: Ah whoops, forgot that! Removed now.

PR #6196, opened 2026-01-14
fix(tests): More authentication test changes (MPP-3505)

jwhitlock:
This makes further changes to the authentication tests to prepare for code changes for MPP-3505.

The `AuthenticationMiscellaneous` class is split into `IntrospectTokenTests` and `GetFxaUidFromOauthTests` classes. The `IntrospectTokenTests` class tests `introspect_token()`, which doesn't use the cache, so the `tearDown` that clears the cache can be removed. The `GetFxaUidFromOauthTests` class tests `get_fxa_uid_from_oauth()`. Since the function name is in the class name, the test function strings don't need them, and can be shortened and standardized. The ruff rule [E501](https://docs.astral.sh/ruff/rules/line-too-long/) does not need to be disabled for the shorter names.

Tests for expected exceptions now use [assertRaisesRegex](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex), which tests the expected exception with the expected message is raised.

Tests are added for a timeout on the FxA introspect call, and to check the cache timeout value for different values of `exp` in the payload. The two functions are now fully covered.

### DIFF
--- a/api/tests/authentication_tests.py
+++ b/api/tests/authentication_tests.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 from typing import Any, Required, TypedDict
+from unittest.mock import patch
 from uuid import uuid4
 
 from django.conf import settings
@@ -9,6 +10,7 @@ from django.test import RequestFactory, TestCase
 import responses
 from allauth.socialaccount.models import SocialAccount
 from model_bakery import baker
+from requests.exceptions import Timeout
 from rest_framework.exceptions import APIException, AuthenticationFailed, NotFound
 from rest_framework.test import APIClient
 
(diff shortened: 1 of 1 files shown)


api/tests/authentication_tests.py
```diff
@@ -91,39 +93,46 @@ def setup_fxa_introspection_failure(
...
 
-        try:
-            introspect_token(invalid_token)
-        except AuthenticationFailed as e:
-            assert str(e.detail) == "JSONDecodeError from FXA introspect response"
-            assert responses.assert_call_count(INTROSPECT_TOKEN_URL, 1) is True
-            return
-        self.fail("Should have raised AuthenticationFailed")
+        expected_err = "JSONDecodeError from FXA introspect response"
+        with self.assertRaisesRegex(AuthenticationFailed, expected_err):
```
groovecoder, line 104: note (non-blocking): this reminded me that I made [a change to the Relay A-S component to return a machine-readable error code so that mobile clients could better detect errors and adjust the mobile UI or behavior for them](https://github.com/mozilla/application-services/commit/d4d571d3ea39430c99047f5b84e8aff0b570b220).

[`relay_exception_handler`](https://github.com/mozilla/fx-private-relay/blob/f5d85e597992e8a998eb6fb9d84f4ccde1ee1150/api/views/__init__.py#L22) ONLY adds `error_code` (defined by `default_code` property of children of `RelayAPIException` to responses when the exception is a `RelayAPIException`. That means these `rest_framework.exceptions` will come across to the mobile clients as just generic errors. Seems fine for now, because I don't think the mobile clients can do anything different between these errors?

At some point, we could maybe change some of these to more specific errors if mobile (or other) clients want to adjust their behavior based on the type of error.

PR #6206, opened 2026-01-20
fix(tests): Update TermsAcceptedUserViewTest, reproduce bugs (MPP-3505, MPP-3928)

jwhitlock:
This PR refactors the test case `TermsAcceptedUserViewTest`, and adds tests to reproduce some unhandled exceptions (MPP-3505 and MPP-3928).

The changes:

* Switch from `TestCase` to `APITestCase`, so `self.client` will be the API test client
* Move constants from `setUp` to the class, and re-use the same token
* Fix test names that say the cache is tested, but don't.
* Add test to demonstrate that a user created during a slow profile fetch is not an issue. I once suspected this was the cause of MPP-3505.
* Add test to show a timeout during a profile fetch is not caught. We might as well catch that too.
* Add tests that create a user during the `django-allauth` `complete_social_app` call in two ways. If the user is added before the `process_auto_signup` call, then a `NoReverseMatch` is raised (MPP-3928). If the user is added after the call, then a `IntegrityError` is raised.

### DIFF
--- a/api/tests/privaterelay_views_tests.py
+++ b/api/tests/privaterelay_views_tests.py
@@ -1,16 +1,23 @@
 """Tests for api/views/privaterelay_views.py"""
 
+import json
+from unittest.mock import patch
+
 from django.conf import LazySettings
 from django.contrib.auth.models import User
 from django.core.cache import cache
-from django.test import RequestFactory, TestCase
+from django.db import IntegrityError
+from django.http import HttpRequest
 from django.test.client import Client
-from django.urls import reverse
+from django.urls import NoReverseMatch, reverse
 
 import pytest
 import responses
-from allauth.socialaccount.models import SocialAccount
-from rest_framework.test import APIClient
+from allauth.socialaccount.internal.flows.signup import process_auto_signup
+from allauth.socialaccount.models import SocialAccount, SocialLogin
+from requests import PreparedRequest
+from requests.exceptions import Timeout
+from rest_framework.test import APIClient, APITestCase
 
 from api.authentication import INTROSPECT_TOKEN_URL, get_cache_key
 from api.tests.authentication_tests import (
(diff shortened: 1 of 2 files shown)


api/tests/privaterelay_views_tests.py
```diff
@@ -363,95 +359,209 @@ def test_no_token_returns_400(self) -> None:
...
-        assert responses.assert_call_count(self.fxa_verify_path, 1) is True
+        assert responses.assert_call_count(INTROSPECT_TOKEN_URL, 1) is True
+
+    @responses.activate
+    def test_account_created_during_profile_request_returns_201(self) -> None:
+        """
+        If the SocialAccount is created during the profile request, returns created.
+
+        The django-allauth flow handles an existing SocialAccount, even if our code
+        thinks it created it.
```
groovecoder: quibble (non-blocking): hard to understand what each "it" means in this context.

api/tests/privaterelay_views_tests.py
```diff
@@ -363,95 +359,209 @@ def test_no_token_returns_400(self) -> None:
...
+            request: PreparedRequest,
+        ) -> tuple[int, dict[str, str], str]:
+            """Create a duplicate SocialAccount during the profile request"""
+            user = User.objects.create(email=self.email)
+            user.profile.created_by = "mocked_function"
+            user.profile.save()
+            SocialAccount.objects.create(provider="fxa", uid=self.uid, user=user)
+            return (200, {}, json.dumps(self.fxa_profile_data))
+
+        responses.add_callback(
```
groovecoder, line 464: praise: cool technique to create the object during the FXA request.

api/tests/privaterelay_views_tests.py
```diff
@@ -363,95 +359,209 @@ def test_no_token_returns_400(self) -> None:
...
+        assert profile.created_by == "firefox_resource"  # Overwritten by our code
+
+    @pytest.mark.xfail(raises=IntegrityError, reason="MPP-3505")
+    @responses.activate
+    def test_account_created_after_user_check_returns_500(self) -> None:
+        """
+        Bug: Parallel creation of SocialAccount can raise IntegrityError
+
+        If the user is created after the django-allauth `process_auto_signup` check,
+        then a duplicate user is created with the same email, a duplicate profile,
```
groovecoder, line 493: question (non-blocking): oh, so only the `SocialAccount` throws the `IntegrityError`!? i.e., we have duplicate user and profile records?

groovecoder (approved):

api/tests/privaterelay_views_tests.py
jwhitlock: Yes, Django's `User` allow duplicates by email, and `Profile` is bound to the `User`, so we get dupes when we hit this situation. I'm planning to wrap the database writing in a transaction so those dupes get rolled back as well.

groovecoder: dang, I didn't realize that. that could be causing some other bugs (or maybe these same bugs) if someone deletes their relay account and/or revokes it from their fxa.

PR #6209, opened 2026-01-22
Update e2e tests to use `data-testid` (MPP-4545)

jwhitlock:
Update the frontend code to emit `data-testid` IDs and the E2E tests to use `page.getByTestId()` instead of CSS class selectors. 

The `webpack` bundler creates class names like `MaskCard_copy-button-wrapper_WocUfa`, so the e2e tests looked for prefixes like `MaskCard_copy-button-wrapper`. The `turbopack` bundler, added in PR #6191, creates class names like `MaskCard-module-scss-module__WocUfa__copy-button-wrapper`, so we'd need to match both the beginning and the end of the class name. Instead, I'm using the Playwright solution of [locating by test id](https://playwright.dev/docs/locators#locate-by-test-id). This requires changing the layout code to include that `data-testid` attribute.

Other changes:

* Disable the [react/react-in-jsx-scope](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md), rather than adding an import. I suspect we need to adjust our eslint rules to match the current guidance, but that seems outside of this scope.
* Update the `useEffect` for `props.isModalOpen` in `Layout.tsx` to wrap in a `setTimeout`. eslint raised an error pointing to https://react.dev/learn/you-might-not-need-an-effect, this cleared it, but might be wrong
* Remove a lot of unused locators from `dashboardPage.ts`, instead of converting them
  * `userMenuPopUp`
  * `userMenuLetter`
  * `userMenuPopEmail`
  * `userMenuButton`
  * `signOutToastAlert`
  * `bottomUpgradeBanner`
  * `dashBoardWithoutMasksEmail`
  * `emailsForwardedAmount`
  * `emailsBlockedAmount`
  * `maskCardExpandButton`
  * `maskCardHeader`
  * `maskCardRepliesAmount`
  * `maskCardBlockedAmount`
  * `dashboardPageWithoutHeader`

## Testing

I ran the e2e suite locally with:

```
E2E_TEST_ENV=local SITE_ORIGIN='http://127.0.0.1:8000/' npx playwright test --project='full-*' --ui --debug
```

However, not everything was testable, and I have a personal AWS account setup so that emails are processed locally. I think this needs to be pushed to dev to be validated.

If this isn't acceptable, I could open a new PR with just the frontend changes, and then we may be able to validate the E2E changes with:

```
E2E_TEST_ENV=dev npx playwright test --project='full-*' --ui --debug
```

### DIFF
--- a/frontend/src/components/layout/Layout.tsx
+++ b/frontend/src/components/layout/Layout.tsx
@@ -152,9 +152,11 @@ export const Layout = (props: Props) => {
       return;
     }
     // When the modal is open, pointer events is disabled for the layout wrapper.
-    setTimeout(() => {
-      setPointerEventsNone(true);
-    }, 0);
+    // TODO MPP-4549: See if useEffect can be removed or altered to avoid re-rendering
+    // without reverting MPP-3701 fix.
+    // https://react.dev/learn/you-might-not-need-an-effect
+    // eslint-disable-next-line react-hooks/set-state-in-effect
+    setPointerEventsNone(true);
   }, [props.isModalOpen]);
 
   return (


e2e-tests/pages/dashboardPage.ts
groovecoder, line 1: praise: woohoo removing so many unused locators!

frontend/.eslintrc.js
```diff
@@ -72,6 +72,7 @@ module.exports = {
         ],
       },
     ],
+    "react/react-in-jsx-scope": "off",
```
groovecoder, line 75: suggestion (non-blocking): dang, [I "fixed" these eslint errors the dumb way by starting to restore `import React from "react";` lines at the top of all of our `.tsx` files](https://github.com/mozilla/fx-private-relay/pull/6148/changes/b9a2912f90c338ee41869b3b6fce7ba86892578c). And then I had the audacity to [suggest that everyone would need to "fix" this for all `.tsx` files we changed going forward](https://mozilla.slack.com/archives/C013CSYEL5T/p1767818864305799). But your approach here [looks like an accepted way to fix this in React 17+ apps](https://kinsta.com/blog/react-must-be-in-scope-when-using-jsx/#2-update-eslint-configuration-fix-for-react-v17-and-higher).

Can you (file a follow-up ticket to) clean out the `import`s that are no longer needed with this eslint config?

jwhitlock: I'm not sure who added the eslint config (@Vinnl?) but it is way off the current recommendations from next.js (https://nextjs.org/docs/pages/api-reference/config/eslint). If I wrote the ticket, it would be "rewrite eslint config and ensure it runs in CI"

frontend/src/components/layout/Layout.tsx
```diff
@@ -152,7 +152,9 @@ export const Layout = (props: Props) => {
       return;
     }
     // When the modal is open, pointer events is disabled for the layout wrapper.
-    setPointerEventsNone(true);
+    setTimeout(() => {
```
vpremamozilla: why is this `setTimeout` added? it might change the modal and pointer behavior if you add this

frontend/.eslintrc.js
jwhitlock: Opened https://mozilla-hub.atlassian.net/browse/MPP-4548

frontend/src/components/layout/Layout.tsx
jwhitlock: Here's the error I get on that line:

```
✖ eslint --fix --config frontend/.eslintrc.js:

/Users/john/src/fx-private-relay/frontend/src/components/layout/Layout.tsx
  155:5  error  Error: Calling setState synchronously within an effect can trigger cascading renders

Effects are intended to synchronize state between React and external systems such as manually updating the DOM, state management libraries, or other platform APIs. In general, the body of an effect should do one or both of the following:
* Update external systems with the latest state from React.
* Subscribe for updates from some external system, calling setState in a callback function when external state changes.

Calling setState synchronously within an effect body causes cascading renders that can hurt performance, and is not recommended. (https://react.dev/learn/you-might-not-need-an-effect).

/Users/john/src/fx-private-relay/frontend/src/components/layout/Layout.tsx:155:5
  153 |     }
  154 |     // When the modal is open, pointer events is disabled for the layout wrapper.
> 155 |     setPointerEventsNone(true);
      |     ^^^^^^^^^^^^^^^^^^^^ Avoid calling setState() directly within an effect
  156 |   }, [props.isModalOpen]);
  157 |
  158 |   return (  react-hooks/set-state-in-effect

✖ 1 problem (1 error, 0 warnings)
```

Wrapping in `setTimeout` made the warning go away. I could try disabling that warning for the one line. Do you have other suggestions for a real fix? https://react.dev/learn/you-might-not-need-an-effect is a stretch for my React level.

frontend/src/components/layout/Layout.tsx
vpremamozilla, line 140: claude recommended something like this instead: 

```suggestion
     const [isDeferredClose, setIsDeferredClose] = useState(false);

  useEffect(() => {
    if (props.isModalOpen === undefined) {
      return;
    }
    if (!props.isModalOpen) {
      setTimeout(() => {
        setIsDeferredClose(false);
      }, 0);
    } else {
      setIsDeferredClose(false);
    }
  }, [props.isModalOpen]);

  const pointerEventsNone = props.isModalOpen || isDeferredClose;
```

which i think could work! bc the setTimeout is going to cause issues with clicking the modal

frontend/src/components/layout/Layout.tsx
vpremamozilla: OR you could disable the esLint error altogether ...
```suggestion
    // eslint-disable-next-line react-hooks/set-state-in-effect
    setPointerEventsNone(true);
```

jwhitlock: Thanks, I think this is the right choice, since changing this code is out-of-scope for fixing the E2E tests. I opened MPP-4549 for trying fixes later.

frontend/.eslintrc.js
Vinnl: Huh, indeed, that hasn't been necessary for a long, long time. I wonder if the package that provides these lint rules if very outdated, i.e. has a years-old version that still includes this in its recommended set? It looks like we didn't explicitly enable it ourselves.

PR #6437, opened 2026-04-09
feat(blender): add auto-merge workflow for safe Dependabot PRs

groovecoder:
How to test:
1. Run dry-run against fx-private-relay:
   `REPO=mozilla/fx-private-relay GH_TOKEN=$(gh auth token) DRY_RUN=true python .github/blender/automerge-dependabot.py`
   * [ ] Verify output shows each PR evaluated against 5 gates
   * [ ] Verify major-bump PRs are skipped (e.g., `eslint`)
   * [ ] Verify patch/minor PRs check compat score and advisories

E.g.:
```
BLEnder automerge-dependabot: repo=mozilla/fx-private-relay dry_run=True
Fetching open Dependabot PRs for mozilla/fx-private-relay...
Found 2 open Dependabot PRs.

--- PR #6436: build(deps): bump cryptography from 46.0.6 to 46.0.7 ---
  Versions: 46.0.6 -> 46.0.7 (patch/minor)
  SKIP: compatibility score 78% < 80%

--- PR #6326: build(deps-dev): bump the eslint group across 1 directory with 2 updates ---
  SKIP: major version bump on eslint

=== Summary ===
Would merge: 0
Skipped: 2

Skip reasons:
  - #6436: compatibility score 78% < 80%
  - #6326: major version bump on eslint
```

Can also test against other repos:
```
     REPO=mozilla-services/guardian-website
     REPO=mozilla/blurts-server
```

### DIFF
--- a/.github/blender/automerge-dependabot.py
+++ b/.github/blender/automerge-dependabot.py
@@ -1,0 +1,483 @@
+#!/usr/bin/env python3
+"""BLEnder automerge-dependabot: auto-merge "safe" Dependabot PRs:
+
+  1. Author is dependabot[bot]
+  2. All CI checks pass
+  3. Update is patch or minor (not major)
+  4. Compatibility score >= 80%
+  5. No security advisories on the new version
+
+Environment variables:
+  REPO      -- GitHub repo, e.g. mozilla/fx-private-relay (required)
+  GH_TOKEN  -- GitHub token with contents:write and pull-requests:write (required)
+  DRY_RUN   -- Set to "true" to check gates without approving or merging (default: true)
+"""
+
+from __future__ import annotations
+
+import os
+import re
+import sys
+from dataclasses import dataclass, field
+from urllib.request import Request, urlopen
+
+import yaml
+from github import Auth, Github
+from github.PullRequest import PullRequest
+from github.Repository import Repository
+from packaging.specifiers import InvalidSpecifier, SpecifierSet
+from packaging.version import InvalidVersion, Version
+
+
+class SkipPR(Exception):
+    """Raised when a gate fails. Message becomes the skip reason."""
+
+
+@dataclass
+class DependencyUpdate:
+    """Structured metadata from Dependabot's commit message YAML."""
+
+    name: str
+    version: str  # new version
+    dependency_type: str  # direct:production, direct:development, indirect
+    update_type: str  # version-update:semver-major, etc.
+    group: str = ""
+
+
+@dataclass
+class PRMetadata:
+    """Parsed metadata for a Dependabot PR."""
+
+    dependencies: list[DependencyUpdate] = field(default_factory=list)
+    ecosystem: str = "unknown"
+    has_major: bool = False
+    old_version: str = ""  # first dependency only, from commit title
+    new_version: str = ""  # first dependency only, from YAML
+
+
+@dataclass
+class Config:
+    repo_name: str
+    token: str
+    dry_run: bool
+
+
+# --- Metadata extraction ---
+
+ECOSYSTEM_MAP = {
+    "npm_and_yarn": "npm",
+    "pip": "pip",
+    "github_actions": "actions",
+    "docker": "docker",
+    "bundler": "rubygems",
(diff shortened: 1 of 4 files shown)


.github/blender/automerge-dependabot.py
```diff
@@ -0,0 +1,483 @@
...
+    return "version-update:semver-patch"
+
+
+def parse_dependabot_yaml(
+    commit_message: str,
+) -> list[DependencyUpdate]:
+    """Extract structured dependency info from Dependabot's commit YAML.
+    The YAML block is between --- and ...
+    """
+    match = re.search(
```
joeherm, line 112: not blocking: can you add a comment on what this regex does? pulling a yaml block?

.github/blender/automerge-dependabot.py
```diff
@@ -0,0 +1,483 @@
...
+# Each gate raises SkipPR on failure.
+
+
+def gate_author(pr: PullRequest) -> None:
+    """Gate 1: Author must be dependabot[bot]."""
+    if pr.user.login != "dependabot[bot]":
+        raise SkipPR(f"author is {pr.user.login}, not dependabot[bot]")
+
+
+def gate_ci(repo: Repository, sha: str) -> None:
```
joeherm, line 243: not blocking: You might be able to skip this check if you use `@dependabot merge` ([docs](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-pull-request-comment-commands)) and replace `pr.enable_automerge(merge_method="MERGE")` with making that comment. Don't have a strong preference, but might reduce the code we need to maintain

joeherm (approved):
Looks good! Could be cool to build a standalone GHA that can be brought into multiple projects if we like how this works

.github/workflows/blender-automerge-dependabot.yml
```diff
@@ -0,0 +1,30 @@
...
+  automerge:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
+
+      - name: Install dependencies
+        run: pip install PyGithub PyYAML packaging
```
mansaj: question: is there a risk here not to pin the version?

.github/blender/automerge-dependabot.py
```diff
@@ -0,0 +1,483 @@
...
+        return ver in spec
+    except (InvalidVersion, InvalidSpecifier):
+        return True
+
+
+# --- Safety gates ---
+# Each gate raises SkipPR on failure.
+
+
+def gate_author(pr: PullRequest) -> None:
```
mansaj, line 237: q: are we double checking here? In `main()` then in `gate_author`(which exe inside of main)?

.github/blender/automerge-dependabot.py
groovecoder: It's the comment above which may be confusing because it's also the function-level doc-block. Should I separate the 2 comments?

.github/blender/automerge-dependabot.py
groovecoder: Yeah. The filter in `main()` is an optimization - it avoids running gates on PRs that will fail gate 1. The `gate_author` gate is the safety check.

.github/blender/automerge-dependabot.py
groovecoder: good idea! that also means we can remove the `contents: write` permission from the token for this.

.github/workflows/blender-automerge-dependabot.yml
groovecoder: Oh dang, good point. 🤔 ... since this whole idea is to keep dependencies up to date, we should really move these dependencies into something that dependabot can see too.

groovecoder: Okay, I moved these dependencies into their own `requirements.txt` so Dependabot can monitor these too.

PR #6601, opened 2026-06-03
fix(masks): fix MPP-4679: skip abuse metric for email-auto-created domain addresses

groovecoder:
Inbound email to a subdomain auto-creates `DomainAddress` records.

Skip `update_abuse_metric` when `first_emailed_at` is set, so only user-initiated mask creation counts toward the abuse threshold.

This PR fixes MPP-4679.

How to test:
  1. Set `MAX_ADDRESS_CREATION_PER_DAY=10` in your environment
  2. Create a premium user with a subdomain via the admin or shell
  3. Send 15+ emails to random addresses at the user's subdomain
     (e.g., random1@subdomain.mozmail.com through random15@...)
     using the SNS email handler or test fixtures
  4. Verify: the user's account is NOT flagged (`profile.is_flagged == False`)
     and all 15 DomainAddress records exist
  5. Now create 10 masks via the API (`POST /api/v1/domainaddresses/`)
  6. Verify: the user's account IS flagged after the 10th API-created mask

### DIFF
--- a/emails/models.py
+++ b/emails/models.py
@@ -268,7 +268,11 @@ class DomainAddress(models.Model):
             ).exists():
                 raise DomainAddrDuplicateException(duplicate_address=self.address)
 
-            self.user.profile.update_abuse_metric(address_created=True)
+            # MPP-4679: Only count toward abuse limits if user creates the mask.
+            # Addresses auto-created by inbound email (first_emailed_at is set)
+            # may not be user-initiated, so shouldn't flag the user's account.
+            if not self.first_emailed_at:
+                self.user.profile.update_abuse_metric(address_created=True)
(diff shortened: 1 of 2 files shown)


joeherm (approved):
LGTM from a code perspective. I wonder if we want to do some follow-up thinking about the autocreation feature and tracking potential abuse. Two things are true:

1. An attacker shouldn't be able pause a Relay user victim's account by sending emails
2. We might still want some way to track rate of address creation / suspicious behavior?

I don't think any of that blocks this PR merging, but should we file some sort of spike and coordinate with product about how to handle this case more broadly?

groovecoder:
I filed https://mozilla-hub.atlassian.net/browse/MPP-4697 to follow-up.

PR #6617, opened 2026-06-04
fix(phones) fix MPP-4682: enforce one RelayNumber per user at database level

groovecoder:
Concurrent `POST` requests to `/api/v1/relaynumber/` could bypass the application-level one-phone-per-user check. The `ForeignKey` on `RelayNumber.user` had no unique constraint. Requests that passed the existence check before any row was committed each inserted a new row.

- Change `RelayNumber.user` from `ForeignKey` to `OneToOneField`
- Wrap `RelayNumberViewSet.create` in `transaction.atomic` with `select_for_update` to serialize concurrent attempts
- Catch `IntegrityError` as a fallback if the constraint fires
- Add model test that bypasses `save()` to verify the DB constraint

This PR fixes MPP-4682.

How to test:
  1. Check out the branch and run `python manage.py migrate`
  2. Verify the migration applies without error
  3. Run `pytest phones/tests/models_tests.py`
     * [ ] All should pass
  5. Run `pytest api/tests/phones_views_tests.py`
     * [ ] All should pass
  7. (Optional - we can ask the reporter to try to reproduce the race fix manually):
     a. Set up a phone-entitled user with a verified RealPhone
     b. Send concurrent `POST /api/v1/relaynumber/` requests with different numbers
     c. Only one should return `201`; the rest should return `400`
     d. Confirm the user has exactly one `RelayNumber` in the database


api/views/phones.py
```diff
@@ -314,10 +315,19 @@ def create(self, request, *args, **kwargs):
...
         """  # noqa: E501  # ignore long line for URL
         incr_if_enabled("phones_RelayNumberViewSet.create")
-        existing_number = RelayNumber.objects.filter(user=request.user)
-        if existing_number:
-            raise exceptions.ValidationError("User already has a RelayNumber.")
-        return super().create(request, *args, **kwargs)
+        # Use a transaction with select_for_update to prevent race conditions.
+        # Without this, concurrent requests can all pass the existence check
+        # before any row is committed, bypassing the one-number-per-user limit.
+        with transaction.atomic():
```
joeherm, line 323: nitpick(non-blocking): I don't think that the `select_for_update()` and atomic transaction are strictly needed if this is a `OneToOne` field, but it will probably catch the error in a more user friendly way. We should throw a DB level integrity error once the migration is run though, even if this code block isn't here.

Probably worth updating the comment since I think it's slightly inaccurate as it stands.

joeherm (approved):
Migration and tests look good- left a small comment about the new logic

PR #6663, opened 2026-06-30
feat(api): add cache-control headers and per-user rate limits to high-traffic endpoints

groovecoder:
Defense-in-depth against client-driven request surges like the FF 151 incident ([IIM-173](https://mozilla-hub.atlassian.net/browse/IIM-173)). Browser-side `Cache-Control` headers prevent re-fetching for 60s even when client-side caching breaks. Per-user throttles (30/min) cap runaway loops at the origin.

- `Profile`, `RelayAddress`, `DomainAddress` `GET`s: `Cache-Control: private, max-age=60`
- `ProfileViewSet`: 30/min per-user throttle via `PROFILE_RATE_LIMIT` (scope `"profile"`)
- `RelayAddress`/`DomainAddress` `ViewSets`: 30/min per-user throttle via `ADDRESS_RATE_LIMIT` (scope `"address"`)
- Distinct throttle scopes keep profile and address limits in separate buckets instead of colliding on the default `"user"` scope
- Rate limits relaxed to 1000/min in test/local/dev environments

`runtime_data` is intentionally left uncached: its response varies per user (waffle flags, request geo) and SWR already memoizes it for the session, so a shared/edge cache would risk cross-user data.

This PR fixes MPP-4692.

How to test:

[IIM-173]: https://mozilla-hub.atlassian.net/browse/IIM-173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/views/privaterelay.py
+++ b/api/views/privaterelay.py
@@ -35,6 +35,7 @@ from rest_framework.status import (
     HTTP_202_ACCEPTED,
     HTTP_400_BAD_REQUEST,
 )
+from rest_framework.throttling import UserRateThrottle
 from rest_framework.viewsets import ModelViewSet
 from sentry_sdk import capture_exception
 from waffle import flag_is_active, get_waffle_flag_model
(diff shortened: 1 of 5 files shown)


privaterelay/settings.py
```diff
@@ -788,6 +788,15 @@ def set_index_cache_control_headers(
 
 VCARD_RATE_LIMIT = config("VCARD_RATE_LIMIT", "10/minute")
 
+if IN_PYTEST or RELAY_CHANNEL in ["local", "dev"]:
+    _DEFAULT_PROFILE_RATE_LIMIT = "1000/minute"
+    _DEFAULT_ADDRESS_RATE_LIMIT = "1000/minute"
+else:
+    _DEFAULT_PROFILE_RATE_LIMIT = "30/minute"
+    _DEFAULT_ADDRESS_RATE_LIMIT = "30/minute"
```
joeherm, line 796: Is this going to break any E2E tests that auto generate FREE_MAX emails?

api/views/privaterelay.py
```diff
@@ -62,19 +63,35 @@
...
+    rate = settings.PROFILE_RATE_LIMIT
+
+
 @extend_schema(tags=["privaterelay"])
 class ProfileViewSet(ModelViewSet):
     """Relay user extended profile data."""
 
     serializer_class = ProfileSerializer
     permission_classes = [IsAuthenticated, IsOwner]
+    throttle_classes = [ProfileRateThrottle]
```
joeherm, line 77: note: I like throttling here as opposed to the WAF, but I wonder if it's worth adding something to the runbook that we can rate-limit in multiple spots.

joeherm (approved):
I took a look at the [FF code](https://github.com/mozilla-firefox/firefox/blob/main/toolkit/components/satchel/integrations/FirefoxRelay.sys.mjs) and it seems like Addresses and Profile are the two endpoints being programmatically fetched with regularity, so I think we've got our bases covered!

privaterelay/settings.py
groovecoder: added `"stage"` to the list of channels that get `1000/minute` limits.

api/views/privaterelay.py
groovecoder: yeah that's fine. you mean https://mozilla-hub.atlassian.net/wiki/spaces/PXI/pages/2166161419/Relay+Alert+Runbook ?

PR #6664, opened 2026-06-30
fix(emails): hide Relay user's real email in reply body

### LINKED ISSUE #1377: Outlook exposes FXA email in quoted text after a reply.
A customer was able to reproduce this by sending an email from a Yahoo address to a masked Outlook email address. They noted that the masked email to the Yahoo email address in the quoted part of the reply. 

1. Send an email to masked Outlook mozmail alias from a Yahoo email address.
2. Reply to the Yahoo email from the masked Outlook web email
3. On Yahoo, observe the quoted text of the response.

Result: The masked outlook email address is visible in the quoted text section.

groovecoder:
Relay rewrites a reply's headers to the mask but sent the body as-is. Clients like Outlook quote the original message, which can include the user's real email in a quoted "To:" line, leaking it to the recipient.

Replace the user's real email with the mask in the plain text and HTML reply bodies, mirroring the get_body/set_content pattern the forwarding flow already uses.

This PR fixes #1377, MPP-1395

How to test (easier on dev server):
1. From an external address, send an email to a (premium) Relay mask.
2. From the account's real inbox (e.g. Outlook web), reply to the forwarded message, leaving the quoted original text in place. (or you can manually force this by just typing the real email address in the reply)
3. View the received reply at the external sender.
   - The reply's From and the quoted "To:" line both show the mask (a1b2c3d4@test.com style), not the user's real email.
   - The user's real email appears nowhere in the message body.
4. Edge case: reply with mixed-case real email in the quote — matching is case-insensitive, so it is still masked.

### DIFF
--- a/emails/tests/views_tests.py
+++ b/emails/tests/views_tests.py
@@ -979,6 +979,28 @@ class SNSNotificationRepliesTest(SNSNotificationTestBase):
             expected_fixture_name="s3_stored_replies",
         )
 
+    def test_reply_hides_relay_user_real_email_in_body(self) -> None:
+        """
+        The user's real email is replaced with the mask in the reply body.
+
+        A client like Outlook quotes the original message, which can include the
+        user's real email (e.g. in a quoted "To:" line). It must not leak to the
+        recipient (MPP-1395 / #1377).
+        """
+        self.mock_get_content.return_value = create_email_from_notification(
+            EMAIL_SNS_BODIES["s3_stored_replies"],
+            text=(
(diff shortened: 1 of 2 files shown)


joeherm (changes requested):
Am I correct that the test only exercises the `plain` body? That's my understanding of how `create_email_from_notification` works. Could we add one more test that exercises the replacement in html bodies?

groovecoder:
Good catch. I added code to test the html too.

joeherm (approved):

PR #6692, opened 2026-07-08
build(deps): bump react, react-dom, react-aria, and react-stately

groovecoder:
Combines three open Dependabot React PRs into one so they can share a single lockfile resolution and go through CI together.

## What this replaces

- #6622 — react and react-dom 19.2.6 -> 19.2.7
- #6667 — react-aria ^3.49.0 -> ^3.50.0
- #6671 — react-stately ^3.46.0 -> ^3.48.0

## Why combine them

The three Dependabot PRs each show the same failing checks (Coveralls, Build Docker Images, Netlify). Those fail because Dependabot PRs run with a restricted token and no secret access, not because of the code. A normal PR gets full secret access, so those checks should pass here.

More importantly, the react bump has a lockfile hazard that Dependabot cannot see. Bumping react to an exact 19.2.7 in the `frontend` workspace makes npm nest react 19.2.7 under `frontend/node_modules` while leaving react 19.2.6 hoisted at the repo root to satisfy peer deps. Two copies of React break the jest suite with "Invalid hook call". Dependabot never catches this because it skips the frontend test job. Running `npm dedupe` collapses react/react-dom to a single hoisted 19.2.7 and fixes the tests.

## Lockfile note

The dedupe also refreshes a handful of unrelated transitive dev deps to newer in-range patch versions (yargs, js-yaml, postcss, tough-cookie, etc.). That churn is unavoidable: CI installs with `npm ci` (`Dockerfile:9`), so the lockfile must be a consistent npm-generated resolution. It cannot be hand-trimmed.

Closes #6622
Closes #6667
Closes #6671


vpremamozilla:
i tested the AppPicker and it seems to show some weird blue rectangle around the first element on hover (see screenshot). not reproducible on dev or prod 

Screenshot 2026-07-08 at 11 27 28 AM

vpremamozilla (approved):

PR #6693, opened 2026-07-08
fix(frontend): bypass browser cache on API reads so mutations refresh

groovecoder:
MPP-4692 added `Cache-Control: max-age=60` to the address and profile list endpoints to protect the origin from the autofill client. That header also made the browser serve a stale list to SWR after a delete or block/forward toggle, so the dashboard did not update until a page refresh. Deleting the same row twice then hit a 404.

Set `cache: "no-store"` on the fetcher so SWR reads always revalidate. The `Cache-Control` header still helps the autofill client, and the per-user throttle still caps the origin.

This PR fixes MPP-4717.

How to test:
1. Sign into the Relay dashboard with a few masks.
2. Expand a mask and delete it. Repeat on another mask.
3. Toggle a mask between forwarding and blocking.
4. Verify the list updates each time with no page refresh.
5. In DevTools Network, confirm `GET /api/v1/relayaddresses/` returns a real 200, not "(from disk cache)".

### DIFF
--- a/frontend/src/hooks/api/api.test.ts
+++ b/frontend/src/hooks/api/api.test.ts
@@ -36,6 +36,23 @@ describe("useApiV1", () => {
     });
   });
 
+  it("fetches with cache: no-store so mutate() bypasses the browser cache (MPP-4717)", async () => {
+    const mockResponse = {
+      ok: true,
+      status: 200,
+      json: async () => ({}),
+    } as Response;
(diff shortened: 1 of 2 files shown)


vpremamozilla:
- code changes make sense 
- tested locally and everything looks good

vpremamozilla (approved):

PR #6695, opened 2026-07-10
fix MPP-4700: surface OAuth token failure on profile refresh

groovecoder:
A stale-scope Mozilla Accounts token made profile refresh fail without telling the user. The error was caught and returned 202, so the browser kept showing old profile data. Now the refresh returns 401 and sends the user to re-authenticate.

This PR fixes MPP-4700.

How to test:
1. Sign up as a new user with Mozilla Accounts. Confirm you land in the dashboard as a signed-in user.
2. Log out. Log back in with the same account. Confirm you return to the dashboard signed in.
3. Close the tab. Open the site again in the same browser. Confirm you return as a signed-in user without re-entering credentials.
4. While signed in, reload the dashboard a few times. Confirm your profile data (email, plan) stays correct and you are never bounced to login.

### DIFF
--- a/frontend/src/hooks/api/profile.test.ts
+++ b/frontend/src/hooks/api/profile.test.ts
@@ -1,5 +1,5 @@
 import { renderHook, waitFor } from "@testing-library/react";
-import { useProfiles } from "./profile";
+import { profileFetcher, useProfiles } from "./profile";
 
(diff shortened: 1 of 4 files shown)


vpremamozilla (approved):
tested locally, looks good

PR #6719, opened 2026-07-21
feat(emails): cap worker DB statement time

groovecoder:
The email worker had no limit on how long a query could run. Set statement_timeout on the worker's Postgres session as a backstop for a slow query.

The row-lock wait is set separately by lock_timeout on the Cloud SQL instance in webservices-infra.

This PR fixes https://mozilla-hub.atlassian.net/browse/MPP-4723.

How to test:
1. Merge this so it landes on the dev server.
2. Check the dev server email job still works.
3. Check the dev server logs show "Message processed" with normal `message_process_time_s` and no new "statement timeout" / OperationalError errors from the relay-emails container.
  * We *DO* expect to see `message_process_time_s` log messages when we get to the next prod surge of emails.

### DIFF
--- a/emails/management/commands/process_emails_from_sqs.py
+++ b/emails/management/commands/process_emails_from_sqs.py
@@ -20,8 +20,10 @@ from typing import Any, cast
 from urllib.parse import urlsplit
 
 from django import setup
+from django.conf import settings
 from django.core.management.base import CommandError
 from django.db import connection
+from django.db.backends.base.base import BaseDatabaseWrapper
 from django.http import HttpResponse
 
 import boto3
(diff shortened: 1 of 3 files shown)


joeherm (approved):
Approving since I don't have any issue with this approach, but a few thoughts:

1. Are we actually having timeout issues that aren't related to locking? I'm wondering if running this `set_config` command might be a bit overkill?
2. An alternative approach here would be to create a separate user/role for email processing that has the timeout permanently set. This would allow us to remove most of this code and reduce the number of DB calls. Not necessary, but just pointing it out for the future.

groovecoder:
Thanks @joeherm good points.

1. Right - we don't have evidence of non-lock timeout issues. Every 120s hang we've traced goes back to the row lock in `update_abuse_metric`, and that's now bounded by `lock_timeout = 5s` on the Cloud SQL instance. https://github.com/mozilla/webservices-infra/pull/11854
   * So, I'll actually just close this PR rather than merge it.
2. I filed https://mozilla-hub.atlassian.net/browse/MPP-4731 for the dedicated-role idea.

PR #6759, opened 2026-08-04
fix(emails) MPP-4739: tracker regex backtracking

groovecoder:
`convert_domains_to_regex_patterns` built a subdomain match from `(\S*\.)*`. `\S*` matches a dot, so a dotted URL could be split into groups exponentially many ways, and the engine tried them all before failing. Cost grew about 4x per added dot.

Every one of the 214 level-one tracker domains paid that cost, because they all fail to match. `remove_trackers` then scanned the content three times, and `_convert_html_content` added two more passes when tracker sampling was on. A 130-byte email holding one link with 17 dots ran past the worker's 120s per-message kill. Prod saw about 1200 kills a day across 400 messages, each retried three times, or roughly 40 pod-hours a day.

Replace `\S*` with `[^\s"'.]*` so a subdomain label cannot contain a dot. The regex now has one way to match a URL instead of many, so it no longer backtracks. The same URLs still match: checked against the old pattern across 17 URL shapes with no difference.

Add a guard test that runs the real tracker list under a 10s alarm. The existing tests stub the list down to three domains, which is fast whatever the pattern does, so they could not catch this.

This PR fixes MPP-4739.

How to test:
1. Confirm tests still pass
2. Run the following code on main and on this branch:
```
       from emails.utils import remove_trackers
       url = "https://click.mailer.example/r?u=" + "seg." * 18 + "end&id=9"
       remove_trackers(f'<a href="{url}">x</a>', "s@example.com", 1700000000000)
```
   * [ ] On this branch, it should return immediately. On `main`, it takes significantly longer.

### DIFF
--- a/emails/tests/utils_tests.py
+++ b/emails/tests/utils_tests.py
@@ -1,8 +1,10 @@
 import base64
 import json
 import random
+import signal
 import zlib
 from base64 import b64encode
+from types import FrameType
 from typing import Literal, TypedDict
 from unittest.mock import patch
 from urllib.parse import quote_plus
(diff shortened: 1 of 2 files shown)


joeherm (approved):
Awesome find with this one

PR #6784, opened 2026-08-11
Fix 6602 review findings

groovecoder:
This PR fixes review comments from @Vinnl: https://github.com/mozilla/fx-private-relay/pull/6602#pullrequestreview-4819510587.

How to test:
  1. `cd frontend && npm run dev:mocked`
  2. Open http://localhost:3000/mock/login
  3. Click any mock account button.
     * [ ] You land on the dashboard and the console shows no "too much recursion" error
  4. Set your browser's preferred language to one Relay translates, such as Dutch.
  5. Reload the site
     * [ ] no React hydration errors in the console

### DIFF
--- a/frontend/src/pages/_app.page.tsx
+++ b/frontend/src/pages/_app.page.tsx
@@ -1,19 +1,14 @@
 import "../styles/globals.scss";
-import {
-  createElement,
-  useEffect,
-  useRef,
-  useState,
-  useSyncExternalStore,
-} from "react";
+import { createElement, useEffect, useRef, useState } from "react";
 import type { AppProps } from "next/app";
 import { useRouter } from "next/router";
-import { LocalizationProvider, ReactLocalization } from "@fluent/react";
+import { LocalizationProvider } from "@fluent/react";
 import { OverlayProvider } from "react-aria";
 import ReactGa from "react-ga";
 import { getL10n } from "../functions/getL10n";
 import { AddonDataContext, useAddonElementWatcher } from "../hooks/addon";
 import { ReactAriaI18nProvider } from "../components/ReactAriaI18nProvider";
+import { useHasRenderedClientSide } from "../hooks/hasRenderedClientSide";
 import { useIsLoggedIn } from "../hooks/session";
 import { useMetrics } from "../hooks/metrics";
 import {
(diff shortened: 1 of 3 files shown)


frontend/src/pages/_app.page.tsx
```diff
@@ -24,15 +43,21 @@ function MyApp({ Component, pageProps }: AppProps) {
...
   // on the client-side yet using @fluent/react, see
   // https://github.com/projectfluent/fluent.js/wiki/ReactLocalization/43a959b35fbf9eea694367f948cfb1387914657c#flexibility
-  const [l10n] = useState<ReactLocalization>(() =>
-    getL10n({ deterministicLocales: typeof window === "undefined" }),
+  const [deterministicL10n] = useState<ReactLocalization>(() =>
+    getL10n({ deterministicLocales: true }),
+  );
+  const [negotiatedL10n] = useState<ReactLocalization>(() =>
+    getL10n({ deterministicLocales: false }),
   );
```
Vinnl: I think these don't need to be wrapped in a `useState`? The values stay constant throughout the component's lifetime, so I think you could even move them outside of the component, to the module's top level.

frontend/src/pages/_app.page.tsx
```diff
@@ -16,6 +22,19 @@ import {
...
+// different value while React is still matching the DOM against the
+// pre-rendered HTML than it returns afterwards. The `useL10n` hook uses the
+// same trick.
+const subscribeToNothing = () => () => {};
+const useIsHydrating = () =>
+  useSyncExternalStore(
+    subscribeToNothing,
+    () => false,
+    () => true,
+  );
```
Vinnl: Ah cool, I guess this is basically a modern version of https://github.com/mozilla/blurts-server/blob/main/src/app/hooks/useHasRenderedClientSide.tsx?

Might be worth extracting that into a separate hook? To reduce the overhead from having to read through the logic here.

(I like how explicit `subscribeToNothing` is btw - I probably would've just put in the anonymous function, which wouldn't be clear to the reader at all.)

Vinnl (approved):

frontend/src/pages/mock/login.page.tsx
```diff
@@ -138,7 +138,7 @@ const MockLogin: NextPage = () => {
 // Wrapped in a function so eslint-plugin-react-hooks doesn't flag it
 // as an impure call during render (it specifically tracks Date.now).
```
Vinnl: Oh, should also note here: this does resolve in a hydration error on the login page, because the timestamp returned during the build will differ from the timestamp on first render on the client-side.

Since this is just the mock login page, that is fine, but we might want to explicitly call that out in a comment.

(And in fact, instead of wrapping this in `getTimestamp`, I would just add an eslint ignore comment and add the remark to that, to save that layer of indirection.)

groovecoder:
Addressed review comments.

PR #6801, opened 2026-08-13
fix(emails): stop silently dropping emails SES rejects

groovecoder:
Relay dropped between 800 and 1,300 emails a day in production, measured across two 24-hour windows. The user never got the mail and was never told. Two defects combined.

Some senders emit a duplicate MIME-Version header. _replace_headers keeps mime-version and every content-* header from the incoming email so the body still parses, but it kept every copy rather than the first. SES refuses a message with a duplicate header, so the forward failed. That was 93% of the volume.

The email code signals a retryable failure by returning 5xx rather than raising. Nothing read the status. success_callback in the SQS worker was an empty stub, so results["success"] stayed True and the batch loop deleted the message. PROCESS_EMAIL_DELETE_FAILED_MESSAGES defaults to False, so the intent was that failures are kept on the queue. The 5xx path defeated that. This defect is the older of the two: the worker has discarded the return value since its first commit in April 2022.

Collapse each duplicated kept header to its first value. Reassign the parsed header object rather than the raw string, because a folded raw value contains newlines and the email policy refuses to store those.

Have success_callback read result.status_code and mark a 5xx as failed. The message then stays on the queue and SQS redrives it, which is what every can_retry=True drop log already claimed was happening. This applies to every 5xx in the email path, not just SES failures: error_from_header and error_storage had the same silent deletion.

Retrying a permanent SES rejection costs three more sends and then dies in the DLQ, which the sqs-dlq cron deletes. Add is_permanent_ses_error, return 400 for those, and increment ses_permanent_rejection so the volume is counted rather than vanishing. The list holds one code, InvalidParameterValue. An unrecognized code stays transient, so an unexpected failure retries and ends up visible instead of discarded. The 400 also lets _sns_notification remove the S3 object, since no retry is coming.

See MPP-4746.

This PR fixes #<issue ID>.

How to test:

- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol / Nebula colors where applicable (see `/frontend/src/styles/colors.scss`).

# New feature description

# Screenshot (if applicable)

Not applicable.

# How to test

- [ ] Customer Experience team has seen or waived a demo of functionality.
- [ ] I've added or updated relevant docs in the docs/ directory
- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol / Nebula colors where applicable (see `/frontend/src/styles/colors.scss`).

### DIFF
--- a/emails/tests/fixtures/duplicate_mime_version_incoming.email
+++ b/emails/tests/fixtures/duplicate_mime_version_incoming.email
@@ -1,0 +1,11 @@
+Subject: Duplicate MIME-Version Email
+From: A server <root@server.example.com>
+To: ebsbdsan7@test.com
+Date: Wed, 27 Sep 2023 16:33:12 +0000
+MIME-Version: 1.0
+Content-Type: text/plain; charset="utf-8"
+Content-Transfer-Encoding: 7bit
+MIME-Version: 1.0
+
+This email has two MIME-Version headers.
+SES rejects a forwarded copy that keeps both.
--- a/emails/management/commands/process_emails_from_sqs.py
+++ b/emails/management/commands/process_emails_from_sqs.py
@@ -374,6 +374,7 @@ class Command(CommandFromDjangoSettings):
         Return is a dict suitable for logging context, with these keys:
         * success: True if message was processed successfully
         * error: The processing error, omitted on success
+        * status_code: The HTTP status from the email code, omitted if it raised
         * message_body_quoted: Set if the message was non-JSON, omitted for valid JSON
         * pause_count: Set to 1 if paused due to temporary error, or omitted
           with no error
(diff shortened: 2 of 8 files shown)


emails/tests/fixtures/duplicate_mime_version_incoming.email
groovecoder, line 1: Note: these emails with duplicate `Mime-Version` headers are 92% of the dropped emails.

groovecoder:
Superseded by https://github.com/mozilla/fx-private-relay/pull/6801 and another incoming PR to split the 2 concerns into separate PRs.

PR #6802, opened 2026-08-13
fix(emails): drop duplicate MIME-Version and Content-Type headers

groovecoder:
Some senders emit the same header twice. SES refuses a message with a duplicate header, so the forward failed and the email was lost. In production that was about 1,170 rejections a day, 93% of every SES rejection Relay saw.

`_replace_headers` keeps `mime-version` and every `content-`* header from the incoming email, because those describe a body Relay reuses rather than rebuilds. Dropping them is not an option: without `Content-Type` the MIME boundary is gone and Python cannot even serialize the message. But the rule only said which headers to keep, never how many of each. Headers outside that set were already collapsed as a side effect, because deleting a header removes every copy of it.

Collapse each duplicated kept header to its first value and record the count in the forwarding issues log. That log entry is the first telemetry Relay has had on duplicate headers.

Reassign the parsed header object rather than the raw string. A folded raw value contains newlines, and the email policy refuses to store those.

The duplicate arrives from outside. Relay never writes `MIME-Version`, and the three places the email library writes it either delete first or check for an existing header. A clean incoming email still produces exactly one top-level `MIME-Version` end to end.

This PR fixes MPP-4746.

How to test:
* `pytest emails/tests/views_tests.py`

### DIFF
--- a/emails/tests/fixtures/README.md
+++ b/emails/tests/fixtures/README.md
@@ -42,6 +42,8 @@ incoming email. Their names all end in `_email_sns_body.json`.
 These fixtures were created outside of an SNS notification, such as exporting from a
 mail client. They are stored in the [Internet Message Format][] (IMF).
 
+- `duplicate_mime_version_incoming.email` - Has two `MIME-Version` headers. SES rejects
+  a forwarded copy that keeps both.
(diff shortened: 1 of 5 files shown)


joeherm (approved):
Looks good. Thanks for the tests. Logic makes sense to me if we want to allow through these emails.

